PR #124, opened 2011-12-07
Translated tip_select_destination_point(_t)

Brainik:


moonandroid:
hi,
i'm using osmand with android.
and faced issue with map rendering

PR #882, opened 2014-09-19
Created method to send notifications to android wear

Bars107:


ronsmits:
Is this version already deployed on the playstore? I walked a route using osmand and my moto 360 did not give me any feedback.

PR #968, opened 2014-11-21
Remove one verification of lanes and turn:lanes

saiarcot895:
This removes one check to verify that the number of lanes based on `lanes` and `turn:lanes` are equal, and instead logs an error.

In addition, `countLanes()` is changed so that it does all the calculations needed to get the lanes (and makes it similar to `getTurnLanesString()`).

This is untested.


vshcherb:
Hi, you might be interested in some weird cases and potentially bugs on our side :) 
https://code.google.com/p/osmand/issues/detail?id=2502

saiarcot895:
Already on it.

PR #1011, opened 2014-12-17
Disambiguate "v1.9 routing" strings in strings.xml

aceman444:
There are some localizers wondering about these strings at
https://hosted.weblate.org/projects/osmand/main/sk/translate/?type=comments
Would this wording be better?


aceman444:
See https://groups.google.com/forum/#!topic/osmand/5FFJ6Hkvx_U for other users potentially having problem with these rules.

vshcherb:
This parameter was deleted

PR #1033, opened 2015-01-08
modernize location sharing

eighthave:
This eliminates the "Share location" submenu, and replaces it with proper Android Intents.  This makes for a more flexible and standard experience.  It also adds a "View in other app" option to the main Context menu.  The URLs are also revamps to preserve privacy by using HTTPS and `geo:` when possible.

It no longer shares the download.osmand.net URL since that just redirects to Google Maps.  :-/

One remaining question is: which branch should my commits be based on?  This pull request is based on master, but these commits should easily apply to r1.9 as well.  Should I submit two pull requests?

_Updated to add a better description:_

The functionality of the current "Share location" menu is included, but it is just implemented using standard Android UX patterns (e.g. the _Share_/`ACTION_SEND` menu).

* to send an email, click "Share location", then choose your email app
* to send an SMS, click "Share location", then choose your SMS app
* to send another kind of message, click "Share location", then choose your messaging app (ChatSecure, Whatsapp, Hangouts, etc. etc.)
* to make a QR Code, click "Share location", then choose "BarCode Scanner"
* to copy to clipboard, click "Share location", then choose "Copy to clipboard"
* to view current location in a different app (aka "geo:" menu item), click "View in other app"

If you want to try these for yourself, I've made a test build.  This is for testing only, it is not a trusted build, only use it on test devices.
https://guardianproject.info/builds/OsmAnd/OsmAnd-debug.apk

_update:_

One thing to note: I did not change net.osmand.plus.activities.actions.ShareDialog.java. It is almost exactly the same of ShareLocation, which this replaces, but I couldn't figure out where in the UI it is used, so I had no way to test it. So I left it as is. Ideally, it would be replaced with the final version of this pull request as well.


OsmAnd-java/src/net/osmand/util/MapUtils.java
```diff
@@ -274,7 +281,29 @@ public int compare(MapObject o1, MapObject o2) {
 			}
 		});
 	}
-	
+
+    /* Share/SEND location to another app for a specific action (message, QRCode, etc) */
+	public static void shareLocation(Context ctx, double latitude, double longitude, int zoom) {
```
vshcherb: This class could not use any Android classes. Also it is not really clear why do we avoid previous Dialog with QR code and direct Email/SMS selection

OsmAnd/src/net/osmand/plus/activities/MapActivityActions.java
```diff
@@ -318,6 +318,8 @@ public void contextMenuPoint(final double latitude, final double longitude, fina
 				R.drawable.ic_action_search_light).reg();
 		adapter.item(R.string.context_menu_item_share_location).icons(
 				R.drawable.ic_action_gshare_dark, R.drawable.ic_action_gshare_light).reg();
+		adapter.item(R.string.context_menu_item_view_location).icons(
```
vshcherb: I wouldn't support idea of having extra item in the location menu list, I would rather prefer to have as a sub selection on share location dialog, next QR code or so.

vshcherb:
Most likely it will be no 1.9 release so it is unnececessary to create additional pull-request

eighthave:
Does that also mean no more 1.8.x releases?

OsmAnd-java/src/net/osmand/util/MapUtils.java
eighthave: Ok, I'll remove Android classes from OsmAnd-java.

About the old menu, first some menu items didn't really work fully.  For example, the SMS menu item didn't work with my SMS app TextSecure, and it caused crashes on tablets with no SMS.  And not all relevant apps were supported using those menu items.

Also, this menu is not a standard design pattern for Android apps.  Share/`ACTION_SEND` menus are provided Android, and they are what users expect to see when clicking on something called "Share" along with the standard "Share" icon.

This new menu item allows the user to choose many different SMS, messaging, email, and QR Code options, all from the standard Share/`ACTION_SEND` menu.

OsmAnd/src/net/osmand/plus/activities/MapActivityActions.java
eighthave: I think the "view" menu item is optional.  It seems handy but not necessary.  Right now, everything but the `geo:` menu item is replaced by a standard Share menu with a list of apps, from which the user chooses where the text gets sent to.  The `geo:` menu item just enables sending a location to a different Map app, or at least that is all I was able to get it to do.  So if you don't want it, I'll just remove it.

vshcherb:
I still see some sense having 1 extra dialog because it provides extra option of QR-code, Clipboard and also it is proper place to Open "geo:" intent. On the other hand I agree that Email/SMS is not so important  and it could be just text. I will consult with other people, but for now I would agree to to replace first 2 items Email, SMS -> Send which will trigger standard Android share.

eighthave:
Looks like I didn't describe the features of this pull request well enough.  The functionality of the current "Share location" menu is included, but it is just implemented using standard Android UX patterns (e.g. the _Share_/`ACTION_SEND` menu).

* to send an email, click "Share location", then choose your email app
* to send an SMS, click "Share location", then choose your SMS app
* to send another kind of message, click "Share location", then choose your messaging app (ChatSecure, Whatsapp, Hangouts, etc. etc.)
* to make a QR Code, click "Share location", then choose "BarCode Scanner"
* to copy to clipboard, click "Share location", then choose "Copy to clipboard"
* to view current location in a different app (aka "geo:" menu item), click "View in other app"

If you want to try these for yourself, I've made a test build.  This is for testing only, it is not a trusted build, only use it on test devices.
https://guardianproject.info/builds/OsmAnd/OsmAnd-debug.apk

On my Samsung tablet and Motorola phone, the use case @sonora described of editing documents and pasting links works fine.  "Copy to clipboard" is a standard feature.  My OmniROM tablet lacks "Copy to clipboard", which is odd.  I blame OmniROM since its more hacker focused. Turns out you can just install an app to do that:
https://play.google.com/store/apps/details?id=com.silentlexx.sharetoclipboard

I also found an issue with sharing via Bluetooth, the HTML doesn't get properly formatted.

I think this particular issue highlights what will need to happen in order to simplify osmand's user experience (which is the #1 request, according to the poll: https://polldaddy.com/poll/8543995/?view=results ).  One good way to simplify the experience is to follow as many of the Android design patterns as possible, then people don't have to learn new interactions to do things.

jkufner:
Hans-Christoph Steiner wrote, on 9.1.2015 10:41:
> With a properly implemented /Share//|ACTION_SEND|, copying to the
> clipboard should no longer be needed since you can just choose the app
> to send the text directly. Can anyone describe a use case for copy to
> clipboard that is not covered by this /Share/ implementation?

I often use geo: link in calendar event location field. It allows me to
exactly specify my destination and include comment. For example:

	geo:50.0800297,14.4297132# Some street, 1st floor

This way the entire string in location field is valid URL, so Osmand
will handle it correctly, and yet it is easy to handle by human.

I think we need two buttons/menu items: One for sharing (as usual on
Android) and another for copying to clipboard.

eighthave:
@jkufner that sounds useful.  I don't understand how you currently do that with OsmAnd.  Or are you saying to add this feature?  Right now in OsmAnd, selecting "Clipboard" gives you the text that is sent via an email.  With my changes, selecting "Share location" then "Copy to clipboard" gives you the same result.

eighthave:
One thing to note: I did not change `net.osmand.plus.activities.actions.ShareDialog.java`. It is almost exactly the same of `ShareLocation`, which this replaces, but I couldn't figure out where in the UI it is used, so I had no way to test it.  So I left it as is.  Ideally, it would be replaced with the final version of this pull request as well.

jkufner:
I copy&paste coordinates from web maps (Google maps or Mapy.cz; using
"What's here" feature) into Thunderbird Lightning calendar on my laptop
and then I let sync to pass everything to my phone before going
somewhere. Later on my phone I can click location in aCalendar and I get
exactly the same point I choosed.

Current "Share" functionality in Osmand is mostly useless and I'm very
happy that something is going on about that :)

eighthave:
I just had a thought: we could make OsmAnd provide an `IntentFilter` that gives an option called "Copy as geo:".  This would use the existing URI `IntentFilter`s for various http:// links from maps (maps.google.com, openstreetmap.org, etc.), and it would grab the latlong from the URL, then put a proper `geo:` link into the Clipboard.  This could either work when a URL is clicked on (i.e. _View_/`ACTION_VIEW`) or as a _Share_/`ACTION_SEND` item that then does a search/replace on the `EXTRA_TEXT` and replaces known URIs with `geo:` URIs.

sonora:
Hans-Christoph,

thanks for the clarifications! My fault entirely, I misread your statement
 "Can anyone describe a use case for copy to clipboard _*that*_ is not
covered by this *Share* implementation?*" *as* "* Can anyone describe a use
case for copy to clipboard _*which*_ is not covered by this *Share*
implementation?", assuming that copy-to-clipboard would entirely not be
supported anymore (and replaced by "select target app" only)  :-)

Mea culpa, no problem then!

Thx -- Hardy

eighthave:
I just checked out the sharing to Bluetooth.  It just takes the whole text and treats it as a URL and wraps it in HTML, so its not really useful.  But the current OsmAnd "Share location" options do the same thing, so I'm going to leave it at that.

@vshcherb I pushed rebased commits that move the stuff out of  `MapUtils` and into `MapActivity`, which makes more sense anyway.  Any new thoughts on the menu?  The "Share location" menu in this pull request does everything that the old one did, and more, except for the `ACTION_VIEW` of the `geo:` link.  But `ACTION_VIEW` and `ACTION_SEND` do not belong on the same submenu, they are not the same category of actions.  I'm open to suggestions for better places to stick the `ACTION_VIEW geo:` button.

eighthave:
I stripped this pull request down to the two commits directly relevant to the discussion.  The others I moved to #1037

eighthave:
rebased on top of today's commits, no conflicts.

eighthave:
I just found a technique that will allow Osmand to have the "View" options on the same Share/`ACTION_SEND` chooser. I have it implemented in a different app and it seems to work fine.  Though I still think it would be better to keep `ACTION_VIEW` and `ACTION_SEND` separate, I would be fine with having the couple `ACTION_VIEW` apps listed in the options when the user clicks this "Share your location" menu item.

Here's how to do it, more or less:
https://stackoverflow.com/questions/11021021/how-to-make-an-intent-with-multiple-actions

eighthave:
@vshcherb this pull request now makes the "Share location" menu also show apps that accept `ACTION_VIEW` of  a `geo:` URI, so it now replaces all of the features of the current "Share location" menu, while working better for most actions.  As far as I've seen, not of the combos I've tried work worse.

eighthave:
I took some screenshots to show what this looks like on various devices, with a wide range of apps installed:
osmand-share-location-few
osmand-share-location-lots
osmand-share-location-some

vshcherb:
Thank you for pull-request, I think now we have all information to make a decision. This is not a bug fix, but it is a functionality change, so it should be taken carefully. Indeed we plan to handle it before 2.0, but most likely closer to the end. I'm not sure yet if that request will be accepted or not, because I don't have designs yet, but I forwarded this request to our designer.

vshcherb:
This change removes quite good options of Qr, Clipboard, geo. I think the only missing option now is "Message" which will open all other options, I refactored original code to include it and exclude duplicate SMS/Email

eighthave:
I think you are misunderstanding this. QR, Clipboard, and geo are all fully supported.  Check the screenshots.  

* Generating a QR Code is supported by choosing a QR Code app, in the screenshot, that's Barcode Scanner, but other apps will work too
* Clipboard is fully supported by selecting "Copy to clipboard"
* `geo:` is fully supported, all apps that can view a `geo:` show up in the screenshot with the words "View in..." before the name of the app

The current QRCode button is hard-coded to Barcode Scanner, this violates one of the core Android design patterns "Don't hardcode links to other apps": https://developer.android.com/design/patterns/pure-android.html  The way it is implemented in this pull request, it will work with any QR app including Barcode Scanner.

This pull request could be modified to detect whether any QR code app is installed, and if not, then use the existing prompt to install Barcode Scanner.

Here are a bunch of examples of what the Share menu should look like:
http://unitid.nl/androidpatterns/uap_pattern/share-button

Here's another approach:
* https://developer.android.com/training/sharing/shareaction.html
* http://android-developers.blogspot.fr/2012/02/share-with-intents.html

The core idea is that a Share button should open a listing of Apps, not generic actions.

vshcherb:
I agree, that current functionality could be done better and potentially there is a misunderstanding. I just have following demands:
- standard messaging intent (covers email, messaging ...)
- clipboard if you type a long message for email
- geo intent ,  there are apps which support geo intent so it will be good to have it coverage
- possibility to display qr code (sending is not needed cause qr code is the link) - may be only if one of the program installed

I'm open for propositions & new pull request which will be mergeable and covers this 4 items.

eighthave:
This pull request already covers all of those demands, so I need to figure out what detail is missing before putting more work into this.  Here is how you complete your demands, referencing the screenshots posted inline above:

* _standard messaging intent_: click Share, then choose any messaging app (you can see Email, Gmail, and TextSecure in the screenshots)
* _clipboard_: click Share, then click "Copy to clipboard" (this is visible in one of the screenshots)
* _geo intent_: click Share, then click any app that says "View in..." (you can see many map apps listed in one screenshot)
* _display QR Code_: click Share, then click any app that can generate QR codes (Barcode Scanner is listed in the screenshots)

What is missing from this pull request is a way to prompt the user to install a QR Code generating app.  That can be added to this style as well.  I am up for doing that once we figure out why you think this pull request is not covering all the needed use cases.

Some ideas:

* The option for Barcode Scanner could be changed to actually say "Generate QR Code" instead of "Barcode Scanner".  This currently just uses what Barcode Scanner provides
* when the user clicks Share, it can do the check for Barcode Scanner, and prompt the user to Install, Ignore, or Never Install.
* if Barcode Scanner is missing, there could be an icon in the Share screen that says "install QR Code generator"
* Osmand could use the Barcode Scanner integration jar to generate QR Codes in a more integrated way

vshcherb:
1. This request still can't be merged automatically. So it is deprecated may be.
2. I'm not sure I saw Copy to Clipboard in standard intent sharing dialog on my phone. 

The better way will be to not *delete* existing code, introduce a static variable *switch* and commit. In that case it will be easy to test and revert if something doesn't work and the code could coexist for a while for testing period.

eighthave:
1. yeah, I'll rebase everything once I get all the issues sorted out.

2. it seems that most, but not all, ROMs provide the "Copy to clipboard".
Perhaps the best way to handle that is to make Osmand provide a generic "Copy
location as geo:" IntentFilter, then there will always be that option when
clicking/sharing a location link.

3. ok, I can do the static variable.

frank334:
check Issue #1312 
https://github.com/osmandapp/Osmand/issues/1312

The geo: url is not decoded in my SMS reader and K9 mail, but that's not an issue of Osmand

But the URL-filter for 
http://osmand.net/go?lat=50&lon=7&z=16
type links does not decode to an intent to call the OsmAnd App. I can only choose to open in Browser to view an online map. That's not what I want.

At the same time, links like
http://maps.google.com/maps?q=50,7
correctly decode and I can chosse to call the OsmAnd App

Unfortunately, location sharing by osmand only creates the first url-type to osmand.net, so location sharing does not work between our phones.

I suppose it's a problem with the "GeoPointParserUtil.java"

Last tested with OsmAnd 2.0.4 from the market on y 4.4.2 Android.

frank334:
Btw, it would also be interesting to share locations with Desktop PC. Unfortunately Desktop Firefox does not understand geo:-URI. There was a handler, but this one needed a service at geourl.org which does not work currently:
https://addons.mozilla.org/de/firefox/addon/geo-uri-scheme-handler/

I checked the sourcecode of this Extension XPI, seems quite easy to extend it to create google maps links. But I'm not familiar with Javascript, and quite busy with a project at the moment. So, I tried a dirty hack to call GoogleEarth from a firefox geo: URI. GoogleEarth is more comfortable than maps, because of the KML markup, setting Bookmarks, polygon paths ...

In a bash script I'm extracting the geo-URI coordinates via regex, creating a KML with a point and then calling GoogleEarth

long=$(echo $1 | sed "s/.*geo:\([-0-9\.]*\),\([-0-9\.]*\).*/\2/g")
 lat=$(echo $1 | sed "s/.*geo:\([-0-9\.]*\),\([-0-9\.]*\).*/\1/g")

And the URI/Protocol is registered with a .reg file:

Windows Registry Editor Version 5.00

[HKEY_CLASSES_ROOT\geo]
@="URL:GEO Protocol"
"URL Protocol"=""

[HKEY_CLASSES_ROOT\geo\DefaultIcon]
@="C:\\Program Files (x86)\\Mozilla Firefox\\firefox.exe,1"

[HKEY_CLASSES_ROOT\geo\shell]
@="open"

[HKEY_CLASSES_ROOT\geo\shell\open]

[HKEY_CLASSES_ROOT\geo\shell\open\command]
@="\"C:\\cygwin\\bin\\bash.exe\" --rcfile /etc/profile -i -c 'BASH_INVOCATION=bat CHERE_INVOKING=true call_geo_uri_googleearth \"%1\"'"

postadelmaga:
please implement this ... the actual location sharing is not useful at all.

PR #1035, opened 2015-01-09
Added favourites updated event

Bars107:


vshcherb:
I think you forget to fire listener in UI thread. app.runInUIThread...

PR #1037, opened 2015-01-13
Update URL parsing to include OSM shortlinks, Baidu, Apple, AutoNavi

eighthave:
This is a big update to the code that handles parsing incoming URLs for location information, and then displaying that location.  This reorganizes the code somewhat to make it easier to add more patterns and services. Then the new services are included, as well as support for decoding OpenStreetMap shortlinks.  There are also some bug fixes here and there, mostly related to bitrot.  Things like updated shortlink parsing (for example, OSM's format now uses `~` instead of `@`).

This includes a couple of commits that were originally included in #1033, but they were moved here since they are more relevant to this pull request.

This is probably of most interest to @alexey-pelykh, @Bars107, and @Zahnstocher since they have most recently worked on this code.

### DIFF
--- a/OsmAnd-java/src/net/osmand/util/MapUtils.java
+++ b/OsmAnd-java/src/net/osmand/util/MapUtils.java
@@ -8,6 +8,7 @@ import java.util.List;
 import net.osmand.data.LatLon;
 import net.osmand.data.MapObject;
 import net.osmand.data.QuadPoint;
+import net.osmand.util.GeoPointParserUtil.GeoParsedPoint;
 
 
 /**
@@ -286,10 +287,10 @@ public class MapUtils {
 //	System.out.println(buildShortOsmUrl(52.30103d, 4.862927d, 18)); // http://osm.org/go/0E4_JiVhs
 //	System.out.println(buildShortOsmUrl(40.59d, -115.213d, 9)); // http://osm.org/go/TelHTB--
 	public static String buildShortOsmUrl(double latitude, double longitude, int zoom){
-        return BASE_SHORT_OSM_URL + createShortLocString(latitude, longitude, zoom) + "?m";
+        return BASE_SHORT_OSM_URL + createShortLinkString(latitude, longitude, zoom) + "?m";
 	}
 
-	public static String createShortLocString(double latitude, double longitude, int zoom) {
+	public static String createShortLinkString(double latitude, double longitude, int zoom) {
 		long lat = (long) (((latitude + 90d)/180d)*(1L << 32));
 		long lon = (long) (((longitude + 180d)/360d)*(1L << 32));
 		long code = interleaveBits(lon, lat);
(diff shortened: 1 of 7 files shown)


OsmAnd-java/src/net/osmand/util/MapUtils.java
```diff
@@ -32,7 +35,7 @@
         'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
         'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm',
         'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
-        '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '_', '@'
+        '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '_', '~'
```
vshcherb: It is not clear why do we change the symbol in lookup table, it could break backward compatibility

vshcherb:
I will close this pull request and I hope it will be reopened after some improvements:
1. We can't change encoding scheme even if it is incorrect , for Audio/Video notes, users might created them already! And they will stop working. So Another solution should be found.
2. This looks like an improvement for our handling scheme, but we made strong decision to refactor it out from regex. It is very unhandy to support regex, better to use java api, substring, indexof ...

Small concerns:
1. This is already quite big pull-request, please focus only on that funcitonality 1 class preferrably, so it will be easier to accept/check it quick.

alexey-pelykh:
Generally, it's bad idea to parse QuetyString using RegExps, since simple position shuffling of arguments will not be parsed. If you're parsing QueryString, first break it into components, and then parse them one by one (regexp here is ok). To get idea what I mean check this change 9678ca155c5c2aa2af8b140daa2dd847d2f17f78

OsmAnd-java/src/net/osmand/util/MapUtils.java
eighthave: Only because that is what OpenStreetMap has done.  If you follow this link included in the commit message:
https://wiki.openstreetmap.org/wiki/Shortlink#Code_for_encoding_.26_decoding

There are a few links to example code that demonstrate this.  Perhaps this one is particularly interesting:
https://github.com/openstreetmap/openstreetmap-website/blob/e84b2bd22f7c92fb7a128a91c999f86e350bf04d/app/assets/javascripts/application.js#L188

But this made me notice that one key change didn't make it into this commit, so I'm adding this snippet and rebasing:

```diff
 	public static GeoParsedPoint decodeShortLinkString(String s) {
+		// convert old shortlink format to current one
+		s = s.replaceAll("@", "~");
```

eighthave:
Oops, I missed 9678ca1.  I think I started working before that commit was pushed.  So I'll incorporate that work into this pull request.  I agree that the old regexs were problematic, I just mostly left them as is.

eighthave:
I checked out 9678ca1, I had seen that before after all. 9678ca1 only affects the `geo:` parsing, and this pull request does not modify the `geo:` except to rename the variable `data` to `uri` to match everywhere else.

I fully agree that the regexp stuff is not a good approach, but this pull request does not change the basic mechanism for parsing, it is just building upon the existing regexp structure and reorganizing it.  This pull request does get the code closer to the goal of eliminating the regexps, but I don't think it makes sense to add even more commits to this pull request to change the regexp stuff.

This pull request does further the goal of using more Java APIs by using `java.net.URL`'s methods more (i.e. `uri.getPath()`, `uri.getHost()`, etc.).

I made the shortlink parsing backwards compatible, following the openstreetmap way, and added some test cases, so I'm reopening this pull request.  Here are the changes I just made (not including rebasing on the latest commits in master):

```diff
diff --git a/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java b/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
index 9268411..4b02062 100644
--- a/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
+++ b/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
@@ -172,6 +172,22 @@ public class GeoPointParserUtil {
 		actual = GeoPointParserUtil.parse(url);
 		assertApproximateGeoPoint(actual, new GeoParsedPoint(dlat, dlon, z));
 
+		// http://osm.org/go/TyFYuF6P~~-?m // current shortlink format with "~"
+		// https://www.openstreetmap.org/?mlat=34.993933029174805&mlon=-106.61568069458008#map=15/34.99393/-106.61568
+		z = 20;
+		url = "http://osm.org/go/TyFYuF6P~~-?m";
+		System.out.println("url: " + url);
+		actual = GeoPointParserUtil.parse(url);
+		assertApproximateGeoPoint(actual, new GeoParsedPoint(longLat, longLon, z));
+
+		// http://osm.org/go/TyFYuF6P@@--?m= // old, deprecated shortlink format with "@"
+		// https://www.openstreetmap.org/?mlat=34.993933029174805&mlon=-106.61568069458008#map=15/34.99393/-106.61568
+		z = 21;
+		url = "http://osm.org/go/TyFYuF6P@@--?m=";
+		System.out.println("url: " + url);
+		actual = GeoPointParserUtil.parse(url);
+		assertApproximateGeoPoint(actual, new GeoParsedPoint(longLat, longLon, z));
+
 		// http://openstreetmap.de/zoom=11&lat=34&lon=-106
 		z = 11;
 		url = "http://openstreetmap.de/zoom=" + z + "&lat=" + ilat + "&lon=" + ilon;
@@ -377,6 +393,10 @@ public class GeoPointParserUtil {
 			"http://osm.org/go/0LQ127-?m",
 			"http://osm.org/go/0EEQjE==",
 			"http://osm.org/go/0EEQjEEb",
+			"http://osm.org/go/0EE~jEEb",
+			"http://osm.org/go/0EE@jEEb",
+			"http://osm.org/go/~~~~",
+			"http://osm.org/go/@@@@",
             "https://www.openstreetmap.org/#map=0/0/0",
             "https://www.openstreetmap.org/#map=0/180/180",
             "https://www.openstreetmap.org/#map=0/-180/-180",
diff --git a/OsmAnd-java/src/net/osmand/util/MapUtils.java b/OsmAnd-java/src/net/osmand/util/MapUtils.java
index 6fa7dd2..f9855ee 100644
--- a/OsmAnd-java/src/net/osmand/util/MapUtils.java
+++ b/OsmAnd-java/src/net/osmand/util/MapUtils.java
@@ -308,6 +308,8 @@ public class MapUtils {
 	}
 	
 	public static GeoParsedPoint decodeShortLinkString(String s) {
+		// convert old shortlink format to current one
+		s = s.replaceAll("@", "~");
 		int i = 0;
 		long x = 0;
 		long y = 0;
```

eighthave:
Turns out that I do not have permission to reopen this pull request, so hopefully someone else will do it.

vshcherb:
I see comments anyway, once you've changed it, just put a comment and we will reopen.

vshcherb:
actually there is another reason (branched was forced-pushed or recreated according to github )

eighthave:
Strange that this can't be reopened.  Rebasing then force pushing is supported by github (and others, like gitlab.com):
https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

But you can also merge these commits in your local git repo:
```
git remote add eighthave https://github.com/eighthave/Osmand
git fetch eighthave
git merge eighthave/update-url-parsing
```

eighthave:
I stuck a direct reference to this pull request in the commit message to keep the record intact (and did a force push to get it here). Let me know if I shouldn't rebase for Osmand pull requests.  I generally do it because it keeps the git commit history cleaner.

vshcherb:
You should keeping rebase but you need to open new pull-request I think

PR #1039, opened 2015-01-18
Update url parsing

eighthave:
This is a big update to the code that handles parsing incoming URLs for location information, and then displaying that location. This reorganizes the code somewhat to make it easier to add more patterns and services. Then the new services are included, as well as support for decoding OpenStreetMap shortlinks. There are also some bug fixes here and there, mostly related to bitrot. Things like updated shortlink parsing (for example, OSM's format now uses ~ instead of @).

The full discussion is in #1037 which was closed and cannot be reopened, so I'm opening this on @vshcherb 's request.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/activities/actions/ShareLocation.java
+++ b/OsmAnd/src/net/osmand/plus/activities/actions/ShareLocation.java
@@ -43,11 +43,12 @@ public class ShareLocation extends OsmAndAction {
 				final double longitude = args.getDouble(MapActivityActions.KEY_LONGITUDE);
 				final int zoom = args.getInt(MapActivityActions.KEY_ZOOM);
 				try {
-					final String shortOsmUrl = MapUtils.buildShortOsmUrl(latitude, longitude, zoom);
 					final String geoUrl = MapUtils.buildGeoUrl(latitude, longitude, zoom);
-					String sms = mapActivity.getString(R.string.send_location_sms_pattern, geoUrl, shortOsmUrl);
+					// TODO change this to HTTPS once it is setup!
+					final String httpUrl = "http://osmand.net/go?lat=" + ((float) latitude) + "&lon=" + ((float) longitude) + "&z=" + zoom;
+					String sms = mapActivity.getString(R.string.send_location_sms_pattern, geoUrl, httpUrl);
 					if (which == 0) {
-						sendEmail(shortOsmUrl, geoUrl);
+						sendEmail(httpUrl, geoUrl);
 					} else if (which == 1) {
 						sendSms(sms);
 					} else if (which == 2) {
@@ -69,8 +70,8 @@ public class ShareLocation extends OsmAndAction {
 	
 
 
-	private void sendEmail(final String shortOsmUrl, final String geoUrl) {
-		String email = mapActivity.getString(R.string.send_location_email_pattern, shortOsmUrl, geoUrl);
+	private void sendEmail(final String httpUrl, final String geoUrl) {
+		String email = mapActivity.getString(R.string.send_location_email_pattern, httpUrl, geoUrl);
 		ShareDialog.sendEmail(mapActivity, email, getString(R.string.send_location));
 	}
 
--- a/OsmAnd/src/net/osmand/plus/activities/actions/ShareLocation.java
+++ b/OsmAnd/src/net/osmand/plus/activities/actions/ShareLocation.java
@@ -44,16 +44,16 @@ public class ShareLocation extends OsmAndAction {
 				final int zoom = args.getInt(MapActivityActions.KEY_ZOOM);
 				try {
 					final String shortOsmUrl = MapUtils.buildShortOsmUrl(latitude, longitude, zoom);
-					final String appLink = "http://download.osmand.net/go?lat=" + ((float) latitude) + "&lon=" + ((float) longitude) + "&z=" + zoom;
-					String sms = mapActivity.getString(R.string.send_location_sms_pattern, shortOsmUrl, appLink);
+					final String geoUrl = MapUtils.buildGeoUrl(latitude, longitude, zoom);
+					String sms = mapActivity.getString(R.string.send_location_sms_pattern, geoUrl, shortOsmUrl);
 					if (which == 0) {
-						sendEmail(shortOsmUrl, appLink);
+						sendEmail(shortOsmUrl, geoUrl);
 					} else if (which == 1) {
 						sendSms(sms);
 					} else if (which == 2) {
 						sendToClipboard(sms);
 					} else if (which == 3) {
-						sendGeoActivity(latitude, longitude, zoom);
+						sendGeoActivity(geoUrl);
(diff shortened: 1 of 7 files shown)


OsmAnd/src/net/osmand/plus/activities/actions/ShareLocation.java
```diff
@@ -44,16 +44,16 @@ public void onClick(DialogInterface dialog, int which) {
 				final int zoom = args.getInt(MapActivityActions.KEY_ZOOM);
 				try {
 					final String shortOsmUrl = MapUtils.buildShortOsmUrl(latitude, longitude, zoom);
-					final String appLink = "http://download.osmand.net/go?lat=" + ((float) latitude) + "&lon=" + ((float) longitude) + "&z=" + zoom;
-					String sms = mapActivity.getString(R.string.send_location_sms_pattern, shortOsmUrl, appLink);
+					final String geoUrl = MapUtils.buildGeoUrl(latitude, longitude, zoom);
+					String sms = mapActivity.getString(R.string.send_location_sms_pattern, geoUrl, shortOsmUrl);
 					if (which == 0) {
-						sendEmail(shortOsmUrl, appLink);
```
vshcherb: This is actually should stay app link because we can't rely that receiver will have OsmAnd install instead we rely on proper internet address

vshcherb:
Thank you for pull-request! Besides the comment above about appLink  and the fact that we still rely too much on regex for parsing urls, the pull-request looks good. Of course we need to double check that tests are still passing and their number increased, which is quite obvious.

eighthave:
The test suite in `GeoPointParserUtil.main()` is passing for me.  Is there any other test suite to run?

OsmAnd/src/net/osmand/plus/activities/actions/ShareLocation.java
eighthave: I tried to figure out what that `appLink` (aka download.osmand.net/go?lat=12.6313&lon=-7.9955&z=8) did, and the only thing I could get it to do was redirect to Google Maps.  Right now, this pull request sends this as the message:

```
Location: geo:48.17396,16.365845?z=12
https://openstreetmap.org/go/0JrID2H--?m
```

So it is the same as the old message, I just replaced the download.osmand.net link with a direct, secure link to OpenStreetMap.

vshcherb:
I like the message and I would like to leave shorten version of osmand.net:

Location: geo:48.17396,16.365845?z=12
http://osmand.net/go?lat=..&lon=..&z=.. 

Soon we will update this web page and I hope quite soon we move to https, the links to google maps and openstreetmap will be on website.

eighthave:
I think having HTTPS on the shared link is important.  Location can divulge very private details of people's lives, and Osmand already has a lot of privacy advantages over all other mapping apps.  I think it can be a big selling point for Osmand: the most private mapping app.

What if there was a preference/setting that allowed the user to choose which map service to share?  The default could even be set automatically, based on what language the phone is in and which apps are installed (e.g. Google Maps).  Or is there also something else that osmand.net/go is meant to do?  The osmand.net/go page and openstreetmap.org could be one of the options, or this preference could either entirely replace the osmand.net/go redirect page.

One thing about the osmand.net/go link.  It could also use the OpenStreetMap shortlink algorithm for parsable shortlinks, i.e.
```
http://osmand.net/go/0JrID2H--?m
```
Or for even shorter, there is http://osm.org/go/0JrID2H--?m It has HTTPS on it already, but they have to update their HTTPS certificate to validate properly when connecting using osm.org:
https://github.com/openstreetmap/operations/issues/2

vshcherb:
The it still should be osmand.net first and second openstreetmap https second. We don't support yet for osmand shortlink and https, but we plan to.

eighthave:
I rebased the commits on the current master and added f5a38287b9c348d1ab507c5d12ba66e1da22a292 to this pull request to change the shared URL to http://osmand.net/go.

I'll do the sharing URL preferences in a separate pull request.

vshcherb:
Investigated deeper, this request could not be accepted as it is. It still operates with regex which should be avoided and it combines all different domains into 1 scope, for example googlecom patters removed or combined with others, which makes code less clear and less maintainable in case google will change format. The unit tests are fine to test , but their coverage is not so high, that's why code clarity is very important.
-			final String[] googleComPattern = {
-					"(?:.*)[@/]([+-]?\\d+(?:\\.\\d+)?),([+-]?\\d+(?:\\.\\d+)?),(\\d{1,2}z)(?:.*)",
-					"(?:.*)ll=([+-]?\\d+(?:\\.\\d+)?),([+-]?\\d+(?:\\.\\d+)?)(?:.+)(z=\\d{1,2})(?:.*)",
-			

share location url, which should a separate pull-request originally consists of 2 urls 1 https and 1 http osmand, so this would be an idea solution for now.

eighthave:
One of the things I changed was changed broken Google links in the tests to make them work, as in you can click on them and Google Maps would show something.  But perhaps I missed something.  Were those non-functional URLs generated by some other apps?  I would be surprised if Google ever supported treating the query string as the path, like I fixed here:

```diff
-		// http://maps.google.com/maps/q=loc:34.99393,-106.61568&z=11
-		url = "http://maps.google.com/maps/q=loc:" + dlat + "," + dlon + "&z=" + z;
+		// http://maps.google.com/maps/?q=loc:34.99393,-106.61568&z=11
+		url = "http://maps.google.com/maps/?q=loc:" + dlat + "," + dlon + "&z=" + z;
 		System.out.println("url: " + url);
 		actual = GeoPointParserUtil.parse(url);
 		assertGeoPoint(actual, new GeoParsedPoint(dlat, dlon, z));
 
 		// whatsapp
-		// http://maps.google.com/maps/q=loc:34,-106 (You)
+		// http://maps.google.com/maps/?q=loc:34,-106 (You)
 		z = GeoParsedPoint.NO_ZOOM;
-		url = "http://maps.google.com/maps/q=loc:" + ilat + "," + ilon + " (You)";
+		url = "http://maps.google.com/maps/?q=loc:" + ilat + "," + ilon + " (You)";
 		System.out.println("url: " + url);
 		actual = GeoPointParserUtil.parse(url);
 		assertGeoPoint(actual, new GeoParsedPoint(ilat, ilon, z));
 
 		// whatsapp
-		// http://maps.google.com/maps/q=loc:34.99393,-106.61568 (You)
+		// http://maps.google.com/maps/?q=loc:34.99393,-106.61568 (You)
 		z = GeoParsedPoint.NO_ZOOM;
-		url = "http://maps.google.com/maps/q=loc:" + dlat + "," + dlon + " (You)";
+		url = "http://maps.google.com/maps/?q=loc:" + dlat + "," + dlon + " (You)";
```

I'll make a new pull request to remove the parts that change the Google stuff. It is rebased these commits on master and these three commits were removed from this pull request:
* also parse maps.apple.com links, the query string includes lat/long/zoom
* support parsing amap.com URLs
* convert all existing URL patterns to new code structure

PR #1041, opened 2015-01-19
Quick git fixes

eighthave:
two quick fixes to the new appcompat stuff

*  remove generated files from git
* git ignore generated files


vshcherb:
Already done

PR #1043, opened 2015-01-20
add support for a proxy and use more HTTPS

eighthave:
This pull request mostly is about adding support for an HTTP proxy for internet connections.  Some networks require proxies in order to access the internet from behind an firewall.  Proxies are also useful for circumvention and privacy tools.

This also changes some connections to use HTTPS (except on `android-9` where it is problematic).  Location-related internet traffic can contain a lot of private information, so it should be protected.

There is also a small fix of a leaked resource that wasn't being closed.

### DIFF
--- a/OsmAnd-java/src/net/osmand/osm/edit/Entity.java
+++ b/OsmAnd-java/src/net/osmand/osm/edit/Entity.java
@@ -67,7 +67,7 @@ public abstract class Entity {
 		}
 
 		public String getOsmUrl() {
-			final String browseUrl = "http://www.openstreetmap.org/browse/";
+			final String browseUrl = "https://www.openstreetmap.org/browse/";
 			if (type == EntityType.NODE)
 				return browseUrl + "node/" + id;
 			if (type == EntityType.WAY)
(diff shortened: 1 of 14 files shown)


vshcherb:
Hi,
Thanks for request. I have nothing against using https, it is good and aligned with our vision. Though I have to admit that I don't understand necessity of proxy. First of all it is not used everywhere.
To be honest I don't remember any specific request for Proxy. So, I would vote against functionality which is not used. Also options take quite a lot of space in General Settings, so I don't really sympathize t hat implementation. If we would like to have it at the end, it is better will be 1 Preference with dialog.

Best Regards,
Victor

eighthave:
My main interest in Osmand is that it has the potential to become the most private map app for Android because it works well offline, and it can download whole regions at a time, rather than specific tiles.  I think this can be achieved without even really being visible, only little things like the proxy preference. One of the key privacy protections is using a anonymity proxy, like I2P or Tor.  There are also proxy services like Psiphon and Ultrasurf that allow.

If I left anywhere out for adding the proxy support, let me know, and I'll add it. This pull request does not cover the OsMo Live Monitoring plugin, even though it should.  The version of Apache HttpClient in Android does not really support a proxy, so that would require changing that code to use java.net.HttpURLConnection like the rest of the OsmAnd code base.

It would also be good to remove the code that is disabling the HTTPS hostname checks so that there is a real, verified HTTPS connection.

vshcherb:
ok, your point is clear. For OsMo you need to contact Konstantin about https on server side and I think there is an option for this. Also there is encryption for OsMo.
Well if proxy goes via https then yes it makes sense.

PR #1052, opened 2015-01-29
GeoParsedPoint.toString() output a geo: URI string

eighthave:
This pull request makes `GeoParsedPoint.toString()` a valid `geo:` URI that represents the data in the `GeoParsedPoint`.  It also tries to include the "query" and "label" parts, i.e. `geo:0,0?q=query(label)` but that part could still be improved.  Previously, `GeoParsedPoint.toString()` returned some debug info that was not really used.

This follows the pattern established by:
* http://docs.oracle.com/javase/7/docs/api/java/net/URL.html#toString()
* https://developer.android.com/reference/android/net/Uri.html#toString()

I am using this a lot in work that will be in coming pull requests, including the "Share location" revamp.

### DIFF
--- a/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
+++ b/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
@@ -846,7 +846,7 @@ public class GeoPointParserUtil {
 
 		private String formatDouble(double d) {
 			if(d == (long) d)
-				return String.format("%d", (long)d);
+				return String.format(Locale.ENGLISH, "%d", (long)d);
 			else
 				return String.format("%s", d);
 		}
--- a/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
+++ b/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
@@ -398,7 +398,6 @@ public class GeoPointParserUtil {
 	private static boolean areCloseEnough(double a, double b, long howClose) {
 		long aRounded = (long) Math.round(a * Math.pow(10, howClose));
 		long bRounded = (long) Math.round(b * Math.pow(10, howClose));
-		System.out.println("areCloseEnough: " + aRounded + ", " + bRounded);
 		return aRounded == bRounded;
 	}
 
--- a/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
+++ b/OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
@@ -1,8 +1,10 @@
+
 package net.osmand.util;
 
 import java.net.URI;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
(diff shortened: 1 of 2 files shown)


OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
```diff
@@ -813,15 +844,57 @@ public boolean isGeoPoint() {
...
 		public boolean isGeoAddress() {
 			return geoAddress;
 		}
 
+		/**
+		 * Generates a URI string according to https://tools.ietf.org/html/rfc5870 and
+		 * https://developer.android.com/guide/components/intents-common.html#Maps
+		 */
 		@Override
 		public String toString() {
```
vshcherb: URI should not be generated in toString() method, there should be a separate method  for this. toString should be used only for debugging feature preferrably

eighthave: I just following the patterns established by Java SE and Android: classes that
represent URIs all use `toString()` to output the complete URI:

* http://docs.oracle.com/javase/7/docs/api/java/net/URL.html#toString()
* http://docs.oracle.com/javase/7/docs/api/java/net/URI.html#toString()
* https://developer.android.com/reference/android/net/Uri.html#toString()

And classes that "build" data classes have a `toString()` method that outputs
a debug string:

* https://developer.android.com/reference/android/net/Uri.Builder.html#toString()

`GeoPointParserUtil` is the "builder", and `GeoParsedPoint` represents data
only, so is like java.net.URL, java.net.URI, and android.net.Uri.

vshcherb:
Hi, 
First of all I needed to redesign the code in order to make it simple and get rid of complex regexes. In case I'm missing any use case please feel free to add it. So this pull-request can't be merged due to conflicts

Your input about URL class is arguable. GeoParsedPoint doesn't represent any URL, first of all is not clear what exactly URL it is missing quite some information, anyway it is not a good idea to convert an entity which was not designed for that, in that case it will be logical if the class will be GeoURL, but it is not. So in that case it should just have method getGeoURL and may print it in toString method but it is not really important.

vshcherb:
I will close this pull-request because it is not ready for merge yet.

eighthave:
(I can't reopen this pull request, so it'll be another split up comment thread unless you can reopen it)

Looks like we duplicated effort, I had already done a lot of the work you did in 567de5806a9c4d45ccd4614f37f0b163c32455d7, I was just breaking it up into easy to review chunks in the pull requests.  Oh well... you can see my work in progress by checking the branches in my fork:  https://github.com/eighthave/Osmand

I'll rebase this, rename the method to `getGeoUrlString()`, and resubmit it.  My point about using a `geo:` URI is that all of the data contained in a `GeoParsedPoint` can be represented in a single `geo:`, so why invent a new way to derive a `String` from it?

PR #1061, opened 2015-02-03
claim all Google map domains

eighthave:
This is all of the possible Google Maps domains.  Unfortunately for this case, Google redirects the .com to the country-specific domain of the detection location of the user.  So if a user copies and pastes a map URL,
then it'll include that user's country domain.  So in order to support catching Google Maps links well, the country domains need to be supported.

https://en.wikipedia.org/wiki/List_of_Google_domains

I don't know how else to do this without having this very long list.


eighthave:
If you want to try an app that has this same list in the manifest,  I have the complete one in my experimental [LocationPrivacy](https://github.com/guardianproject/LocationPrivacy) app.  It is free on "Google Play":https://play.google.com/store/apps/details?id=info.guardianproject.locationprivacy

eighthave:
@vshcherb any word on this?  I think its a low risk addition that will make Osmand work with all Google location URLs.

vshcherb:
I don't really like that mess, it would be interesting if google maps handle it in that way.

eighthave:
yeah, it is not pretty, but it is the only way that anyone seems to have figured out, including Google.  It is very important since all of the world except for the US will default to the country specific domains.  So right now in Osmand, Google Maps URLs only really work for people in the US, where .com is the default. Google Maps' `AndroidManifest.xml` is worse than this, so much so that I can't even paste the revelant section since it is longer than Github's 256k message limit. I emailed the file to you directly.

vshcherb:
May be a little bit more important, I've got a request for supporting which some external organization use a standard: 
Intent intent = new Intent (android.content.Intent.ACTION_VIEW, Uri.parse (String.format ("google.navigation: q =% s,% s", lt, lng)));

eighthave:
yeah, there is a lot of stuff in Google Maps manifest.  This pull request is really just the most obvious ones.  I think it makes sense to include things like this as there is a use case for them, rather than just trying to take everything from the Google Maps manifest.  For what its worth, Google supports `google.navigation:` with its own filter:

```xml
            <intent-filter android:label="@string/MAPS_APP_NAME">
                <action android:name="android.intent.action.VIEW"/>
                <category android:name="android.intent.category.DEFAULT"/>
                <category android:name="android.intent.category.BROWSABLE"/>
                <data android:scheme="google.navigation"/>
            </intent-filter>
```

vshcherb:
To be honest I didn't hear any request (except from your side about supporting google maps locale), but about google.navigation sounds reasonable. I will quote it a bit 

As a navigation app we would like OsmAnd use Android, as we predict and navigation capabilities of your app. In our immediate vicinity several fire departments to equip to this alarm system and want to fall back on the navigation option on OsmAnd. However, in the transfer of alarm addresses of the alarm app to OsmAnd problems. The Alarm App used for address transfer and routing command the standard "Intent google.navigation" because it is used by multiple navigation apps, the way it looks, but not your app. Is it possible that you implement the above standard in your app?
An example of a call from the alarm app:
Intent intent = new Intent (android.content.Intent.ACTION_VIEW, Uri.parse (String.format ("google.navigation: q =% s,% s", lt, lng)));

eighthave:
Supporting `google.navigation:` makes sense.  I haven't seen any documentation on the URL format.

I think most people do not realize that clicking a Google URL could take them to the location in Osmand.  People are used to thinking that a Google URL takes you to Google, a Facebook URL to Facebook, Yandex URL to Yandex, etc.  This is really a new feature, and if we make Osmand claim as many location URLs as possible, people will discover this feature in their normal usage patterns.  And when people find that they are going to Osmand over other map apps without thinking about it, then we'll see a lot more people using Osmand instead of others.

eighthave:
I wonder why this is marked "Nice to Have" yet it is implemented here in this pull request, and it is implemented just like how the Google Maps app implements it.

vshcherb:
1) Main issue related to this pull-request, it's quite aggressive use of non-standard intents. 
2) We don't support full functionality of google maps intents, so it could be very annoying to users to  get as an option but see no results

eighthave:
1) these `IntentFilters` 100% standard, plain old `VIEW Intents` for `http` and `https` schemes with Google domain names.  I can't think of anything even close to non-standard about the URLs, the `Intents` received, or the `IntentFilters`

2) I agree it is important to not create an annoyance.  These aren't `IntentFilters` for `Intents` specific to Google Maps, but just the Google Maps URLs.  When the user is prompted because of these `IntentFilters`, they can choose Google Maps, then click "Always", and they will never see the prompt again for that domain name.  So however annoying this might be, it'll be only once the vast majority of the time.

Spacefish:
+1

Norbert80:
Related to issue #2584?

eighthave:
yup, its definitely related.

eighthave:
Thanks for merging!  If anyone is interested in making Osmand parse all Google Maps links, I have working code that first tries to parse the location from the URL, then if that fails, it downloads the HTML and parses it from there:
https://github.com/guardianproject/LocationPrivacy/blob/0.3/src/info/guardianproject/locationprivacy/GoogleUriActivity.java

PR #1066, opened 2015-02-06
Fix work with SQLiteDB

ilasica:
1. Add ellipsoid projection (Yandex). Use new optional field ELLIPSOID in table INFO. 1 means ellipsoid. If there is no field ELLIPSOID or its value differs from 1, used default spheroid projection.
2. Add beanshell use. Use new optional field RULE in table INFO. Value "beanshell" makes to use interpreter for processing of a script which put into the field URL. If there is no field RULE or its value differs from "beanshell", content of the field URL used as before.
3. Enable load map to db in case of SQLiteDBSource. 
4. Saving imidiatelly to db (no temp file)  in case of SQLiteDBSource.
5. Fixed urlTemplate wildcards replacing for SQLiteDB field URL and metainfo field urlTemplate in case of not using beanshell. {$z}->{0}, {$x}->{1}, {$y}->{2}

### DIFF
--- a/OsmAnd-java/src/net/osmand/map/MapTileDownloader.java
+++ b/OsmAnd-java/src/net/osmand/map/MapTileDownloader.java
@@ -4,6 +4,8 @@ import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.UnknownHostException;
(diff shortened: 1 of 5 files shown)


vshcherb:
Too many changes in 1 pull -request and too much commented code. I would advise to split pull-requests by small commits which will be easy to code review, for now it will require too much testing and reviewing, the value is also unclear because ellipsoid was supported as far as I remember. So let's break it into small pieces and review it again.

Thanks for your efforts

vshcherb:
You can git cherry-pick commits and create new pull request

PR #1070, opened 2015-02-10
Continue work on BeanShell in SQLiteDB

ilasica:
Add beanshell use. Use new optional field RULE in table INFO. Value "beanshell" makes to use interpreter for processing of a script which put into the field URL. If there is no field RULE or its value differs from "beanshell", content of the field URL used as before.


ilasica:
> Not really clear yet, because column is not added. So the code is not really used at all.

Really. Sorry. I divided pull request #1066 into too small parts.

PR #1071, opened 2015-02-11
Saving tiles into sqlitedb

ilasica:
In case of SQLiteTileSource using:
Changed loading tiles into sqlitedb when using "Download map" from context menu. Before now in such cases tiles were saved in files in TEMP folder
Changed saving tiles directly into sqlitedb when browsing new area. Before now in such cases 
intermediate temp files were used.


vshcherb:
Is this part tested? I looked at the code and it looks ok, but testing would be good

ilasica:
Yes. I`ve tested it. And it tested by community from 4pda.

PR #1083, opened 2015-02-21
Update phrases.xml

aceman444:
Bugs:
shop=chemist is not "Drug store" (that would be a pharmacy).
shop=kitchen is for a "kitchen studio", i.e. furnishing,architecting. Kitchenware (cooking stuff, etc) is explicitly under shop=houseware.

### DIFF
--- a/OsmAnd/res/values/phrases.xml
+++ b/OsmAnd/res/values/phrases.xml
@@ -52,7 +52,7 @@
 	<string name="poi_bed_shop">Bed room furnishings</string>
 	<string name="poi_boutique">Fashion boutique</string>
 	<string name="poi_carpet_shop">Carpet shop</string>
-	<string name="poi_chemist_shop">Drug store</string>
+	<string name="poi_chemist_shop">Chemist shop</string>
 	<string name="poi_clothes_shop">Apparel store</string>
 	<string name="poi_child_clothes_shop">Childrens apparel</string>
 	<string name="poi_shoes">Shoe store</string>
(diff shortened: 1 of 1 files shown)


sonora:
Aceman,
Regarding "Drug store":

While we do not have a separate version form British English, I had been
trying to keep things consistent in US English. And there "Drug store" is
correct, as is also referenced here:
http://wiki.openstreetmap.org/wiki/Tag:shop%3Dchemist

Ok?

I have made the other 2 changes you prepared manually, thanks!

Hardy

aceman444:
No I do not agree.

I read the sentence as "For prescription drugs, ... . (in) U.S. drug store, see amenity=pharmacy".

So "drug store" is not chemist, it is pharmacy. Yes there is a dot before U.S. which probably should have been a comma on the wiki. See also http://wiki.openstreetmap.org/wiki/Shop where this whole sentence is in brackets so again "drug store" means pharmacy.

sonora:
Well, but using "Chemist's shop" in US English would be rather unusual, for
a discussion see eg http://en.wikipedia.org/wiki/Pharmacy. Quote from the
3rd paragraph is entirely true in my experience: "In the United States and
Canada, *drugstores* commonly sell drugs, as well as miscellaneous items
such as confectionary, cosmetics
<http://en.wikipedia.org/wiki/Cosmetics>, office
supplies <http://en.wikipedia.org/wiki/Office_supplies>, and magazines
<http://en.wikipedia.org/wiki/Magazine> and occasionally refreshments and
groceries.". Most big chains call themselves "drug store", and if they also
sell prescription drugs, they would explicitly say "with pharmacy".

So I am really not very happy with "chemist's", but also have no better
suggestion than drugstore...

aceman444:
Yeah:) I just do not want that translators see "drug store" and translate it as pharmacy as the wiki suggests at some places. I could see this catch as I was looking at real tags you have behind the texts, but not everybody will do that.

sonora:
That's true ....  Is there maybe a way to add a comment in the translators'
tool (I am not too familiar with that) ?

aceman444:
Good idea, I have put the comment into the Weblate translation system to the poi_chemist_shop string. Not sure if anybody notices it, but that is what can be done in the system.

PR #1084, opened 2015-02-22
Change the hostname back to osm.org

Zahnstocher:
Change the hostname back to osm.org, because it now supports HTTPS.


Zahnstocher:
OSM sysadmins switched back to old certificates (which don't support osm.org) because of problems with JOSM.

PR #1093, opened 2015-03-03
Sqlite speed up

ilasica:
DO NOT MERGE! Pull request created for discussion.
This attempt to remove unnecessary delays.


vshcherb:
It is very hard to say! We need to recover the logic why it was written like this, I doubt this complexity was created for no reason.

ilasica:
I agree with you. Usually coder is too lazy to create unnecessay code. Therefore I rised this request pull. I found deleys help to manipulate with progress bar (I should say a little bit incorrectly) and to limit number of threads in queue. I tried app with removed sleeps on my phone and I have problem only with progress bar, it is easy to fix. We can ask 4pda-community to test it.

I see we could improve performance of tile downloading. We could try to remove dalays and use transaction for inserts into DB. Also we should consider that higher preformance will increase workload on servers of map provider.

vshcherb:
it will be beneficial to find when this code was introduced, git annotate?

ilasica:
I could not find when it was introduced. At least, first time DownloadTilesDialog.java was committed 13 Apr 2011 (6ac259b5e443ba9b9854643b4aba854d774e7ee5) and already had these sleeps.

vshcherb:
git annotate could really show the source, because the history of this file
starts from renaming

ilasica:
It was created in https://github.com/osmandapp/Osmand/commit/b180b6bd706932dccae3f552da03ec582a9b7ec8.

No special comments.

vshcherb:
Not it is clear

while (instance.isSomethingBeingDownloaded()) {
+											try {
+												Thread.sleep(500);

This was introduced when we have more than limit requests. I don't think the target code uses the same logic. Original code is clear to me.

vshcherb:
sleep(750); could be reduced to 50, this was done to not overwhelm devices with background thread. Anyway requests accumulated. But we don't want to let the thread run all the time without sleep. Or we need to use wait/notify technique. Like ExecutorService which should be better.

vshcherb:
https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/activities/DownloadTilesDialog.java#L179

This is fine code, I don't see any issue. Of course we could reduce it to 200, but anyway I don't think it is important. This is done, to not show too much progress, while it mostly downloads tiles, i.e. progress bar 100% but the tiles are not downloaded yet.

PR #1101, opened 2015-03-06
Added parking plugin card, added refresh for dashboard cards.

Bars107:


vshcherb:
Parking should be redone, everything else is deprecated

PR #1103, opened 2015-03-10
Parking card and osmo edits.

Bars107:


OsmAnd/res/layout/dash_parking_fragment.xml
```diff
@@ -1,53 +1,82 @@
...
+			android:layout_height="wrap_content"
+			android:layout_gravity="center_vertical"
+			android:layout_weight="1"
+			android:orientation="vertical">
+
+			<TextView
+				android:id="@+id/distance"
+				android:textSize="@dimen/dash_parking_text_size"
+				android:layout_width="match_parent"
+				android:layout_height="wrap_content"
```
vshcherb: no text color white! ?attr/textPrimaryColor

OsmAnd/res/layout/dash_parking_fragment.xml
```diff
@@ -1,53 +1,82 @@
...
+				android:textColor="@color/color_white"
+				tools:text="@string/app_version"/>
+
+			<TextView
+				android:id="@+id/name"
+				android:layout_width="wrap_content"
+				android:layout_height="wrap_content"
+				android:textColor="@color/color_white"
```
vshcherb: Same issue, no text colros!

OsmAnd/src/net/osmand/plus/views/MapControlsLayer.java
```diff
@@ -125,6 +125,13 @@ public void initLayer(final OsmandMapTileView view) {
 		initTransparencyBar(view, parent);
 	}
 
+	public void startNavigation() {
```
vshcherb: doesn't look like a good solution... though it does exactly as we press on the button, that doesn't the code should be in the button. Better place to extract from the button and put it somewhere MapActivityActions. And we don't go from MapActivity to a specific layer

OsmAnd/src/net/osmand/plus/osmedit/LocalOpenstreetmapActivity.java
```diff
@@ -424,8 +424,6 @@ protected void onProgressUpdate(OsmPoint... points) {
 		public LocalOpenstreetmapAdapter() {
 			super(LocalOpenstreetmapActivity.this, net.osmand.plus.R.layout.local_openstreetmap_list_item);
 		}
-		
-
 
 		public void delete(OsmPoint i) {
```
vshcherb: Later we need to delete that activity from settings

vshcherb:
Use shared strings as much as possible shared_string_show_on_map! For example
Actions To export all icons are missing

vshcherb:
Use shared_string_delete instead of Delete modification.
No bulk upload to OSM

PR #1115, opened 2015-03-30
Update phrases.xml

Barrocas:
Change POI names back to OSM naming.
Please maintain OSM naming (as much as possible), to keep an easy work between Osmand and OSM mapping/tagging.

### DIFF
--- a/OsmAnd/res/values/phrases.xml
+++ b/OsmAnd/res/values/phrases.xml
@@ -54,7 +54,7 @@
 	<string name="poi_greengrocer">Greengrocery</string>
 	<string name="poi_seafood">Seafood store</string>
 	<string name="poi_confectionery">Confectionery</string>
-	<string name="poi_ice_cream">Ice cream parlor</string>
+	<string name="poi_ice_cream">Ice cream shop</string>
 	<string name="poi_supermarket">Supermarket</string>
 	<string name="poi_shop_tea">Tea shop</string>
 	<string name="poi_dairy">Dairy shop</string>
@@ -73,8 +73,8 @@
 	<string name="poi_boutique">Fashion boutique</string>
 	<string name="poi_carpet">Carpet shop</string>
 	<string name="poi_chemist">Drug store</string>
-	<string name="poi_clothes">Apparel store</string>
-	<string name="poi_child_clothes">Childrens apparel</string>
+	<string name="poi_clothes">Clothes shop</string>
+	<string name="poi_child_clothes">Child clothes shop</string>
 	<string name="poi_shoes">Shoe store</string>
 	<string name="poi_computer">Computer store</string>
 	<string name="poi_copyshop">Copy shop</string>
@@ -82,9 +82,9 @@
 	<string name="poi_fabric">Drapery shop</string>
 	<string name="poi_bedding">Bedding store</string>
 	<string name="poi_dive">Diving equipment</string>
-	<string name="poi_doityourself">Home improvement store</string>
+	<string name="poi_doityourself">Do-It-Yourself store</string>
 	<string name="poi_erotic">Erotic shop</string>
-	<string name="poi_fishing">Fishing tackle</string>
+	<string name="poi_fishing">Fishing store</string>
 	<string name="poi_florist">Flower shop</string>
 	<string name="poi_frame">Picture frame shop</string>
 	<string name="poi_furniture">Furniture store</string>
@@ -100,15 +100,15 @@
 	<string name="poi_houseware">Household supply store</string>
 	<string name="poi_hunting">Hunting equipment</string>
 	<string name="poi_interior_decoration">Interior decoration store</string>
-	<string name="poi_jewelry">Jewelry store</string>
+	<string name="poi_jewelry">Jewelry</string>
 	<string name="poi_kiosk">Kiosk</string>
 	<string name="poi_kitchen">Kitchen furnishings</string>
 	<string name="poi_medical_supply">Medical supply</string>
-	<string name="poi_mobile_phone">Cell phone store</string>
+	<string name="poi_mobile_phone">Mobile phone store</string>
 	<string name="poi_motorcycle">Motorcycle shop</string>
 	<string name="poi_musical_instrument">Musical instruments</string>
-	<string name="poi_newsagent">Newsagency</string>
-	<string name="poi_optician">Optometrist</string>
+	<string name="poi_newsagent">News Agency</string>
+	<string name="poi_optician">Optician</string>
 	<string name="poi_organic">Organic food</string>
 	<string name="poi_outdoor">Outdoor store</string>
 	<string name="poi_paint">Paint shop</string>
@@ -116,7 +116,7 @@
 	<string name="poi_radiotechnics">Radiotechnics store</string>
 	<string name="poi_second_hand">Second hand store</string>
 	<string name="poi_ship_chandler">Ship chandler</string>
-	<string name="poi_sports">Sporting goods</string>
+	<string name="poi_sports">Sports shop</string>
 	<string name="poi_stationery">Stationery shop</string>
 	<string name="poi_tableware">Tableware shop</string>
 	<string name="poi_ticket">Ticket shop</string>
@@ -131,10 +131,10 @@
 	<string name="poi_electronics">Electronics store</string>
 	<string name="poi_car">Car shop</string>
 	<string name="poi_car_parts">Car parts</string>
-	<string name="poi_atv">Quad store</string>
+	<string name="poi_atv">ATV store</string>
 	<string name="poi_cosmetics">Cosmetics</string>
 	<string name="poi_marketplace">Marketplace</string>
-	<string name="poi_sewing">Dry goods store</string>
+	<string name="poi_sewing">Sewing store</string>
 
 	<string name="poi_police">Police</string>
 	<string name="poi_fire_station">Fire station</string>
@@ -169,7 +169,7 @@
 	<string name="poi_service_tyres">Tires</string>
 	<string name="poi_vehicle_inspection">Vehicle inspection</string>
 	<string name="poi_car_wash">Car wash</string>
-	<string name="poi_fuel">Gas station</string>
+	<string name="poi_fuel">Fuel station</string>
 	<string name="poi_charging_station">Charging station</string>
 	<string name="poi_vehicle_ramp">Vehicle ramp</string>
 	<string name="poi_compressed_air">Compressed air</string>


sonora:
So far, we do in OsmAnd not have a American English vs. British English (or any other variant) resource file. And since most of the en maintenance for years had been done by me, I tried to keep everything consistent in AE for now. I think this commit would break this consistency, hope you understand. If you disagree, please say so.

Thx - Hardy

Barrocas:
Hi Hardy,

I do not want to go against your work, by the contrary, I like to help
wenever possible.

But I believe that the recent POI naming changes did break OSM-OSmand
relationship, just because the previous POI naming was more consistent
with OSM.

And I'm not meaning at all to make a "war" between AE or BE (I'm
Portuguese, to me it really doesn't make a difference), but I would
prefer to have the same names in Osmand as in OSM.

I have the exact same problems with Portuguese vs
Brazilian-Portuguese, so I understand the BE vs AE situation.

In my opinion, Osmand (and OSM) should be international.
British English is the primary language for that use, an is what is
spread around the world.
American English is not the "default", and is not teached in
International schools.

For example, in OSM we have amenity=fuel, and in Osmand we now have
Gas Station (with undernote OSM amenity=fuel).

Fuel is more internationally understandable.
Gas is a term more adequate for US citizens, I can recognize the usage
by American movies, but for me Gas is another very different thing.
By using the term "Gas Station", Osmand is being "Americanized".

Apparel? I didn't even know that term... I expected Clothes shop, as in OSM.

The new names used are American related.
If you are American, and you did the changes, that's fine for you :)

But International users do not recognize them as you do, and other OSM
users (like me) will be looking for different names.

For the ones who do OSM mapping, it does make a difference, and it
would also be a better guide for Osmand users to further integrate the
OSM community.

With better linkage from OSmand to OSM (and vice-versa), both parts win.

Again, I'd like to make clear that I have nothing against you, or
Americans or American English, and I respect and appreciate your work
and dedication to Osmand, but I hope that you can see my (non-US
citizen) point of view.

Keep up the good work,
Cheers mate!



2015-03-30 14:25 GMT, Hardy <notifications@github.com>:
> So far, we do in OsmAnd not have a American English vs. British English (or
> any other variant) resource file. And since most of the en maintenance for
> years had been done by me, I tried to keep everything consistent in AE for
> now. I think this commit would break this consistency, hope you understand.
> If you disagree, please say so.
>
> Thx - Hardy
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/osmandapp/Osmand/pull/1115#issuecomment-87700648


-- 
Cumprimentos,
D.er Hugo Barrocas

sonora:
Hi Barrocas,

thank you for your reply, not a problem at all!  :-)

You may actually have a point.
I admit I have not researched in what places the content of our phrases.xml
takes shape. Wherever OSM tags are addressed, we need to stick with their
original text 100%, that would be e.g. in our OSM edit module.

Wherever the captions on the map or on search screens are affected, I have
mixed feelings: I think we should not create a mixture of different
language variants, it would look rather unprofessional, but rather stay
consistent in one language. And if we can afford open a second resource
file at some point for a second variant.

But I will include Victor in this conversation, so he can utter his
opinion. And maybe answer off the top of his head where our phrases.xml
comes to life, I think it is in the POI search screen, but not in the OSM
edit module?

Barrocas, thanks for your points and speaking up, anyway!

Best,
Hardy

vshcherb:
I think our phrases.xml came originally from special_phrases which came from OSM Nominatim, but because there were duplicate translations and so on, the wrong & non standard term was chosen, so we need now to find and update to most standard translation

sonora:
Yes, I am confused here, too, with our different xml files.

OSM tags are well defined, so that's a no-brainer. But it seems here we
need to cater for what the normal (non-OSM-savvy) user understands, and
particularly when different types of stores (or should I say shops?  :-) )
are concerned, this may differ significantly in the Americas, Europe, or
Asia. So it is a bit tricky... And I dare say a transatlantic mixture is no
good solution here ... any suggestions?

vshcherb:
So far no, we don't support alliases, which might be good, but it is not considered as short term solution

Barrocas:
Well, I only started adding stuff in OSM because of Osmand, that I
discovered around v1.6, until today.

Now I use Osmand nightly's, daily (lol), to check how the builds go,
to help test them in the real world, provide feedback (if it matters)
and add more POI info to OSM, as well as personal notes to later help
me editing the map.

I learned how to add POI in Osmand, and as I typed, suggestion came up. Nice! :)
Then in OSM wiki I started learning the naming and reading the explanations.
All went fine, since OSM tagging was exactly as in Osmand.

When going from Osmand to OSM, and back, things were predictable and
there was an obvious connection.

Then, somewere around mid February 2015, the POI names started being
modified, and I had some trouble trying to find POI names I was used
to add frequently:
https://groups.google.com/forum/#!topic/osmand/OVPWceV_-lQ

If the standard is the OSM nominatim, as in Osmand versions before and
until the 2.0 versions were modified, why change it?

If a normal user wants to add information or POI to the OSM map, he
will have to search for what he wants to add, either by clicking in
Osmand categories button (dead easy), or in the writen pop-up
suggestions (later on), or in OSM wiki (advanced).
That's what happened to me.

Now when in doubt, I still open up the OSM Wiki and start searching/reading.
That all goes down if Osmand takes a different naming route :(

If Osmand list the POI with different names, it's a mess :P

Instead of an obvious connection between OSM and Osmand, any user will
have to learn and decorate different tags/names for the same things :(

We all know Osmand is being worked up to be easier and simpler.
So why complicate what is already pre-defined and established?

Japanese saying: "If it works, don't fix it!" :)

option 1 - Ditch the phrases.xml completely (my vote).
Primary Rule = Keep it simple.
Stick to OSM naming absolutely.
Less trouble, more time for other work, bugs or UI.

option 2 - Modify the names to be as OSM-alike as possible. Will have
to go International/regional, and that also differs from translator to
translator.

option 3 - keep everything as is, kick me out and call me an idiot
(one who has ideas) :)

Shop or store can be (and was) added to the names, it makes sense, but
does it make any difference/improvement?
That would only work, if there were regional phrases for each
country/language, so instead of OSM "Fuel" or Osmand "Gas Station",
then I would be searching for "Bomba de Basolina", or "Posto de
Abastecimento".
But those names should be exactly as in any localized OSM or
JOSM-localized too...
Here in Portugal we have some discussions because someone edited JOSM
language to PT, and some users liked it, others differ and would have
preffered other naming, etc.
Way too messy. See my point?

It could help the normal (non-OSM-savvy) user, but it will not matter,
if it's not fully localized, and even then, there are tons of people
that can hardly read/write correctly in their own home language, let
alone in either British or American English.

-- 
Cumprimentos,
D.er Hugo Barrocas

Barrocas:
Hardy, the content of the phrases.xml is used everywere.

Example: amenity=fuel

In rendered POI description it only shows "Gas Station", as in POI
adding/modifying, and also in search results.

I cannot go add a POI, start typing "fuel" and expect to see that tag.
Now only "Gas Station" is valid as an option.

The only way to see "amenity=fuel", is in a small note under the POI
category button, when adding a POI and writing/selecting "Gas
Station".


-- 
Cumprimentos,
D.er Hugo Barrocas

sonora:
Barrocas, yes, that seems the actual problem, that the phrases.xml is now
used to provide the OSM keys for POI edits also, producing bogus things
like amenity=fuel.

It looks to me like whatever happened in February was not changes in
translation, but rather changes in what the phrases file is now wrongfully
used for ... and that needs fixing. After all, the phrases file is
translated, while the OSM tags should _not_ be, so that seems
another side of this same issue.

Maybe Victor can comment if I overlook something here. If not, we need to
start investigating what exactly changed. or which file to use to actually
supply the official (non-translatable) OSM tags for the edit module. Maybe
the issue is actually just that we started translating the tags for OSM
editor, which would be very questionable practice ...

Best, Hardy

sonora:
Thinking about this again, I guess the way to go could be to use a
localized translation, and then add the exact OSM key/tag behind it in
parenthesis. But this makes sense only in the OSM edit module. Only the
local translation should be used for search screen POI categories, etc.

Victor, should we go this route? I guess we would simply add the exact OSM
tags to no_translate, then?

Best, Hardy

Barrocas:
Hardy, that is exactly how Osmand is at the moment:

Transportation=Gas Station
(amenity=fuel)

The phrases are acting as a translated Osm tag file, in edit, map and search.

Selectable Categories are indeed modified in Osmand, they are
displayed as Transportation=Gas Station, and then internally
sub-converted to Osm amenity=fuel, in order to be sent to Osm.

In edit module, the real basic Osm tag is only displayed below the
category button.

For that to be correct, we would need translations for all countries
and languages, and there would have to be a choice of terms to be used
for each one. That's some big work to do.

But I think it could be the other way around, just maintain the stock
Osm categories and tags untouched, and if you find it necessary,
provide a translated sub-note to help in understanding the tags.

That way, Osmand would still use regular Osm tags, mappers would have
no problems using it, and the non-pro user would also have a
help/guide/translator for better osm tagging understanding, like this:

amenity=fuel
(Gas Station)

Even if you typed "Gas Station", it could also be a valid (pop-up)
choice, but the end result should be amenity=fuel (Gas Station).

That would be good for pro's and for starters.
But a lot of translations would become necessary...

-- 
Cumprimentos,
D.er Hugo Barrocas

sonora:
Hugo,

thanks for your suggestion, makes sense to me!

Victor, if you agree, that's what we could do, should not be too tricky to
implement, essentially we just swap what we now have in parentheses and
what outside?

Best,
Hardy

Barrocas:
I think you didn't quite understand...

Actually there are no ().

In the Fuel example, the selectable category to add/modify a POI in
Osmand is "Transportation", and the value is "Gas Station".

It should be (as it is in Osm, and also in Osmand in 2014) category
"amenity", and value "fuel".

When we select transportation=gas station, the edit mode displays a
(new) smaller text underneath the category button, with
"amenity=fuel".

Maybe I should add a picture for a clearer understanding?


-- 
Cumprimentos,
D.er Hugo Barrocas

sonora:
I admit  I rarely look at that corner of our code, (and I do my OSM edits,
if any, on a PC) :-)

But looking at it just now, we have all the data (correct OSM names are in
our string names and poi_types). So yes, could you, please show or explain
how things are today, and how you think they should be (preferably also for
a language other than English?

Thx - Hardy

Barrocas:
Hi,

I made a composition of what I was talking about, using a screenshot
of a nightly from day 27, because today's nightly do not allow to edit
POI, they crash instantly if I press "Modify POI".

Please see attached pic.

Red is what I think is wrong.
Green is how I think it should be (and most of it was like this, back in 2014)

Any doubt, please ask :)

Barrocas:
I replyed by e-mail and the pic was discarded.
Here it is:
osmand-edit_mode

Please comment and state your opinions.

sonora:
Hello Victor, please see below screenshots and suggestions, I think I can
fully agree with what Hugo suggests in green.

What do you think?

also: Looks like there is currently some ongoing work, if we have errors in
that module?

Thanks, Hardy


Hi,

I made a composition of what I was talking about, using a screenshot
of a nightly from day 27, because today's nightly do not allow to edit
POI, they crash instantly if I press "Modify POI".

Please see attached pic.

Red is what I think is wrong.
Green is how I think it should be (and most of it was like this, back in
2014)

Any doubt, please ask :)

Barrocas:
Any development in this area?

Are my sugestions acceptable?

Also, please notice that from day 27-03-2015 until today, night builds
keep on crashing when trying to "modify POI" :(

-- 
Cumprimentos,
D.er Hugo Barrocas

Barrocas:
Hi, is this subject going anywhere?

Night builds edit module is still in the same situation as when this
issue was first discussed.

Or should I open up this topic somewere else?

-- 
Cumprimentos,
D.er Hugo Barrocas

DFyson:
I see this issue as one of the key weak points of the search function. I've been noticing how I can't find certain POI but remember that the terminology is different. For example in Canada, _grocery store_ is pretty much synonymous with _supermarket_. It's frustrating for me but I know enough to try searching for synonyms, whereas someone less understanding or patient will likely give up in frustration.
So I think creating an alias/synonym lookup-table is important. And I think a goldmine for that is wikidata. For example check out all the 'also known as' in various languages for [gas station](https://www.wikidata.org/wiki/Q205495).

sonora:
Yes, I agree, some sort of automatic "aka"or synonyms search would be a much needed addition to our search interface. I have no clear picture yet how it could be achieved, but as soon as there are some tangible ideas I would say it is worth opening a new issue for this... (@vshcherb )

DFyson:
How about making a relational database of the OSM tags and all their synonyms using wikidata? Then get the search to include the database. The tedious part would be assigning wikidata IDs to each OSM tag (eg amenaty=fuel will be [Q205495](https://www.wikidata.org/wiki/Q205495)), then the rest can be automated to build the synonym database.

sonora:
I'd like to get Victor's opinion here. ( @vshcherb ) . The current search has some more open issues, so this needs to be looked at with broader perspective.

vshcherb:
Synonyms are needed of course and we used some synonyms in English for search but we have synonyms in every language. I think it is worth to open a separate issue though we need to think how to organize it. 

I think one of the alternative is too support it natively in phrases.xml but we need to implement it first.

sonora:
Yes, agree. By supporting synonyms in phrases.xml we could offload things to our translators, which is a very flexible way to handle this.

PR #1116, opened 2015-03-30
Fix poi_attraction string in phrases.xml

aceman444:
Somehow it got renamed to French in 
https://github.com/osmandapp/Osmand/commit/48e362328b19c750bc23e25623f75ffcdb78bd6b


sonora:
already fixed manually :-)

PR #1123, opened 2015-04-18
Fix crash

Zahnstocher:
Fix issue #2629:
https://code.google.com/p/osmand/issues/detail?id=2629


vshcherb:
it is really not clear what was wrong (may be == comparision with long) I fixed it differently

PR #1128, opened 2015-04-24
Don't copy unused specialphrases

Zahnstocher:
It seems that specialphrases were migrated to phrases.xml.


vshcherb:
True, we even can delete them from resources

PR #1150, opened 2015-05-01
Update phrases.xml

GiZig:
Better translations


OsmAnd/res/values-fr/phrases.xml
```diff
@@ -662,19 +662,19 @@
 	<string name="poi_internet_access_service">Accès Internet : service</string>
 	
 	<string name="poi_railway_buffer_stop">Heurtoir ferroviaire</string>
-	<string name="poi_shop_food">Magasin alimentaire</string>
+	<string name="poi_shop_food">Commerce alimentaire</string>
```
sbrunner: ??? not french

OsmAnd/res/values-fr/phrases.xml
```diff
@@ -633,13 +633,13 @@
 	<string name="poi_transportation">Transport</string>
 	<string name="poi_man_made">Infrastructure</string>
 	<string name="poi_education">Éducation</string>
-	<string name="poi_administrative">Territoire</string>
+	<string name="poi_administrative">Administration</string>
 	<string name="poi_healthcare">Santé</string>
-	<string name="poi_office">Service</string>
+	<string name="poi_office">Tertiaire</string>
```
sbrunner: Is it relay a good translation ?

GiZig: It is the better one I've found.
"Bureau" is a literal translation, but it is really very strange for 
French (I assume you speak French, and so you'll agree).
"Service" is not so bad, but it is used for… "Service" POI !
As every POI in this "office" category belongs to tertiary sector, I 
think it is the best way to tag it.

sbrunner: Why not `Bureau` ?

GiZig: Don't you think "Bureau" is very strange? I think nobody can understand 
which POI would be in such a category "Bureau": shops where you can buy 
pens? :-)

sbrunner:
Then I mis understand the usage of office do you have more information about that? For me a shops where you can buy pens is a `Papeterie`!

GiZig:
"office"(en) category contains POI such as "Agence de voyages", "Agence 
pour l'emploi", "ONG", "mairie", etc.
I said "bureau"(fr) is unclear. French people would think it contains 
POI such "Papeterie" (which sells pens for office).

sbrunner:
Wow it's a real global word, It's clear that we don't have any word for all those things in french... :-(

PR #1151, opened 2015-05-01
Update strings.xml

GiZig:
Better translation


sbrunner:
The new sentence has very strange grammar and ti breaks the XML ...

GiZig:
I don't find it is so strange (and I'm French).
I think this way is better for usability, since on a narrow screen you 
see "Plus proches points…" (which user can uderstead) instead of "Points 
d'intérêts les…" (which cannot be understood).
Maybe "Les plus proches points d'intérêts" is better. If you prefer it, 
I can change, and so fix XML.

sbrunner:
OK I understand your opinion about the small screens, than leave it as it :-)

PR #1188, opened 2015-05-08
Try to support maxheight in routing.

aceman444:
I'd like to support maxheight limits for routing.

I'll try to put more files into this pull request. Let's see how it goes in this web interface :)


aceman444:
Or should I just use the existing type="length" also for maxheight? The tag value syntax seems to be the same.

aceman444:
I think that would be better to not duplicate code.

vshcherb:
type = length is definitely enough, it would better called metric, but that is the same meaning

aceman444:
OK, thanks.

PR #1189, opened 2015-05-09
Update parseLength to support inches

aceman444:
Update parseLength to be more strict and precise and also support inches when feet are specified. Inches are actually mandatory per the wiki, but not all mappers give them. The new function supports them when they are specified (so handles them as optional).

Sorry, I can't build OsmAnd so this patch is untested. Please try it out and then merge. Thanks.

### DIFF
--- a/OsmAnd-java/src/net/osmand/binary/RouteDataObject.java
+++ b/OsmAnd-java/src/net/osmand/binary/RouteDataObject.java
@@ -187,15 +187,61 @@ public class RouteDataObject {
 	}
 	
 	public static float parseLength(String v, float def) {
-		// 14"10' not supported
+		// Supported formats:
+		// 10 m
+		// 10m
+		// 14'10"
+		// 14.5'
+		// 15ft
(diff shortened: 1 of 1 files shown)


vshcherb:
You can right a test similar to https://github.com/osmandapp/Osmand/blob/master/OsmAnd-java/src/net/osmand/util/OpeningHoursParser.java

aceman444:
Yes, I would love to write tests. But I wouldn't know where to put them. So I'll put the test expressions here and you can put them into the right function.

vshcherb:
Usually we just create main methods in the classes with very simple tests, like unit test. We don't have elaborate structure but the tests are more valuable than structure. So please create the simple test like I send you rather putting in comments what supported or not

aceman444:
I see you were faster to create the tests:) But also removed all my code :( I see you choose to be less strict and allow even invalid values (invalid per the wiki, but not logically). I need to see how your code works now (it is again without comments) and if it doesn't allow also logically invalid values.
And why do the tests not throw exception, while the OpeningHoursParser.java they do?

aceman444:
So I have made the file changes for this in pull request #1208.

Issue #1154, opened 2015-05-01
Opening Hours Parse error


midnightcomm:
POI Seems to have correct format for opening_hours, however, OSMAND is not parsing the tag as expected. Showing the place as closed when it is actually open.

Might have something to do with use of commas/semi-colon.

Mo-Th 09:00-21:00, Fr-Sa 09:00-17:00, Su off

[Screenshot](https://imgur.com/3kJwxO0)
[POI Example](https://www.openstreetmap.org/way/189795777)
[Validation Result](http://ypid.de/~osm/evaluation_tool/?EXP=Mo-Th%2009%3A00-21%3A00%2C%20Fr-Sa%2009%3A00-17%3A00%2C%20Su%20off&DATE=1430508420000&lat=41.1457707&lon=-95.9753071&mode=0)

[Related Issue](https://github.com/osmandapp/Osmand/issues/473)

ypid:
Actually, the use of [\<additional_rule_separator\>](https://wiki.openstreetmap.org/w/index.php?title=Key:opening_hours/specification#additional_rule_separator), which is not supported by OSMand, is not recommended for this value. Rewrite as "Mo-Th 09:00-21:00; Fr-Sa 09:00-17:00; Su off".
Parsing the \<additional_rule_separator\> is difficult …

That said, the I would also like to see issue #473 closed.

vshcherb:
I don't expect any time soon to use javascript library especially we used this function in a critical way for search. 
I expect 
"Mo-Th 09:00-03:00, Fr-Sa 09:00-04:00; Su,PH off", 
that this will fail with PH detecting but won't fail with other values

ypid:
Wrong. The value "Tu-Th 07:00-2:00, Fr 17:00-4:00, Sa 18:00-05:00; Su,Mo off" throws an exception:
Exception in thread "main" java.lang.NullPointerException
        at OpeningHoursParser.testOpened(OpeningHoursParser.java:700)
        at OpeningHoursParser.main(OpeningHoursParser.java:800)

I intend to bundle the javascript library into a Java library (.jar) in the future. Would you consider using it then?

midnightcomm:
Made a note on OSM Wiki for future editors/users.

https://wiki.openstreetmap.org/wiki/Key:opening_hours#OsmAnd

vshcherb:
To avoid sort of confusion.
[...]

PR #1192, opened 2015-05-09
Added test for "Mo-Th 09:00-03:00, Fr-Sa 09:00-04:00; Su off". Not evaluated correctly. Related to #1154.

ypid:


vshcherb:
testOpened("16.05.2015 03:59", hours, true); // FIXME: Wrong
I think this is correct
Fr-Sa 09:00-04:00 it stays that Saturday will be open till 4 o'clock, for me it is hard to image if I see 
Fr-Sa 09:00-04:00, Su off this, I would think other way

ypid:
I am not sure if I understood you correctly. Please check on the [evaluation tool](http://openingh.openstreetmap.de/evaluation_tool/?EXP=Mo-Th%2009%3A00-03%3A00%2C%20Fr-Sa%2009%3A00-04%3A00%3B%20Su%20off&DATE=1431213240000&lat=48.7769&lon=9.1844&mode=0). Is the value evaluated there as you expect it?
2015-05-16 03:59 (Sa) should be open because of the rule: "Fr-Sa 09:00-04:00"

vshcherb:
https://github.com/osmandapp/Osmand/commit/5cd40baff32df901437ff2db48ba19287fd42563

vshcherb:
Yes, it should be open. I didn't understand what is the failure I did only small change "," -> ';' , seemingly osmand doesn't parse ',' correctly I think in the following examples it should be just replaced with ';'.

Please submit other failing tests or point to any wrong test

ypid:
> Please submit other failing tests or point to any wrong test

For starters, you can have a look at the test framework of opening_hours.js :smile:
https://github.com/ypid/opening_hours.js/blob/master/test.js

> [I intend to bundle the javascript library into a Java library (.jar) in the future. Would you consider using it then?](https://github.com/osmandapp/Osmand/issues/1154#issuecomment-100557718)

vshcherb:
I never doubt to use something standard or specially purposed. I even think to start with it in iOS, though for existing solutions I have really strong criteria it shouldn't be worse by any meaning:
- I see js more powerful
- I'm not sure about performance, it could be critical because we have search POI that are opened, so the numbers of POI starts with 1K and this is on Android devices (!) of course we expect this to be quick
- It is used for maxspeed:conditional routing times and there it could even more critical, but I hope it is not used directly in routing

Current solution solves most of the opening times I know, new UI is made in a such a way that it shows original string even if it is wrong, the user could see the times. I see quite a big problem that OSM data is going to a format which user might not understand, which means that it should be specially formatted and so on.

So it is a hard choice, but if you could help us, you are welcome & we are ready to try it.

vshcherb:
I would also think something JS -> Java automatic translation.

ypid:
OK, good to hear that. It will probably take me some time before the Java library is ready and tested. I will keep you posted.
The performance question is an open question but it should work. Maybe some caching can reduce the performance problem.
An automatic translation of the JavaScript code into Java Code is [not practical/doable](https://stackoverflow.com/questions/10370384/how-to-convert-javascript-to-java).

PR #1208, opened 2015-05-11
tighten parseLength() again

aceman444:
I wonder what the test for "-" does. There can't be any negative values in maxlength or maxheight.
It seems your version also accepts values like "1 km 1 km" (meaning 2km) which is invalid per the wiki, but probably shouldn't harm. I'm fine with that. But I think it also accepted "1 mile" and thinking it was 1meter. You can't just test unit.contains("m") as that would also match "mile", "cm", etc. That is not good so I tightened it a bit.

So I haven't compiled this again, please tell me if all the tests pass.

### DIFF
--- a/OsmAnd-java/src/net/osmand/binary/RouteDataObject.java
+++ b/OsmAnd-java/src/net/osmand/binary/RouteDataObject.java
@@ -192,26 +192,37 @@ public class RouteDataObject {
 		int i = Algorithms.findFirstNumberEndIndex(v);
 		if (i > 0) {
 			f += Float.parseFloat(v.substring(0, i));
-			String pref = v.substring(i, v.length()).trim();
+			String unit = v.substring(i).trim();
 			float add = 0;
-			for(int ik = 0; ik < pref.length(); ik++) {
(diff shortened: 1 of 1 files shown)


aceman444:
What does the label mean? Can you build OsmAnd with the patch and say if it works and all the tests pass? It will be useful for the maxheight support in routing.

aceman444:
Can you please look at the patch?

vshcherb:
I don't really understand the change and I think the current solution works fine. May be it is good but it requires quite some investigation, last time it took quite a lot for that small feature to analyze

aceman444:
OK, take your time:) I explained why the change is needed (what is wrong with current state) and also added more tests so it can be verified if the code works right.

aceman444:
So what has changed? Why is the patch bad?

vshcherb:
I added the test cases and they are all passing (https://github.com/osmandapp/Osmand/commit/c792360fdd4f8db7155375d4f4c5c09fe3ba9162) except few with bad value. So I don't see a reason to change code.

aceman444:
Yes, because you do not test what happens on invalid values.

PR #1214, opened 2015-05-12
Orthographe

GiZig:
picnic (en) = pique-nique (fr)


vshcherb:
I think it is better to use https://hosted.weblate.org/projects/osmand/ (to have review and discussions there)

GiZig:
OK, I didn't know it was better to use weblate. I'll do it now.

PR #1218, opened 2015-05-13
Add "more" to "search villages/postcodes"

stephan75:
each month we have users who cannot find places in offline address search because they do not press the button "search villages/postcode" because instant search has no result for smaller places.

see https://groups.google.com/forum/#!topic/osmand/IDm-pPb4IpI again.

So users need to know that this button does a further deeper search.

Alternatively add a text like "Press above button when place name is not found instantly."

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -949,7 +949,7 @@
     <string name="monitoring_control_start">GPX</string>
     <string name="no_buildings_found">No buildings found.</string>
     <string name="incremental_search_city">Search city incrementally</string>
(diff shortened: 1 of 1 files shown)


sonora:
I had been discussing this with many users over several years. I think it is worth for us to think about removing this button completely, and _always_ use a complete search automatically. Some years back this was _very_ slow, but devices and our code has now improved significantly.
Best, Hardy

vshcherb:
Nope I would disagree, actually it still slow on my device. And it is not good, also don't forget it could cause out of memory which could be a serious issue. 
I think we should use same search as on iOS devices, which uses different file section, but this task is pending from 2.0, cause we want to change to one line search anyway

PR #1219, opened 2015-05-13
Use building entrance as a building location (handy for routing)

Larry0ua:
In case when only one entrance is present it is nice to route user directly to that entrance. 

Logic: if only one main entrance is present - it's chosen. If only one entrance in a building - it's chosen. Otherwise, default behavior is using weighted center.

### DIFF
--- a/OsmAnd-java/src/net/osmand/osm/edit/EntityParser.java
+++ b/OsmAnd-java/src/net/osmand/osm/edit/EntityParser.java
@@ -1,22 +1,17 @@
 package net.osmand.osm.edit;
 
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-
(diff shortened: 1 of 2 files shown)


CJTmmr:
Hi,

Suggestion:
Always when a destination (POI on a node or adress-node) is within a building, the route could go to the entrance of the building (not only when the POI is on the building).

Try to route (car-mode) to the adress:
Country: The Netherlands,
City: Bodegraven, 
Street: Rijngaarde
Housenumber: 220



OsmAnd Offline route takes you to Dronensingel, close to the appartment, but on that side of the building there is no entrance (and even a ditch between Dronensingel and the building..)

In future Osmand could route to the entrance (tagged with entrance=main) on the other side of the building.
Compare with walking-route:


Good luck

CJTmmr

PR #1243, opened 2015-05-17
Update favorite_edit_dialog.xml

Edward17:
I propose to make some changes in Favourite edit dialog:
Delete unused LinearLayout and android:minLines="2", add ScrollView, other small fixes

### DIFF
--- a/OsmAnd/res/layout/favorite_edit_dialog.xml
+++ b/OsmAnd/res/layout/favorite_edit_dialog.xml
@@ -48,6 +48,7 @@
 			android:layout_width="fill_parent"
 			android:layout_height="wrap_content"
 			android:inputType="textMultiLine|textCapSentences"
+			android:minLines="2"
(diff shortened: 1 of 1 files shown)


vshcherb:
Would be better to see screenshots. And I'm not sure minlines should be deleted because it shows multiline structure as it supposed to be for description

Edward17:
@vshcherb , OK, android:minLines="2" can be kept on
> because it shows multiline structure

Why I propose to add ScrollView? In landscape orientation "Description" fielв is not visible and there isn't any ability to scroll down and fill this field:

screenshot

A display isn't small: 4,3 ", 720x1280 px

vshcherb:
Yes good idea. Please put minLines=2 and we will merge immediately.

Edward17:
I have added minLines=2.

Edward17:
Thanks!

Issue #1198, opened 2015-05-10
Display full information about turn lanes


vshcherb:
Offline Routing: Display full information about turn lanes

vshcherb:
https://code.google.com/p/osmand/issues/detail?id=2592

vshcherb:
Implemented

PR #1273, opened 2015-05-25
Add small left turn and right turn icons. Work on #1198.

saiarcot895:
If `true` is passed in into the `smallIcon` parameter, and a left or right turn is to be drawn, then it will draw a smaller version of those icons. These icons _should_ be centered on the straight arrow (haven't verified this), so they can be overlaid directly on that arrow.

Currently, these icons aren't used yet, but are planned to be used in drawing the turn lanes.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/TurnPathHelper.java
+++ b/OsmAnd/src/net/osmand/plus/views/TurnPathHelper.java
@@ -17,7 +17,7 @@ import android.util.FloatMath;
 public class TurnPathHelper {
 
 	// 72x72
-	public static void calcTurnPath(Path pathForTurn, TurnType turnType, Matrix transform) {
+	public static void calcTurnPath(Path pathForTurn, TurnType turnType, Matrix transform, boolean smallIcon) {
 		if(turnType == null){
 			return;
 		}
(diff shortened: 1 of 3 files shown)


vshcherb:
Would be great to see the screenshot

saiarcot895:
smallarrow

(This is after setting the `false` to `true` for the `smallIcon` parameter. With the code change in the commit, there should be no difference in what is displayed.)

Note that the tip of the arrow is cut off due to the width of the image. I'm hoping that, in displaying the turn lanes, this won't have an adverse effect.

vshcherb:
I'm slightly missing the target of that pull request. Anyway you could test the lanes easily calculating route  and using next turn indicator (you will see the lanes as well), so you don't need to travel to see how it looks

saiarcot895:
The reason for these icons is that, if I try to draw the left turn arrow and the straight arrow for a single lane, then it won't look good. It will look something like the following:

overlaid turns

I could shift the straight arrow to the side so that it lines up with the upward portion of the left turn arrow, but the shift will have to be done for each turn type.

overlaid shifted turns

Instead, what I'm proposing is to create smaller, centered icons for each turn type, so that the icons can be just overlaid one on top of another. That way, no manual shifting will be needed for each turn pair, and the code can be kept (fairly) clean, while having the icons still look good.

As for the testing, I can't do that yet because the code there:
1. Only displays one turn per lane.
2. Uses the large icons (by default).

I don't have the code ready to display two turns per lane.

saiarcot895:
I just finished up the code to show two turns per lane, and it doesn't look too bad.

smallicon1
smallicon2

The tip of the arrow is being cut off here as well. Is that acceptable, or should the arrow be shrunk down so that it would fit?

vshcherb:
Doesn't seem good though, I think we need to improve arrows slightly. Will come back to that issue later and I hope to provide new geometry as soon as I have it.

PR #1277, opened 2015-05-27
If adding a left/right turn to a lane, make sure that that lane is ma…

saiarcot895:
…rked as selected.

If a segment has something like `turn:lanes=left|through`, and a right turn is being made, then the left arrow followed by the right arrow appear (correctly), but both arrows are gray. This change makes sure that the right-most lane (or left-most lane, in the case of a left turn) is selected to be highlighted.


saiarcot895:
Before change:
laneaddedbefore

After change:
laneaddedafter

saiarcot895:
Ping.

vshcherb:
This one I guess is easy. I just want to discuss about unit-testing that functionality I already have list of intersections, but it is impossible to overview already the algorithm. It is critical we start writing tests for all implemented cases.

PR #1374, opened 2015-06-21
Small refactoring

Edward17:
I propose to change a name of file res/navigate_point.xml in order to make the finding of this file easier.


vshcherb:
Makes perfect sense, I hope it was tested

Edward17:
@vshcherb 
no, I haven't tested it :)

I just searched files in this repository which link to R.layout.navigate_point and changed them. So it must be tested.

For example, _gen/R.java_ class was not changed. However it will be changed during compilation.

PR #1405, opened 2015-06-25
Fixed a crash due to errors in format strings

mariyan-borisov:
When you set the language to Bulgarian and then you go to Manage map files (Управление на карти) the application crashes.


vshcherb:
Could you please that change to r2.1 branch too?

mariyan-borisov:
Done... https://github.com/osmandapp/Osmand/pull/1406

PR #1482, opened 2015-07-19
Support zooming via mouse wheel.

rdoeffinger:
Behaviour should possibly be fine-tuned to use a
smooth instead of animated zooming and to keep
the cursor position as a fixed point instead of
moving it to the center.
Still, this should be a small improvement for usage
with mouse.


vshcherb:
Thanks please test it.

PR #1487, opened 2015-07-20
Rate us dash added

GaidamakUA:
Some imports organized
String resource name changed to more appropriate
Style "OsmandLight" subheader color changed to fit the design.


GaidamakUA:
Removed logging
Fixed issue with email

PR #1495, opened 2015-07-23
Notification with buttons and Bookmark dialog enchantment.

GaidamakUA:


vshcherb:
Can you please describe what is doing? Also please I recall other issues with notification, should they also be in that pull-request

PR #1596, opened 2015-08-25
POI - added single tap

crimean:


OsmAnd/src/net/osmand/plus/views/OsmandMapLayer.java
```diff
@@ -30,6 +30,23 @@ public void onRetainNonConfigurationInstance(Map<String, Object> map) {}
 	
 	public void populateObjectContextMenu(Object o, ContextMenuAdapter adapter) {}
 
+	/**
+	 * return true when this layer should handle single tap instead of context layer
+	 * @return
+	 */
+	public boolean isSingleTapResponder() {
```
vshcherb: We can remove that method if we just follow onSingleTap single action click. onSingleTap returns true

OsmAnd/src/net/osmand/plus/views/OsmandMapLayer.java
```diff
@@ -30,6 +30,23 @@ public void onRetainNonConfigurationInstance(Map<String, Object> map) {}
...
+		return false;
+	}
+
+	/**
+	 * return true when context layer should open context menu of the long pressed point
+	 * (instead of common behaviour when lat/lon text opens on long press)
+	 * @return
+	 */
+	public boolean shouldProcessLongPress() {
```
vshcherb: Similar suggestion to isSingleTapResponder(), so we don't overload map layer with methods.

crimean: If we do so - Measurement plugin will broke since it's layer lies under Context Menu layer.

vshcherb:
can't we change order of layers?

PR #1687, opened 2015-09-13
Small changes in structure of address- and coordinates search window

Edward17:
With this changes User Interface will be more logical.


vshcherb:
Can you please provide screenshots to understand the change.

Edward17:
Here:

vshcherb:
My concern is only that keyboard can hide Latitude on small screens and I believe I saw it couple of times. For the rest I'm fine, let's see if Dmitriy can say something.

vshcherb:
Unfortunately no action was made since pull request and right now it's going to deprecate with new search.

PR #1712, opened 2015-09-18
Remove "avoid" prefix in the "Avoid..." menu

Debeet:
It's unnecessary to have "avoid" before every option in "Avoid..." menu. It's wasting screen space and making the app a little less readable.


aceman444:
Aren't those strings also used directly in the route planning toolbar and there there is not leaving "avoid..." as in the configuration menu? In that place the "avoid" prefixes are needed as not all routing options are "avoid...". Did you create these other versions of the strings, or they already exist? Other than that, I like this deduplication. There seems to be a similar case in the "Announce..." menu. Can you do that too?

Debeet:
> There seems to be a similar case in the "Announce..." menu. Can you do that too?

Sure, I just submitted a new PR (#1718).

As for the route planning toolbar... to be honest I'm not really sure what it is so I can't answer this question.

aceman444:
You set a destination and then click on the curved arrow on the main map. That switches on the "routing toolbar" (my name, I don't know how it is called officially). Is has the x icon for stopping routing and the "go" arrow as last icon. The third icon is a cogwheel, where there are options for the routing. A menu expands and there are the "avoid options" and some others. The "avoid options" do not have a label of "avoid..." to group them so each of them needs the "avoid" prefix in the string itself.

Debeet:
Ah, I see. Yes, after this change it may sound confusing. I have a simple solution to this problem so we don't need to double these strings - negate these options in this toolbar. If for example "unpaved roads" option is checked, it means that user wants "unpaved roads". Same situation with other options.

aceman444:
These options are handled in the routing.xml file and it is not so easy to invert their meaning. The routing file format is much better at negative options (disallowing of roads). Also, till recently there were 2 options "avoid highways" and "prefer highways". It would get strange if you shorten the first to "Highways".

vshcherb:
Thanks aceman, you are right we can't remove these strings as they are. They are used in route preparation screen (bottom on the map settings), I think we need to add new strings if we want to change settings screen.
I reverted previous commit.

aceman444:
But then how do you tie the options in the settings screen to the route preparation options? I always thought somehow the options from routing.xml are also displayed in the settings screen. As the strings are the same at both places, the tie (link) would be preserved. But how would it work with different strings?

vshcherb:
True, I think it is a bigger change than I though, so possibly we can leave these duplicate as it is today.

PR #1750, opened 2015-09-26
add more variants of times into testing voice commands

aceman444:
I try to update sk/ttsconfig.p to support more of the existing features of voice commands. As there are several forms of plurals, I update the route times to be more different.

I still do not understand why in the phrase "onto 'Market Square'", the internal command is 'street(p,"", "", "Market Square")' (market square as 3rd argument to street()).
In the phrase "onto 'Bridge Avenue'" the internal command is 'street(p, "Bridge Avenue")' (Bridge avenue is first argument).
What is the difference? It seems to me the 2nd argument is road ref, first could be the street name. 3rd could be the destination (something a bit away, not current street/road). But these assumptions are not true at these commands, and also at the "onto 'SR 80'" (it looks like a ref, but is used in 1st argument). So what are the rules?

Also I'd like if there was an example using e.g. bearLeft() including a street name. I don't know how the street name is used, in which context. Is it the street on which you should be keeping left, or is it the one where you get after you turn left (bearing on left before that)?

### DIFF
--- a/OsmAnd/src/net/osmand/plus/development/TestVoiceActivity.java
+++ b/OsmAnd/src/net/osmand/plus/development/TestVoiceActivity.java
@@ -167,14 +167,14 @@ public class TestVoiceActivity extends OsmandActionBarActivity {
 	}
 	
 	private void addButtons(final LinearLayout ll, CommandPlayer p) {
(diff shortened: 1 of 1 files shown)


vshcherb:
I don't really see bear left only A_LEFT_KEEP and slight turn left.

aceman444:
I think sonora knows what I mean, see pull #349. Can we get him to comment here? :)

PR #1902, opened 2015-11-11
AndroidManifest.xml: add intent-filter to match Android Download Manager

bagage:
Download Manager uses content data type when requesting file opening. Currently, it fails opening `GPX` file because it cannot find anything handling it.


bagage:
Any comment?

vshcherb:
1. This branch has conflicts that must be resolved

2. +                <data android:scheme="content" android:host="*" android:mimeType="text/xml"/>
It looks like OsmAnd will try to open any file, which I'm afraid could be wrong because we support only GPX and KML partially, for that purpose we need to think about validation or less mimeTypes.

bagage:
# Done
# Since both gpx and kml can use `application/xml` mime type, I do not think we can filter valid documents within intent filter. I will add some code in java to not crash when input cannot be parsed due to invalid type (it currently crashes).

vshcherb:
I guess the issue was fixed with mimeType gpx+xml

bagage:
Sorry but that's not the case.

bagage:
edit: after checking again, it seems that behavior has changed since my PR and mimeType is indeed set to `gpx+xml` which is great. Tried master and it works ok, this PR is indeed irrelevant now.
Thanks @vshcherb!

PR #1954, opened 2015-11-25
Plugin for Android Smartwatches

and01:
This is an integration for the plugin called SmartNaviWatch. It enables the user to see the current location or the current navigation route on an Android smartwatch. The SmartNaviWatch library is used for messaging.

### DIFF
--- a/OsmAnd/build.gradle
+++ b/OsmAnd/build.gradle
@@ -19,7 +19,7 @@ apply plugin: 'com.android.application'
 
 android {
 	compileSdkVersion 21
-	buildToolsVersion "23.0.1"
+	buildToolsVersion "21.1.2"
 
 	signingConfigs {
 		development {
@@ -38,12 +38,14 @@ android {
 	}
 
 	defaultConfig {
-		minSdkVersion System.getenv("MIN_SDK_VERSION") ? System.getenv("MIN_SDK_VERSION").toInteger() : 14
+		minSdkVersion System.getenv("MIN_SDK_VERSION") ? System.getenv("MIN_SDK_VERSION").toInteger() : 19
 		targetSdkVersion 21
+		multiDexEnabled true
 
 		versionCode System.getenv("APK_NUMBER_VERSION") ? System.getenv("APK_NUMBER_VERSION").toInteger() : versionCode
 		//versionName already assigned in code
 		//versionName System.getenv("APK_VERSION")? System.getenv("APK_VERSION").toString(): versionName
+
 	}
 
 	lintOptions {
@@ -59,7 +61,8 @@ android {
 	}
 
 	dexOptions {
-		jumboMode = true
+		incremental true
+		javaMaxHeapSize "4g"
 	}
 
 	sourceSets {
@@ -320,6 +323,10 @@ repositories {
 	// mavenCentral()
 }
 
+configurations {
+	all*.exclude group: 'com.android.support', module: 'support-v4'
+}
+
 dependencies {
 	compile project(path: ":OsmAnd-java", configuration: "android")
 	compile project(":eclipse-compile:design")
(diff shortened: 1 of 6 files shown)


vshcherb:
Though the plugin doesn't seem to be very big, but it will take quite effort to merge it properly.
1. We already upgraded to SDK 23 , it would be great if you could rebase your changes or merge with master.
2. It would be not great to increase minsdk version to 14, I think we should workaround it and android provides possibility how to enable some features for certain versions. For example everything that requires 19 api, should be gathered in one class. I think it is fine, to create a special OsmandPlugin Smart Watch integration.

OsmAnd/src/net/osmand/plus/OsmandPlugin.java
```diff
@@ -116,6 +117,7 @@ public static void initPlugins(OsmandApplication app) {
 		allPlugins.add(new AccessibilityPlugin(app));
 		allPlugins.add(new OsmEditingPlugin(app));
 		allPlugins.add(new OsmandDevelopmentPlugin(app));
+		allPlugins.add(new SmartNaviWatchPlugin(app));
```
vshcherb: here it could be if(Android.SDK_BUILD >= 19) {
static function to create that class (and specify full classpath to the class without imports)
}

OsmAnd/build.gradle
```diff
@@ -320,6 +323,10 @@ repositories {
 	// mavenCentral()
 }
 
+configurations {
```
vshcherb: Why do we want to exclude it?

OsmAnd/src/net/osmand/plus/smartnaviwatch/SmartNaviWatchPlugin.java
```diff
@@ -0,0 +1,501 @@
...
+
+        return c;
+    }
+
+    /**
+     * Converts a MapDataObject into a polygon
+     * @param obj    Map Data Object
+     * @return       Polygon
+     */
+    private MapPolygon polygonFromDataObject(BinaryMapDataObject obj) {
```
vshcherb: Is it simplified rendering for watch? Did you consider to create special rendering style for that.

OsmAnd/src/net/osmand/plus/OsmandPlugin.java
and01: Agreed. I changed this.

OsmAnd/build.gradle
and01: We don’t. I got commited by mistake..

OsmAnd/src/net/osmand/plus/smartnaviwatch/SmartNaviWatchPlugin.java
and01: Yes, it is simplified. In order to minimze the data to be transferred to the watch (and for presenting a clean and crisp looking map) we just transfer the map data and not a prerendered image.

vshcherb:
Could you please provide more information what library is used on the smartwatch? Why did you decide to split communication code in 2 projects.

and01:
The library is a generic one for sending navigation information to the wear app. Available here:
https://github.com/and01/SmartNaviWatch

The UML looks like this:

[cid:D6BD4F04-FA29-47AB-9950-F69A42686CA7@local]




On 25 Nov 2015, at 16:52, vshcherb <notifications@github.com<mailto:notifications@github.com>> wrote:

Could you please provide more information what library is used on the smartwatch? Why did you decide to split communication code in 2 projects.

JiK:
Any news on this?

idtownie:
(bump?)

OsmAnd/src/net/osmand/plus/smartnaviwatch/SmartNaviWatchPlugin.java
```diff
@@ -0,0 +1,501 @@
...
+        }
+
+        /**
+         * Navigation has ended
+         */
+        @Override
+        public void routeWasCancelled() {
+            cleanRouteData();
+
+            //sendMessage(MessageTypes.CancellationMessage, null);
```
GaspardPO: there is still some commented code.
does it has a meaning, or it is a oversight?

StalenBas: Seems like the original coder has abandoned this project, seeing as it's been three years. Does that mean this will not get added any time soon? Would be a shame, since the bulk of the coding seems to have been done already.

vshcherb:
The code was reviewed by the team and it is quite complex, we are not ready to pick it up & maintain. So very likely it could exist as a side project.

Issue #2019, opened 2015-12-03
Fix for Persian and Urdu language maps


AmZaf:
I have fixed a bug related to Arabic shaping which causes incorrect rendering of place names on Persian and Urdu language maps. The fix involves the icu4j library on OsmAnd/libs/icu4j and OsmAnd-java/libs/icu4j path. I have filed a ticket upstream at http://bugs.icu-project.org/trac/ticket/12021. The fix is slated for the next major release of the library. 
In the mean time, what is the best way I can contribute to Osmand? Will you accept a patched version of icu4j? I reckon the issue deserves high priority.

AmZaf:
Edit:The original link to the icu4j bug report page was incorrect and has been corrected.

sonora:
See my comment here!  https://github.com/osmandapp/Osmand/commit/d73f34d9d341fffa0d51c2c698dc8666acb806c3

PR #2036, opened 2015-12-08
Patched Osmand icu4j to support Persian and Urdu #2019

AmZaf:
As mentioned by @sonora the commit d73f34d caused bloated apk. I have backported my changes to the original Osmand library so that the file size remains virtually unchanged and the issue #2019 gets resolved.


vshcherb:
The only one place it is used is in the class Reshaper, is it possible to add a simple test in main method? So next time we know what is expected value.

AmZaf:
Yes, it's possible. You mean like the one commented out for Hebrew?

vshcherb:
Yes, I don't why I commented it out, but at least serves the purpose. More or less reasonable example of tests are here https://github.com/osmandapp/Osmand/blob/master/OsmAnd-java/src/net/osmand/util/OpeningHoursParser.java#L879

AmZaf:
Here is a simple Arabic shaping test case:

	public static void main(String[] args) {
		String s = "گچ پژ نمکی باللغة العربي";
		String reshape = reshape(s);

		if(!reshape.equals("ﻲﺑﺮﻌﻟﺍ ﺔﻐﻠﻟﺎﺑ ﯽﮑﻤﻧ ﮋﭘ ﭻﮔ")) {
			throw new IllegalArgumentException("BUG!!!");
		}
	}

AmZaf:
Why is there a size discrepency in the the displayed apk size? http://download.osmand.net/latest-night-build/

PR #2055, opened 2015-12-13
Losslessly shrink png files.

rdoeffinger:
Done using trimage (wrapper around tools like pngcrush),
applied to all files even when it is barely worth it.
Overall savings 2012147 bytes.

Signed-off-by: Reimar Döffinger <Reimar.Doeffinger@gmx.de>


vshcherb:
Thanks! I think this happens automatically every buidl using pngcrunch. I'm afraid this is not that important because it will just require extra testing and there is no clear value.

rdoeffinger:
I understand your point about testing, but can you point me to where this optimization happens at build?
I can't find anything like that in the gradle files.
Also note that trimage does further optimizations like reducing palette which pngcrush can not and are sometimes particularly effective, though I have not compared for these files specifically.

vshcherb:
http://tools.android.com/tech-docs/new-build-system ( crunch png)

PR #2167, opened 2016-01-17
Add 'Postcode' to address search

keunes:
After some searching on the website I figured out that search based on postcode *is* possible. Rather than giving explanation on the website, it's more convenient to novice users to just print the info in the search screen.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -1997,7 +1997,7 @@ Afghanistan, Albania, Algeria, Andorra, Angola, Anguilla, Antigua and Barbuda, A
     <string name="navigate_point_format_DMS">DDD MM SS.SS</string>
     <string name="search_address_top_text">Select address</string>
     <string name="search_address_region">Region</string>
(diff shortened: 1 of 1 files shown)


vshcherb:
Unfortunately it doesn't fit UI

keunes:
Doesn't fit as in? (is it a matter of being ugly or really not fitting? I suspect it's the former as to me there seems enough space to accommodate the word postcode, which is just one letterspace longer than 'building')

vshcherb:
Doesn't fit in the search address screen on the left column

noci2012:
Zipcode?

vshcherb:
The issue will be anyway present with translation, I think we need to add postcode in the text box hint when you enter the city, but I'm afraid it is already there?

keunes:
> I think we need to add postcode in the text box hint when you enter the city, but I'm afraid it is already there?

That's a good one. Currently when the 'City' text field is empty, it displays 'CHOOSE CITY' we could append ... OR POSTCODE (or replace OR with / if needed)
Seems there's enough space there, and it would help a lot as well.

PR #2174, opened 2016-01-18
Add 'or postcode' address search

keunes:
Within the address search, there is no hint that the user can search for a postcode. Adding this text in the 'City' field (displayed when no input given) helps the user a lot.


keunes:
As alternative to #2167

PR #2245, opened 2016-02-15
Create values-es-rUS to Latin America

franco999:
Considering that most of smartphones in Latin America only have the option of Spanish (Spain) => [es] and Spanish (United States) => [es-US], some words do not mean the same in Spain as in Latin America, that the language in the app is changed depending on the system language and the Spanish (Spain) option and is used with "values-es", I have created the "values-en-rUS" folder for neutral translation into Spanish in Latin America.

idioma


vshcherb:
Thanks!

PR #2252, opened 2016-02-18
Properly calculate the disk usage change for downloads of updates, improve messages to user

stefanhoehne:
# Motivation
Currently, OsmAnd checks the free space against the cumulated size of the downloaded files.
This works well with new downloads, but results in irritating messages when downloading map updates.

# Solution
Properly calculate the changed disk usage and max temp usage and show it to the user.
Don't allow the download if the free space will not be sufficient.


vshcherb:
Thanks good wor, though we need to test it.

Issue #2193, opened 2016-01-23
[Accented Search] search names including their diacritics alternatives


okias:
When entering name of street, please consider searching for "č" when entered "c" etc. This could be really helpful and navs like Waze does it.

My proposal is first list names specific to entered letter, in second pass check for all accented letters and print them below first search result.

Of course, this should work even when street with non-accented letter is found, because for example, in Czech republic, there is majority of streets which containts accented letters and lot of similiar ones.

Some resources:
[1] http://stackoverflow.com/questions/14083446/accented-search-in-sqlite-android
[2] http://stackoverflow.com/questions/16282083/how-to-ignore-accent-in-sqlite-query-android

hakuchi:
this sold work with all diacritics, not only "č", "ø", "ü" etc. 

here a list of diacritics:
[...]

PR #2254, opened 2016-02-19
support diacritics symbols also with czech language

okias:
Search without diacritics will work properly in Czech Republic, Slovakia or Romania only when phone UI is set to these languages.

This is temporary solution.This should be handled better.

Addition to bug: https://github.com/osmandapp/Osmand/issues/2193


vshcherb:
Please test it.

PR #2297, opened 2016-03-02
Remove icu4j and reshaper as they are no longer in use

ebraminio:
I was investigating a bug on osmand on Persian and with code grepping reached to the fact you no longer use reshape anywhere on the code. Even so, I wanted to find the place you are using it and remove it from there as you are not supporting Android lower than SDK14 and this is totally unnecessary if not making other issues by its own.


vshcherb:
They are in used and they are used by native engine via jni calls

ebraminio:
Where is that native engine? The point is it should be just removed, the Android version your are targeting doesn't need it (if you using right calls on the system that I can investigate if I could see the native engine)

ebraminio:
And the reason for my insist is the fact it is still buggy and users are complaining about it and if there is no way to avoid simply I know more tested and very lightweight alternative that I use on my project.

ebraminio:
Ok. I found the native engine but https://github.com/osmandapp/OsmAnd-core/blob/21d847cd511ac5003d12e77fde01b7708b189dc7/src/ICU.cpp#L126 is directly using ICU4C and not Java library imported here, ICU4J.

PR #2351, opened 2016-03-20
Reworked GPX display.

andrew-davie:
1. GPX track colour can come from .gpx file (as before) via \<extensions> block. However, this was previously
 on an entire-track basis. Since the reader code also worked when an \<extensions> block was placed inside a \<trk> or
\<trkseg> block, I have honoured this implicit arrangement.  So, if an extension containing \<color> is inside a \<trkseg>
then that segment will be drawn in the selected colour. If there is no colour specified for a segment, it will draw
in it's parent track's colour. If the parent's colour is not specified, it will draw in the GPX file's colour. If
none is specified then the default OsmAnd colour is used.

2. Track width is also configurable on per-segment. I added \<zoom>1.0\</zoom> into the \<extensions>, so if you put
that in either a gpx \<trk\> or \<trkseg> then you will get scaling of the width of the drawn segments. The zoom is
relative to default size, so 1.0 is 100%. To draw double-thick lines, use 2.0, or half-width 0.5.

With (1) and (2) you can use one or both inside the \<extensions> block.

The reasoning for the above is that you may wish to have some tracks marked with easy and difficult sections, or
by different surface types, etc., By putting colouring capability in segments you can do all of that.

3) Reduced points for speedups.  Huge GPX files with hundreds of thousands of points could potentially slow the UI down.
I have implemented point-reduction at runtime so that tracks are displayed using the minimum number of points required
to maintain visual integrity. Code is inside GPXLayer.java and uses Ramer-Douglas-Peucer algorithm. Selection criteria
for culling points is based on view zoom level.  The culled point set is cached, so the line reduction only happens when
there is a change in zoom, or number of points in the original line. The reduction could theoretically replace the
"offscreen" culling code, because that code has a number of bugs;

  - e.g., 1) lines just offscreen are clipped but you can see curved ends on them when they shouldn't be.
  2) some lines that should be onsceen are totally culled. I have a sample track demonstrating this bug.

My preference would be to simply rely on the point culling to give great perfomance and drop the segment onscreen checks.

4) Scaling based on view zoom - the line widths are modified according to the view zoom modified by (2) the track
zoom. Now when you zoom right out, you don't get a large block of "mess" - in fact tracks are legible right down to very
small sizes (zooming right out).  It's a lovely thing.

Andrew Davie
andrew@taswegian.com

### DIFF
--- a/OsmAnd/src/net/osmand/plus/GPXUtilities.java
+++ b/OsmAnd/src/net/osmand/plus/GPXUtilities.java
@@ -59,7 +59,22 @@ public class GPXUtilities {
 			}
 			return extensions;
 		}
-		
+
+		public float getGpxZoom(float defaultGpxZoom) {
+			if(extensions != null && extensions.containsKey("zoom")) {
+				try {
+					defaultGpxZoom = Float.parseFloat(extensions.get("zoom"));
+				} catch (NumberFormatException e) {
+					e.printStackTrace();
+				}
+			}
+			return defaultGpxZoom;
+		}
+
+		public void setGpxZoom(float gpxZoom) {
+			getExtensionsToWrite().put("zoom", Float.toString(gpxZoom));
+		}
+
 		public int getColor(int defColor) {
 			if(extensions != null && extensions.containsKey("color")) {
 				try {
@@ -166,7 +181,9 @@ public class GPXUtilities {
 
 	public static class TrkSegment extends GPXExtensions {
 		public List<WptPt> points = new ArrayList<WptPt>();
-		
+		public List<WptPt> culledPoints = null;
+		public int culledFingerprint = -1;
+
 		public List<GPXTrackAnalysis> splitByDistance(double meters) {
 			return split(getDistanceMetric(), getTimeSplit(), meters);
 		}
@@ -692,12 +709,14 @@ public class GPXUtilities {
 			List<TrkSegment> tpoints = new ArrayList<TrkSegment>();
 			for (Track t : tracks) {
 				int trackColor = t.getColor(getColor(0));
+				float trackZoom = t.getGpxZoom(getGpxZoom(1.0f));
 				for (TrkSegment ts : t.segments) {
 					if (ts.points.size() > 0) {
 						TrkSegment sgmt = new TrkSegment();
 						tpoints.add(sgmt);
 						sgmt.points.addAll(ts.points);
-						sgmt.setColor(trackColor);
+						sgmt.setColor(ts.getColor(trackColor));
+						sgmt.setGpxZoom(ts.getGpxZoom(trackZoom));
 					}
 				}
 			}
@@ -964,6 +983,7 @@ public class GPXUtilities {
 		SimpleDateFormat format = new SimpleDateFormat(GPX_TIME_FORMAT, Locale.US);
 		format.setTimeZone(TimeZone.getTimeZone("UTC"));
 		try {
+
 			XmlPullParser parser = PlatformUtil.newXMLPullParser();
 			parser.setInput(getUTF8Reader(f)); //$NON-NLS-1$
 			Stack<GPXExtensions> parserState = new Stack<GPXExtensions>();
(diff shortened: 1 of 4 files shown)


andrew-davie:
Sample .gpx with zoom and colour tags on the per-segment level for use with this mod.

[wworks.gpx.zip](https://github.com/osmandapp/Osmand/files/181256/wworks.gpx.zip)

OsmAnd/src/net/osmand/plus/GPXUtilities.java
```diff
@@ -989,87 +1009,44 @@ public static GPXFile loadGPXFile(Context ctx, InputStream f) {
 					} else if (parse instanceof GPXExtensions && tag.equals("extensions")) {
 						extensionReadMode = true;
 					} else {
-						if (parse instanceof GPXFile) {
```
vshcherb: This is quite a big change and because of the permutation of the lines. Will it be possible to reduce the diff so we can check semantically the changes rather than deep testing.

andrew-davie: You can discard this part of the change.
There is/was a slight bug in the original code which checked “tag” instead of “getName” but probably wasn’t that important.
To be clear - I don’t think I actually added any functionality in loadGPXFile - I just cleaned it up.  So, leave this change OUT.
Cheers
A

vshcherb:
If you have time can you please make it. In case to do it myself I need first accept pull-request or do it in a different branch.

vshcherb:
Thanks for your pull-request it looks attractive change, we are compiling apk to test how it looks

andrew-davie:
OK.  I will do it now.

andrew-davie:
I am new to "git". I hope I have done the right thing to help you review my code.
I modified loadGPX back to original version, and done a new add/commit/push/"pull request".
Let me know if this is OK.

xmd5a2:
I have tested it. Mostly I like it but 4) needs tweaking. At high zooms (16-19) GPX width is unnecessary bold.
https://dl.dropboxusercontent.com/u/4624786/osm/Screenshots/Screenshot_2016-03-20-20-55-22.png
Please limit width in these conditions to default value. I think GPX width change depending of zoom should be nonlinear - normal at high and middle zooms, thin at low zooms.

xmd5a2:
And it seems that you broke manual GPX width setting here
https://github.com/osmandapp/OsmAnd-resources/blob/master/rendering_styles/UniRS.render.xml#L35
and here
https://github.com/osmandapp/OsmAnd-resources/blob/master/rendering_styles/default.render.xml#L177
We need a way to manually control GPX width and this should be somehow solved before merging because this is a regression.

andrew-davie:
OK. Original code had an array with the magnifications for each zoom level, but I went the formula way as it was more elegant.
I will revert to a hand-crafted set of zooms to fix the width.  I thought I had it so it was reasonably closely mirroring the road widths up to about 22.
I will post a revised version in about a day.
Cheers
A

andrew-davie:
OK, I’ll look into this.   There are effectively TWO “manually controlled GPX widths” now.
1 is the original system, but 2 is now inside the .gpx file.
I think if I just multiply them, with a bit of mangling, that may work.
The issue is that even with manually controlled widths, the relative zooming specified by the .gpx mods should also be apparent.
That’s my thinking - I’ll try it anyway and see how it looks.
Cheers
A

andrew-davie:
I am still learning git and github.  I *think* you already have access/notifications of my recent upload that consists of the work requested below.
I think it was a positive change, and zooming looks good to me. I also implemented 0-size tracks, which means that at low zoom levels, GPX are not displayed.
Easy to adjust/configure via the table giving zoom values - well enough commented that anyone can adjust.
Please let me know if the revision is suitable. I will address the other issues shortly.

vshcherb:
Yes we see all the changes immediately. Looks good and good that you documented 3 TODO

xmd5a2:
I added "4) Scaling based on view zoom" functionality to rendering styles: https://github.com/osmandapp/OsmAnd-resources/commit/1e6d07847a903e62038393bf5e3a66dc628dbc8e
Please modify your pull request to allow GPX scaling only with track width inside them. (<zoom>1.0</zoom>)

andrew-davie:
I think I understand this request.
You’re wanting NO scaling to happen UNLESS the GPX file specifies zooming on a segment.
You’re going to implement a zoom-based scale via UI configuration. That’s completely bypassing the whole auto-scale thing.
I think that’s a wrong call, but it’s your baby. I’ll kill the pull request.

xmd5a2:
No, I don't want "a zoom-based scale via UI configuration". I want to select GPX width that will be visible at 12+ zooms and which gets thicker when lowering zoom (3-11). It will be used as default width and you can modify it using 'zoom' modificators inside GPX. If we put scale-zoom modifications to java code it will not possible to users to change it.
But your pull request not only related to that and I think you should reopen it. I think we can merge at least 1) and especially 3).

vshcherb:
Hi Andrew,

Could you please explain your point it is not very clear what do you mean by "auto-scale". In your code there is a big map of 2ˆ(18-zoom) manually configured coefficients and what kind of "auto-scale" you expect?

Could you also explain "zoom-based scale via UI configuration" because I'm not sure that @xmd5a2 intention.

andrew-davie:
There’s a 16 element array (one double for each zoom level) - that’s not very big!  :)
You’re moving that from “hardwired” into the UI - that’s what I think you’re saying. Pretty sure.
In my code now, I use the zoom level (say, 0-31) to look up the zoom coefficient in that array. That’s all.
You’re replacing a single brush width (as it exists now) with multiple, based on zoom.
I used the single brush width (as it exists now) and modified it based on zoom.

You are basically replacing the last line above, and putting the zoom table (16 entry) into the UI.
As I said, I understand.

vshcherb:
There is a conceptual difference what we call UI. 
For me UI is a dialog "Configure Map" where you can select options as a User! Changing rendering style is not a UI, it is application config file which you can change as a Rendering style editor. If we agree on what we call "UI" it will be easier to express what are the pros & cons.

andrew-davie:
OK. we agree.  How you actually change the values is unimportant. I thought via UI in Osmand was your suggestion.  Via UI in a style editor is similar too.
The point is, the values are set by the user, or set for the user by a rendering style XML definition. The user can select the style, or copy the style file… whatever.
Basically the zoom multipliers for each level are thus configurable to/for the user.  I’m sure we are talking the same thing, if not the same terms.

I am not arguing for my version to be kept. I have killed the pull request already.

vshcherb:
It still varies who you call a User but essentially we can call him Advanced User. I think we have common understanding and my point of view that Advanced User should be able to change rendering style and displaying elements on the map exactly how he wants (may be not in the program itself). 
My main argument it is very friendly for end user and for customizations and for custom rendering style which we developed and would like to maintain. It works great for colors, for widths and elements I would say it works reasonable.

If there is no contr-argument except for developers it is easier to keep numbers in the code, I don't think we should make exception how do we determine GPX. 

I just found in code there is a special setting for GPX zoom, but  I put that Q in a different thread, so we can continue our discussion there

PR #2360, opened 2016-03-23
GPX redux

andrew-davie:
Still learning git/github - I can't see how to re-open my previous request.
We do seem to have a bit of a language/terminology/understanding issue.

What I've done here is honour all the requests you have made, as far as I understand them.

Specifically, the custom .gpx "zoom" is honoured with auto-scaling - those tracks which specify "zoom" will automatically become big/small as you zoom in and out. All other tracks do not - the track width is left to whatever the rest of the code has set it to (ie., either a single width for all zooms, or different width for each zoom - via an XML description).

However, there is a switch to enable auto-scaling on all tracks. This would be useful, for example, in the case where there is just the one width set for all zoom levels and you WANT it to scale. It's useful to ME because I can continue testing/reviewing the GPX track stuff. It defaults to OFF.

Please note the comment about why scaling doesn't work when you change track width - the hash on updatePaints preventing that code running more than once, and thus preventing knowledge about the width change getting to my code. Point is moot, really - as you will not be running my code in that situation, anyway (well, unless you turn on the autoscaling). But that's why it didn't seem to work when you tested.

For the record, I'm planning to do some experiments with colouring of tracks
 - bringing in elevation data from the GPX
 - bringing in time data 
 - using these to indicate altitude/speed via colour changes


xmd5a2:
Looks like you have broken GPX simplification.
https://dl.dropboxusercontent.com/u/4624786/osm/Screenshots/Screenshot_2016-03-23-21-29-02.png
This is how this track should look like:
https://dl.dropboxusercontent.com/u/4624786/osm/Screenshots/Screenshot_2016-03-23-21-35-25.png

andrew-davie:
I disagree!
If you choose points evenly spaced every 150 m along that track, that is EXACTLY what it should look like.
The paint is, the algorithm is choosing points, not creating a new track
In actual usage, you would resample at a higher point rate than the track - say, every 1m.
The sample was just testing that the code works. I think it does.
Try changing it to 10m instead of 160m.  It’s a generic routine for track resampling - not for display (necessarily) but for effects/other use.
I’m sure it works - you need to understand what you’re looking at.
Wait till I get some animation going, or put some icons there on the original track at those resample points, and you’ll see :)
Cheers
A

andrew-davie:
Maybe you missed the comment on check-in…

"I have tested this - you can see the test code still in there where the culledPoints list is generated. I'm displaying the resampled track instead of the culled track, and it works just fine.
“
This is not code to replace displaying the track!  The test call should be removed. It was just showing that the subroutine is correctly choosing a sparse number of points evenly spaced along the original track.  The next step is to use that for other things such as colour cycling, or speed colouring, or altitude colouring, etc.  As I said, watch what comes next.

vshcherb:
Hi Andrew, 
I definitely see the progress you make, congratulations. Though I'm concerned that not everybody can follow your pace, me for instance. I know it sounds hard but merging is always quite a big process and bigger the merge gets harder is to merge. 

For example, you've started with really nice set of missing features. Let's take Douglas-Peucker algorithm for tracks, it would worth to merge just this part and avoid other discussions about width of track and everything else. As I understand today we are not discussing should we change the width of track automatically, we are discussing where that code should be which is perfectly fine, but I don't think it should stop merging parts which we agreed on. 

I know that's not the easiest to work on multiple branches, but that's a nice way to get something merged quicker. You don't have to please one though we need to find a common understanding there are multiple ways how to do stuff together, we just need to pay enough attention to each other. 

Victor

vshcherb:
Q: Found in the code some comments (IFF the GPX specifies a zoom (i.e., getGpxZoom != 1.0))
Could you please explain why GPX should specify a zoom, I can't recall any GPX with zoom and I barely understand what is the intent to specify a zoom

andrew-davie:
Sure. We have discussed this before.
If the .gpx has an extensions grouping in the top level, or in a trk or a trkseg ** as already allowed by the original code**
THEN, I implemented a “zoom” key/value pair in the extensions to allow the specification of zooming for that gpx or track or segment
also, I used “color” key/value (which already existed, as I recall) for the same, if present. I refer to the GPX documentation for extensions where it says "You can add extend GPX by adding your own elements from another schema here.” This lets GPX files be edited to highlight or colour individual segments. If I want to mark a track as narrow, or in some other way then I can edit the extensions tag in the .gpx. The comment is referring to this capability.
The code handles both zoom and color on a per-segment basis, if specified in the gpx file. 
As to use-case - well there could be any number of uses. Marking different difficulty of riding along a track - via width, or colour. Marking the actual width of the track(!!) for example like with roads where we have wide roads and narrow roads, likewise gpx can specify the relative width of the track itself.
Look to be honest, you guys don’t seem too open to these things and that’s OK. I think my code and ideas are sound, and the ideas are interesting. I’ll just keep pottering away and doing my stuff - it works for me and I’m happy to have my own version of OsmAnd.  I’ll keep pointing you to the improvements/changes I make and you can review them as you wish. However, I do feel that you’re not really paying much attention to what I’m actually trying to do. Perhaps it’s a language issue - in any case as I said it’s your baby. I’ll be happy enough to just point you to videos in future to show you what these changes are for, because it doesn’t seem you’re understanding what I’m trying to do from the source code alone.

andrew-davie:
Re: the below…
I think the best way is for me to just do my experiments and program what I think and see how it works. That will make me happy - I enjoy this sort of work/play.
I will show you things that I think are useful or good, when they are finished, and you decide if they fit with what you want. If they are useful but need to be modified slightly to fit with how you need things to work, that’s OK - I’m happy to do that. I am always strong in expressing my opinion, but always if I say “I agree” then that’s the end of it. I’m just trying to make improvements - if my stuff isn’t used, that’s perfectly OK too.
The DPA algorithm is good, and nicely written :)  but I suspect it might eventually be discarded as the resampling offers better possibilities.
First thing, though, I’ll get it working asynchronously so there is no impact on the UI speed.
Cheers
A

andrew-davie:
I am going to break down these new features into smaller and easy to install updates via new pull requests.

vshcherb:
Really nice to hear from you, Andrew, smaller pull-requests easier to review and adopt. Unfortunately nowadays very busy with 2.3 release, reviewing, testing all the bug reports people send around. Hope to come back to your many pull-requests soon.

PR #2379, opened 2016-03-28
New GPX track rendering and animation/overlay package.

andrew-davie:
WHAT THIS PROVIDES:

A generic rendering/draw of GPX tracks that allows multiple 'renders' to be included as a part of the draw of any TrkSegment. These 'renders' include the basic track drawing, but also other options such as dashed lines, rainbow-coloured altitude indication, speed indication, arrows to point direction, animating 'lines' showing direction of movement, and dots/text showing 1km (or any other) distances along tracks.  It's very extendable and incredibly cheap in terms of processing speed.

For a view of some of these operating on this version of the code, see [https://youtu.be/aRGCNLmBAlk] this video.

`Additionally,` all of the above have automatic asynchronous track resampling - either via line culling of
Ramer-Douglas-Peucer (implemented for the base track draw at different zooms), or an actual resampler that takes a distance and steps off and creates a new track with points spaced exactly that distance apart along the original track. The asynchronous resampling/culling willl automatically enable the new (optimal) track display when the background task has completed.

This is completely up to date with the master branch as of an hour or so ago!

Two modified files
 - GPXUtilities
   Added some fields to WptPt to enable distance measurement on tracks and colouring for altitude/speed
 - GPXLayer
   Installed the new track rendering with examples (commented/out)

Two new files
 - AsynchronousResampler.java
    Implements line resampling and culling asynchrnonously for all line drawing
 - Renderable.java
    Set of classes for drawing different kinds of gpx 'renders'
     - normal with automatic Ramer-Douglas-Peucer line culling
     - conveyor-belt type animation of segments on a path
     - altitude colouring of a path
     - speed colouring of a path
     - distance based waypoint/marker drawing


vshcherb:
Just watched video. Lots of features are really cool though I'm not a fan of lots of small ants moving on the screen :) I think it should be optional and by default disabled (moving animation).

OsmAnd/src/net/osmand/plus/GPXUtilities.java
```diff
@@ -181,6 +207,57 @@ public boolean equals(Object obj) {
...
+		}
+
+
+		public void recalculateRenderScales(OsmandMapTileView view) {
+			for (Renderable.RenderableSegment rs : renders)
+				rs.recalculateRenderScale(view);
+		}
+
+		public void drawRenderers(Paint p, Canvas c, RotatedTileBox tb) {
+			for (Renderable.RenderableSegment rs : renders)
```
vshcherb: Just a small remark. Is it possible to put { braces around for, I'm trying to ask people follow java guidelines so it easier for others to follow if it is consistent everywhere.

OsmAnd/src/net/osmand/plus/views/Renderable.java
```diff
@@ -0,0 +1,480 @@
...
+                    if (isIn(x, y, w, h)) {
+
+                        px.setColor(0xFF000000);
+                        px.setStrokeWidth(ds + 2);
+                        canvas.drawPoint(x, y, px);
+                        px.setStrokeWidth(ds);
+                        px.setColor(0xFFFFFFFF);
+                        canvas.drawPoint(x, y, px);
+
+                        //TODO: I do not know how to correctly handle screen density!
```
vshcherb: There is method view.getScaleCoefficient it is better to use it because it is also takes into account tablets (we enlarge for tablets 1.5 times everything what's visible on the map screen)

OsmAnd/src/net/osmand/plus/views/Renderable.java
```diff
@@ -0,0 +1,480 @@
...
+                        //TODO: modify the text size based on density calculations!!
+
+                        if (sw > 6) {
+                            px.setColor(Color.BLACK);
+                            px.setStrokeWidth(1);
+                            px.setTextSize(sw*2f); //<<< TODO fix
```
vshcherb: Yes please fix it :)

OsmAnd/src/net/osmand/plus/views/Renderable.java
```diff
@@ -0,0 +1,480 @@
...
+            TIntArrayList ty = new TIntArrayList();
+            canvas.rotate(-tb.getRotate(), tb.getCenterPixelX(), tb.getCenterPixelY());
+            Path path = new Path();
+            for (int i = startIndex; i <= endIndex; i++) {
+                GPXUtilities.WptPt p = pts.get(i);
+                tx.add((int)(tb.getPixXFromLatLon(p.lat, p.lon) + 0.5));
+                ty.add((int)(tb.getPixYFromLatLon(p.lat, p.lon) + 0.5));
+            }
+
+//TODO: colour
```
vshcherb: Not clear what that todo means...

OsmAnd/src/net/osmand/plus/views/AsynchronousResampler.java
```diff
@@ -0,0 +1,282 @@
...
+        this.view = view;
+        this.rs = rs;
+        this.renderType = rt;
+
+        this.param1 = param1;
+        this.param2 = param2;
+    }
+
+    public int getColor2(double percent) {
+
```
vshcherb: It is better to have that piece of code somewhere in Algorithms class with proper explanation what is doing it looks very generic.

OsmAnd/src/net/osmand/plus/views/GPXLayer.java
```diff
@@ -326,10 +327,57 @@ private void drawSelectedFilesPoints(Canvas canvas, RotatedTileBox tileBox, List
...
 		for (SelectedGpxFile g : selectedGPXFiles) {
-			List<TrkSegment> points = g.getPointsToDisplay();
-			boolean routePoints = g.isRoutePoints();
-			updatePaints(g.getColor(), routePoints, g.isShowCurrentTrack(), settings, tileBox);
-			drawSegments(canvas, tileBox, points);
+			List<TrkSegment> segments = g.getPointsToDisplay();
+			for (TrkSegment ts : segments) {
+
+				// TODO: Select/install renderables via UI or external process!  (See comments below)
+
```
vshcherb: This is quite a long text and doesn't look to be at the right place in sense I see the mixture of what is going on in development how it could work or how it should not work. It might not be the right place to have here.
I think if somebody will need to dive to that code he just need to understand exactly what's going on he doesn't need to understand the plans of future implementation. 
Personally I'm against comments because I think it is always possible to refactor the code that it doesn't have comments. For example, TrkSegment has one or more 'Renderables' - it could be clear from the code itself (1 to many relations). What is Renderable is much better to describe in the Renderable class header itself. Second part Renderables are processed asynchronously is kind of obvious because we see AsyncTask.
"Note" is a good to know for us in pull request but not in the code may be.

OsmAnd/src/net/osmand/plus/views/GPXLayer.java
```diff
@@ -326,10 +327,57 @@ private void drawSelectedFilesPoints(Canvas canvas, RotatedTileBox tileBox, List
...
+					// type of system. Please leave the code alone, and I will implement the arrows after this code is approved!
+
+					// You can, of course, comment out the following...
+
+					ts.addRenderable(view, Renderable.RenderType.CONVEYOR, 10, 250);        	// conveyor belt animation (m,refresh(ms))
+					//ts.addRenderable(view, Renderable.RenderType.ALTITUDE, 25, 128);			// an altitude display (m,alpha) << IMPORTANT: See [note 1] below
+					//ts.addRenderable(view, Renderable.RenderType.DISTANCE, 1000, 1);        	// 1km markings (m,size)
+					//ts.addRenderable(view, Renderable.RenderType.SPEED, 20, 255);        		// a speed display (m,alpha) << IMPORTANT: See [Note 1]
+
+					// [Note 1]: The altitude and speed renders (only if enabled, above) have a bug that can crash OsmAnd
```
vshcherb: It requires just more testing but it is fine of course would be good to see exception stack trace, I don't think emulator has something very different.

OsmAnd/src/net/osmand/plus/views/Renderable.java
```diff
@@ -0,0 +1,480 @@
...
+                            px.setStrokeWidth(1);
+                            px.setTextSize(sw*2f); //<<< TODO fix
+                            canvas.drawText(getLabel(pt.getDistance()), x + ds / 2, y + ds / 2, px);
+                        }
+                    }
+                    canvas.rotate(tileBox.getRotate(), tileBox.getCenterPixelX(), tileBox.getCenterPixelY());
+                }
+            } catch (Exception e) {
+                String exception = e.getMessage();
+                Throwable cause = e.getCause();
```
vshcherb: Looks bad, please delete catch exception.

OsmAnd/src/net/osmand/plus/views/AsynchronousResampler.java
```diff
@@ -0,0 +1,282 @@
...
+        GPXUtilities.WptPt lastPt = track.get(0);
+        lastPt.speed = 0;
+
+        // calculate speeds
+        for (int i=1; i<track.size(); i++) {
+            GPXUtilities.WptPt pt = track.get(i);
+            double delta = pt.time - lastPt.time;
+            if (delta>0)
+                pt.speed = MapUtils.getDistance(pt.getLatitude(),pt.getLongitude(),
```
vshcherb: As pointed earlier please try to put curve braces around if/else/for. Thanks in advance.

vshcherb:
Hi Andrew,

The code looks very good, I think if we don't give up 2-3 iterations it will be perfect to merge. And we need to start may be more complex conversation how do we see UI to enable options. For example, to enable altitude rendering, speed rendering, to enable labels (time/distance split), to enable arrows on the track (though you replaced with animation) but I think it is a different feature.

andrew-davie:
I don’t like ants either.  The difficulty for me is to demonstrate what the code does and make sure you see it.
For the final version, no — no ants.  But the fact that the track is resampled (asynchronously so the user sees no lag) is important for future capability.
By default it should not be there at all - agreed. It is easy to add/remove - we will discuss its use later.

OsmAnd/src/net/osmand/plus/GPXUtilities.java
andrew-davie: Yes, I am happy to adhere to this coding convention.
I will make the changes and re-upload.
I have also made major changes to more object-oriented in the rendering and resampling.  It is much cleaner - no “type” being passed around anymore.  So it’s good for me to have a chance to update before you seriously consider including my code.

OsmAnd/src/net/osmand/plus/views/Renderable.java
andrew-davie: Great - I will look at this.  Some things are impossible for me to find easily in such a large code base. Leave the fix for me to do.

OsmAnd/src/net/osmand/plus/views/Renderable.java
andrew-davie: No, it isn’t :)
OK, the code there does a black border around tracks (another option). I think you may call this “shadow” but I’m not really sure - so it was just a note to indicate that that’s where the shadows go, and this should somehow be selectable via UI or variable.  This is something we need to resolve later.

OsmAnd/src/net/osmand/plus/views/AsynchronousResampler.java
andrew-davie: OK, I will do that. I have already rewritten it to look nice.
Essentially it will, given an input between 0 and 1 (double), return a colour of the rainbow from purple to red.
Great for colour-coding a range of values such as altitude or speed.

OsmAnd/src/net/osmand/plus/views/GPXLayer.java
andrew-davie: Yes, too long for final code. I just wanted to make SURE that you gent understood the capability and why.
I wanted you to experiment with adding removing renderable when you viewed/tested so that you could see how it works.
Definitely not to be in the included code.

And YES!!!  It is in the wrong place, You have perfectly understood.
There will be some way for the user via interface or some render XML to decide which ‘renderables’ they want to see.
It might be the altitude band, or the arrows, or the ants… whatever.  It needs to b added to the TrkSegment list somewhere elsewhere.
But for TESTING, and demonstration, without UI … this was a perfect place for me to put it.

MOST of these comments are for you to accept/understand my submission. The comments like this one would NOT be required in the final code.
As to putting them in the pull request, yes OK - but sometimes there are many different areas with such a big pull request, and it’s easier for me to base the notes with the part of the code like this example. It will be a bit of a balance, but I will try to adjust.

I do disagree with your “I’m against comments” philosophy.  If you understood how many HOURS I have spent trying to follow and understand code to find something I need, where simple comments would have made that process much easier. I just cannot agree that comments are a bad thing.

OsmAnd/src/net/osmand/plus/views/GPXLayer.java
andrew-davie: The try/catch exception I put in NEVER CATCHES ANYTHING.
However, on the emulator the render of altitude/speed sometimes crash. I am now certain this is an emulator GPU bug.
The emulator documentation specifically states that this (GPU host) is experimental and not to be trusted.
The code has never crashed on my actual phone.  Next time I see a crash on the emulator I will press the “send to developer” button?
Because, when running under Android Studio - there is no stack trace or exception - the emulator just crashes OsmAnd, and that’s the end of it.
Nothing in the output/log window, no exceptions, no stack trace - impossible to diagnose. I just know that it’s to do with the drawing of altitudes/speeds.
And that drawing is GPU intensive with alpha blending of a canvas onto another canvas and lots of alpha thick line draws. So.

OsmAnd/src/net/osmand/plus/views/Renderable.java
andrew-davie: This was my attempts to find the GPU emulator bug previously mentioned.
This catch never got executed, and simply shows that its not a catchable exception that’s the problem.
I will take it out. it is doing nothing anyway.

andrew-davie:
Thank you - I understand you are busy with a pending release.  I don’t quite know how to fix the problem that I am working on LOTS of stuff and GitHub does not allow more than one fork.  So, if you leave my code for later, there may be a lot more changes elsewhere to discuss/merge, too.

I agree with you about UI discussions. The point of my work was to offer options and then to discuss if or how they would be useful to a user.  The arrows, for example, they do look good, but perhaps they’re only going to be useful for the routing - an area I would like to look at because the track resampling makes the display of arrows on the roadway MUCH easier.

One other thing - the current code ENABLES ALL GPX TRACKS when you click to select a track to display.  That is, ALL tracks have a tick.  This is a major pain when you have dozens of tracks (as I do, for testing).  I have to go through and un-tick every one, every single time.  I think/hope it was a debugging feature someone put in, because it’s awful as a final version :)

Now I have covered all your comments, I will make some changes and upload revised version.

Phyxion:
Nice work andrew, I personally don't like the moving dots/bars on screen but I can see it's a lot faster than the current implementation :+1:

vshcherb:
Hi Andrew, you don't need to create new fork. You can create new branch! And you can create pull-requests from any branch!
Also if you continue changing master branch the pull-request will be updated automatically.

vshcherb:
>> One other thing - the current code ENABLES ALL GPX TRACKS 
That's bug you can may be open a github issue so we don't forget to fix it.

andrew-davie:
Got it, thanks!  I’ll learn how branches work. That will save me a lot of mucking about.

andrew-davie:
Thanks!  The moving dots/bars are really just utility capability. The bar concept is extended to draw arrows as shown on some of the videos I have been uploading. 
For example, in the routing now it can be used by

a) creating a TrkSegment with the route points
b) adding a renderer to draw arrows

That’s all you have to do!
Of course the arrows are prototypes. I will make them really pretty and configurable soon.
The real benefit is that it is extremely cheap to use in terms of code and processing time.
You can draw the arrows/dots as a one-off, so they are fixed in position.
Or, you can have animation with them moving along the track.

The whole point being - it was never the intent to have the dots moving like that, unless the user chooses it, or unless it is required for functionality.
I’m just adding capability.

andrew-davie:
ttps://youtu.be/2kTnU-4-rhU

Phyxion:
Is it possible to disable the coloring, arrows, etc? I'm personally not that interested in it. Also, I think the coloring looks ok in the app but the style of the arrows/etc doesn't really match the other design .

andrew-davie:
Yes, of course. All configurable. I was just showing the bells and whistles. As always with these things, try not to look at the specific implementation of colouring or arrow style - that’s window dressing. It’s the *capability* that is important. These options open up opportunity for much improved route following display, for example - think arrows exactly placed on the track/road you’re following, for example. It’s very modular code - you simply add a new renderer if/as you want it. A renderer being stuff like arrows, distance markings, conveyor belt, speed colour display, altitude colour display. How those displays are actually DONE (visually) that can be changed. But the basic system has the capability. So, that’s the important bit.

The code resamples the tracks, producing much smaller tracks in terms of point count without visible differences. This reduces the point count to a minimum and so increases rendering speed to a maximum.
The draw has many optimisations including…
(1) a fast “no draw” check if none of the track would be on the screen
 (2) a super-fast “sub-track” rendering capability which divides large tracks into many small tracks. Thus, to draw a track you can perform (1) on each sub-track. This saves iterating through (say) 100,000 points to draw just a small number of them on the screen. Instead you iterate though (say) 100 sub-tracks and only draw 1 of them on the screen (after fast-culling the 99 others). It’s all automatic - user/programmer does nothing, it just all happens.
There’s lots of other stuff going on like cached resampled line management to reduce memory usage.  That is, resampled tracks are discarded if not viewed for 10 seconds or so. And if they’re needed again then a background asynchronous task regenerates them.

Lots of stuff going on in the background. The “style” is neither here nor there - that’s easy to change. What wasn’t easy was making it possible.

andrew-davie:
Some code might make my answer clearer…

if (g.isShowCurrentTrack()) {
   ts.renders.add(new Renderable.CurrentTrack(ts.points));
} else {
   ts.renders.add(new Renderable.Altitude(ts.points, 10));
   ts.renders.add(new Renderable.StandardTrack(ts.points, 17.2));
   ts.renders.add(new Renderable.Conveyor(ts.points, view, 5, 250));
   ts.renders.add(new Renderable.Arrows(ts.points, view, 10, 250));
   ts.renders.add(new Renderable.DistanceMarker(ts.points, 1000));
}


If you don’t want arrows, comment out the Arrows line.
If you don’t want the conveyor belt, comment out the conveyor belt…. etc.
The parameters on the end there control various stuff like spacing, animation speed, etc.
It’s all highly configurable.

Selection of these options (which things to show) would ideally be placed in the GUI when you select a GPX track you could also select things like colour, width, and accessory displays like altitude, speed, distance markers - all of which I think would be super-useful. The arrows less so, and the conveyor not at all.  But the arrow code, at least, will live on in the route display if I have my way with that ;)

Phyxion:
> Selection of these options (which things to show) would ideally be placed in the GUI.

Yes, that's what I meant indeed, perhaps somewhere in the options. The performance improvements are awesome as it was sometimes very slow.

PR #2413, opened 2016-04-03
GPX asynchronous line drawing - super quick MINIMAL code set

andrew-davie:
Here is a minimal subset of the GPX code I've been working on. It's setup so that newer stuff (distance markers, rainbow altitude, speed, etc.) can later be 'plugged in' as new classes, inheriting all the good stuff from this minimal set.  The set includes asynchronous line culling, and supports 'current Track' (which should not be culled). 

Features are basically asynchronous line reduction to minimise points and improve visuals at lower zooms, and very very very quick screen drawing of tracks. Did I say very quick?

I'm pretty much done. I don't expect to touch this unless there are bugs. 

Edit:  Oh, 250 lines of code including whitespace!  Amazing.


vshcherb:
Hi here is the build http://download.osmand.net/latest-night-build/OsmAnd-design.apk. 
I think the code is good, so we started manual testing. So far we were quite surprised but we didn't see much difference yet speed wise. @xmd5a2 could provide some screenshots. Basically we enabled Test Rendering checkbox and FPS widget (2nd number if fps related to tracks)

I know that code is better structured and ready for next improvements, but I just want to be clear if we aimed here for faster rendering than we should be able to test it.

Great work!

andrew-davie:
Try it with a GPX track of 100,000 points and see how the frame rate differs. I expect it to improve, but proof is in the testing.
Try it with many GPX tracks, say 250,000 points….  I think the old one would slow down and the new one won’t.
Cheers
A

vshcherb:
Yes, good point. I think small tracks (points < 100-500) won't give any difference.

andrew-davie:
For smaller tracks, it will still be many times faster, but if the rendering took just 0.1s anyway then who’s going to notice the difference :) 
However, now the user can load their entire library of (say) 100 tracks - memory permitting - and see them all displayed at once.
Actually I’d be really interested in figures of the speedup. I’m going to be brave and say 100x faster if we test with 100,000 points.  :)  !

andrew-davie:
ALSO, isn’t the frame rate limited by the UI code?
That is, If you are limiting the refresh rate to 5Hz or whatever, then my code isn’t going to run faster than 5 Hz if timed :)
You really need a start/stop time  in the code bracketing the draw for a fair comparison!

xmd5a2:
I have tested this with 108mb of 241 tracks. Release version loads them with no problem and renders them in equal conditions slightly faster then with this pull request (especially faster when changing zoom levels).
Here is average rendering time changing zoom from 10 to 11 (rendering settings are equal):
version with this pull request: https://dl.dropboxusercontent.com/u/4624786/osm/Screenshots/Screenshot_2016-04-04-15-20-18.png
release version: https://dl.dropboxusercontent.com/u/4624786/osm/Screenshots/Screenshot_2016-04-04-15-19-47.png

andrew-davie:
Not quite clear what you’re testing here, but to be clear on my thoughts…

1) consider zoom lowest (say, 6 or so) then my system has resampled the track and is rendering perhaps 50 points total.
   - original system is still rendering 70,000 points.
  - you say original is slightly faster. Bizarre.

2) New system calls an asynchronous re-sample of all tracks that are visible when zoom level changes.
 - after that is it significantly quicker.  If you understood what’s happening this would not be in dispute.
- I have put in many many optimisations.  I pre-cull the entire list before drawing. I resize the points to significantly smaller when zooming.
I have an extremely rapid screen-based cull BEFORE doing any lat/lon to X/Y calculations. It’s really really really quick compared to the old version.

I just can’t see how you can actually time this and have it come out slower.
I know exactly what code I removed, and that it is now rendering a very small fraction of the number of points.

So, what’s happening here?

andrew-davie:
These images do not look like GPX tracks to me… ?  What are all those big red dots???

xmd5a2:
It is wpt. Ok, I see difference in FPS when testing 38000 points track at zoom 6.

andrew-davie:
Ah. Well, WPTs are not tracks, are they :) None of the optimisations apply to WPT drawing at all; are they even drawn by my code?  And in any case the time difference then becomes just a measure of the difference in drawing big red dots which are pretty much twice the size between the two versions you posted images of.
I would be very interested, please, if you could forward test results on the differences in time between versions actually rendering large tracks.

xmd5a2:
WPT icons are the same, this screenshot was done during rendering process and shows rescaled image.
I don't have tracks with more than 38000 points.

vshcherb:
I think we've found cases were it gets extremely fast though Andrew you could agree that there are some spaces to improve. For example huge tracks on zoom 17-18 there is no pretty much difference even other way around it will be a bit slower in the first seconds because of presampling algorithm. The idea that cycle over 38000 points is not an efficient way. In that case using quad-tree/r-tree to find which subsegment should be displayed is more efficient. On the other hand I must say we should not really invest much in that because these are very specific cases.

vshcherb:
Let's move forward and try to integrate altitude/speed display.

andrew-davie:
I don’t know how to measure the time here.
However, the “resampling” - not quite resampling but let’s call it that - is *asynchronous*.  It effectively costs zero time - there should be no UI speed issue caused by that. Huge tracks - let’s say they are not “resampled” I still expect there to be a HUGE difference in rendering speed because
a) I pre-cull each point using a significantly quicker piece of code than is used in the old version.  Could we please have some actual times to discuss.
38000 points is fine.  Iterating/cycling over 38000 points is not ideal, no - but my code only does this if a part of the track is onscreen (or potentially). It pre-culls entire tracks based on bounding box. So load a number of GPX tracks zoom in a bit to show one or two, my system will only iterate/cycle over the potentially visible ones. Original system will iterate/cycle over them ALL.  That’s one optimisation.  In the new system even when iterating/cycling over 38000 points it is probably 10x quicker or more than the original system. My draw cull is a simple comparison on the lat/lon values - the original system had to convert every single point into a X/Y coordinate which was quite significant. You re right that encoding the track as a tree would improve efficiency even further. But I don’t feel you yet have a clear handle on how much quicker the new code is, yet. First you were telling me it was slower!!! I would really like to see some actual and fair comparisons.
Just load a large plain vanilla GPX track and do some timing on that.

andrew-davie:
These (altitude/speed) are pretty much ready to go.  I should pull the latest codebase, then insert speed/altitude and do a new pull request?
It’s pretty much just adding two new classes to the Renderers, a new resampler, and working on the GUI to select to display speed/altitude.
I don’t know how to do the GUI part, sorry - but I suggest that when you click on a GPX track name you open a settings dialog for it
and in that you can select track colour, width multipler, altitude renderer, speed renderer (probably mutually exclusive as they use the same rainbow effect).
If you want them both at the same time, then we have to think about how to display - perhaps a band-within-a-band (outer being speed, inner being altitude, and inside that the actual track).

PR #2446, opened 2016-04-09
GPX tracks - Altitude and speed and distance along track

andrew-davie:
All these are optimised with asynchronous line culling to reduce the number of points to display.
Also handles 'current track' with addition of points being gracefully catered and resampled in the background. Lots of options;  band width for speed and altitude, and the distance renderer has inbuilt units conversion/display for metres, kilometres, miles, yards and feet.  Display of markers is limited to sensible zoom ranges where they will not clutter the display.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/GPXUtilities.java
+++ b/OsmAnd/src/net/osmand/plus/GPXUtilities.java
@@ -1036,17 +1036,17 @@ public class GPXUtilities {
 							if (parser.getName().equals("gpx")) {
 								((GPXFile) parse).author = parser.getAttributeValue("", "creator");
 							}
-							if (parser.getName().equals("trk")) {
+							else if (parser.getName().equals("trk")) {
 								Track track = new Track();
 								((GPXFile) parse).tracks.add(track);
 								parserState.push(track);
 							}
-							if (parser.getName().equals("rte")) {
+							else if (parser.getName().equals("rte")) {
 								Route route = new Route();
 								((GPXFile) parse).routes.add(route);
 								parserState.push(route);
 							}
-							if (parser.getName().equals("wpt")) {
+							else if (parser.getName().equals("wpt")) {
 								WptPt wptPt = parseWptAttributes(parser);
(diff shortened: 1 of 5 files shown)


vshcherb:
Sorry coming later to this request as expected. So, I have some questions related to it.
1) What it is exactly about (altitude, speed - distance), from the code I don't see how it will be configurable (color, style, gradient...)
2) Why did you refactor so much from the code you wrote first time? This refactoring doesn't look as nice as your previous one.

andrew-davie:
Hi. Thanks for getting to the code to review.

1) Altitude/speed - these auto-configure by scanning the entire track to determine the range of values, then map that range onto the entire rainbow.  If you think carefully about any other way to do it, then you have a problem in that you need a minimum and maximum speed/altitude to represent the extremes of the rainbow colours available. Some people may have a track at sea level with a maximum height of (say) 100m… so it would pretty much be totally the same colour (purple) if you allowed the rainbow range to specify all altitudes available - for example it might have to represent up to 2.5 km maximum -because some people will be hiking in mountains. For any particular track, the rainbow is best to represent the variation in altitude over that track - and match the lowest colour to the lowest point on THAT track, and the highest colour to the highest colour on THAT track. Otherwise, you will have most tracks with almost no colour variation, because they would only use a very small sub-range of the rainbow.  To emphasise - if a track has 100 m variation in altitude but the rainbow needs to cater for 3 km of variation, then the colours on that track are going to be pretty much the same.  But if the rainbow for THAT track also has to only cater for that same 100 m variation, then the track is nicely coloured. I thought a long time about this, and it seems clear that a rainbow range specific to each track is the way to go.

And so, it auto-configures, as noted.  Same for speed.  It configures to the rainbow palette.  Currently there is no support/options for any other palette style, but if you would like to suggest something I would be happy to implement.

2) Code refactor - it is MUCH more object oriented and re-uses shared functionality wherever possible. I disagree with your comment that it does not “look as nice” but perhaps I am not looking in the same places as you. Could you please give an example of where it is worse, and I will be happy to explain the changes and why.

Cheers
A

PR #2476, opened 2016-04-18
Add erase points button

sdesimeur:
When there is a lot of intermediates points, it take a long time to erase all these. So the added button give the ability to erase all the points in one time.


vshcherb:
Can you please provide screenshots how does it look

sdesimeur:
In OsmAnd/src/net/osmand/plus/activities/MapActivityActions.java
When I stop "navigation", the waypoint stay in.
I add the ability to erase waypoints in the route without erase markers.
You can see the new button (second button at the bottom) in the screenshot :
screenshot_2016-04-27-10-14-54

vshcherb:
Unfortunately in some languages that wouldn't work (Go button) and also there was clear intention to make it bigger. I will forward that pull-request to our designer so he could come up with alternative solutions.

PR #2477, opened 2016-04-19
Route display - alternate

andrew-davie:
This branch isn't really a pull-request; more a work in progress to share with the OsmAnd team so that I could at least make the code/changes available if someone wants to review in future.

It relies on some modifications in the existing pull request #2446 but that's been sitting 10 days waiting for something to happen, and I'm just moving a lot quicker than this cycle time.  Basically this uses the asynchronous resampling to provide nicely spaced arrows and route markers - start a route, and zoom in a bit - you should see orange arrows and purple marker arrows over a track, dynamically adjusting as you change zoom level. The code is heaps smaller than the existing code, but there are a few things I have disabled/not implemented - specifically the different paint contexts for shadows and the like, which I do not understand.

I am aware that these changes have come along at a busy/inconvenient time for the project, and that nobody really has the time/inclination to use what I'm doing. I really needed a bit more interactivity and feedback to keep up the enthusiasm; it's just too difficult to work with, without someone to assist... but I guess it's just not the right time as there are other things happening.

So, here's the code as I have left it. Hopefully it will help the project at some time in the future.


andrew-davie:
https://youtu.be/0OQ7uR3TtRU

vshcherb:
Looks promising.

vshcherb:
My first deep look, the changes are really cool and huge. 
I'm afraid I can't review validity of the changes that's why I will need to ask as many people as possible to test it. The build is available here http://download.osmand.net/latest-night-build/OsmAnd-design.apk.

I think to move settings how to colourize route line into Development Settings, so it is possible to test and configure and will also display that's an incubating beta feature.

P.S.: I hope you won't mind of merging this pull-request even if it is closed. Please let us know.

andrew-davie:
> P.S.: I hope you won't mind of merging this pull-request even if it is closed. Please let us know.
> 

That’s OK with me.
I am away at a conference right now, but will put a pull request for my latest version of the code in a day or two.
There will be two - one for the Altitude/Speed, and a second for the Route stuff.

vshcherb:
Great to hear! I just tested I think it looks nicer overall though there is a functional issue with turns. They are displayed not as they were.Only if there is a turn in the point, the yellow arrow should be displayed at all zooms and should have at least 30 meters long, if yellow arrows overlap they are displayed as one line. There is also specifics for roundabouts.

andrew-davie:
I thought I'd add some generic notes on how my system works, so we can have common ground when discussing the changes and any problems that show up.

Essentially, when you consider a 'track' - that is, any series of lat/lon points, these are generally unevenly spaced. You may have dozens or hundreds of points around the same area when there is no movement, and then inconsistent spacing when there is movement.  In short, the points are evenly spaced *in time*.  Typically recorded with "every 5 seconds" or something like that.  This means that when you display at different zoom levels, you are always drawing all the points - even though many of those points may be on top of each other.  

So, what my system does is resample the track from the time domain into the distance domain. Instead of having points "every 5 seconds" it resamples to produce points "every N metres".  And the "N" is based on the zoom level.  So, a resample of the entire track is retriggered whenever the zoom changes. And the resampling is such that the SPACING on the screen is constant, regardless of the zoom level.  This is all handled transparently and asynchronously as a background task. It's all very clean, and elegant.

Once we have an evenly-spaced in distance track, now we can use those evenly spaced points as a basis for drawing tracks.  The points, as they are resampled from the original time-based track, are coincident with the original track (that is, the new points are guaranteed to be on the line joining points of the original track).  The only difference is the spacing/resolution. This is adjustable in the code - the greater the distance we have between points, the less points to draw, but the more 'chunky' it looks when representing curves/corners.

In any case, drawing things like the route arrow markers (the black arrow things) becomes extremely simple - because we already have consistent marker positions along the entire track. We simply have to choose every 4th... or 5th... whatever... and draw an arrow there.  The nice bit about all this is that the arrows are locked to the track, not to the screen - and so they do not jiggle around as we scroll around.

vshcherb:
I don't think I fully understood the message. I think there are 2 arrows: direction arrow (small repeating arrows on the track) and turn arrow (yellow lines on the turn directions). From my perspective the direction arrow is better displayed in your approach, but the turn arrows are broken in your approach. They are randomly appearing and disappearing at zoom levels.

P.S. Also the changes are really huge and I think it is still possible to simplify the code for general understanding. Such classes as WptPt2 doesn't seem reasonable.

andrew-davie:
Let’s clarify so we agree what we are discussing.

1) black direction arrows - working OK.
2) yellow arrows on the route itself - ‘broken’
3) the signposted direction arrow in the top corner

Since I had *no idea* how the 2) yellow arrows  were calculated, here’s how I did it…

a. first calculate the angle between every (resampled) point and its neighbour
b. sum the delta angles for each point in the route, using N points before, and N points after (so, it might sum 10 points to get the total “twistyness” near each point)
c. if the sum exceeds some threshold (say, 90 degrees) then mark that we should draw a yellow arrow here

The draw itself guarantees that yellow arrows are a certain length, and that there is reasonable spacing between all arrow types.


Basically the above functionality may not be what the arrows were meant for - I did not see them as “turn directions” but more as “twisty bits here”.
So let’s say I don’t understand how/what the original arrows were FOR, or how they were calculated. I would be happy to fix this in the new version so it behaves correctly.

I will say, though, that because they are drawn using the resampled points, AND because the points are resampled at each zoom level so that they are EVENLY SPACED according to the eye, then it will be impossible to have the arrows looking exactly the same at each zoom level (because the resampled points are not exactly the same at each zoom level).  In any case, I do not believe it correct behaviour to keep them the same.  The arrows should look good and give appropriate “directions” based on what you see on the screen, not some “common” look and feel over all zoom levels.  Let’s discuss this more?



Your second point about WptPt2 - I disagree with you. This is a *memory usage* issue and in my opinion, important.  I have added a new minimal class that ONLY has the variables needed for the route/track display so that it minimises memory and have NOT added those variables to WptPt so it also minimises memory. A generic class that is good for all uses is possible, but a terrible waste of resources especially when you start loading hundreds of thousands of track points.

Essentially if you’re dealing with a few hundred thousands of points, then if you use WptPt you are wasting LOTS of memory for the resampled tracks, and for th original track, especially as the functionality I have added requires extra variables (e.g, speed, color, distance, angle). 

If you use WptPt for tracks, then you have the unused baggage variables  name, link, category, desc, hop, deleted….  per point of the resampled tracks. A HUGE waste of memory. If instead you add the new resampled-track variables to WptPt (instead of using WptPt2) then you add colourARGB, distance, angle to every instance of WptPt.  Again, a huge waste of space because those vars are only used for the resampled tracks.

Essentially this is just being prudent and efficient with the memory usage. Seems like a no-brainer to me.

andrew-davie:
I don't think the route code I have written is quite ready for the big time yet. The yellow arrows, in particular, need a second person to cast their eye over what I've done.  In any case, what's not working for me at ALL is the delays between my submissions and any feedback. I am not seeing a lot of love, basically.  I think the speed and altitude display, AND the distance display code for GPX tracks is fantastic and should be incorporated. The route code is a work in progress that needs some assistance from others. But I just can't work with the complete lack of feedback and long delays between drinks, so to speak. So, I'm pulling this code soon and basically abandoning contributing to the project. I think it's a great project, but I'm a bit incompatible with the whole thing. I don't want to be a PITA as I know there's lots of work going on elsewhere. But yeah, feeling like a second class citizen...

PR #2500, opened 2016-04-27
Always show gpx track settings

HattinkT:
I've enabled the routing settings again when a different navigation service than osmand is used for gpx tracks, which are 'pass along entire track' and 'reverse GPX direction'. I think this is allowed as gpx navigation always uses osmand for calculating the first and last segments and recalculations when deviating from the gpx track.

I've also removed the option to 'calculate osmand route without internet' when brouter is used, as it is an offline service as well and should always succeed.

Please review carefully. I'm still familiarizing with the sources and I could have missed something when implementing these fixes.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/mapcontextmenu/other/RoutePreferencesMenu.java
+++ b/OsmAnd/src/net/osmand/plus/mapcontextmenu/other/RoutePreferencesMenu.java
@@ -562,10 +562,13 @@ public class RoutePreferencesMenu {
 		List<LocalRoutingParameter> list = new ArrayList<LocalRoutingParameter>();
 		RouteProvider.GPXRouteParamsBuilder rparams = mapActivity.getRoutingHelper().getCurrentGPXRoute();
 		boolean osmandRouter = settings.ROUTER_SERVICE.get() == RouteProvider.RouteService.OSMAND;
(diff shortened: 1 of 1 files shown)


vshcherb:
Can you please provide screenshots?

HattinkT:
Settings as shown when osrm (online service) is selected:
settings-osrm

Settings as shown when brouter (offline service) is selected:
settings-brouter

Settings as shown when brouter is selected and a gpx track is loaded:
settings-brouter-with-gpx

vshcherb:
The value of this pull request is not clear, though the code looks fine. These options are not relevant without GPX and could only confuse people.

HattinkT:
I agree that those options are not relevant without GPX. I think my naming of this patch is unclear in that regard. I've changed the name of the merge request to better reflect my intention.

So the GPX related settings are only shown when an GPX track is selected for navigation. But before the patch they were also only shown when osmand was selected as general navigation service. The patch makes sure the GPX related settings are also shown when another navigation service (like brouter) is selected. I think these GPX settings are always relevant as long as a GPX track is selected for navigation.

I've done multiple test runs and I always use these settings, and I have brouter selected for calculating routes when only using a destination without GPX.

I hope this makes it more clear what my intention is for this merge request.

PR #2512, opened 2016-05-03
Out-of-date transifex translations

unaizalakain:
Before Osmand began to use Weblate, I used Transifex to help translate Osmand to
Basque. I used Transifex as a one-off solution, and I forgot about it after
making the Basque translations.

Since then, some people continued to use Transifex without me noticing. The
translations are therefore quite old and some of them are probably out of date
as newer ones could have been made using Weblate.

This `transifex_translations` branch exists so that current translators at Weblate
can still use possibly valid translations made using Transifex.


unaizalakain:
I merged the Basque translations using Weblate's file uploading mechanism. It is possible to only add those translation strings that are in the uploaded file but not on Weblate, which is very neat.

vshcherb:
Thanks for you work though it is hard to say for each language where is more complete. It is easy to see that for lang-mr transifex is leading, though for lang-fr it looks vice versa. I will keep it open, may be somehow later we manage to iterate and split for each language to review.

PR #2537, opened 2016-05-11
Fix bug causing first segment of drawn path to be not drawn

HattinkT:
When a new segment of a path is drawn the first line was not drawn when the start coordinates were visible but the second set was not.

Screenshot with navigation path not being drawn:
device-2016-05-11-195708

Screenshot after fix:
device-2016-05-11-200244


scaidermern:
I did encounter this issue multiple times while navigating along a GPX file. Would be great to have it fixed.

vshcherb:
Many thanks, will test it and give feedback here.

PR #2608, opened 2016-05-27
speedAlarm: pass maxSpeed and speed

sonora:
Victor, can you please take a look if we want to do this, and double-check the implementation. Thx - Hardy


sonora:
Tested. Think it works!

vshcherb:
good

PR #2671, opened 2016-06-07
Only show "find parking" if destination is reached by car-derived modes

njohnston:
When reaching your destination in OsmAnd, the context menu shows a "find
parking" button to help you find a parking space. This might be useful if
you're driving (or using an application mode derived from car) but isn't
useful if you've just walked up to the top of a hill or you're cycling.

With this change, the button is shown only if you've reached the
destination by car (or a car-derived mode).


geodesicgnomes:
Don't forget about [bicycle parking](https://wiki.openstreetmap.org/wiki/Bicycle_parking). It is useful, especially in urban areas.

njohnston:
@geodesicgnomes: so you'd like the parking button to be shown at the end of cycle routing? From looking at [poi.xml in OsmAnd's resources](https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/poi_types.xml#L1457), `bicycle_parking` is indeed included in the parking POI filter.

PR #2903, opened 2016-08-04
CollatorStringMatcher.java fix

iPuzzle:


vshcherb:
Gridlayout and some others eclipse-compile are not used for real compilation. None of them is used in the main build or build for Android Studio (it is purely gradle version)

PR #3076, opened 2016-09-14
 Fix SQLite open flags

jmakovicka:
app.openOrCreateDatabase expects flags from android.content.Context
which are completely different

### DIFF
--- a/OsmAnd/src/net/osmand/plus/api/SQLiteAPIImpl.java
+++ b/OsmAnd/src/net/osmand/plus/api/SQLiteAPIImpl.java
@@ -14,8 +14,9 @@ public class SQLiteAPIImpl implements SQLiteAPI {
 
 	@Override
 	public SQLiteConnection getOrCreateDatabase(String name, boolean readOnly) {
-		android.database.sqlite.SQLiteDatabase db = app.openOrCreateDatabase(name,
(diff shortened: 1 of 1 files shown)


jmakovicka:
Hi,

I spotted this issue when trying to build Osmand for Android 24 API, wondering why Android keeps opening the database world readable. It turned out that the mode flags got misinterpreted.

vshcherb:
Thanks

crimean:
Unfortunately it causes error - http://pastebin.com/Uz5wj7yj
Rolled back.

jmakovicka:
I will look into that. It's possible that the DB is not created when opened read-only, although CREATE_IF_NECESSARY is set.

The stack trace is from a clean install? Could you check if /data/data/net.osmand.plus/databases/poi_filters existed at the time of openDatabase call?

crimean:
I tried CREATE_IF_NECESSARY without read-only flag - same thing.
This is from clean install. The poi_filters does not exist.

jmakovicka:
I reproduced the bug on Marshallow. The problem is that the last subdir ("database") does not exist yet at 1st startup. So the proper fix is either
- adding mkdir in case the directory does not exist
- or giving up on readonly/read-write flag and opening always read-write using app.openOrCreateDatabase.

Also, I am not sure about the db == null error checks - according to docs, the db opening functions should either return a valid db handle, or throw SQLiteException.

PR #3104, opened 2016-09-21
OSRM now has precision 5, Fix #3055

### LINKED ISSUE #3055: OSRM routing error: Calculated route is empty
> What version of OsmAnd are you using, on what device and Android/iOS version?

OsmAnd~ 2.4.4, Android 5.1

> Do you use the offline maps offered within the OsmAnd app for download, or online (tile / raster) maps?

Offline maps

>__*In case of wrong routing:*__
> Tell us whether you have used OsmAnd's in-app offline routing, or any online routing provider like YOURS, OpenRouteService or OSRM.

OSRM gives me the error 'Calculated route is empty' all the time. YOURS works, and so does the offline/built-in routing.

> What routing profile is choosen in OsmAnd app? (car, bike, pedestrian, fastest or shortest)

Car, fastest.

> Please specify as exactly as possible the start and end point of your route by telling us city name and street name so that we can find it via in-app address search easily. Also a permalink from openstreetmap.org can be helpful.

Start: Timișoara, Timiș, Romania
End: Pensiunea Melba, DN57, Dubova, Mehedinți, 227170, Romania

Routing on openstreetmap.org and map.project-osrm.org works:
http://www.openstreetmap.org/directions?engine=osrm_car&route=45.7538%2C21.2257%3B44.6303%2C22.2629#map=9/45.2275/21.8188
http://map.project-osrm.org/?z=9&center=45.225118%2C21.817835&loc=45.753836%2C21.225747&loc=44.630343%2C22.262939&hl=en&alt=0

> Tell us your expected routing, and how OsmAnd routes, or add screenshots here.

OsmAnd fails to find a route with OSRM. OSRM used to work in the past though, don't know if its a problem due to new OsmAnd version or due to server-side changes.

This is from `adb logcat`:

```
E/net.osmand(12822): RouteProvider Failed to find route 
E/net.osmand(12822): org.json.JSONException: Value The of type java.lang.String cannot be converted to JSONObject
E/net.osmand(12822): 	at org.json.JSON.typeMismatch(JSON.java:111)
E/net.osmand(12822): 	at org.json.JSONObject.<init>(JSONObject.java:160)
E/net.osmand(12822): 	at org.json.JSONObject.<init>(JSONObject.java:173)
E/net.osmand(12822): 	at net.osmand.plus.routing.RouteProvider.findOSRMRoute(RouteProvider.java:1200)
E/net.osmand(12822): 	at net.osmand.plus.routing.RouteProvider.calculateRouteImpl(RouteProvider.java:318)
E/net.osmand(12822): 	at net.osmand.plus.routing.RoutingHelper$RouteRecalculationThread.run(RoutingHelper.java:833)
```

Zahnstocher:
https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md

### DIFF
--- a/OsmAnd-java/src/net/osmand/util/GeoPolylineParserUtil.java
+++ b/OsmAnd-java/src/net/osmand/util/GeoPolylineParserUtil.java
@@ -15,7 +15,7 @@ public class GeoPolylineParserUtil {
 		List<LatLon> track = new ArrayList<LatLon>();
 		int index = 0;
 		int lat = 0, lng = 0;
-		double precision = 1E6;
+		double precision = 1E5;
 
 		while (index < encoded.length()) {
 			int b, shift = 0, result = 0;


Zahnstocher:
It is tested.
Old OSRM had precision 6, now it is 5.
Without this change, the decoded coordinates will be very wrong.

PR #3195, opened 2016-10-22
translation and typo in phrases.xml

hakuchi:
there are some translation and grammatical errors inside this file.
I fixed this with this edit.


stephan75:
Why not doing this via the weblate service?

hakuchi:
It is easier to find errors in a single file than to check one value next by next.

PR #3202, opened 2016-10-25
Sort pois in phrases.xml

aceman444:
Sort some POI strings in phrases.xml so that similar classes are together (e.g. shops, cuisines, wikis). It is useful for consistent translation.


aceman444:
Thanks.

PR #3219, opened 2016-11-01
Remove resource_cereal from phrases.xml

aceman444:
There is no resource_cereal in poi_types.xml (only crop_cereal) and it is not documented on the wiki for what it means.


sonora:
In rendering_types.xml we have
		`<entity_convert pattern="tag_transform" from_tag="resource" from_value="cereal" to_tag1="crop"/>`

@xmd5a2 should have a look at it. If this is by mistake, we should adjust all localized phrases.xml.

PR #3268, opened 2016-11-15
Restored string that was lost somehow

poretsky:
This string is still used, so should have a default.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -262,6 +262,7 @@
     <string name="download_files_question_space">Download {0} file(s)?
         Storage space used is {1} MB.
         (Available space is {2} MB.)</string>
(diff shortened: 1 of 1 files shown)


stephan75:
That string needs tuning ... how can a note be anonymus when using your OSM account????

poretsky:
Ok, but it should present anyway, shouldn't it?

sonora:
We have 2 separate dialogs now: upload an OSM edit, and upload a note.

In sync with OSM rules, only the latter one offers an option to upload anonymously. From what I remember, this string is not in use any more (neither did it male any sense).

poretsky:
Sorry, but this string is referenced from the layout file
send_poi_dialog.xml which in turn is referenced from
net/osmand/plus/osmedit/dialogs/SendPoiDialogFragment.java.

sonora:
We need to double-check, I think the dialog was cleaned up by @crimean

Issue #3398, opened 2017-01-07
Tapping the Quick action button crashes OsmAnd


sonora:
Log attached: [exception.log.txt](https://github.com/osmandapp/Osmand/files/691727/exception.log.txt)

sonora:
Also: Tapping Quick action too early after app start can produce this NPE:  [exception.log.txt](https://github.com/osmandapp/Osmand/files/692512/exception.log.txt)

sonora:
And: [exception.log.txt](https://github.com/osmandapp/Osmand/files/694852/exception.log.txt)

vshcherb:
Fixed

PR #3409, opened 2017-01-10
quick action part 2

rostdev:


sonora:
This brings back at least the first crash documented in #3398

PR #3434, opened 2017-01-19
Added support for attrFloatValue

Rofang:
Allows rendering attributes to be defined using "attrFloatValue" parameter (such as for dynamic stroke widths)


vshcherb:
That's definitely not enough, core is the most important bit. Something like this would be needed anyway that's a complex request and it requires testing (some reference https://github.com/osmandapp/OsmAnd-core/commit/6961724ce7a76aad30388252a390d33e240b4137)

PR #3436, opened 2017-01-20
Update settings.xml (spanish - Spain)

OpenSourceForEver:
Better traslation of 725 2295 and 2298.


OpenSourceForEver:
I think this traslation is better but it would be better if anyone could check it

PR #3437, opened 2017-01-20
Make all recording parameters global, due to ...

sonora:
- The issue that parameters for an ongoing recording currently change unnoticed (behind the scenes) when peoplechange app profile to check something on the map
- Starting a recording from the new notification uses undefined/not evident parameterization
- and the fact that we have no interface to access the parameters for the default profile (serious flaw!)


vshcherb:
I feel pros/cons but I can't make that call. @xmd5a2 is better one cause he might use more often (mode specific settings)

xmd5a2:
I think that this pull request makes sense. Indeed, linking these settings to profiles can confuse users.

PR #3486, opened 2017-02-06
srtm_plugin_description: specify that relief data have to be downloaded before using plugin

brunetton:
I didn't find obvious that relief data have to be downloaded separately before using this plugin. People may think that relief data is automatically downloaded when online, like online maps tiles for example.
Translation is not complete for other languages.


sonora:
Thanks! I have put it slightly different, see https://github.com/osmandapp/Osmand/commit/c9a813e1a759b244b38bbcaf7cd81f9a0b408042.

brunetton:
Good to read

PR #3499, opened 2017-02-09
Waypoints announcement initialization fix

poretsky:
I cannot imagine the reason these strings should be commented out for,
but not having the list properly initialized it is impossible to go
through the same route again hearing the announcements without force
reload.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
+++ b/OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
@@ -443,7 +443,7 @@ public class WaypointHelper {
 	}
 
 	public void clearAllVisiblePoints() {
-		//this.locationPointsStates.clear();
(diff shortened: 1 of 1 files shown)


vshcherb:
Looks strange the array was never cleared.

vshcherb:
@sonora  may be you have any idea? https://github.com/osmandapp/Osmand/commit/bbf426e08a348cc947a2140c14ec17ca0340ba2f

sonora:
No immediate idea ..."Updated fix to waypoint announced many times" sounds like that commit tried to fix something about repeated announcements of waypoints after route-recalculations (which possibly occur or occurred often in the background).

Unless I am mistaken, we never call `clearAllVisiblePoints()` anywhere in our code anyway, so the question is limited to `setLocationPoints(`, which we only call from `setNewRoute(`. I'd say let's test it.

PR #3571, opened 2017-03-21
Show paused state more clearly on track recording notification

njohnston:
Previously, the track recording notification displayed "Pause" when recording was paused. This was confusing because when recording, the notification also shows "Pause" to allow recording to be paused.

With this change, "Paused" (rather than "Pause") is shown in the notification when recording is paused.


sonora:
Thank you!

PR #3587, opened 2017-03-27
Extracts elevation data from BRouter.

saemy:
Makes use of BRouter's capability of returning elevation information for a calculated route. This is helpful since OsmAnd's SRTM data is not considered when using BRouter.


vshcherb:
Please test it cause OsmAnd 2.6 is going to be released soon.

saemy:
I ran the unit tests and tested it in the app. Is there anything else I can do? Are there  some tricky edge cases?

vshcherb:
Nope that's enough. The issue is that I mostly checked the code and the functionality.
Thanks for implementation!

PR #3618, opened 2017-04-03
Fixes #3617.

### LINKED ISSUE #3617: "Analyze on map" does not always work
Clicking `ANALYZE ON MAP` in `Route details` does not work if we arrived there by `☰ -> Dashboard -> [Current route] INFO`. Rather than showing the elevation graph on the map, it jumps back to the dashboard.

It does work if we arrived there by the route info menu.

saemy:
When showing the elevation details, we need to distinguish between the user arriving by the dashboard or the route info menu. Only in the latter case the menu needs to be closed.


saemy:
I am still not fully satisfied, as hitting the back button does not return to `Route details` but simply closes the elevation graph. But this is a different issue.

PR #3639, opened 2017-04-08
Minor string fixes and improvements

njohnston:
Also added missing strings for the British English / UK regional locale.

Hopefully there's still time to get this into OsmAnd 2.6 :)

### DIFF
--- a/OsmAnd/res/values-en-rGB/strings.xml
+++ b/OsmAnd/res/values-en-rGB/strings.xml
@@ -4,6 +4,8 @@
 		This file is manually maintained in res/values-en-rGB and should ONLY contain strings where a different spelling should be used for English as used in the United Kingdom
 	-->
 
+	<string name="change_color">Change colour</string>
+	<string name="srtm_color_scheme">Colour scheme</string>
 	<string name="towards">towards</string>
 	<string name="access_default_color">Default colour</string>
 	<string name="rendering_value_walkingRoutesScopeOSMC_name">Colouring according to route scope</string>
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -15,12 +15,12 @@
     <string name="quick_action_showhide_osmbugs_descr">Tapping the action button will show or hide OSM Notes on the map.</string>
     <string name="sorted_by_distance">Sorted by distance</string>
     <string name="search_favorites">Search favorites</string>
-    <string name="hillshade_menu_download_descr">For view hillshade on the map, you need to download special map of this region.</string>
-    <string name="hillshade_purchase_header">To see hillshade on the map, you need to buy and install plugin</string>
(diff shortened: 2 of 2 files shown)


sonora:
@njohnston I have fixed your 2 en-GB additions manually here: https://github.com/osmandapp/Osmand/commit/b00af7ed230d647e2d90cb2eb889629debf29b4c

The rest had already been superseded by a larger improvement I had tried this morning concerning many recent new strings. Maybe you feel like double-checking https://github.com/osmandapp/Osmand/commit/f87818d837425399e2d0eb7861fd7f5616cb068f.

I have to admit that in a few cases I am a little at loss with what exactly we mean by some of them ..  :) and did not have the time to perform a full code review in all cases too find out.

2 things I remember:

- The whole "Plains vs. Hills" preferences story: I have tried to resolve by " Prioritize leveled vs. hilly" (another case for an en-GB spelling correction if we leave it at that). Also not sure about the "scale", I have put "Most leveled, leveled, hilly" for now.
- I am not sure in string do_not_send_anonymous_app_usage_desc what information is actually routinely collected vs. what not. "information about screen used" seems a little vague concerning such a sensitive area. We may have to ask @vshcherb here to become more precise, like explaining if we only collect which widgets and dialogs are displayed/used, or also what map regions or locations are viewed

njohnston:
@sonora thanks for adding the en-GB strings.

I've added some comments on your commit f87818d.

vshcherb:
We only collect which screens are used and how often, no geolocation data at all, no any input data and not even search terms (though it would really help us). That information is only collected in the free version.

sonora:
Victor, so we can rephrase to make it clear we also do not track which map section is viewed, right? This is the major ambiguity when we talk about "screen data".

njohnston:
@sonora @vshcherb I suggested an improvement for that string in a comment on commit f87818d, please take a look. Thanks.

sonora:
@njohnston; Thank you, good work! I have left the Plugin name in place, may reduce ambiguity.

I guess we need some help from Victor regarding Location "Animation" ... I do not see what it does at first sight.

All the rest I implemented, good suggestions! Thanks again! (https://github.com/osmandapp/Osmand/commit/286bccce8e23fc322b2117ff5bf6628f6dda896a)

Thx - Hardy

njohnston:
@sonora: thanks for making the changes. A few points though:

* There's a stray " just before `</string>` for do_not_send_anonymous_app_usage_desc.
* For relief_smoothness_factor_descr, I think an *or* is necessary: "Prefer flat hilly" doesn't make much sense. Maybe "Prefer flat *or* hilly", or "Preferred terrain: flat *or* hilly".

sonora:
Yes, true. One I had already fixed, the other one now. Thx!

njohnston:
Thanks for the fixes :)

PR #3766, opened 2017-05-11
Calculate speed for GPX tracks which don't contain speed

### LINKED ISSUE #3702: Calculate speed for tracks without <speed>
Many GPX tracks don't have the `<speed>` tag: only position (`<trkpt>` or `<rtept>`) and an associated timestamp. For these tracks, OsmAnd shows no speed data:

track_without_speed

As [this track](https://drive.google.com/file/d/0ByKX_Fubbb4Qby1ORnN5RmNGSm8/view) has a timestamp associated with each position/point, it is possible to calculate the speed, albeit slightly approximately depending on the geoid used.

Please calculate speed for such tracks.

njohnston:
Most GPX tracks don't contain speed (GPX has no standard way of including it), but do include time for each point, so speed can be calculated.

Partly fixes #3702--ideally tracks where speed is calculated should be indicated slightly differently in the UI.

Includes a trivial refactor of this code too:

```java
          if (speed > 0) {
            totalSpeedSum += speed;
            maxSpeed = Math.max(speed, maxSpeed);
            speedCount++;

            speed1.speed = speed;
          } else {
            speed1.speed = 0;
          }
```

Since `speed1.speed` is always assigned to, it can be moved out of "if", and the "else" can be removed, because if `speed` is 0 then assigning `speed` to `speed1.speed` will have the desired effect. No need to explicitly assign 0.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/GPXUtilities.java
+++ b/OsmAnd/src/net/osmand/plus/GPXUtilities.java
@@ -312,6 +312,7 @@ public class GPXUtilities {
 
 		public boolean hasElevationData;
 		public boolean hasSpeedData;
+		public boolean hasSpeedInTrack;
 
 		public boolean isSpeedSpecified() {
 			return avgSpeed > 0;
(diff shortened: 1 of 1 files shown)


sonora:
Currently out traveling, will check when I am back. At first sight it looks like we need a slight refactoring or else this affects timeMoving and totalDistanceMoving calclulations (which I had introduced to filter for points recorded while the GPS chipset reports motion, a case not applicable to the 'speedless' tracks you are considering here)?

vshcherb:
Let's wait @sonora, from my side everything looks good (we can merge)

sonora:
I have slightly resorted this and committed here https://github.com/osmandapp/Osmand/commit/0ccdc225c6743ce8109904d084a7cccf74010651, I think like this it should be ok and can be tested now.

In a second commit I will change the color for speed data derived by this displacement method from orange to red so we have transparency what type of values we are looking at.

PR #3872, opened 2017-06-03
[Review] Request from 'poretsky' @ 'poretsky/Osmand/proposal2'

poretsky:
In accessibility context it's just important for now to have an option
that allows to switch map markers mode off.

Please review the following changes:
  * 6e34ca4 Moved map markers enabling option to the accessibility plugin
  * 62e384c Proper conditionals for map marker mode depending quick action


vshcherb:
This request is not clear, "map markers" are going to be completely removed cause there is no way we could support 2 similar functionalities Markers and Destination points. That was intermediate solution before release 2.5 and it is deprecated. Second commit only indicates the problem of supporting various concepts across application.

PR #3891, opened 2017-06-08
Add ruler

crimean:


sonora:
Hint: The new ruler does not work if you put the widget on the map screen as collapsable widget, it only works in "always show" widget mode.

PR #3957, opened 2017-06-23
Add a delay before the line between fingers disappears

AlexSyt:


OsmAnd/src/net/osmand/plus/views/RulerControlLayer.java
```diff
@@ -98,7 +99,10 @@ public void onDraw(Canvas canvas, RotatedTileBox tb, DrawSettings settings) {
             final QuadPoint center = tb.getCenterPixelPoint();
             final RulerMode mode = app.getSettings().RULER_MODE.get();
 
-            if (view.isMultiTouch()) {
+            if (cacheMultiTouchTime != view.getMultiTouchTime()) {
+                cacheMultiTouchTime = view.getMultiTouchTime();
+            }
+            if (view.isMultiTouch() || System.currentTimeMillis() - cacheMultiTouchTime < 3000) {
```
vshcherb: Should be constant instead of Number!

PR #3968, opened 2017-06-25
fix testTooManyCapitalLetters wrt unicameral scripts

mbethke:
In v2.6.3, Osmand would keep complaining about too many capital letters in a name when using any script that doesn't distinguish between upper- and lowercase such as the CJK family, Arabic, Thai/Lao etc.
This is because the `Character.toUpperCase(c) != c` test could never be true for those scripts so all characters were classified as uppercase. The builtin Character.isUpperCase() handles this correctly and is much clearer, too.


vshcherb:
Just to speed up approval case did you test it for English based languages?

mbethke:
It returns true for "FOO" and false for "foo" and "ສະບາຍດີ", the latter having returned true in the original version - but that's as far as I went with "unit testing" :) Should be fine though as the logic regarding non-alphabetic characters hasn't changed.

PR #3971, opened 2017-06-26
Add small improvements

AlexSyt:


sonora:
Not sure this was a good improvement. Looks like now the appearance of the Two-finger ruler is flaky, it e.g disappears after the slightest even unintentional pinch-zoom gesture whereas before it remained visible OK and its value displayed even during full pinch-zooming and/or two-finger panning?

PR #3987, opened 2017-06-28
Add current point and previous point

PavelRatushnyi:


vshcherb:
I hope it is tested cause I mostly verify the code logic and code structure

PR #3996, opened 2017-06-29
Gpx improvements

PavelRatushnyi:


vshcherb:
Let's see if there is no bugs with introducing new setting type. Though it is still to double check who was using that setting as LatLon (now it also could give QuadRect and cause CastException)

PR #4111, opened 2017-07-14
Alex branch

AlexSyt:


OsmAnd/src/net/osmand/plus/views/MapControlsLayer.java
```diff
@@ -350,6 +350,10 @@ protected void clickRouteWaypoints() {
 		}
 	}
 
+	public void stopNavigation() {
+		clickRouteCancel();
```
crimean: Extract clickRouteCancel() to stopNavigation() and call stopNavigation() from clickRouteCancel()

OsmAnd/src/net/osmand/plus/views/MapControlsLayer.java
```diff
@@ -440,6 +444,10 @@ public void doNavigate() {
 		startNavigation();
 	}
 
+	public void showNavigationDialog() {
+		onNavigationClick();
```
crimean: Why you do not call doRoute(boolean hasTargets) instead?

OsmAnd/src/net/osmand/plus/quickaction/QuickActionFactory.java
```diff
@@ -46,7 +46,7 @@
 
     public String quickActionListToString(List<QuickAction> quickActions) {
 
-		return new Gson().toJson(quickActions);
+        return new Gson().toJson(quickActions);
```
crimean: Fixed with spaces?

AlexSyt: yes

OsmAnd/src/net/osmand/plus/views/MapControlsLayer.java
AlexSyt: Dima said that the behavior when clicking on the navigation button and navigation in quick actions should be identical. Therefore I use the same method.

OsmAnd/src/net/osmand/plus/views/MapControlsLayer.java
AlexSyt: Fixed.

PR #4128, opened 2017-07-17
Fix bugs

PavelRatushnyi:


OsmAnd/src/net/osmand/plus/myplaces/TrackSegmentFragment.java
```diff
@@ -279,12 +278,12 @@ private void updateHeader() {
 		imageView.setOnClickListener(new View.OnClickListener() {
 			@Override
 			public void onClick(View view) {
-				LatLon location = new LatLon(generalDisplayItem.locationStart.lat,
-						generalDisplayItem.locationStart.lon);
+				LatLon location = new LatLon(getGpx().findPointToShow().getLatitude(),
```
vshcherb: Should be check if findPointToShow null?

PR #4139, opened 2017-07-18
Fix #4027

### LINKED ISSUE #4027: Multiple POI entries does not appear on map.
When selecting multiple POI categories to be shown on the map (Gas Station and Parking) only Parking is being displayed.

AlexSyt:


vshcherb:
Todo reintroduce caching though it is needed to make PoiUIFilter immutable first.

PR #4148, opened 2017-07-19
Format default comment

PavelRatushnyi:


OsmAnd/res/values/strings.xml
```diff
@@ -2642,4 +2642,5 @@
     <string name="quick_action_btn_tutorial_title">Change button position</string>
     <string name="quick_action_btn_tutorial_descr">Long tap and drag the button to change its position on the screen</string>
     <string name="shared_string_action_name">Action name</string>
+    <string name="items_modified">items modified</string>
```
vshcherb: It should be in no_translate.xml file otherwise people will translate it.

OsmAnd/src/net/osmand/plus/osmedit/EditPoiDialogFragment.java
```diff
@@ -449,7 +449,8 @@ private void save() {
 			if (offlineEdit && !Algorithms.isEmpty(poiTypeTag)) {
 				node.putTagNoLC(EditPoiData.POI_TYPE_TAG, poiTypeTag);
 			}
-			comment = OsmPoint.displayAction.get(action) + " " + poiTypeTag;
+			String actionString = action == OsmPoint.Action.CREATE ? getString(R.string.shared_string_add) : getString(R.string.shared_string_edit);
```
vshcherb: It should be in no_translate.xml file otherwise people will translate it. We should have only in english in default comment so we should not use shared_string... which is translatable

PR #4155, opened 2017-07-20
Alex fixes

AlexSyt:


OsmAnd/src/net/osmand/plus/views/mapwidgets/MapMarkersWidgetsFactory.java
```diff
@@ -189,9 +190,16 @@ public void updateInfo(LatLon customLocation, int zoom) {
 			Location l = map.getMapViewTrackingUtilities().getMyLocation();
 			if (l != null) {
 				loc = new LatLon(l.getLatitude(), l.getLongitude());
-				lastKnownPosition = loc;
+				lastKnownPosition = l;
+			} else if (lastKnownPosition == null) {
+				loc = null;
 			} else {
-				loc = lastKnownPosition;
+				if (System.currentTimeMillis() - lastKnownPosition.getTime() < OsmAndLocationProvider.LOCATION_SHELF_LIFE) {
```
vshcherb: Why here is needed?

OsmAnd/src/net/osmand/plus/views/PointLocationLayer.java
```diff
@@ -53,38 +56,48 @@ private void initUI() {
...
+
 	@Override
 	public void initLayer(OsmandMapTileView view) {
 		this.view = view;
 		initUI();
 	}
 
+	private boolean isLocationOutdated() {
+		if (locationProvider.getLastKnownLocation() != null) {
+			return System.currentTimeMillis() - locationProvider.getLastKnownLocation().getTime() >
```
vshcherb: Do we need to call it every time ? System.currentTimeMillis() ? Practically 60 times a second.

OsmAnd/src/net/osmand/plus/views/mapwidgets/MapMarkersWidgetsFactory.java
vshcherb: Do we need to call it every time ? System.currentTimeMillis() ? Practically 60 times a second.

vshcherb: May be it should be generical functionality that would cover all cases not only 1 widget?

PR #4384, opened 2017-09-01
Add setting for speed

PavelRatushnyi:


vshcherb:
@sonora we have a special "sponsored" request for that setting. So in theory OsmAnd should behave as follow today in Bearing mode if the speed is too low it should "always" and in all modes show the view angle (we should validate it actually).

The request is following: instead of showing the view angle just to switch rotating map to the compass. Which to be honest could give bad results on some devices but it is also useful for some people. This setting is present in other apps.

We're still very confused how to properly name it that's pull request is not accepted.

PR #4404, opened 2017-09-05
Fix #4394

### LINKED ISSUE #4394: Unable to properly open links
OsmAnd v2.7.5 (from F-Droid repository)
Android 5.1.1 on Xiaomi Mi-4c

When clicking on a link like http://www.openstreetmap.org/search?query=lat%20lon (which works fine on a browser), OsmAnd will be suggested to open the link, however it will just copy it to the search bar and return 0N 0W as location regardless of the query, meaning the user must edit the link manually until lat lon is left on the search bar for the app to show the proper result.

PavelRatushnyi:


OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
```diff
@@ -220,6 +220,16 @@ public static void main(String[] args) {
 		actual = GeoPointParserUtil.parse(url);
 		assertGeoPoint(actual, new GeoParsedPoint(ilat, ilon, z));
 
+		url = "http://www.openstreetmap.org/search?query=" + qlat + "%2C" + qlon;
```
vshcherb: We should also support "http://www.openstreetmap.org/search?query=Amsterdam"

OsmAnd-java/src/net/osmand/util/GeoPointParserUtil.java
```diff
@@ -928,6 +953,31 @@ public static GeoParsedPoint parse(final String uriString) {
 								lat = parseSilentDouble(vls[1]);
 								lon = parseSilentDouble(vls[2]);
 							}
+						} else if (queryMap != null) {
+							String queryStr = queryMap.get("query");
+							if (queryStr != null) {
+								String[] vls = null;
```
vshcherb: just put return new GeoParsedPoint(queryStr.replace("+", " ").replace("/|,", " "));

PR #4406, opened 2017-09-05
Sasha pasha branch

AlexSyt:


polarbearing:
Hi @AlexSyt - good to see all the progress here. I am testing nightly builds, so it would be even more helpful if there were some hints about what features are being implemented, and what bugs are being fixed. So I could keep my eyes in target when testing them.

AlexSyt:
Hi, @polarbearing.
Thank you for your attention to our work.
We are engaged in improvement and redesign of Map markers. But at the moment the development is carried out in a separate branch, so the latest changes are not yet available in the nightly builds.

OsmAnd/src/net/osmand/plus/mapmarkers/MapMarkersDbHelper.java
```diff
@@ -0,0 +1,511 @@
...
+		return conn;
+	}
+
+	private void onCreate(SQLiteConnection db) {
+		db.execSQL(MARKERS_TABLE_CREATE);
+		db.execSQL(GROUPS_TABLE_CREATE);
+		saveExistingMarkersToDb();
+	}
+
+	private void onUpgrade(SQLiteConnection db, int oldVersion, int newVersion) {
```
crimean: So after upgrade we drop all existing markers?

AlexSyt: This is an option for the development time.

PR #4451, opened 2017-09-18
Sasha pasha branch

AlexSyt:


OsmAnd/src/net/osmand/plus/MapMarkersHelper.java
```diff
@@ -132,34 +137,37 @@ public int hashCode() {
...
+				R.color.marker_blue,
+				R.color.marker_green,
+				R.color.marker_orange,
+				R.color.marker_red,
+				R.color.marker_yellow,
+				R.color.marker_teal,
+				R.color.marker_purple
+		};
+
+		public static int[] getColors(Context context) {
```
crimean: This array must be calculated one time only

OsmAnd/src/net/osmand/plus/MapMarkersHelper.java
```diff
@@ -132,34 +137,37 @@ public int hashCode() {
...
-				default:
-					colorId = R.color.marker_blue;
+			return (colorIndex >= 0 && colorIndex < colors.length) ? colors[colorIndex] : colors[0];
+		}
+
+		public static int getColorIndex(Context context, int colorId) {
+			int[] colors = getColors(context);
+			for (int i = 0; i < colors.length; i++) {
+				int color = colors[i];
+				if (color == colorId) {
```
crimean: Why do you compare color and colorId?

AlexSyt: Fixed

OsmAnd/src/net/osmand/plus/MapMarkersHelper.java
AlexSyt: Fixed

Issue #4447, opened 2017-09-16
OSM Live map files are never deleted when their main map file is removed


aceman444:
1. have some maps downloaded
2. then download OSM Live updates for them
3. remove one of the maps (normally from inside OsmAnd UI).

Result:
look into OsmAnd data folder/live subfolder. live update files for the removed map (country) are still there.
For other maps, the live update files are removed when the main map is updated from the monthly full update files. But the removed maps never will have their live files cleared.

PaulStets:
Fixed with 5182a72

aceman444:
Thanks!

PR #4456, opened 2017-09-19
Fixed issue #4447

PaulStets:


OsmAnd/src/net/osmand/plus/download/ui/LocalIndexesFragment.java
```diff
@@ -465,6 +467,13 @@ protected String doInBackground(LocalIndexInfo... params) {
 					if (operation == DELETE_OPERATION) {
 						File f = new File(info.getPathToData());
 						successfull = Algorithms.removeAllFiles(f);
+						if (InAppHelper.isSubscribedToLiveUpdates()) {
```
crimean: Better to use getSettings().LIVE_UPDATES_PURCHASED.get() here

PR #4474, opened 2017-09-24
add more constructed languages as selectable map-languages

verdulo:
I added Interlingua, Ido and Lojban as selectable map-languages. To make this work I also need to edit some files as in https://github.com/osmandapp/OsmAnd-tools/commit/46feb32a1c68d3587135bcd44e406d55c04c0369, yes?


verdulo:
@vshcherb Is there any chance to add these languages as map-languages?

vshcherb:
Small chance these languages don't have 1000 names according to taginfo. 
@sonora might have another opinion

sonora:
Fully concur, it seems really pointless to accomodate map languages which have neither a regional importance nor are used to a significant extent.

PR #4566, opened 2017-10-10
Fix #4445

### LINKED ISSUE #4445: OSMand crashes, when Search function is triggered soon after start-up
OSMand consistently (i.e. 100% reproducibly) crashes, when its Search function is triggered soon after start-up.  It does not make any difference, if the Search function is triggered by the "magnifying glass" icon on OSMand's primary screen or from the menu entry "Search". 
On my (rather old and slow) device, it takes about 15 seconds *after* the map rendering on OSMand's primary screen has finished for OSMand to settle, in order not to crash when starting a Search.  Hence one should be able to reproduce this issue on faster devices by triggering a Search *while* the map is rendered immediately after start-up. 

Tested with OSMand~ v2.7.5, v2.6.5, v2.6.2, v2.5.4 under AOSP4.1.2 (API16). 
I already sent in a crash report (including an exception log) for v2.7.5 via email (directly out of OSMand 2.7.5: a really nice improvement!), here are the exception logs for [v2.6.5](https://github.com/osmandapp/Osmand/files/1308503/exception_v265.log.txt), [v2.6.2](https://github.com/osmandapp/Osmand/files/1308504/exception_v262.log.txt) and [v2.5.4](https://github.com/osmandapp/Osmand/files/1308505/exception_v254.log.txt); they seem to be cumulative, so [the last one](https://github.com/osmandapp/Osmand/files/1308505/exception_v254.log.txt) should cover all crashes (throughout my testing) related to this issue. 

Note, that "Searching immediately after OSMand's start-up" is a common use pattern, and when testing this (thus quickly hitting the Search icon after start-up), one easily believes that OSMand is always crashing when triggering a Search, as it takes some patience to wait for OSMand to have settled (when testing with a  slow device). 

Kudos to you, OSMand developers, for creating simply the best maps application, superseding all proprietary applications in this field.

PavelRatushnyi:


vshcherb:
Will it be refreshed later if it is null? Or we will miss completely this entry??

PR #4572, opened 2017-10-11
Add location listener

AlexSyt:


vshcherb:
Recalculation should happen only if my location changed by 50 meters ( don't see any threshold at all)

AlexSyt:
ok

PR #4643, opened 2017-10-24
Online search

AlexSyt:


njohnston:
Thanks for your work on this--would it be possible to support [Photon](http://photon.komoot.de/) as well as Nominatim? Photon supports search-as-you-type, as well as typo tolerance.

PR #4673, opened 2017-10-31
cosmetics: spelling fix and remove double semicolon.

rdoeffinger:
Minor things I stumbled over. For you to decide if it's worth taking/fixing or you'd rather leave it.


sonora:
Thanks!

PR #4706, opened 2017-11-07
Add small fixes; do not hide status bar with map controls

AlexSyt:


sonora:
I really liked the **complete** full screen mode including hiding the status bar as well, but I guess this caused some issues?

Issue #4712, opened 2017-11-08
Replace prebuilt jar files with Maven dependencies


yourealwaysbe:
I have been working on adding the new version of OsmAnd to the F-Droid repository.  A problem here is that F-Droid does not accept prebuilt binaries.  Our discussion can be followed [here](https://gitlab.com/fdroid/fdroiddata/issues/959).

One hopefully easy to fix issue is the number of libraries distributed with OsmAnd that could be replaced by Maven dependencies.  The purpose of this post is to ask: 

- would OsmAnd be willing to accept a patch/pull request to replace these dependencies?
- are our chosen replacements for version mismatches benign?

We have been able to remove all jar files and still compile OsmAnd with the following replacements.  Our simple testing seems to show no problems.  I will list the replacements in full before discussing mismatches.

In OsmAnd/build.gradle, the fileTree dependencies can be replaced with
```
compile 'commons-logging:commons-logging-api:1.1'
compile 'it.unibo.alice.tuprolog:tuprolog:3.2.1'
compile 'org.beanshell:bsh-core:2.0b4'
compile 'com.ibm.icu:icu4j:50.1'
compile 'fr.univ-valenciennes:bzip2:1.0'
compile 'com.moparisthebest:junidecode:0.1.1'
compile 'net.sf.trove4j:trove4j:3.0.3'
compile 'org.immutables:gson:2.5.0'
compile 'com.vividsolutions:jts-core:1.14.0'
```

In OsmAnd-java/build.gradle, the dependencies can be replaced with
```
compile 'commons-logging:commons-logging:1.1.1'
compile 'net.sf.trove4j:trove4j:3.0.3'
compile 'org.json:json:20090211'
compile 'it.unibo.alice.tuprolog:tuprolog:3.2.1'
compile 'net.sf.kxml:kxml2:2.3.0'
compile 'org.beanshell:bsh-core:2.0b4'
compile 'com.ibm.icu:icu4j:50.1'
compile 'fr.univ-valenciennes:bzip2:1.0'
compile 'com.moparisthebest:junidecode:0.1.1'
compile 'com.vividsolutions:jts-core:1.14.0'
```

There are the following mismatches:

-  gnu-trove-osmand.jar replaced with net.sf.trove4j:trove4j:3.0.3. What is gnu-trove-osmand?  If it is a bespoke version of GNU Trove, where can the source be found?
[...]

PR #4728, opened 2017-11-10
Replace prebuilt jar dependencies with Maven alternatives (legacy-build)

yourealwaysbe:
NOTE: does not compile using master branch (./gradlew assembleFullLegacyFat) due to "incorrect translation", though this is unrelated to the pull request.  Commit compiles when cherry-picked against stable v2.8.2 commit.

RELATED: Issue #4712.

COMMIT MESSAGE:

Primarily to aid F-Droid build, but also to avoid distribution of
unknown code.

There are the following mismatches:

gnu-trove-osmand.jar replaced with net.sf.trove4j:trove4j:3.0.3.  Latest
version.  Unsure what gnu-trove-osmand is.

tuprolog.jar replaced with it.unibo.alice.tuprolog:tuprolog:3.2.1.
Latest version, no version specified in jar name.

icu4j-49_1_patched.jar replaced with compile 'com.ibm.icu:icu4j:50.1' in
the hope that whatever patch was applied appears also in 50.1.

bzip2-20090327.jar replaced with compile
'fr.univ-valenciennes:bzip2:1.0'.  Latest version taken since none close
to 2009 release date available.

junidecode-0.1.jar replaced with compile
'com.moparisthebest:junidecode:0.1.1'. There is a .1 version difference.

simple-logging.jar not replaced in OsmAnd-java (it was not needed). It
was replaced by commons-logging:commons-logging-api:1.1 in OsmAnd. I
could not find any sources for simple-logging.jar, but its contents are
similar to commons-logging-api. The main difference is that the provided
Log class requires a few more abstract methods to be implemented. I
added the required code to OsmandLogImplementation in PlatformUtil.java.


vshcherb:
There are 2 questions we should be attentive to this pull request:
1. Whether everything remains functional for every jar we could 1-2 places where it is used & test
2. What is the final size of application (so it shouldn't grow). 
   I would focus on 2 jars that were shrinked 
	'com.ibm.icu:icu4j:50.1' & gnu-trove.

sonora:
I will gladly assist with the testing. Can we please produce a common design build which includes this commit so we all test the same thing?

PS: @vshcherb apk size: A significant portion of our large apk size is fonts. It deserves a separate investigation if we reaaly need all fonts in the quality we include them.

vshcherb:
@sonora  I remember the moment we were including them & yes we need them all. Otherwise people will need to download the separately as today for Japanese/Chinese which is counter-user intuitive.

vshcherb:
@sonora  I merged to a branch - https://github.com/osmandapp/Osmand/tree/yourealwaysbe/maven-deps so we can try to test in the Jenkins design build

sonora:
I find no functional flaw so far, but the apk size increases by 9 MB which seems hardly acceptable ...

Bubu:
@sonora Where can I find the build you are referring to and where can I find a comparison apk not including this change? I'd like to look into what causes the size increase.

yourealwaysbe:
@sonara @Bubu i don't have access to my good laptop at the moment to experiment with real builds, but comparing the OsmAnd/libs jar file size with the size of replacements from Maven, it looks like the main culprits are:

* [icu4j-49_1_patched](https://github.com/osmandapp/Osmand/blob/master/OsmAnd/libs/icu4j-49_1_patched.jar) is 2.7mb, while [icu4j:50.1](http://repo1.maven.org/maven2/com/ibm/icu/icu4j/50.1/) is 9.2mb. [icu:49.1](http://repo1.maven.org/maven2/com/ibm/icu/icu4j/49.1/) as distributed by maven is 7.4mb.
* [gnu-trove-osmand](https://github.com/osmandapp/Osmand/blob/master/OsmAnd/libs/gnu-trove-osmand.jar) is 320k while [trove4j](http://repo1.maven.org/maven2/net/sf/trove4j/trove4j/3.0.3/) is 2.5mb.

I assume, then, that both of these libraries have been substantially modified from the standard distribution in some way.  Is it known what changes were made?

sonora:
@Bubu Build is currently here http://download.osmand.net/latest-night-build/OsmAnd-design.apk, build number should end with Des3049. Please be quick downloading it, it may be ovrrerwitten by other tests. Compare vs. latest release, I would say.

sonora:
@vshcherb  I almost think we should temporarily commit this to master to get a much more solid test from more people and on more devices. We could always back out later, at least the 2 huge jars?

Bubu:
Android studio has a nice apk comparison view:

image

So yes, icu is by far the largest contributor to the size increase. followed by gnu trove.
Is their some method to only include parts of a jar file?

vshcherb:
I remember we trimmed icu manually cause we needed only *transliteration* & *reshaping* data where icu contains much more than that. Anyway I think we could possibly make pull-request & delete all other jars except these 2.

yourealwaysbe:
Do you know if [Proguard](https://www.guardsquare.com/en/proguard) would perform the required optimisation?  There are already Proguard configuration lines commented out in OsmAnd/build.gradle.  My initial experiments, after configuring it to ignore a lot of warnings, built an 80mb apk with assembleFullLegacyFat, which is the same as without Proguard.  So either Proguard is not a solution or i don't know how to use it.

Another alternative is to write Gradle tasks to strip everything we don't need out of the jars.  This sounds horribly hacky though...

Bubu:
The icu jar includes a lot of unneded ressources, not sure if proguard can do anything about this.

yourealwaysbe:
There's a shrinkResources option, but it did not seem to make a difference.

Zahnstocher:
> I remember we trimmed icu manually cause we needed only transliteration & reshaping data where icu contains much more than that. Anyway I think we could possibly make pull-request & delete all other jars except these 2.

I think icu jar could also be deleted, because it is possible to remove unneeded data with gradle, just add this to OsmAnd/build.gradle (this is for icu4j 50.1):

>        packagingOptions {
>                 exclude 'com/ibm/icu/impl/data/icudt50b/brkitr/**'
>                 exclude 'com/ibm/icu/impl/data/icudt50b/coll/**'
>                 exclude 'com/ibm/icu/impl/data/icudt50b/curr/**'
>                 exclude 'com/ibm/icu/impl/data/icudt50b/lang/**'
>                 exclude 'com/ibm/icu/impl/data/icudt50b/rbnf/**'
>                 exclude 'com/ibm/icu/impl/data/icudt50b/region/**'
>                 exclude 'com/ibm/icu/impl/data/icudt50b/translit/**'
>                 exclude 'com/ibm/icu/impl/data/icudt50b/zone/**'
>         }

> There's a shrinkResources option, but it did not seem to make a difference.

This is only for /res folder.

yourealwaysbe:
@Zahnstocher thanks for this.  It works for me.  The debug jar is 6mb smaller (74mb).  I tested it for a minute and didn't see any immediate problems.

Bubu:
@yourealwaysbe can you adjust the PR here?

yourealwaysbe:
@Bubu -- will do, i am testing trimming trove4j using @Zahstocher's trick first.

yourealwaysbe:
@Bubu -- hopefully this is now done

Bubu:
Hey, any news on this? Can we make another testing build so that we see what the size difference is now with the amended PR?

AlexSyt:
I merged all changes from this PR and new changes from master to this branch:
https://github.com/osmandapp/Osmand/tree/maven-deps

The APK size is still larger by 3.2 mb.

AlexSyt:
To accept these changes it is necessary that the size of APK does not increase by more than 1 megabyte

matthewhague:
Thanks for this.  I will report back if/when i am able to reduce the apk size.

vshcherb:
Cause this pull request doesn't have any functionality added it would be great to not get increase by 3 MB, on the other hand of course we are fine to make the dependencies conditionalized or forked but again it is not the most preferrable option in sense of the clean code.

Please don't look at 70 MB full options build cause Google Play build which we are targetting to have the smallest size has smaller size (30-40 MB)

vshcherb:
BTW 
Some work was already done in background i.e. 
https://github.com/osmandapp/Osmand/commit/f7f3e2448d856b86fc2c08dd2b9ae73bc27f312e
https://github.com/osmandapp/Osmand/commit/a00ad99449607a2d4cfc1d5b8809c1c36a139b27

The size increase was negligible.
@yourealwaysbe Please adjust pull request for gnu-trove and icu.

yourealwaysbe:
@vshcherb thanks for this. I'll take a look when i can (been super busy lately).

vshcherb:
Major part of this PR is implemented we have left only the big jars trim which is not clear how to proceed so far

PR #4752, opened 2017-11-14
Remove probably superfluous commas in strings.xml

aceman444:
Remove probably superfluous commas in strings.xml


sonora:
http://www.getitwriteonline.com/archive/052709serialcomma.htm

aceman444:
Interesting, thanks.

aceman444:
But somebody has done this in strings for iOS version and it was accepted :)

sonora:
Yeah, hard to keep things in sync, and I really hardly have enough time to maintain some consistency in our Android code ... :)

Fact remains that in English it is conventional (I think almost mandatory) to have the serial comma, while I am aware that is some other languages like e.g. German it is definitely the opposite.

PR #4793, opened 2017-11-21
Change dashboard animations

PavelRatushnyi:


vshcherb:
Some issues here:
A) open animation is completely different than close animation
B) Top toolbar probably should not be animated or should be animated by sliding from top to down (so animation should consist of 2 parts probably)
C) It would be great to consolidate with Search screen animation or other screens opening from the Map screen

PavelRatushnyi:
Thank you for feedback, I'll try to fix these issues

Issue #4804, opened 2017-11-25
User Story: Terrain (Ascent) Aware Hiking Time


sonora:
As a spin-off of Issue #4455, the idea is for pedestrian based profiles to factor any ascent into the 'travel time needed' estimate. Naismith's rule of adding 1 hour of walking time per every 600 m of ascent is a well accepted lower limit for the actual walking time required in hilly terrain.

Basic idea is to use the `heightDistanceArray` for every `RouteDataObject` to calculate the ascent (or any remaining ascent) on the segment, then apply Naismith's rule to add an ascent penalty to `getLeftTime` in `RouteCalculationResult`. To be investigated if this can be done with simple methods  or needs CPU intensive GPX analysis code each time (`analysis.diffElevationUp`).

Then, as per Zahnstocher: The only change needed for routing engine is a change of routing.xml, just change `obstacle_srtm_alt_speed` section of pedestrian to:

```
  <way attribute="obstacle_srtm_alt_speed">
  	<if param="height_obstacles">
  		<gt value1="0" value2=":incline">
  			<select value="0"/>
  		</gt>
  		<select value="6"/>
  	</if>
  </way>
```
Now you get 6 seconds penalty per ascent meter (= Naismith, 1h per 600m). Inclines smaller 1% are filtered out.

Zahnstocher:
> To be investigated if this can be done with simple methods or needs CPU intensive GPX analysis code each time (analysis.diffElevationUp).

I did an implementation of Naismith's rule for OsmAnd time calculation (ETA) for pedestrian mode.
It doesn't have a noticeable effect on the computation time (CPU).

I did some routing tests:
12km
30km
31km

With Naismith's rule for ETA computation:
25s
59s
46s

Without Naismith's rule for ETA computation:
25s
60s
46s

There are only some random time differences, less than one second.
Thus Naismith's rule for ETA computation is nearly for free.

sonora:
Ok, cool! Do you have a pull request yet, it would save me the time next weekend ... :)

Zahnstocher:
https://github.com/osmandapp/Osmand/pull/4805
It is hard coded, thus there is room for improvement. :)

sonora:
@Zahnstocher  Had a look and I think this should work ok. I have shortened it somewhat and committed to master, so we can test.

Thanks again!!

PR #4805, opened 2017-11-26
ascent aware hiking time (#4804)

Zahnstocher:


sonora:
Ok, thanks. This does not use the noise reduction and channel analysis I apply for the GPX track evaluation. But maybe it is not needed here as we are dealing with non-recorded data, so this could possibly be just the simplification I was looking for. Need to find some time to play with it.

PR #4827, opened 2017-12-04
add Lao language code

mbethke:
A trivial little change that would make my life a lot easier: add the ISO-639-1 code for Lao


sonora:
I have added the corresponding string, see https://github.com/osmandapp/Osmand/commit/db05fbac56ff586a79bcc89df15b5a98972512ca.

PR #4863, opened 2017-12-18
Fixed duplicated search results

PaulStets:


OsmAnd-java/src/net/osmand/binary/BinaryMapAddressReaderAdapter.java
```diff
@@ -707,7 +709,9 @@ public boolean matches(City city) {
 									}
 								}
 								if (matches) {
-									req.publish(s);
+									if (added.add(s.getLocation())) {
```
GaspardPO: Here, in the second if there is a the same time a test (is the Street already in the set) and an action (add a Street to a set).

Will it change a lot in term of performance if it is split in 2:

```java
if (matches &&  ! added.contains(s)) {
    added.add(s);
    req.publish(s);
```

OsmAnd-java/src/net/osmand/binary/BinaryMapAddressReaderAdapter.java
```diff
@@ -593,6 +594,7 @@ protected Building readBuilding(int fileOffset, int street24X, int street24Y, Li
 	}
 
 	public void searchAddressDataByName(AddressRegion reg, SearchRequest<MapObject> req, List<Integer> typeFilter) throws IOException {
+		Set<LatLon> added = new HashSet<>();
```
GaspardPO: Maybe the name "added" can be a little more explicit.
What does it really means?
(it contains the list of LatLong that were already found ?)

OsmAnd-java/src/net/osmand/binary/BinaryMapAddressReaderAdapter.java
PaulStets: Hi, according to the HashSet JavaDoc the ```add()``` method: Returns:
true if this set did not already contain the specified element. Querying the set would add an unnecessary extra step.

GaspardPO: Unnecessary for the compiler, but maybe necessary for the readers / maintainers.

If performance is an issue, it is better to use add.
But if not, it is probably better to have something easy to read and understand for further modifications. especially if this part of code does not have tests.

PR #4973, opened 2018-02-01
Pauls branch

PaulStets:


OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
```diff
@@ -202,9 +204,32 @@ public AlarmInfo getMostImportantAlarm(MetricsConstants mc, boolean showCameras)
...
+			inf.setFloatValue(calculateDistance(segments, mc));
+			int vl = inf.updateDistanceAndGetPriority(time, d);
+			if (vl < value && (showCameras || inf.getType() != AlarmInfoType.SPEED_CAMERA)) {
+				mostImportant = inf;
+			}
+		}
 		return mostImportant;
 	}
 
+	private float calculateDistance(List<RouteSegmentResult> segments, MetricsConstants mc) {
```
vshcherb: We have many MetricsConstants and km/m is only 1, this check is incorrect in case we will introduce MetricsConstants.METERS. The formatting is also in the wrong place should be in OsmAndFormatter

OsmAnd/src/net/osmand/plus/views/mapwidgets/RouteInfoWidgetsFactory.java
```diff
@@ -1271,6 +1274,15 @@ public boolean updateInfo(DrawSettings drawSettings) {
 						} else {
 							locimgId = R.drawable.warnings_pedestrian;
 						}
+					} else if(alarm.getType() == AlarmInfoType.TUNNEL) {
+						DecimalFormat df = new DecimalFormat("0.#");
+						if(settings.DRIVING_REGION.get().americanSigns){
```
vshcherb: Should be in OsmAndFormatter. alarm.getFloatValue() - should be always in meters!

OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
```diff
@@ -202,9 +204,32 @@ public AlarmInfo getMostImportantAlarm(MetricsConstants mc, boolean showCameras)
 				kIterator++;
 			}
 		}
+		List<RouteSegmentResult> segments = route.getUpcomingTunnel(500);
```
vshcherb: Why length of tunnel everytime is recalculated? Why that AlarmInfo is dynamic??? It means it is not displayed in the waypoints alarm list before you get close to the tunnel

PR #4975, opened 2018-02-01
Fix #4740

### LINKED ISSUE #4740: show osm notes later/higher zoom (configurable)
I can overlay osm notes. 
But they are shown much too early and clutter the map.
Personaly i want them at the 500m level or even later.

AlexSyt:


OsmAnd/res/values/strings.xml
```diff
@@ -9,6 +9,8 @@
          3. All your modified/created strings are in the top of the file (to make easier find what\'s translated).
     PLEASE: Have a look at http://code.google.com/p/osmand/wiki/UIConsistency, it may really improve your and our work  :-)  Thx - Hardy
     -->
+    <string name="show_closed_notes">Show closed notes</string>
+    <string name="switch_osm_notes_visibility_desc">Show/Hide OSM Notes on the map.</string>
```
HolgerJeromin: Am i wrong or is this string definition not matching the code?

OsmAnd/src/net/osmand/plus/OsmandSettings.java
```diff
@@ -1155,6 +1155,8 @@ protected Integer getDefaultValue() {
 
 	// this value string is synchronized with settings_pref.xml preference name
 	public final OsmandPreference<Boolean> SHOW_OSM_BUGS = new BooleanPreference("show_osm_bugs", false).makeGlobal();
+	public final CommonPreference<Boolean> SHOW_CLOSED_OSM_BUGS = new BooleanPreference("show_closed_osm_bugs", false).makeGlobal();
+	public final CommonPreference<Integer> SHOW_OSM_BUGS_MIN_ZOOM = new IntPreference("show_osm_bugs_min_zoom", 8).makeGlobal();
```
HolgerJeromin: Settings code here without matching string.xml?

OsmAnd/res/values/strings.xml
AlexSyt: These are the strings for UI.

OsmAnd/src/net/osmand/plus/OsmandSettings.java
AlexSyt: These settings should not match anything. I'll add an empty line to separate them.

HolgerJeromin: I thought the first parameter is used to find the appropriate localization (and therefor has to match name in string.xml).
But ok. I am no android expert

PR #5048, opened 2018-02-21
added transport icons by stop type inside Transport Context Menu.

Chumva:
inside TransportStopsLayer made a methods getIconIdByTypeOfStop and overlayBitmapToCenter

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/TransportStopsLayer.java
+++ b/OsmAnd/src/net/osmand/plus/views/TransportStopsLayer.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Canvas;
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/views/TransportStopsLayer.java
```diff
@@ -48,6 +53,10 @@
 
 	private boolean showTransportStops;
 	private Path path;
+	private int backgroundIconWidth;
```
AlexSyt: Can be local variable

OsmAnd/src/net/osmand/plus/views/TransportStopsLayer.java
```diff
@@ -48,6 +53,10 @@
...
+	private int backgroundIconHeight;
```
AlexSyt: Can be local variable

PR #5162, opened 2018-03-24
Spelling: Language rework

comradekingu:


sonora:
Thanks for your work! A lot of it makes sense. But some of it changes some long- used expressions and that may need a lot more amalysis, also checks with web pages and documentation. I have made a quick corrective review and will merge this lighter version for now.

Things I removed is gratis for free, libre for open-souce. Debatable is all replacement of screen center by middle (as our center may not be the middle). Renaming app profiles to Driving/Cycling/Walking may well work, not sure if it csuses complications.

OsmAnd/res/values/strings.xml
```diff
@@ -289,9 +289,9 @@
     <string name="quick_action_add_destination">Add destination</string>
     <string name="quick_action_replace_destination">Replace destination</string>
     <string name="quick_action_add_first_intermediate">Add first intermediate</string>
-    <string name="quick_action_add_destination_desc">Tap the action button to make your previous destination into an intermediary waypoint to a new destination in the centre of the screen.</string>
-    <string name="quick_action_replace_destination_desc">Center the view away from the destination and onto yourself by tapping the action button.</string>
-    <string name="quick_action_add_first_intermediate_desc">Add an intermediate waypoint before your destination in the middle of the screen by tapping the action button.</string>
+    <string name="quick_action_add_destination_desc">Tap the action button to add a destination at the screen center location. The previously set destination becomes the last intermediate destination.</string>
```
comradekingu: This is two unlinked sentences, I would think it to be harder to understand.

OsmAnd/res/values/strings.xml
```diff
@@ -289,9 +289,9 @@
...
+    <string name="quick_action_replace_destination_desc">Tap the action button to replace the destination with the screen center location.</string>
```
comradekingu: "Tap the action button" is not the info that is conveyed. I see I did the same above, but nontheless. It is hard for me to read this sentence. It sounds a bit like the destination is moved, is that the case?

OsmAnd/res/values/strings.xml
```diff
@@ -289,9 +289,9 @@
...
+    <string name="quick_action_add_first_intermediate_desc">Tap the action button to add a first intermediate point at the screen center location.</string>
```
comradekingu: Very hard to understand this sentence, not sure I even did. These are the questions that need answering, and they are not answered until the end.
1. Why should I tap the action button?
2. Are there any second intermediate points?
3 Intermediate to what?
4. Don't know what "the screen center location" is clear about, but it sounds clear somehow.

OsmAnd/res/values/strings.xml
```diff
@@ -343,9 +343,9 @@
     <string name="shared_string_tracks">Tracks</string>
     <string name="routing_attr_driving_style_name">Driving style</string>
     <string name="route_altitude">Route elevation</string>
-    <string name="altitude_descent">◺</string>
-    <string name="altitude_ascent">◿</string>
-    <string name="altitude_range">◿◺◺◿</string>
+    <string name="altitude_descent">Descent</string>
```
comradekingu: I seem to condescend on the notion of whether maybe the combination "◺ Descent"makes it easier to understand?

OsmAnd/res/values/strings.xml
```diff
@@ -443,8 +443,8 @@
     <string name="save_track_precision">Logging minimum accuracy</string>
     <string name="save_track_precision_descr">Filter: No logging unless this accuracy is reached.</string>
     <string name="christmas_poi">Christmas POI</string>
-    <string name="christmas_desc">Anticipating Yuletide and New Year holidays, you can choose to display the POI associated with Yule: Yule-trees, markets, etc.</string>
-    <string name="christmas_desc_q">Show Yuletide POI?</string>
+    <string name="christmas_desc">Anticipating Christmas and New Year holidays, you can choose to display the POI associated with Christmas: Christmas trees, markets, etc.</string>
```
comradekingu: Never! :)

OsmAnd/res/values/strings.xml
```diff
@@ -543,9 +543,9 @@
     <string name="map_widget_magnetic_bearing">Magnetic bearing</string>
     <string name="map_widget_bearing">Relative bearing</string>
     <string name="access_disable_offroute_recalc">No route recalculation after leaving it</string>
-    <string name="access_disable_offroute_recalc_descr">Keeps the same route when you go offroad.</string>
+    <string name="access_disable_offroute_recalc_descr">No automatic route recalculation after going off route.</string>
```
comradekingu: Offroad isnt the word, but still this could be easier to understand.

OsmAnd/res/values/strings.xml
```diff
@@ -543,9 +543,9 @@
...
     <string name="access_disable_wrong_direction_recalc">No route recalculation for opposite direction</string>
-    <string name="access_disable_wrong_direction_recalc_descr">Keeps the same route when you move in the fully wrong direction.</string>
+    <string name="access_disable_wrong_direction_recalc_descr">Prevents automatic route recalculation while just moving in the opposite direction.</string>
```
comradekingu: "Preventing automatic route recalculation while" is more complex than
"Keeps the same route when"

comradekingu:
@sonora :)  Did the whole thing in one go, trying to remember what greivances ive had with the app as I did, so quite taxing. Made some comments for the extra potential to be found.

Using "free" without "for free" or "free of charge", in an app that is Free Software, makes for errors in translation, and a lot of confusion as to what is going on.
"Libre software" fixes this, in the same way the duality doesn't exist for other languages. Also libre focuses on the added freedoms, for the user. "Open Source" is selling it short IMO.

Could you explain why "middle of the screen" is not the same as "screen centre"?
Is it because of software buttons? Or maybe the side-menu?
In the middle of the visible map? Colour me confused.

Pedestrian is harder to understand than walking, that was what I was going for.
There is no benefit to the more complex sentences that I can see, and it makes it hard for children to understand what is going on, when they otherwise would have no problem.

File name/Filename  as per https://en.wikipedia.org/wiki/Filename
Bug fix/Bugfix as per https://en.wikipedia.org/wiki/Patch_(computing)

Going with respectively "Filename" and "Bugfix" avoids compound errors in translation for other languages.

WiFi/Wi-Fi as per https://en.wikipedia.org/wiki/Wi-Fi

Since one never sees "WiFi", it strikes me as wrong.

sonora:
Thanks! Will review next week, currently out travelling. Please note that spellingwise this is the American English file, there is a file for any British English deviations. Looks like filename, bugfix may be more BE?

Screen 'center' for movement direction may be off the screen middle towards the bottom, hard to find good wording here, as center is also not clear, it is the logical rather than the geometric meaning.

Profiles: We have more, there are/will be also truck, motorcycle, bus, ship etc.

The destnation options are about where in the order the nee destination is added, and what happens to existing ones. Concur that brevity is key here.

Ascent: your special characters seem somehow swallowed by the web interface?

sonora:
I have taken a shot at it, maybe you want to review https://github.com/osmandapp/Osmand/commit/8ad13b3d3e8b8d40c876a23bba00aaab4c7a678c. I hope it fixes more issues than it creates.  :)

Regarding some of the other ideas you had, I have edited my post above. Ok?

For new ideas it is probably better to start a new thread for clarity. But any past and future feedback is much appreciated, most people underestimate the power of good wording for app clarity.   :)

PR #5281, opened 2018-04-25
Add map widget to show arrival time at first intermediate point

schrod:
This should probably get new icons as well, but I don't have any artistic skill.


AlexSyt:
Thank you for your work, i will add new icons.

PR #5303, opened 2018-04-29
Language rework 3

comradekingu:


sonora:
Thanks for your work! Quickly reviewed and like95 percent. Here are some points I noted for a second opinion or double check:

 - Use of libre, copyleft, gratis, I leave this to @vshcherb to judge. I was not familiar with that convention.
 - You removed string 'Remove the tag', have not checked if this is not needed
 - Where Point(s) was used, it referred to singular or plural
  - `<string name="empty_state_markers_history" translatable="false">`: not sure if this works
 - 'tracking method 'instead of 'location provider' may cause ambiguities, not sure
 - 'Favorites' sometimes changed to lower case, in other places exactly the other way round? We mostly used upper case for expressions with specific meaning or menu in the app

vshcherb:
I'm not even sure if copyleft concept is legal in some countries but it is not applied. OsmAnd is distributed as completely open-source software but it has some limitation so it couldn't be called copyleft (there is even copyright for changes done by OsmAnd developers) . The license it is distributed quite permissive GPLv3 but it is not libre in some sense.

comradekingu:
@goedson 
The string says plural, so i imagined there was a singular version too. The whole point is to avoid (s). If it is plural, the user should see the plural form. If it is singular, it should show as singular.
translatable="false" is how to not allow translations. It is used for names that aren't meant to be changed, etc. I don't know what the "empty_state_markers_history" string is, but changing it looks like trouble.
min/max is shorter, and in tune with using "misc", etc.
Can't remember why I went with "tracking method".

@vshcherb 
GPLv3+, _is_ copyleft, which is a libre software license. (Which creators still hold copyright to). Libre means nothing other than "free as in freedom software", otherwise invoking the ambiguity of "free software".
That is a crucial level of freedoms to have 0-3. GPLv3+ is even 0-4.
When people translate that as "gratis software", a great point is lost, to the tune of seeming like freeware, and that is not very accurate, nor the intention.
The amount of freedoms the user gets, ranges from permissive (to remove user freedoms), to restrictive licenses (protections of user freedom). With respectively BSD-type things on one side, and more freedoms and protections with AGPLv3+.
Fully open source, means strictly nothing more than the ability to see the code, freedom 1, that is not pointless, but very much short of ideal.
What the license for the map edits are for OSM, is not part of the license for OsmAnd.
You could substitute a different map service, and it would still work.

goedson:
@comradekingu I'm missing context on why I've been mentioned here. Did I change this file in the past? Or maybe translated it?

comradekingu:
@goedson Ooops, went with the wrong nick, sorry about that.

PR #5370, opened 2018-05-10
New strings reworked 2

comradekingu:
What does "aerial ways" mean, and what type of skiing is it?
Edit: Can the wiki be imported here?


sonora:
Thanks!

Side note: For BE spelling I once created a separate _enGB language file where deviating spelling can be maintained.

sonora:
PS: 'aerial ways' as in the OSM terminolog, I assume, please check their wiki on related tags.
PS2: I guess you are talking about what we have on code.google.com from years back? I will take a look and maybe transfer to osmand.net what seems still useful.

PR #5398, opened 2018-05-13
Pauls branch

PaulStets:


OsmAnd-java/src/net/osmand/map/OsmandRegions.java
```diff
@@ -695,11 +694,53 @@ private void sortSubregions(WorldRegion region, Comparator<WorldRegion> comparat
 		}
 	}
 
+	public List<WorldRegion> getWoldRegions(LatLon latLon) {
+		List<WorldRegion> result = new ArrayList<>();
+		List<BinaryMapDataObject> mapDataObjects = null;
+		try {
+			mapDataObjects = getBinaryMapDataObjects(latLon);
+		} catch (IOException e) {
+			e.printStackTrace();
```
vshcherb: Android application should never use e.printStackTrace(), only Log.error

OsmAnd-java/src/net/osmand/map/OsmandRegions.java
```diff
@@ -695,11 +694,53 @@ private void sortSubregions(WorldRegion region, Comparator<WorldRegion> comparat
 		}
 	}
 
+	public List<WorldRegion> getWoldRegions(LatLon latLon) {
```
vshcherb: This method doesn't look correct it and almost copy/paste findBinaryMapDataObject

vshcherb:
The code looks quite unstructured and looks like the bug flow will continue

PR #5424, opened 2018-05-17
Do not save city_id for wikivoyage search history

AlexSyt:


OsmAnd/src/net/osmand/plus/wikivoyage/data/TravelLocalDataHelper.java
```diff
@@ -35,8 +35,16 @@ public void setListener(Listener listener) {
 		dbHelper = new WikivoyageLocalDataDbHelper(app);
 	}
 
-	void refreshCachedData() {
-		historyMap = dbHelper.getAllHistoryMap();
+	void refreshCachedData(TravelDbHelper travelDbHelper) {
+		List<WikivoyageSearchHistoryItem> history = dbHelper.getAllHistory();
+		historyMap = new TLongObjectHashMap<>(history.size());
+		for (WikivoyageSearchHistoryItem item : history) {
+			long id = travelDbHelper.getArticleId(item.articleTitle, item.lang);
```
vshcherb: Doesn't look good and definitely won't work fast. We shouldn't use tripId at all, I guess

PR #5455, opened 2018-05-21
Add pathes for fileprovider

Chumva:


OsmAnd/src/net/osmand/AndroidUtils.java
```diff
@@ -112,7 +112,7 @@ public static String trimExtension(String src) {
 	}
 
 	public static Uri getUriForFile(Context context, File file) {
-		return FileProvider.getUriForFile(context, context.getPackageName() + ".fileprovider", file);
+		return FileProvider.getUriForFile(context, "net.osmand.plus.fileprovider", file);
```
vshcherb: Doesn't look correct at all

PR #5483, opened 2018-05-24
Wikivoyage sharing links

Chumva:


OsmAnd/AndroidManifest.xml
```diff
@@ -756,7 +756,35 @@
 		</receiver>
         <activity android:name=".liveupdates.OsmLiveActivity"
             android:label="@string/osm_live"/>
-        <activity android:name=".wikivoyage.explore.WikivoyageExploreActivity"/>
+		<activity android:name=".wikivoyage.explore.WikivoyageExploreActivity">
+
+			<intent-filter>
```
vshcherb: Do we need 2 intent-filter or can we combine into one with 2 android:scheme

OsmAnd/res/values/strings.xml
```diff
@@ -10,6 +10,7 @@
 	- For wording and consistency, please note http://osmand.net/help-online?id=technical-articles#Creating_a_Consistent_User_Experience
 	Thx - Hardy
 -->
+    <string name="share_article">Share article</string>
```
vshcherb: Let's use just string Share.

vshcherb:
@PaulStets should create php script to redirect from https://osmand.net/travel? ... to wikivoyage.
So people who don't have OsmAnd still could see.

OsmAnd/src/net/osmand/plus/wikipedia/WikiArticleHelper.java
```diff
@@ -324,4 +324,9 @@ public static String getPartialContent(String content) {
 
 		return res.toString();
 	}
+
+	public static String buildTravelUrl(String url, String lang) {
+		String title = url.substring(url.indexOf(':') + 1);
+		return "https://osmand.net/travel?query=" + title + "&lang=" + lang;
```
vshcherb: There is no proper Url.Encode double check with @PaulStets how to encode

OsmAnd/src/net/osmand/plus/helpers/DiscountHelper.java
```diff
@@ -100,6 +104,8 @@ private static void processDiscountResponse(String response, MapActivity mapActi
 			String description = obj.getString("description");
 			String icon = obj.getString("icon");
 			String url = parseUrl(app, obj.getString("url"));
+			String travelUrl = parseUrl(app, obj.getString("travel"));
```
vshcherb: These fields to delete

PR #5571, opened 2018-06-13
English string correctness, typo and style fixes

### LINKED ISSUE #5566: Please revert wording from "trespass private roads" back to "allow access to private roads"
What version of OsmAnd are you using, on what device and Android/iOS version?

OsmAnd+ 3.0.3, released 2018-06-02, Samsung Galaxy S5, Android 5.0.

Previous versions of OsmAnd would prompt the user to "allow access to private roads" when the destination required traveling on a road with access=private (or possibly also when it was necessary to pass through a gate that was access=private).  Apparently the wording has been changed to "trespass private roads."  Please consider changing this back to the earlier wording.  The English word "trespass" means to enter property without permission, and carries with it the strong connotation of an illegal act (the crime of "criminal trespass" is in the law of every U.S. state, for example).  The earlier wording was "nicer" and didn't make the user think he would be breaking the law :-) .

njohnston:
Commits e14d7518 and 554e433c changed many English strings. Many of the changes (use of active voice, clearer phrasing, some simplification) were good, but others introduced unusual style, inconsistencies in the UI, and strange language.

Some examples:

* import_gpx_file_description: was "can be imported as Favorites, or as track file.", changed to "can be imported as favorite, or track file.". The new phrase doesn't take into account that OsmAnd has a concept or feature called "Favorites", so when referring to that, an initial capital letter should be used, and that a track can have many waypoints, so it's more appropriate to use "Favorites" (plural).
* Several strings containing colons have been changed such that the initial letter of the word immediately following the colon is in uppercase. For example, osm_live_banner_desc now contains "once a month: Weekly, daily, or hourly". Capitalising the first letter in "Weekly" is not correct.
* Inconsistencies have been introduced, for example changing "underlay" (as in "underlay map" to "underlaying"). The "Configure map" menu now is now inconsistent as it contains "Overlay map" and "Underlaying map".
* Typos have been introduced, for example "doens\'t" instead of "doesn\'t".
* "Start time" and "End time" (used in GPX analysis, for example) are now "Departure" and "Arrival" which makes little sense in this context.

Also fixes #5566.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -54,8 +54,8 @@
     <string name="welcome_to_open_beta">Welcome to the open beta</string>
     <string name="wikivoyage_travel_guide">Travel Guides</string>
     <string name="wikivoyage_travel_guide_descr">Guides to the most interesting places on the planet, inside OsmAnd, without a connection to the Internet.</string>
-    <string name="monthly_map_updates">Map updates: <b>Every month</b></string>
-    <string name="daily_map_updates">Map updates: <b>Every hour</b></string>
+    <string name="monthly_map_updates">Map updates: <b>every month</b></string>
+    <string name="daily_map_updates">Map updates: <b>every hour</b></string>
     <string name="in_app_purchase">In-app purchase</string>
     <string name="in_app_purchase_desc">One-time payment</string>
     <string name="in_app_purchase_desc_ex">Once purchased, it will be permanently available to you.</string>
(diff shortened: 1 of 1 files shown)


sonora:
I tend to agree that while there were a number of valid improvements and corrections, some of the overall consistency we had built up over the years was sacrificed in those commits. Let me merge your work, and should there be concerns, let us tackle that one by one subsequently. Also, can I please once again point out that many years ago already I had introduced a language file for British English wording, spelling, and capitaization, so let us please keep the main en file for American English and use en-GB for any deviations. Thx!

PR #5611, opened 2018-06-24
Voice guidance from "General" to "Navigation settings"

comradekingu:
As reported by https://hosted.weblate.org/user/franco999/ :)

Edit: If https://hosted.weblate.org/user/kingu/ is added as administrator for OsmAnd on Weblate, I can clean up old issues that I have tended to already.


franco999:
In addition, in https://github.com/osmandapp/Osmand/blob/master/OsmAnd/res/values/strings.xml#L326 it should be "Distance travelled" instead of "Distance corrected".

sonora:
"Distance traveled" seems kind of overspecified. What this value really reports is a "corrected" distance to the effect that displacements are only counted while the GPS chipset reports the device being in motion.

sonora:
It is already fixed, thanks!

PR #5624, opened 2018-07-01
Updated sv (Swedish) translation

nyqvist:
Commit made via [Stringlate](https://lonamiwebs.github.io/stringlate/)


comradekingu:
@nyqvist OsmAnd is translated via https://hosted.weblate.org/projects/osmand/main/sv/
I imagine strings will be overwritten from there if updated via Git.
If need be I can update the Swedish strings on Weblate wrt. to the changes made here.

nyqvist:
Hi,

OK I see! Thanks for the web link. I will make my updates there instead.

PR #5714, opened 2018-07-20
Fix_5651

Chumva:


vshcherb:
Pull request looks quite complex but logically good. If it is tested and we don't expect any NPE in the middle we should merge it

PR #5746, opened 2018-07-29
String typo fix, and add some missing en-rGB strings

njohnston:


njohnston:
@vshcherb: thanks for merging. Could you please merge it into the r3.1 branch too? :) Thanks

PR #5766, opened 2018-08-01
Language rework 4

comradekingu:


OsmAnd/res/values/strings.xml
```diff
@@ -12,21 +12,21 @@
...
 
     <string name="thank_you_for_feedback">Thank you for feedback</string>
-    <string name="poi_cannot_be_found">Node or way cannot be found.</string>
-    <string name="search_no_results_feedback">No search results?\nGive us feedback</string>
+    <string name="poi_cannot_be_found">Could not find node or waypoint.</string>
+    <string name="search_no_results_feedback">No search results?\nProvide feedback</string>
     <string name="release_3_1">
     • Navigation: Fix progress bar, fast swapping of the start and end point of the route\n\n
-    • Map markers: fix turn on/off groups, ability to hide markers from the map\n\n
+    • Map markers: Fix turn on/off groups, ability to hide markers from the map\n\n
```
njohnston: Why include an initial uppercase letter on the first word after a colon? This is not common style.

OsmAnd/res/values/strings.xml
```diff
@@ -12,21 +12,21 @@
...
+    • Boat navigation: Support for waterway fairway\n\n
     • Other bug fixes\n\n
     </string>
-    <string name="commiting_way">Committing way…</string>
-    <string name="increase_search_radius_to">Increase search radius to %1$s</string>
-    <string name="send_search_query_description"><![CDATA[We will send your search query: <b>\"%1$s\"</b>, as well as your location.<br/><br/>
-        We do not collect personal information, we only need search data to improve the search algorithm.<br/>]]></string>
+    <string name="commiting_way">Committing waypoint…</string>
+    <string name="increase_search_radius_to">Extend search radius to %1$s</string>
+    <string name="send_search_query_description"><![CDATA[Your search query is sent: <b>\"%1$s\"</b>, as well as your location.<br/><br/>
```
njohnston: This changes the meaning of the message. If I understand the code correctly, the message is shown *before* the search query is sent, so the previous message was more correct.

OsmAnd/res/values/strings.xml
```diff
@@ -36,7 +36,7 @@
     <string name="coord_input_save_as_track_descr">You added %1$s points. Type a filename and tap \"Save\".</string>
     <string name="error_notification_desc">Please send a screenshot of this notification to support@osmand.net</string>
     <string name="quick_action_edit_actions">Edit actions</string>
-    <string name="get_osmand_live">Get OsmAnd Live to unlock all features: Daily map updates with unlimited downloads, all paid and free plugins, Wikipedia, Wikivoyage and much more.</string>
+    <string name="get_osmand_live">Get OsmAnd Live to unlock all features: Daily map updates with unlimited downloads, all paid and gratis plugins, Wikipedia, Wikivoyage and much more.</string>
```
njohnston: "free" is much more commonly used than "gratis" in English.

OsmAnd/res/values/strings.xml
```diff
@@ -686,7 +686,7 @@
     <string name="rendering_value_bold_name">Bold</string>
     <string name="no_map_markers_found">Please add map markers via map</string>
     <string name="no_waypoints_found">No waypoints found</string>
-    <string name="anonymous_user_hint">An anonymous user cannot:\n- Create groups;\n- Sync groups and devices with the server;\n- Manage groups and devices in a personal dashboard on the website.</string>
+    <string name="anonymous_user_hint">Anonymous users can not:\n- Create groups;\n- Sync groups and devices with the server;\n- Manage groups and devices in a personal dashboard on the website.</string>
```
njohnston: "can not" should be "cannot". See the [Oxford English Dictionary page about this](https://en.oxforddictionaries.com/usage/cannot-or-can-not).

OsmAnd/res/values/strings.xml
```diff
@@ -715,7 +715,7 @@
     <string name="osm_editors_ranking">OSM Editors ranking</string>
     <string name="osm_live_subscription">OsmAnd Live subscription</string>
     <string name="osm_live_subscribe_btn">Subscribe</string>
-    <string name="osm_live_email_desc">Needed to provide you information about contributions.</string>
+    <string name="osm_live_email_desc">Needed to provide you info about contributions.</string>
```
njohnston: For a message like this, I think using the full word "information" is more appropriate.

OsmAnd/res/values/strings.xml
```diff
@@ -1508,7 +1508,7 @@
     <string name="routing_attr_weight_description">Specify permitted vehicle weight limit on routes.</string>
     <string name="routing_attr_height_name">Height limit</string>
     <string name="routing_attr_height_description">Specify vehicle height to be permitted on routes.</string>
-    <string name="android_19_location_disabled">Since Android version 4.4 (KitKat) you cannot download and update maps to the old storage folder (%s). Copy all OsmAnd files to new storage location?
+    <string name="android_19_location_disabled">Android 4.4 (KitKat) and later versions can not download or update maps to the old storage folder (%s). Copy all OsmAnd files to new storage location?
```
njohnston: This changes the meaning of the message. It implies that Android itself cannot download files to the old storage folder, when the intent of the message is to explain that due to changes in Android, *OsmAnd* can't download files to the old storage folder. And again, "can not" should be "cannot".

OsmAnd/res/values/strings.xml
comradekingu: Both are allowed in English, whereas in other languages smaller versal following colon is an active error. In effect doing it this way produces fewer errors in translation.

OsmAnd/res/values/strings.xml
comradekingu: Free is ambiguous, gratis is not. An error in translation here is on the contrary wrong 100% of the time.

OsmAnd/res/values/strings.xml
njohnston: From what I've read, an initial uppercase letter after a colon is generally only used occasionally. I think it's poor style, because it removes the ability to indicate proper nouns.

I don't think it's acceptable to change English strings to worse ones to ensure better translation. If you're concerned at translators taking the English strings too literally, perhaps more guidance could be given, but why should the English strings suffer?

OsmAnd/res/values/strings.xml
njohnston: Again, I don't think it's acceptable to use "gratis" here just to ensure that there are better translations. I strongly suspect any native English speaker would use "free" here. And while "free" has multiple meanings, the meaning here is clear from the context.

It's also important the interface uses terms which users are familiar with. In previous pull requests you changed "Christmas" to "Yuletide". Thankfully this was reverted before it was included in a release. Had it not been, I imagine OsmAnd users would be somewhat perplexed.

Using "gratis" here is not a good idea.

OsmAnd/res/values/strings.xml
comradekingu: Proper nouns are still in proper form that way, with the possibility of being misunderstood, for a somewhat slippery slope of allowing both forms in a language, or clear error where adherence to consistency in using only colon bigversal is maintained.(Like with OsmAnd at large, this change now reflecting that.)
In any event care could be taken when making a proper noun, or incorporating a badly made one in a sentence. I would argue big versal following colon isn't much of an issue in English stylistically, whereas it really is in other languages.
The sentence reads "Fix turn on/off groups", it should imo. read "turning", and it is a bigger fish to fry.

njohnston: With your style, proper nouns are in proper form, but only as a consequence of the blind rule that the first letter of a word following a colon should be in uppercase. How can you tell that the word is a proper noun?

If you don't use an initial uppercase letter for the first word following a colon, you're:

* following the most common style in English
* preserving the ability to indicate a proper noun

As such, I see little argument for using an initial uppercase letter for the first word after a colon. Leave the translations to experts in other languages. If someone makes a mistake there, others can always suggest an improvement. Please don't make the English strings worse out of a well-intentioned idea for helping translators. I'd argue that the best translations will come from looking at the source code and application itself to see how the strings are used.

OsmAnd/res/values/strings.xml
comradekingu: The choice of precision is clearly one made by some English speakers, otherwise the use of "gratis" would end, notwithstanding, the word itself seems to be understood by at-least an acceptable demographic. The feeling of familiarity is a valid concern, but pales in comparison to the potential error. The context is a piece of software that is free software, a quality plugins could or could not have. In this case it isn't very likely, so no solution is actually ideal. Other than potentially just "all plugins", which would be my choice. Maybe making a comment if not so.

In changing "Christmas" to "Yule" (an English word derived from the indigenous season called, "yuletide") in English, consistency is found with the OsmAnd instance of "a market that takes place during this time", corrected to the effect of "yule-market", is the correct use for that particular facet of it. Thus it would be both correct and consistent. Similarly, the tale of Jesus in the stable is not one of the Yule-stories, it is a Christmas story. My rationale is representing peoples traditions accurately. "Somewhat perplexed" trumps misrepresenting culture no matter how common the error.

I make the argument, and grant changes I don't agree with.

OsmAnd/res/values/strings.xml
comradekingu: You could put it in italics, quote it, or just write your way around ( on the assumption that neither is correct. Never had the problem to begin with.

The "this is not the most common style in English", is not a major upset, whereas seeing a small versal following a colon in other indo-european languages is. Thankfully it is consistent, so this error can be avoided altogether by just opting for colon bigversal.

Translations in this case are just equally good, in being correct, as opposed to being, as a causal effect, fully wrong. The libre software translators are not experts, nor am I. The result is invariably, always the same. It is a trap that works.

Yes, it could change in time, but doesn't. This level of consistency is predicated on innate consideration. Give it enough time, and that makes for an academic argument. Meanwhile, it is a small concession to make, which increases the perceived quality of the product by a lot in languages other than English. At no point in time should anyone be exposed to it.

Also, it is how every other string is styled already. If you think about it, I am that person that comes along and ensures consistency. This was the case before I made any changes too. Now nobody has to see errors, because this hasn't landed yet.
In any event, doing both in one corpus is clearly worse than either.

sonora:
Thank you for the discussion! This article summarizes pretty well what I have seen on the subject of capitalization after a colon over the years:  https://www.grammarly.com/blog/capitalization-after-colons/

Looks to me like the line referenced above is a rather clear case for lower case? And yes, there are some inconsistencies here in our current American English strings.xml file, I would love to see a pull request cleaning this up if anyone can find the time.

OsmAnd/res/values/strings.xml
```diff
@@ -1691,14 +1691,14 @@
...
-    <string name="recording_can_not_be_played">Recording can not be played</string>
+    <string name="recording_can_not_be_played">Could not play recording.</string>
     <string name="recording_context_menu_delete">Delete recording</string>
     <string name="recording_context_menu_play">Play</string>
     <string name="recording_description">Recording %1$s %3$s %2$s</string>
     <string name="recording_default_name">Recording</string>
     <string name="map_widget_av_notes">Audio/video notes</string>
     <string name="map_widget_distancemeasurement">Distance measurement</string>
-    <string name="audionotes_location_not_defined">Location to associate with the note is not defined yet. \"Use location…\" to assign a note to the location specified.</string>
+    <string name="audionotes_location_not_defined">Tap \'Use location…\' to add a location to the note.</string>
```
comradekingu: It is not clear whether a note is added to a location or vice versa. Original does both.

comradekingu:
Found the commit that introduced inconsistencies as I removed a big/small versal error following a colon in translation.

comradekingu:
1. @sonora Favorites is a general feature in any program. As with any other title, like GPX files, it should be capitalized. (I think I am wrong there, but that is the concept, and from what I remember it wasn't consistent) "Copy Point/POI name" should either be "Copy point/POI name" or "Copy Point/POI Name".
2. "GPX waypoint" should not be "GPX Waypoint" outside of titles.
3. 
4. OSM note is already only "note", I could make it OSM note everywhere, though I don't know what extra clarity it brings over just "note".
5. https://en.wiktionary.org/wiki/favorited is a word.
6. The problem with "internet connection", is that the convention for capitalization is unclear across languages. Possibly another solution can be found, lets try to find it.
7. "addomg" shouldn't be there.
8. 
9. What is in your view a non-issue? Here are some solved issues.
10. One way to communicate non-possible actions, "could not" over "cannot" and "can not".
11. One way to do colon big/smallversal, with most instances avoided.
12. 2x "Coordinates search" fixed.
13. No "information", always "info", or "inform"
14. Consistent use of periods after sentences communicating something to the user.
15. Avoiding redundancy to reduce string length.
16. Consistent use of apostrophes ', over ". Alternatively, it is also possible to use " for concepts and ' for actual things.
17. Consistent use of apostrophes for quoting referenced menu items.
18. No unmarked noun sentences using presence. "Intermediate destination %1$s is too far from the nearest road."
19. Similarly "Download of country or region maps directly" fixed.
20. Unclarity about what respect OsmAnd is free in, fixed.
21. "points of interest" over "Points Of Interest, and "Point of interests"
22. "Underlay map" changed to "underlaying map". It is dubious originally, and now clearly not a verb.
23. Consistent use of "near"/"nearest" instead of "close"/"closest."
24. Some repetition avoided. Acknowledgning dialog with actions, rather than references. "Delete POI" → "Deleted"
25.

sonora:
As I said, I am lenient with some of your solutions, but by far not all, and this thread has become unmanageable. The road to progress is to please slice the elephant by opening small pull requests adressing one topic area at a time. Then people can reply and discuss just that topic in one place each, and it can be merged after agreement. Much better than choking by many diverse replies on a mammoth request. :)

PR #6066, opened 2018-09-19
Fix 5975

Chumva:


vshcherb:
We need to check if no documentation needs to be updated in https://github.com/osmandapp/osmandapp.github.io (please search for poi_types.xml)

PR #6100, opened 2018-09-26
Fixed #5921

### LINKED ISSUE #5921: Searching for a POI with 2 words leads to a very long list of results
Searching for POI ("Bank ABCD" as an example) with 2 words will show a very long list of results with "Bank ABCD" within a 500km radius. And Starting from 0 meters again, till 500km, with all the results with either the word "Bank" **or** "ABCD" too.

The problem with this is that if you tap on -Show results on map-:
1- you will get hundreds of banks that you are **not** searching for, starting with 0 meter radius
2- and the bank you searching for, but with 500km radius, which is not directly necessary

Suggestion for the above:
1- can OSMAND show -**only**- the results of the search inquiry "Bank" **AND** "ABCD" ?
2- and reduce the search radius to 1km, with the possibility to enlarge to x km?

Tested with OSMAND 3.1.6

PaulStets:


OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -61,14 +61,14 @@
 
 	// context less
 	public static final int SEARCH_LOCATION_PRIORITY = 0;
-	public static final int SEARCH_AMENITY_TYPE_PRIORITY = 100;
```
vshcherb: doesn't look correct.

OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -83,6 +83,8 @@
 	protected static final double SEARCH_AMENITY_BY_NAME_CITY_PRIORITY_DISTANCE = 0.001;
 	protected static final double SEARCH_AMENITY_BY_NAME_TOWN_PRIORITY_DISTANCE = 0.005;
 
+	static boolean USE_AMENITY_BY_NAME;
```
vshcherb: Doesn't look correct

PR #6106, opened 2018-09-28
avoid test failing on non en-US default-locale

ntruchsess:
OpeningHoursParserTest fails when default-locale is something that is incompatible with en-US.


ntruchsess:
closing as fix is integrated now

PR #6153, opened 2018-10-08
Move sharing settings to shareInfo

Chumva:


OsmAnd-telegram/src/net/osmand/telegram/helpers/TelegramHelper.kt
```diff
@@ -653,28 +637,63 @@ class TelegramHelper private constructor() {
...
-					return@forEach
+	private fun recreateLiveLocationMessage(chatId: Long, msgId: Long, content: TdApi.InputMessageLocation) {
+		if (msgId != -1L) {
+			log.info("recreateLiveLocationMessage - $msgId")
+			val array = LongArray(1)
+			array[0] = msgId
+			client?.send(TdApi.DeleteMessages(chatId, array, true)) { obj ->
+				when (obj.constructor) {
+					TdApi.Ok.CONSTRUCTOR -> {
+						sendNewLiveLocationMessage(chatId, content)
```
vshcherb: What if delete message will not run callback? Will new Location message be sent anyway?

Chumva: yes, new location message will be sent anyway in sendLiveLocationImpl, because of current message id made to  -1 before recreating the new message. I made a similar setting to -1 in onDeleteMessages if this callback fails or if a user deletes a message from the telegram

Chumva: onDeleteMessages deletes only if currentMessageId among deleted messages

PR #6262, opened 2018-11-08
Language rework 5

comradekingu:


OsmAnd/res/values/strings.xml
```diff
@@ -824,9 +824,10 @@
     <string name="select_voice_provider_descr">Select or download voice guidance for your language.</string>
     <string name="impassable_road_desc">Select roads you want to avoid during navigation.</string>
     <string name="shared_string_sound">Sound</string>
-    <string name="no_location_permission">App has no permission to access location data.</string>
-    <string name="no_camera_permission">App has no permission to access the camera.</string>
-    <string name="no_microphone_permission">App has no permission to access the microphone.</string>
+    <string name="no_location_permission">OsmAnd needs location data access.</string>
```
vshcherb: Not sure which line here is better cause it needs to be checked in the context. If it is an error message it is better to say why things don't work, in case it is a question to user then 2nd option is better. So, I think context need to be checked.

comradekingu: There is one string ala: "Grant OsmAnd * to *", which is even more to the point.

comradekingu:
@vshcherb I could go through the others and comment on why I think they are needed, there is some thinking behind each and every one.

PR #6419, opened 2019-01-04
Updated some names for clarity

ADepic:
if diet:xyz = yes
The name was changed from xyz to Some xyz options
eg
diet:vegetarian=yes
name was "vegetarian"
now is "some vegetarian options"
Same done for takeaway and delivery

Also Bicycle Parking name improvements

Stands ---> Sheffield Stands
Wall Loops ---> Wheel Benders

These names are more common among cyclists.

### DIFF
--- a/OsmAnd/res/values/phrases.xml
+++ b/OsmAnd/res/values/phrases.xml
@@ -1359,8 +1359,8 @@
 	<string name="poi_parking_multi_storey">Multi-storey</string>
 
 	<string name="poi_bicycle_parking_anchors">Anchors</string>
-	<string name="poi_bicycle_parking_stands">Stands</string>
-	<string name="poi_bicycle_parking_wall_loops">Wall loops</string>
+	<string name="poi_bicycle_parking_stands">Sheffield Stands</string>
+	<string name="poi_bicycle_parking_wall_loops">Wheel Benders</string>
 	<string name="poi_bicycle_parking_rack">Rack</string>
 	<string name="poi_bicycle_parking_building">Special building</string>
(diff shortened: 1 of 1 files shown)


OsmAnd/res/values/phrases.xml
```diff
@@ -1820,9 +1820,9 @@
 
 	<string name="poi_public_bath">Public bath</string>
 
-	<string name="poi_male_yes">Male</string>
+	<string name="poi_male_yes">Male Only</string>
```
vshcherb: I don't think this is correct cause there are toiltets with male/yes + female/yes

ADepic:
Directly from the OSM wiki:
Key:female

A value other than no implies no for all other gender-based access tags, but this can be overridden by adding more access role tags. In case of conflict, resolve in favour of more access.

Key:male

A value other than no is interpreted as yes, and implies no for all other unspecified gender-based access tags, but this can be overridden by adding more access role tags. In case of conflict, resolve in favour of more access.

vshcherb:
Please revert those changes cause it will take quite a lot of time and it is a complex techinical task to provide translations for a tag combinations

OsmAnd/res/values/phrases.xml
jaller94: The OSM wiki actually uses the strict word "only" as well. But as mentioned, multiple gender-keys can be `yes` on one POI.
https://wiki.openstreetmap.org/wiki/Key:male

For an app like OSMAnd, I suggest: Male allowed / Female allowed

Issue #6466, opened 2019-01-22
Two AIDL items in the drawer


dmpr0:
In the latest nightly, I have two OsmAnd Telegram items.
![photo_2019-01-22 10 32 35](https://user-images.githubusercontent.com/9365308/51522074-0fdec800-1e31-11e9-9334-5a7a7c4082a3.jpeg)

MadWasp79:
closed with https://github.com/osmandapp/Osmand/commit/07f839d4081a4a81beef71fd50901f1d9ae53496

MadWasp79:
closed with  https://github.com/osmandapp/Osmand/commit/f33f4eed5f516792dfac5f436f4aea2503cb3a66

Issue #6465, opened 2019-01-22
Delete selected favorites bug


dmpr0:
Try to delete the group of favorites by check group checkbox, then unselect a few points all group will be deleted anyway.
![screenshot_20190122-102828](https://user-images.githubusercontent.com/9365308/51521981-cdb58680-1e30-11e9-8ce6-3b42d1225ff1.png)
![screenshot_20190122-102833](https://user-images.githubusercontent.com/9365308/51521982-cdb58680-1e30-11e9-93c2-ce9ba440e1e7.png)
![screenshot_20190122-102843](https://user-images.githubusercontent.com/9365308/51521983-cdb58680-1e30-11e9-92d6-b664c83fd5f6.png)
![screenshot_20190122-102848](https://user-images.githubusercontent.com/9365308/51521984-cdb58680-1e30-11e9-952b-7c420352e1e9.png)

MadWasp79:
Closed with https://github.com/osmandapp/Osmand/commit/6ccc7b50970072eb2a5bf73caa6f64bce3358cbc

PR #6481, opened 2019-01-28
Kolomiets branch

MadWasp79:
Bug fix - Delete selected favorites bug #6465
Bugfix (Two AIDL items in the drawer) #6466

### DIFF
--- a/OsmAnd/src/net/osmand/aidl/OsmandAidlApi.java
+++ b/OsmAnd/src/net/osmand/aidl/OsmandAidlApi.java
@@ -1715,21 +1715,18 @@ public class OsmandAidlApi {
 						intent.addFlags(item.flags);
 					}
 					final Intent finalIntent = intent;
-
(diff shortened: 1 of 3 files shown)


OsmAnd/src/net/osmand/aidl/OsmandAidlApi.java
```diff
@@ -1703,6 +1703,7 @@ public void registerNavDrawerItems(final Activity activity, ContextMenuAdapter a
 			String appPackage = entry.getKey();
 			for (NavDrawerItem item : entry.getValue()) {
 				Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(item.uri));
+				boolean isCopy = false;
```
crimean: api should not limit items to be displayed even with same names. you fixed result, but not source of the issue.
looks like something wrong on telegram side.

PR #6482, opened 2019-01-28
New strings reworked

comradekingu:


vshcherb:
That would need an extra thinking, it was done with intention to have different words for "destination" and "waypoint" to not confuse one entities with anothers

OsmAnd/res/values/strings.xml
```diff
@@ -1779,31 +1779,31 @@
...
+    <string name="add_as_first_destination_point">Add as first intermediate waypoint</string>
     <string name="replace_destination_point">Replace the destination</string>
     <string name="new_destination_point_dialog">You have already set a destination:</string>
     <string name="shared_string_target_points">Destinations</string>
-    <string name="intermediate_point_too_far">Intermediate destination %1$s is too far from the nearest road.</string>
-    <string name="arrived_at_intermediate_point">You have arrived at your intermediate destination</string>
-    <string name="context_menu_item_intermediate_point">Add as intermediate destination</string>
-    <string name="map_widget_intermediate_distance">Intermediate destination</string>
+    <string name="intermediate_point_too_far">Intermediate waypoint %1$s is too far from the nearest road.</string>
+    <string name="arrived_at_intermediate_point">Intermediate waypoint reached</string>
```
comradekingu: "Last intermediate waypoint reached"?

comradekingu:
I still think "You have arrived at your intermediate destination" is strange when there can be multiple intermediate destinations, but as that has been reverted, the rest looks good to me.

sonora:
Looks fine to me as it stands now.

Only small issue I still see is this: Over the years I had made an effort to use capitalization wherever we refer to an OsmAnd feature, to distinguish form the common meaning of the expression. While we still follow that principle for e.g. 'Favorites', it looks like we have abandoned it for 'Map markers'. Probably no big deal in that case, though.

comradekingu:
@sonora I see the concern, and I think the issue I have with it is solved by using "Map markers".
"Favourites" falls in a category to where it is generic enough to be equivalent to "Bookmarks". Everyone has those, so they are unspecific to OsmAnd. Also it looses part of the tangible audience on "favorite/favourite". "Map markers" is a lot better, because it is always "Map markers", and descriptive at that, and gives it more of an OsmAnd twist, warranting the capitalization. "Map Markers", would IMO. be too much ®™© pat. pending.

How things translate, "In the settings" vs "In "Settings" you will find" and what Google generally does "In Settings you will find", must be taken into account.

I'd say it is an issue for another day, I need to make a list of terms in OsmAnd, and a strategy for using them, as there is quite a bit of duplication and design-wins to be made as of right now.

sonora:
Regarding favorites vs. favourites: You are probably aware we use American spelling in the default en file, while we have an en-GB file for any British spelling deviations. :)

comradekingu:
@sonora Yes, the issue arises when British person shows American person their phone, and it looks unfamiliar for no good reason.

vshcherb:
@sonora  please feel free to merge

sonora:
Regarding waypoints vs. destinations, I have added some background here: https://github.com/osmandapp/Osmand/issues/6606#issuecomment-469153546

PR #6627, opened 2019-03-06
Latlon parsing issues

MadWasp79:
Couple of issues with latlon parsing (see commit titles)


OsmAnd-java/src/test/java/net/osmand/util/GeoPointParserUtilTest.java
```diff
@@ -269,7 +269,7 @@ public void testGeoPoint() {
 		url = "http://maps.google.com/maps?q=N" + dlat + ",E" + Math.abs(dlon);
```
sonora: Our example in line 268 seems wrong, http://maps.google.com/maps?q=N34.939,E-106 is not a valid google URL.

Issue #6345, opened 2018-12-06
Feature request: show azimuth when measuring distance


alyonapinchuk:
Please add the feature to show azimuth (with distance) when the user draws lines by Measure distance tool.

MadWasp79:
closed with https://github.com/osmandapp/Osmand/commit/c452d8429dbd34d9463788c4ae3a79033235ffcc

breversa:
From what I can gather from the pull request, the azimuth will be shown in degrees.

Would it be possible to display it as milliradians (https://en.wikipedia.org/wiki/Milliradian), with 360° = 6400mil ?
Along with UTM/MGRS, it's highly useful for land navigation in a military context.

vshcherb:
I think we can display it when UTM is enabled

MadWasp79:
We'll add angular measurement unit selection to settings.

PR #6705, opened 2019-03-18
Add azimuth to measurement tool #6345

MadWasp79:


OsmAnd/src/net/osmand/plus/measurementtool/adapter/MeasurementToolAdapter.java
```diff
@@ -88,15 +90,27 @@ public boolean onTouch(View view, MotionEvent event) {
...
+				if (mapActivity.getMyApplication().getLocationProvider().getLastKnownLocation() != null) {
+					azimuth = OsmAndFormatter.getFormattedAzimuth(MapUtils.getBearingToPoint(l1, points.get(0).lat, points.get(0).lon));
+					text = text + BULLET + azimuth;
+				}
+				holder.descr.setText(text);
 			} else {
 				float dist = 0;
 				for (int i = 1; i <= pos; i++) {
 					dist += MapUtils.getDistance(points.get(i - 1).lat, points.get(i - 1).lon,
 							points.get(i).lat, points.get(i).lon);
```
vshcherb: distance and bearing could be measured in 1 call  l1.bearingTo(l2);  l1.distTo(l2);

OsmAnd/src/net/osmand/plus/measurementtool/adapter/MeasurementToolAdapter.java
```diff
@@ -88,15 +90,27 @@ public boolean onTouch(View view, MotionEvent event) {
...
 			holder.descr.setText(pointDesc);
 		} else {
+			String azimuth;
+			String text = "";
 			if (pos < 1) {
-				holder.descr.setText(mapActivity.getString(R.string.shared_string_control_start));
+				text = mapActivity.getString(R.string.shared_string_control_start);
+				Location l1 = mapActivity.getMyApplication().getLocationProvider().getLastStaleKnownLocation();
+				if (mapActivity.getMyApplication().getLocationProvider().getLastKnownLocation() != null) {
+					azimuth = OsmAndFormatter.getFormattedAzimuth(MapUtils.getBearingToPoint(l1, points.get(0).lat, points.get(0).lon));
```
vshcherb: No distance?

OsmAnd/src/net/osmand/plus/OsmAndFormatter.java
```diff
@@ -165,7 +165,17 @@ public static String getFormattedAlarmInfoDistance(OsmandApplication app, float
 
 		return df.format(meters / mainUnitInMeters) + " " + app.getString(mainUnitStr);
 	}
-	
+
+	public static String getFormattedAzimuth(float bearing) {
```
vshcherb: Not sure about formatting, needs to be checked how it is usually formatted

MadWasp79: From Wiki: "Today, the reference plane for an azimuth is typically true north, measured as a 0° azimuth, though other angular units (grad, mil) can be used. Moving clockwise on a 360 degree circle, east has azimuth 90°, south 180°, and west 270°."

OsmAnd/src/net/osmand/plus/measurementtool/adapter/MeasurementToolAdapter.java
MadWasp79: Yes, it's the first point. Azimuth there is calculated from your current location

OsmAnd/src/net/osmand/plus/measurementtool/adapter/MeasurementToolAdapter.java
MadWasp79: Dunno how it will help, but added one more method to MapUtils which returns distance and bearing.

Issue #5021, opened 2018-02-13
Ski profile for planning routes


rindlerblabla:
For planning ski routes it would be nice with a profile which accepts piste:type and aerialway. At least one other user agree on that, hehe, https://groups.google.com/forum/m/#!topic/osmand/Xp-tsITE8GI

Personally I don't need advanced routing rules but rather to be able to use the snap-function when using the measure tool.

xmd5a2:
After some investigation I've found that OSM data related to ski aren't routable. In most cases there is no connection between aerialway stations and pistes. Also some pistes are mapped as areas hence we can't calculate route via them.

rindlerblabla:
Okey, it was the other user in the Google group that suggested for other types of piste types than nordic. At least for piste:type=nordic it's clearly usable and would be nice to have as feature.

scaidermern:
> In most cases there is no connection between aerialway stations and pistes.

There are several ski areas where pistes and aerialway stations do have connections. But I remember a discussion somewhere about how these connections should be tagged. Maybe someone is able to find it.

pietervdvn:
Hey,

It is correct that most ski pistes are connected badly. However, I've been skiing and cleaned two ski areas (a small one [here](https://www.openstreetmap.org/relation/902341#map=14/47.4699/13.2850&layers=N) and [here](https://www.openstreetmap.org/search?query=serre%20chevalier#map=13/44.9286/6.5939&layers=N))

If you would include them into the map builds; I'll gladly create the routing.xml for skiing. Also, as soon as this is in beta; everyone OSM'er with Osmand Live who will go skiing next year; will gladly fix this up. Just communicate clearly that this feature is in beta.

rindlerblabla:
@pitchblack1 But not only pistes but also aerialways should be connected to be able to use routing engine (?).  As far as I could see from the areas you had corrected still the aerialways are not connected to the piste? Think an unnamed piste:type=downhill way should be just fine.

pietervdvn:
Hey,

I had a look again to Werfenweng (it's been a few years). Someone added a new chairlift (for which I don't know where it ends). The rest should be still properly connected (or I fixed it now).

The other area (Serre Chevalier) was more or less fixed last week.

scaidermern:
@rindlerblabla Yes, exactly. Thanks :)

mahlzahn:
Any news on that? With
```
[...]

PR #6718, opened 2019-03-20
Ski profile for planning routes (#5021)

mahlzahn:
See discussion thread in #5021.


pietervdvn:
As the ski profiles are already in the production version, I think that this PR should be closed.

mahlzahn:
Thanks for putting it in!

pietervdvn:
Yep, it was a cool project! Thanks for helping out with it, I look forward to test it on the pistes this winter.

If you feel like it, come skiing early february in La Plagne, Fance and we might meet :p

PR #6842, opened 2019-04-19
Aidl new method

MadWasp79:


OsmAnd/src/net/osmand/aidl/OsmandAidlService.java
```diff
@@ -121,14 +127,52 @@ public IBinder onBind(Intent intent) {
...
+		return aidlCallbackId.get();
+	}
+
+	private long getAndIncrementCallbackId() {
+		return aidlCallbackId.getAndIncrement();
+	}
+
+	@Override
+	public long addAidlCallback(IOsmAndAidlCallback callback, int key) {
+		callbacks.put(getAndIncrementCallbackId(), new AidlCallbackParams(callback, key));
```
vshcherb: concurrency issue, callback could change in between

Issue #5632, opened 2018-07-04
Track recording doesn't record in the background (Android 8, EMUI 8)


vshcherb:
I noticed a bad bug in current release: during navigation Osmand very frequently shut down when running in background. I have installed Osmand+ 3.0.4 on Huawei P10 with Android 8.0.

The issue is that during navigation after some time Osmand close instead than keep running in background and consequently the navigation stops. I do not think it is related with strict battery optimization policy of Huawei phones, because it never happen before.

I started to notice this only with last version.

sonora:
Yes, agree, I guess we can cross-link this to issues like #5578?

MarkusTV:
I don't know if it can be useful for investigation, anyway I noticed that when I reopen Osmand after the crash in background it shows a 5 seconds countdown message asking if continue or not the interrupted route navigation.
It means that Osmand realizes that something happen and try to resume.

sonora:
Yes, that is a standard feature we have if OsmAnd was closed while a navigation was ongoing. It is a result of the issue, not the cause of it.

MarkusTV:
Ok got it, thank you.
My question is if Osmand crashes itself or it is forced closed by o.s  (Android or Huawei EMUI maybe for battery optimization purposes).
I do not think so because the issue is new, appeared in this last version for first time only.

Marco

Da: Hardy <notifications@github.com>
Inviato: sabato 7 luglio 2018 22:18
A: osmandapp/Osmand <Osmand@noreply.github.com>
Cc: MarkusTV <marco.brun@outlook.com>; Manual <manual@noreply.github.com>
Oggetto: Re: [osmandapp/Osmand] OsmAnd very frequently shut down when running in background. (#5632)


Yes, that is a standard feature we have if OsmAnd was closed while a navigation was ongoing. It is a result of the issue, not the cause of it.

sonora:
Thar's interesting, so far I had attributed it to OS behavior, but you are saying it could be an OsmAnd regression?

MarkusTV:
No idea, I cannot say that, I see that Osmand closes but I don't know if closes itself or it is killed by the OS.

Marco
________________________________
Da: Hardy <notifications@github.com>
Inviato: domenica 8 luglio 2018 10:01
A: osmandapp/Osmand
Cc: MarkusTV; Manual
Oggetto: Re: [osmandapp/Osmand] OsmAnd very frequently shut down when running in background. (#5632)


[...]

PR #6903, opened 2019-05-04
Use setExact for long-interval recording

sonora:
Already solves issues on some devices, and should create the basis for solving 5532.


vshcherb:
#5632 https://github.com/osmandapp/Osmand/issues/5632#issuecomment-489350120

PR #6953, opened 2019-05-19
Turn instructions and travel time from BRouter

vodie:
Get turn instructions and travel time estimation from BRouter.
The default travel time calculation was wrong if BRouter is the Navigation service and has no times (e.g. older BRouter version).

### DIFF
--- a/OsmAnd/src/net/osmand/plus/routing/RouteProvider.java
+++ b/OsmAnd/src/net/osmand/plus/routing/RouteProvider.java
@@ -1139,7 +1139,6 @@ public class RouteProvider {
 		double[] lons = new double[numpoints];
 		int index = 0;
 		String mode;
-		float defSpeed;
 		boolean addMissingTurns = true;
 		lats[index] = params.start.getLatitude();
 		lons[index] = params.start.getLongitude();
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/routing/RouteProvider.java
```diff
@@ -1149,20 +1155,25 @@ protected RouteCalculationResult findBROUTERRoute(RouteCalculationParams params)
 		lons[index] = params.end.getLongitude();
 		if (ApplicationMode.PEDESTRIAN == params.mode) {
 			mode = "foot"; //$NON-NLS-1$
+			defSpeed = (float) (5/3.6);
```
vshcherb: @MadWasp79 Here we should use default speed defined in the routing configuration

vodie: With the new application profiles you cant select BRouter as routing service for the base profiles (driving, cycling and walking) that are used by BRouter for Profile selection. Is it possible to select BRouter for the base profiles again? Otherwise this section needs a rework to use other BRouter modes than the "motorcar" with the non base profiles.

I deleted my "defspeed" and use "params.mode.getDefaultSpeed()" for the default speed in "parseOsmAndGPXRoute"

The default speed for the new non base cycling profiles is 10 m/s, pretty fast for a normal bicycle!

PR #6959, opened 2019-05-20
Library compilation fixes

Codename26:
Changed all switch cases to if\else, because library compilation failed with switches.


alexey-pelykh:
@Codename26 also please add gradle file that builds osmand as library

vshcherb:
Please also keep in touch in Telegram group cause the issues could be delayed with merges

Issue #7051, opened 2019-06-19
Expand coordnates


vshcherb:
We can build expandable context menu and display there more option of location i.e.
Lat xxx Lon yyy: >
 - xxx, yyyy
 - OLC location 
 - osmand.net link to location
 - UTM

sonora:
We have always had requests to also show the hdop value, maybe this can be added here, too?

PR #7066, opened 2019-06-20
7051 expand coordnates

MadWasp79:
issue #7051


OsmAnd/src/net/osmand/plus/mapcontextmenu/MenuBuilder.java
```diff
@@ -389,6 +389,10 @@ protected void buildNearestWikiRow(View view) {
 		}
 	}
 
+	protected void buildCoordinatesRow(View view) {
```
crimean: ?

MadWasp79: oops. trash, cleared.

PR #7157, opened 2019-07-05
New strings reworked, all WL comments sorted

comradekingu:
Fixes spotted by @critdroid @yarons @franco999 and @fastestnoob on Weblate
Will annotate later.


OsmAnd/res/values/strings.xml
```diff
@@ -661,9 +661,9 @@
     <string name="srtm_purchase_header">To see contour lines on the map, you need to buy and install the \'Contour Lines\' plugin</string>
     <string name="srtm_color_scheme">Color scheme</string>
     <string name="show_from_zoom_level">Display from zoom level</string>
-    <string name="routing_attr_allow_private_name">Allow private access</string>
+    <string name="routing_attr_allow_private_name">Allow trespassing</string>
```
njohnston: You suggested this type of change before. "trespassing" is not a good word to use here. The option is to allow access to private roads. For example, suppose you're staying at a hotel and it has a private access road. As a guest of the hotel, you have permission (perhaps implied, but nevertheless permission) to use the road, and therefore it is not trespassing.

vshcherb: Agree, private roads is quite easy to understand, while trespassing is something related to violation

OsmAnd/res/values/strings.xml
```diff
@@ -321,7 +321,7 @@
     <string name="quick_action_edit_actions">Edit actions</string>
     <string name="get_osmand_live">Get OsmAnd Live to unlock all features: Daily map updates with unlimited downloads, all paid and free plugins, Wikipedia, Wikivoyage and much more.</string>
     <string name="unirs_render_descr">Modification of the default style to increase contrast of pedestrian and bicycle roads. Uses legacy Mapnik colors.</string>
-    <string name="shared_string_bookmark">Bookmark</string>
+    <string name="shared_string_bookmark">Add bookmark</string>
```
vshcherb: This is a wrong change, shared_string_bookmark implies that it will stay as bookmark in case "add bookmark" fits better it requires new string

vshcherb:
I do like most of the changes but 2-3 needs to be pulled out into separate discussion

OsmAnd/res/values/strings.xml
comradekingu: Thought I might have. It is a foreign concept to me, where I live a private road is not private  until there is a gate on it.
I don't think it is always a question of being offroad when using small paths, (not roads) some of which may be private ones?

OsmAnd/res/values/strings.xml
comradekingu: It is difficult to decide whether to translate it as a noun or verb, the string name gives no further clues either.

OsmAnd/res/values/strings.xml
```diff
@@ -73,7 +73,7 @@
     <string name="app_mode_subway">Subway</string>
     <string name="app_mode_horse">Horse</string>
     <string name="app_mode_helicopter">Helicopter</string>
-    <string name="osmand_routing_promo">You can add your own modified version of the file routing.xml in ..osmand/routing</string>
+    <string name="osmand_routing_promo">Optionally modify routing.xml in ..osmand/routing</string>
```
comradekingu: Is this meant to be "..osmand"?

OsmAnd/res/values/strings.xml
```diff
@@ -111,21 +111,21 @@
...
     <string name="dont_move_maps">Don\'t move</string>
-    <string name="public_transport_ped_route_title">The route by foot is approximately %1$s and may be faster than via public transport</string>
-    <string name="public_transport_no_route_title">Unfortunately, OsmAnd could not find a route suitable for your settings.</string>
-    <string name="public_transport_try_ped">Try the pedestrian navigation.</string>
+    <string name="public_transport_ped_route_title">The route on foot is about %1$s, possibly faster than public transport.</string>
+    <string name="public_transport_no_route_title">Could not find a route corresponding with your settings.</string>
+    <string name="public_transport_try_ped">Try navigation on foot.</string>
     <string name="public_transport_try_change_settings">Try changing the settings.</string>
-    <string name="public_transport_calc_pedestrian">Calculate pedestrian route</string>
+    <string name="public_transport_calc_pedestrian">Calculate route on foot</string>
```
comradekingu: "Calculate" is a bit redundant here.

OsmAnd/res/values/strings.xml
```diff
@@ -560,9 +560,9 @@
     <string name="show_passed">Show passed</string>
     <string name="hide_passed">Hide passed</string>
     <string name="remove_from_map_markers">Remove from map markers\'</string>
-    <string name="descendingly">descending</string>
-    <string name="ascendingly">ascending</string>
-    <string name="date_added">Date added</string>
+    <string name="ascendingly">A-Z</string>
```
comradekingu: Not 100% sure these are always letters, but then again the point gets across that A= 0 if it is sometimes numbers

OsmAnd/res/values/strings.xml
```diff
@@ -661,9 +661,9 @@
     <string name="srtm_purchase_header">To see contour lines on the map, you need to buy and install the \'Contour Lines\' plugin</string>
     <string name="srtm_color_scheme">Color scheme</string>
     <string name="show_from_zoom_level">Display from zoom level</string>
-    <string name="routing_attr_allow_private_name">Allow private access</string>
+    <string name="routing_attr_allow_private_name">Allow trespassing</string>
     <string name="routing_attr_allow_private_description">Allow access to private areas.</string>
-    <string name="display_zoom_level">Display zoom level: %1$s</string>
+    <string name="display_zoom_level">Zoom level: %1$s</string>
```
comradekingu: I think this may be a change I have attempted before, and can't remember the reasoning behind.

OsmAnd/res/values/strings.xml
```diff
@@ -705,7 +705,7 @@
     <string name="shared_string_visible">Visible</string>
     <string name="restore_purchases">Restore purchases</string>
     <string name="fonts_header">Map fonts</string>
-    <string name="right_side_navigation">Right-hand driving</string>
+    <string name="right_side_navigation">Right-hand traffic</string>
```
comradekingu: This should be next to the "Left-hand traffic" string?

OsmAnd/res/values/strings.xml
```diff
@@ -705,7 +705,7 @@
...
     <string name="driving_region_automatic">Automatic</string>
     <string name="do_not_send_anonymous_app_usage">Don\'t send anonymous app usage statistics</string>
     <string name="do_not_send_anonymous_app_usage_desc">OsmAnd collects info about which parts of the app you open. Your location is never sent, nor is anything you enter into the app or details of areas you view, search, or download.</string>
```
comradekingu: This string is very much a copy of "do_not_send_anonymous_app_usage_desc"

OsmAnd/res/values/strings.xml
```diff
@@ -41,29 +41,29 @@
...
+    <string name="rate_dialog_descr">Please share your feedback and rate our work on Google Play.</string>
     <string name="button_rate">Rate</string>
     <string name="shared_string_privacy_policy">Privacy Policy</string>
-    <string name="help_us_make_osmand_better">Help us make OsmAnd better!</string>
-    <string name="make_osmand_better_descr">Allow OsmAnd to collect and process anonymous application usage data. We do not collect or store data about your position, or about any locations you view on the map.\n\nYou can change your selection at any time in Settings > Privacy and Security.</string>
-    <string name="choose_data_to_share">Choose what type of data you want to share:</string>
+    <string name="help_us_make_osmand_better">Help make OsmAnd better!</string>
+    <string name="make_osmand_better_descr">Allow OsmAnd to collect and process anonymous app usage data. Your position is not collected nor stored, neither is any locations you view on the map.\n\nChange your choice any time in \'Settings\' → \'Privacy and Security\'.</string>
```
comradekingu: The first part of this is very much a copy of "do_not_send_anonymous_app_usage_desc".

OsmAnd/res/values/strings.xml
sonora: 'Trespassing' is wrong here.

OsmAnd/res/values/strings.xml
sonora: Yes, it is the folder name. And I would leave in the word 'file' because this really relates to a file modification.

sonora:
I think there are quite a few more issues here, see my comments entered under "changes".

Many changes seem to disregard the actual environment where the strings occur! Let's please be careful before merging. I think smaller change packages would make things easier.

OsmAnd/res/values/strings.xml
```diff
@@ -627,7 +627,7 @@
     <string name="map_widget_ruler_control">Radius ruler</string>
     <string name="shared_string_permissions">Permissions</string>
     <string name="import_gpx_failed_descr">OsmAnd cannot import the file. Please check if OsmAnd has permission to read file from its location.</string>
-    <string name="distance_moving">Distance corrected</string>
+    <string name="distance_moving">Distance travelled</string>
```
comradekingu: Not 100% sure here, could this be explained?

PR #7187, opened 2019-07-08
AIDL updates updateMapMarker, removeMapMarker, removeAllMarkers

fsbugreporter:
- introduction of an option to ignore coordinates in updateMapMarker and removeMapMarker AIDL functions; default behaviour is that marker name and coordinates have to match for update/remove functions. The new option allows to ignore the coordinate values, so only the name has to match.

- added new AIDL function removeAllMarkers which removes all map markers


vshcherb:
Probably we need to merge in 3.4 as well. @crimean to check

vshcherb:
Let's rename removeAllMapMarkers - markAllActiveMapMarkersAsPassed

vshcherb:
And merge to 3.4

vshcherb:
https://github.com/osmandapp/Osmand/commit/5b9d3cc12624f2bf0c6a53572fc28e48f5aa27ee

PR #7653, opened 2019-10-09
Added option to save gpx files in daily folders fixes #6831 and #7514

### LINKED ISSUE #6831: [Feature Request] GPX Tracks: Create folder every day
Hello,

for people like me, who use the gpx tracking very often and have a lot of tracks on some days, it would be nice, if there where an option to create a folder for every day, not only for every month.

Thanks in advance for reading and thank you all for the great work and the great app.

### LINKED ISSUE #7514: New dialog after saving a GPX is not working when "Store recorded tracks in monthly folders" ist activated
Hi,
I use:

OsmAnd+ 3.4.8
Moto G5 Plus
Android 8.1

the offline maps offered within the OsmAnd app for download.

When I recorded a GPX track and save it, the new dialog with "open track" and "show track on map" ("Auf Karte anzeigen") is displayed at the bottom of the app, but neither the "open track" nor the "show track" ("Auf Karte anzeigen") button works for me.
When I choose "show track" ("Auf Karte anzeigen") the dialog vanishes and nothing happens, when I choose "open track" it looks like this: 
Screenshot_OsmAnd+_20190904-081444
Screenshot_OsmAnd+_20190904-081454

and when I try to share it, the app I share it with gets opened, but the gpx-file is missing. All other functions (i.e. show track on map, delete etc.) are also not working.

When I go to "my places" -> tracks ("Meine Orte" -> Tracks) the track is there and everything works as expected.

jensMF:
I added the option to save files not only in monthly, but also in daily folders as I suggested in #6831
I also changed the saveDataToGpx method to save the whole filename in SaveGpxResult.filenames. That Fixes #7514 too.


jensMF:
I realized there is a getTracksDir() function. So I Think the approach to create and choose the folder in the saveDataToGpx(File dir) function of the SavingTrackHelper is not good, because it brakes the getTracksDir() function. That means I need to change the way monthly and daily folders are implemented at the moment.

jensMF:
I had some more time to think about it and spoke with a colleague and now I think getTracksDir() might work as intended. It returns the base directory of the recording section.
Is it intended to be like that? If yes, this pull-request it ok.

jensMF:
I mess with my branches.

PR #7800, opened 2019-10-30
Fix #7612

### LINKED ISSUE #7612: Swap start and destination bold font issue
The algorithm does not take the word order into consideration.
image

Chumva:


max-klaus:
We have https://github.com/osmandapp/Osmand/blob/4a8bf89556ad2f7068f4a4281ea5dcfc42d01a94/OsmAnd/res/values/strings.xml#L544
Could it be used here?

Chumva:
> We have
> 
> https://github.com/osmandapp/Osmand/blob/4a8bf89556ad2f7068f4a4281ea5dcfc42d01a94/OsmAnd/res/values/strings.xml#L544
> 
> 
> Could it be used here?

Yes, It is possible to use same formatting here, but this will add default bold typeface, not our Roboto-Medium. The difference almost invisible for average human eye, but it exists.

Issue #7745, opened 2019-10-24
Speed limit widget can't be deactivated in browse map profile


TomTurnschuh:
In the ‹browse map› profile the ‹speed limit› widget is not available for (de)activation. It is always activated when centered the map to current position.

It would be nice, if this widget could be included in the corresponding widget list of the ‹browse map› profile.

Regards

vshcherb:
Yes actually the widget is part of the problem, we don't scan user street name and road he is located cause it is impossible to do for generic profile we can do it only if we know navigation type. So, that's why widget could be visible but no information will be there. This fucntionality is only available in car, bicycle, pedestrian

PR #7802, opened 2019-10-30
Fix#7745 Add <speed limit> widget in the list of widgets in the <browse map> profile

Dima-1:


vshcherb:
Unfortunately it is a widget of MAX_SPEED and user asks to disable just Speed widget.

Dima-1:
Maybe I misunderstood, but there is widget MAX_SPEED called "Speed limit" on the right panel of Configure screen. It's exactly the same widgets name as user asks in issue #7745. This widget exist in the CAR profile only. Availability for activation/deactivation Speed widget exist in the all profiles by default.

vshcherb:
replied in #7745

PR #7811, opened 2019-10-31
Can't backup a temp file.

Dmole:
https://github.com/osmandapp/Osmand/blob/8fc456f3ecab9927ca0e65eaa10ffa3ffcead28a/OsmAnd/src/net/osmand/plus/AppInitializer.java#L750


vshcherb:
https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/FavouritesDbHelper.java#L153
BAK is correct name

PR #7815, opened 2019-11-01
Strings reworked 6

comradekingu:


OsmAnd/res/values/strings.xml
```diff
@@ -1767,24 +1766,24 @@
...
     <string name="save_track_to_gpx_descrp">A GPX track is automatically saved to the tracks folder during navigation.</string>
     <string name="save_track_interval_globally">Logging interval</string>
     <string name="save_track_interval">Logging interval during navigation</string>
-    <string name="save_track_interval_descr">Choose logging interval for track recording during navigation.</string>
+    <string name="save_track_interval_descr">Choose logging interval for track recording during navigation</string>
     <string name="voice_provider_descr">Select the voice guidance for navigation.</string>
     <string name="voice_provider">Voice guidance</string>
     <string name="enable_proxy_title">Enable HTTP proxy</string>
-    <string name="enable_proxy_descr">Configure an HTTP proxy for all network requests.</string>
+    <string name="enable_proxy_descr">Set up a HTTP proxy for all network requests.</string>
```
dmpr0: Hello! 
Could you please tell, whey "Set up" is better than "Configure"?

sonora:
@comradekingu 

**I made modifications to facilitate merging**

Very large PR, maybe in the future you can create smaller packages bearing smaller risk to fatfinger things before a release.

In order to be able to merge the majority of what is ok, I have reverted your suggestions regarding:
- "Configure" vs. "Set up": Set up is from scratch, as in when you select from where to where you want to be routed, while "Configure" denotes the settings to customize already existing things or principles, like the appearance of the map, the appearance of the map screen, etc. Changes here are rather far-reaching, and would also affect major portions of our documentation,as well as affect experienced users.
- I could not agree with your use of "Choose"(usually decide that a particular thing is the one that you want) vs. "Select" (pick something from a group or number of alternatives), hence tried to revert most of your suggestions.
- "start" and "end" is not always equivalent to departure and destination 
- We need to make a distinction between GPX files and tracks in some cases
- I have accepted "announcements" vs. "voice prompts", but am not convinced there is a clear benefit. May have to revisit.

Anything no accepted now,please make it a new ans small package so we can discuss.

Thx - Hardy

OsmAnd/res/values/strings.xml
sonora: Yes, I also do not agree with the suggested modifications of replacing "Configure" by "Set up". Will remove for now, unless where it is clearly ok.

comradekingu: The idea was to arrive at some consistency. Configure/configuration, vs set up/setup. The latter is marginally easier to understand, I think easier to translate, and a bit shorter.

comradekingu:
@sonora Wow, that was quick :)
It is going to be short and sweet from now on. This was the everything PR. I have a textfile with all unresolved issues. Didn't get around to presenting the rationale.
Btw, don't want any MIT-licensing on anything I do, given the new contributor licensing file.

I suppose you are right about initial configuration more-so denotes "set up". "Choose" was an effort to see how much "select" there really was. I think most can be dropped, and "pick" is also in use.
Choose/choice vs select/option, but it didn't arrive anywhere near that being the most pressing issue. Renaming everything "GPX files" left some "tracks", and omitted it where it makes more sense. "Record a GPX track" gives some idea of what GPX is.
Thank you for merging, will get to translating soon. The (semi-consistent) uppercase "Favorites" is also quirky.

PR #7821, opened 2019-11-01
Fix 7235

Chumva:


OsmAnd/src/net/osmand/plus/OsmandSettings.java
```diff
@@ -1550,9 +1550,9 @@ protected boolean setValue(Object prefs, Boolean val) {
 	public static final Integer MONTHLY_DIRECTORY = 1;
 	public static final Integer DAILY_DIRECTORY = 2;
 
-	public final CommonPreference<Boolean> DISABLE_RECORDING_ONCE_APP_KILLED = new BooleanPreference("disable_recording_once_app_killed", false).makeGlobal();
+	public final CommonPreference<Boolean> DISABLE_RECORDING_ONCE_APP_KILLED = new BooleanPreference("disable_recording_once_app_killed", false).makeProfile().makeGeneral();
 
-	public final CommonPreference<Integer> TRACK_STORAGE_DIRECTORY = new IntPreference("track_storage_directory", 0).makeGlobal();
+	public final CommonPreference<Integer> TRACK_STORAGE_DIRECTORY = new IntPreference("track_storage_directory", 0).makeProfile().makeGeneral();
```
sonora: @vshcherb Do we have a concept how things should actually behave? Isn't it very weird if we even give up that e.g. DISABLE_RECORDING_ONCE_APP_KILLED and TRACK_STORAGE_DIRECTORY are global?

PR #7837, opened 2019-11-04
Strings reworked 7

comradekingu:


OsmAnd/res/values/strings.xml
```diff
@@ -13,26 +13,26 @@
...
+    <string name="quick_action_hillshade_descr">Button showing or hiding hillshade on the map.</string>
     <string name="quick_action_hillshade_show">Show hillshade</string>
     <string name="quick_action_hillshade_hide">Hide hillshade</string>
     <string name="quick_action_show_hide_hillshade">Show/hide hillshade</string>
-    <string name="apply_preference_to_all_profiles">You can apply this change to all profiles or only to the currently selected one.</string>
+    <string name="apply_preference_to_all_profiles">You can apply this change to all or only the selected profile.</string>
     <string name="shared_preference">Shared</string>
-    <string name="routing_attr_driving_style_prefer_unpaved_name">Prefer unpaved roads</string>
-    <string name="routing_attr_driving_style_prefer_unpaved_description">Prefer unpaved roads.</string>
+    <string name="routing_attr_driving_style_prefer_unpaved_name">Prefer unpaved routes</string>
```
comradekingu: Paths are not roads, which may still be part of a calculated route(?)

OsmAnd/res/values/strings.xml
```diff
@@ -77,8 +77,8 @@
     <string name="paste_Osmand_data_folder_path">Paste path to the folder with OsmAnd data</string>
     <string name="change_osmand_data_folder_question">Change OsmAnd data folder?</string>
     <string name="move_maps_to_new_destination">Move to the new destination</string>
-    <string name="internal_app_storage_description">Internal storage, hidden from user and other apps, exclusively accessible to OsmAnd</string>
-    <string name="change_data_storage_folder">Change data storage folder</string>
+    <string name="internal_app_storage_description">Internal storage for OsmAnd, (hidden from users and other apps).</string>
```
comradekingu: Is it always on internal memory, or could it be "Internal OsmAnd storage"?

OsmAnd/res/values/strings.xml
```diff
@@ -77,8 +77,8 @@
...
+    <string name="change_data_storage_folder">Change data folder</string>
```
comradekingu: Called "data folder" elsewhere

OsmAnd/res/values/strings.xml
```diff
@@ -285,7 +285,7 @@
     <string name="base_profile_descr_car">Car, truck, motorcycle</string>
     <string name="base_profile_descr_bicycle">Mountain bike, moped, horse</string>
     <string name="base_profile_descr_pedestrian">Walking, hiking, running</string>
-    <string name="base_profile_descr_public_transport">All PT types</string>
```
comradekingu: "PT" (or simila)r showed up in translations where it doesn't belong.

OsmAnd/res/values/strings.xml
```diff
@@ -418,7 +417,7 @@
     <string name="show_more">Show more</string>
     <string name="tracks_on_map">Displayed tracks</string>
     <string name="quick_action_show_hide_gpx_tracks">Show/hide GPX tracks</string>
-    <string name="quick_action_show_hide_gpx_tracks_descr">Tapping this action button shows or hides selected GPX tracks on the map</string>
+    <string name="quick_action_show_hide_gpx_tracks_descr">Button showing or hiding selected GPX tracks on the map.</string>
```
comradekingu: The idea of a button is to push/tap it. It needs not be explained.

OsmAnd/res/values/strings.xml
```diff
@@ -497,8 +496,8 @@
     • Other bugfixes\n\n
     </string -->
     <string name="increase_search_radius_to">Extend search radius to %1$s</string>
-    <string name="send_search_query_description"><![CDATA[Your search query will be sent to: <b>\"%1$s\"</b>, along with your location.<br/><br/>
-        Personal info is not collected, only search data needed to improve the search.<br/>]]></string>
+    <string name="send_search_query_description"><![CDATA[Your search query will be sent to: <b>\"%1$s\"</b>, along with your location.\n\n
```
comradekingu: Not 100% sure here, but it is consistent.

sonora:
This overlaps with multiple changes which had happendd over the weekend and some stuff I had just committed, can you maybe re-baseline or re-publish to see what's left?

Regarding 'data' folder: There are numerous folders on any Android system named 'data' per se, so having storage may be a little clearer.

Unpaved roads is a fixed term in thd US, part of many  car rental contracts, would hate to change.

OsmAnd/res/values/strings.xml
sonora: memory, I think

OsmAnd/res/values/strings.xml
sonora: Yeah, this overlapped, had already shortened it.

OsmAnd/res/values/strings.xml
sonora: I think \n won't work in this case

aceman444:
Yes, these changes are strange.
The physical roads can be unpaved, not whole routes (a collection of roads).
Also the PT is unreadable. I also don't agree with the 'storage' removal.

I don't see anything useful in this patch.

sonora:
This is outdated and has many conflicts, with no update in over a month - let's close here.

Issue #6598, opened 2019-03-02
"Can not read GPX data" when opening a GPX file out of Signal messenger


jensMF:
When I try to open a gpx file from within the Signal messenger app, I get the error message "Could not read GPX data".
If I open the GPX-file for the first time, the track is sometimes shown on the map, but the file is not being imported.
![screenshot_20190302-004938](https://user-images.githubusercontent.com/1620034/53673204-748a0f80-3c86-11e9-8a77-85dd7946a7ae.png)

![screenshot_20190302-004953](https://user-images.githubusercontent.com/1620034/53673309-e5312c00-3c86-11e9-93e5-cfa20b6efcd6.png)

It is not in the My Places Tracks tab either.
If I try it with the same GPX-file again, the track is not even shown on the map.
For other apps, it is possible to open the GPX-files directly out of the Signal messenger app.

I am using OsmAnd+ 3.2.7 on android 7.1.2 (omnirom). The device is an Oppo find7.
But I also tried it with a Samsung galaxy S4 mini on android 7.1.2 (LineageOS) same app version and another Samsung device with official rom. I don't know the android Version and device name, but its the same app version (except it's the free version).

In all test the offline maps offered within OsmAnd were used.
Maybe it is related to issue  #3840

melitele:
I am experiencing the same issue when opening a file from Files/Downloads app with Osmand+ 3.4.4 on Moto Z3 Play Android 9. Moreover I have received a report from user of my website about the same issue appearing with Samsung Galaxy S7 and Samsung Galaxy S10.
Placing the file directly under Osmand import folder - https://osmand.net/help/faq.html#importing_gpx - and checking My Places > Tracks > Import works.

vshcherb:
I think we fixed that issue with 3.4.5. It is going to be this week released

jensMF:
> I think we fixed that issue with 3.4.5. It is going to be this week released

Great news, thank you!

When will it be in the nightly?

jensMF:
I am on version 3.4.8 now and it is still not working.

jensMF:
#7624 is a duplicate of this one, because it is not solved and did not get reopened.

Mahmoud-emar:
I had the same issue and the easiest of them all was: 
1- Uploading the KML file that can't read on OSMAND on https://mymaps.google.com.
Click on Create new map 
and Click on Import 
then Drag the KML file of your route 
[...]

PR #7846, opened 2019-11-04
added workaround to import files if the fileName is null fixes #7624 and #6598

### LINKED ISSUE #7624: "Can not read GPX data" when opening a GPX file out of Signal messenger Duplicate of #6598
I would have preferred to continue in #6598 but it did get not reopened.

OsmAnd+ 3.4.8 and OsmAnd 3.4.8
in various devices from Android 6 to Android 9
i.e. Samsung Galaxy S4 Mini, Moto G5 plus, OnePlus 5T and others

We all use Offline Maps.
  
When one tries to open a gpx file from within the Signal messenger app,  the error message "Could not read GPX data" is shown and it does not get imported nor shown.

For the apps "The Mantrailing App" or "GPX Viewer", it is possible to open the GPX-files directly out of the Signal messenger app.

jensMF:


jensMF:
Found out the real fix. Will open a new pr today or tomorrow.

PR #7847, opened 2019-11-04
Folder per day as subfolder

jensMF:
The daily folders are now subdirectories of the monthly folders and the setting is now also used when importing a gpx-track


vshcherb:
This is a discussible feature, we need to open an issue to make decision first

jensMF:
Ok, I thought it could be discussed in the pull request as well, but then I would open issues for both changes?

PR #7992, opened 2019-11-27
Fix #7394

### LINKED ISSUE #7394: [Feature request] Calculate and store bearing (aka heading/direction/course) in gpx when using track recording plugin
Hi,
I am using OsmAnd android app to record gpx tracks, using 'Track recording plugin'.
Recorded tracks are used for geotagging of the images for mapillary.
Essential data required is coodinates, date&time and heading (direction).

OsmAnd track recording plugin does not store heading at the moment which makes batch processing a bit more complicated, while exiftool does not support heading interpolation from gpx either.

There are various tools available to do the job, but for the sake of workflow simplicity and speed, it would be great if OsmAnd will add heading support to its 'track recording' plugin.
Exiftool could pickup heading from gpx (looking up  <course> tags)

Sample trackpoint:

```
<trkpt lat="48.225437" lon="16.370304">
				<ele>4.76</ele>
				<time>2012-07-03T19:32:34+02:00</time>
				<course>45.16</course>
				<roll>-3.3</roll>
				<pitch>-3.26</pitch>
</trkpt>

```

nazar-kutz:


OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
```diff
@@ -187,6 +187,7 @@ private int parseColor(String colorString, int defColor) {
 		public double ele = Double.NaN;
 		public double speed = 0;
 		public double hdop = Double.NaN;
+		public float heading = -1.0f;
```
max-klaus: Use Double.NaN here

OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
```diff
@@ -1606,6 +1616,9 @@ private static void writeWpt(SimpleDateFormat format, XmlSerializer serializer,
 		if (p.speed > 0) {
 			p.getExtensionsToWrite().put("speed", decimalFormat.format(p.speed));
 		}
+		if (p.heading >= 0) {
+			p.getExtensionsToWrite().put("heading", decimalFormat.format(p.heading));
```
max-klaus: We do not need heading as decimal. Use integer.

OsmAnd-java/src/main/java/net/osmand/util/MapUtils.java
```diff
@@ -410,6 +410,16 @@ public static float unifyRotationTo360(float rotate) {
 		}
 		return rotate;
 	}
+	
+	public static float formatDegrees360(float degrees) {
```
max-klaus: formatDegrees360 > normalizeDegrees360

OsmAnd/src/net/osmand/plus/OsmAndLocationProvider.java
```diff
@@ -771,7 +771,7 @@ public void setLocationFromService(net.osmand.Location location, boolean continu
 		if (continuous) {
 			scheduleCheckIfGpsLost(location);
 		}
-		app.getSavingTrackHelper().updateLocation(location);
+		app.getSavingTrackHelper().updateLocation(location, app.getLocationProvider().heading);
```
max-klaus: app.getLocationProvider() ?

OsmAnd/src/net/osmand/plus/activities/SavingTrackHelper.java
```diff
@@ -402,14 +409,19 @@ private void collectDBTracks(SQLiteDatabase db, Map<String, GPXFile> dataTracks)
 	public void startNewSegment() {
 		lastTimeUpdated = 0;
 		lastPoint = null;
-		execWithClose(updateScript, new Object[] { 0, 0, 0, 0, 0, System.currentTimeMillis()});
+		execWithClose(updateScript, new Object[] { 0, 0, 0, 0, 0, System.currentTimeMillis(), -1.0f});
```
max-klaus: Create constant for -1.0

PR #8017, opened 2019-12-01
migration to update file paths in the db

jensMF:
added migration to update file paths in the db according to the new 'getFileDir' method

### DIFF
--- a/OsmAnd/src/net/osmand/plus/GPXDatabase.java
+++ b/OsmAnd/src/net/osmand/plus/GPXDatabase.java
@@ -343,12 +343,26 @@ public class GPXDatabase {
 		}
 		if (oldVersion < 11) {
 			File lookupDir = context.getAppPath(IndexConstants.GPX_INDEX_DIR);
-			List<File> toUpdate = allFilesInDir(lookupDir);
-			for (File file : toUpdate) {
-				db.execSQL("UPDATE " + GPX_TABLE_NAME +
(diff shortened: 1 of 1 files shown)


max-klaus:
I do not think we need such migration. cleanUpDatabase will be enough.

jensMF:
I am not sure about that.
This migration makes sure, that all gpx files in the Tracks folder and all its sub-folders are in the database and have the correct fileDir. It also makes sure, that there are only entries corresponding to a file.
The cleanUpDatabase only deletes entries without a correct path. So it deletes all entries with files in sub/sub folders of Tracks created before the change of the getFileDir method.
In my case and the case of my friends (because we all used monthly and then daily folders) nearly the whole database is deleted. And that means, that the entries are created, when the folder is unfolded in My Places/Tracks, what needs a long time on older and slower devices, and changes all the color settings.

jensMF:
The idea of deleting all entries and recreate them for every file does not work, because there is no method to insert a file.

max-klaus:
New getFileDir method is released already. So migration not make sense.

jensMF:
Well of course the new getFileDir method is released, but every database entry that existed before, is still in the database, and every track that was not already shown in the My Places/Tracks ui view after that release, does not have a correct entry. Both is corrected by this migration.

max-klaus:
New correct entries for tracks will be added anyway. Its not a problem.

PR #8018, opened 2019-12-01
GPXDatabase clean up function

jensMF:
added function to clean up gpx db an running it in 'openConnection' method

### DIFF
--- a/OsmAnd/src/net/osmand/plus/GPXDatabase.java
+++ b/OsmAnd/src/net/osmand/plus/GPXDatabase.java
@@ -258,7 +258,40 @@ public class GPXDatabase {
 				onUpgrade(conn, version, DB_VERSION);
 			}
 		}
+		cleanUpDatabase(conn);
 		return conn;
+	}
+
+	private void cleanUpDatabase(SQLiteConnection db) {
+		SQLiteCursor query = db.rawQuery("SELECT " + GPX_COL_NAME + ", " +
+				GPX_COL_DIR +
+				" FROM " + GPX_TABLE_NAME, null);
+		if (query != null) {
+			try {
+				if (query.moveToFirst()) {
+					do {
+						String fileName = query.getString(0);
+						String fileDir = query.getString(1);
+						File dir;
+						if (!Algorithms.isEmpty(fileDir)) {
+							dir = new File(context.getAppPath(IndexConstants.GPX_INDEX_DIR), fileDir);
+						} else {
+							dir = context.getAppPath(IndexConstants.GPX_INDEX_DIR);
+						}
+						File toCheck = new File(dir, fileName);
+						if (!toCheck.exists()) {
+							db.execSQL("DELETE FROM " + GPX_TABLE_NAME +
+											" WHERE " +
+											GPX_COL_DIR + " = ? AND " +
+											GPX_COL_NAME + " = ?",
+									new Object[]{fileDir, fileName});
+						}
+					} while (query.moveToNext());
+				}
+			} finally {
+				query.close();
+			}
+		}
 	}
 
 	private void onCreate(SQLiteConnection db) {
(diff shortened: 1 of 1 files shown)


jensMF:
Is it possible, that the file name is legally null or empty in the database?
If so we should add
```
" WHERE " + GPX_COL_NAME + " IS NOT null AND " + GPX_COL_NAME + " <> '' " +
```
between the FROM and the GROUP BY lines.

vshcherb:
This is not really clear why it needs to be cleaned up on every connection, if it is one time cleaning it should be related to version. If cleaning is needed regularly then it is probably it is better still do it on application upgrade or past some time though we probably need to fix the origin of the problem

OsmAnd/src/net/osmand/plus/GPXDatabase.java
```diff
@@ -258,9 +258,47 @@ SQLiteConnection openConnection(boolean readonly) {
 				onUpgrade(conn, version, DB_VERSION);
 			}
 		}
+		cleanUpDatabase(conn);
```
max-klaus: It is too often. Please move cleanUpDatabase to constructor (239 line)

OsmAnd/src/net/osmand/plus/GPXDatabase.java
```diff
@@ -258,9 +258,47 @@ SQLiteConnection openConnection(boolean readonly) {
...
 		return conn;
 	}
 
+	private void cleanUpDatabase(SQLiteConnection db) {
+        db.execSQL("DELETE FROM " + GPX_TABLE_NAME +
```
max-klaus: What purpose of this query?

jensMF: This query deletes all duplicates, where fileDir and filName are the same (that was possible because of the old getFileDir method, if we have the migration, this is not needed, I think)

OsmAnd/src/net/osmand/plus/GPXDatabase.java
jensMF: Done

OsmAnd/src/net/osmand/plus/GPXDatabase.java
jensMF: As the decision is made, that the migration is not needed, this just speeds up the cleanup at the first time, because there are tens of thousands of duplicate database entries, if one has hundreds of gpx files in sub-sub-folders of rec (I had about 250 files in sub-sub-folders and about 12000 database entries, all duplicates of this 250 files).

vshcherb:
This is really great that you have made this change and at least simplified life to yourself. We definitely are going to work with this database more & more. For example we have plans to build Heatmap for all tracks or tracks by period (for month / year) and also calculates various reports. Probably we can store an image of the track in this database to visualize it and improve searching facilities

vshcherb:
In case you need one of this function and willing to contribute, you are welcome!

jensMF:
> This is really great that you have made this change and at least simplified life to yourself. We definitely are going to work with this database more & more. For example we have plans to build Heatmap for all tracks or tracks by period (for month / year) and also calculates various reports. Probably we can store an image of the track in this database to visualize it and improve searching facilities

I think it is a great Idea to store/create/calculate more Information about the Track (maybe some day also weather data), but I dont't know what you want to tell me the rest of your comment.
The possibility to store Tracks in Monthly folders was in the app before I started to contribute. The possibility to store Tracks in daily folders, added by a pull request I created, did work the same way. Everyone who used one of this possibilities before the change of the getFileDir method in the GXPDatabase class by commit 58d56a6d24639f9b743e40771a7515066d3979d0, has multiple entries for every file in a Monthly or Daily folder (or another sub-sub folder of the tracks directory). Every time a track analysis was needed a new entry was created. In addition, if someone deletes tracks not through OsmAnd, nothing gets deleted in the database. So especially when the Database should be used more, I think it should be cleaned up some times.
If you think the query to delete entries with the same file name AND the same directory name could break things in the future, I can remove it of course. As I mentioned, it would only speed up the cleaning the first time it runs and only if there are such duplicates.
I moved the cleaning to the constructor, as @max-klaus asked me.  If you think this is still too often, I can think of a way to reduce it further.
If there is an alternative plan to keep the database clean and consistent, I will stop talking about it, so just let me know.

Issue #7437, opened 2019-08-18
Move screenshots and metadata for app stores upstream


PanderMusubi:
For automated adding of screenshots and metadata to app stores, several mechanisms exist such as metadata and fastlane. This saves the maintainer a lot of manual work uploading screenshots to app stores when new screenshots are available. This is a widely used mechanism helping releasing new versions of apps.

For the F-Droid app store, screenshots and metadata have been created to facilitate this. These can be found at https://gitlab.com/fdroid/fdroiddata/tree/master/metadata/net.osmand.plus/en-US

The request is to move this data upstream from the F-Droid repo to this repo. If it is desired, I can make a pull request for this. Let me know if this is welcomed. This will move screenshots to a more generic location allowing maintenance in one place for the benefit of all app stores.

PanderMusubi:
This issue is part of an effort to remove custom screenshots from the F-Droid data repo, see also https://gitlab.com/fdroid/fdroiddata/issues/1413#note_193403754

IzzySoft:
For details, please also see: https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/#in-the-apps-source-repository

Note that this gives you a bunch of advantages as a welcome side effect:

* you can easily update/replace screenshots without the need of opening MRs at fdroiddata
* you can easily maintain app descriptions (including multi-locale) directly in your app's repo
* you can use the same for F-Droid and Google Play (so no redundant double-work)
* you can also include per-release changelogs (`changelogs/<versionCode>.txt`) which are displayed inline with the app description on both, F-Droid and Google Play – which in the F-Droid client make your app appear on the featured tab (still called "new") if you have screenshots along those changelogs

Thus resulting in a clear win-win-win (less work for you, less work for us, better information for the user). So this move is strongly encouraged :wink:

vshcherb:
Yes it sounds good, though we use 2 different repos for images and for text / translations. Though for sure it would be good to be maintained by us

PanderMusubi:
Super. Let us know when you have support for this, so that metadata can get removed from the F-Droid repo.

vshcherb:
@PanderMusubi please create pull-request to this repo (https://github.com/osmandapp/Osmand-misc) to one of the subfolder

Concern: if you have automation and audience that can translate the app name / descriptions that probably will lack temporarily.

Once you create pull-request we will know that you are ready to support that location and we will take over.

PanderMusubi:
[...]

PR #8047, opened 2019-12-06
fastlane images, optimized with optipng

PanderMusubi:


mikehgentry:
Image 02 has a typo - "you" should be "your"

vshcherb:
@PanderMusubi  is it fine to store it here https://github.com/osmandapp/Osmand-misc ? This repository is already quite huge to store resources - same folder fastline is fine

PanderMusubi:
Yes of course

vshcherb:
Great looking forward! We will update screenshots with the most recent version and different translations

shuvashish76:
@vshcherb
But as Izzy mentioned [here](https://github.com/osmandapp/OsmAnd/issues/7437#issuecomment-578325416): _They are in a separate repo, they won't show up e.g. at F-Droid, which would look for Fastlane data in the source repo (which would be here)._

So [those files](https://github.com/osmandapp/OsmAnd-misc/tree/master/fastlane/metadata/android) are laying there with no use to F-Droid. Would you rather consider including metadata here?

PR #8126, opened 2019-12-20
Settings for buffer

veliymolfar:


OsmAnd-telegram/src/net/osmand/telegram/helpers/LocationMessages.kt
```diff
@@ -140,8 +147,27 @@ class LocationMessages(val app: TelegramApplication) {
 		dbHelper.removeBufferedMessage(message)
 	}
 
+	private fun removeOldBufferedMessages() {
+		if (this.bufferedMessages.isNotEmpty()) {
```
max-klaus: Request not more often than 1 minute. It also could simplified using .filter. Copy the list, retrieve only expired messages using filter and delete them from db and cache. Use mutableListOf<T>().apply { addAll(this) } for creating mutable copy. Also it would be good to replace mutableListOf(*this.bufferedMessages.toTypedArray()) with addAll.

OsmAnd-telegram/res/values/strings.xml
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+    <string name="buffer_time_descr">Time after which buffered messages will be deleted</string>
```
max-klaus: Are strings agreed?

Issue #8127, opened 2019-12-21
[Feature request] LiveMonitoringHelper responseBody should be parsed for OsmandAidlServiceV2 commands


goatchurchprime:
With some coding in a [Node-RED](https://nodered.org/), this would provide a method to share locations in realtime (Issue #7469), as well as other interesting features (see below).    

The online tracking feature works by calling the function [`LiveMonitoringHelper.senddata(LiveMonitoringData data)`](https://github.com/osmandapp/Osmand/blob/8030287b880a084125829acda0ed3776165a0b96/OsmAnd/src/net/osmand/plus/monitoring/LiveMonitoringHelper.java#L130) which populates the user-specified `live_monitoring_url` format string with positional data and sends out a GET request to the internet.  The response string, which is received, is discarded.  

(The LiveMonitoringData objects are actually local and are created and queued up in the function `LiveMonitoringHelper.updateLocation(net.osmand.Location location)`.)

Meanwhile there is an API used deep within the Telegram sharing app for moving and changing the state of a set of favourite markers.  The function which sends out the positions is  [TelegramNotification.refreshNotification()](https://github.com/osmandapp/Osmand/blob/27469a9e7e4d022a312002e8780ba290d8467a0a/OsmAnd-telegram/src/net/osmand/telegram/notifications/TelegramNotification.kt#L83) and marker updates are done through the [telegram.helpers.OsmandAidlHelper.updateFavorite()](https://github.com/osmandapp/Osmand/blob/27469a9e7e4d022a312002e8780ba290d8467a0a/OsmAnd-telegram/src/net/osmand/telegram/helpers/OsmandAidlHelper.kt#L388) function.  

The same interface exist at [OsmandAidlServiceV2](https://github.com/osmandapp/Osmand/blob/355f474d4ed6059b799478d09cfb9d8678ac7a5d/OsmAnd/src/net/osmand/aidl/OsmandAidlServiceV2.java) ("Aild" stands for "Android Interface Definition Language").

The interesting functions in this interface are: addFavoriteGroup(), removeFavoriteGroup(), updateFavoriteGroup(), addFavorite(), updateFavorite(), removeFavorite() (Question: is "category string" the same as  "favourite group"?)

**Proposal:** If the return string from `LiveMonitoringHelper.updateLocation()` were parsed for function calls and parameters in that interface (maybe it's in JSON format), then this could go a very long way towards enabling live tracking and some productive geocaching work.

[...]

PR #8178, opened 2020-01-05
Use the response text from live monitoring to send commands to the API

goatchurchprime:
This purpose of this change is to enable live-tracking (and other sophisticated features) without a dependence on Telegram.

I opened this pull-request following a [comment](https://github.com/osmandapp/Osmand/issues/8127#issuecomment-569610900) that it would make it easier for the other developers to understand what I'm trying to do.  

Instead of Telegram, we can use Node-RED as the location broker, which can easily be programmed to implement live-tracking.  (The same system will also be able to push non-player targets to the participants as well.)

I have made an example node-RED editor at https://osmtest1.eu-gb.mybluemix.net/red/ following [these instructions](https://nodered.org/docs/getting-started/ibmcloud).  

The first node is listening for any "http in" signals on the path "/osmand1", the second function node decodes the lat,lon,bearing parameters and adds a random position from them, and the final "http response" node returns the json-encoded osmapicommand response of the form `{"osmandcommands":["osmand.api://add_map_marker?name=nodred-3&lat=52.19&lon=0.10"]}`.

Once you have this pull-request compiled, you need to enable the location tracking to the correct URL with the following steps:

* On Osmand go to plugins -> Trip recording -> Settings -> Cycling -> Online tracking web address and set to:

* http://osmtest1.eu-gb.mybluemix.net/osmand1?lat={0}&lon={1}&timestamp={2}&hop={3}&altitude={4}&speed={5}&bearing={6}

* Click "Apply to all profiles"

* Then click "Online tracking (GPX required)"  and Apply to all profiles

Now on the main map click on "Start GPX logging"

If you have the node-RED editor up, then you click on the bug symbol near the top right corner to see the messages that are coming in and going out.

There are three issues to consider concerning this pull-request:

1. The response is in json form "{ osmandapicmds: [ apistring1, apistring2, ...]}" because there is a json decoder in java, and it's future-proof -- eg extra parameters can be added and ignored.

2. The function [ExternalApiHelper.processApiRequest()](https://github.com/osmandapp/Osmand/blob/21ac829a3da5d2b8ebeee6dd67c08a90a604c824/OsmAnd/src/net/osmand/plus/helpers/ExternalApiHelper.java#L174) is incomplete, because it implements "add_favourite" but not "update_favourite".  (Also, the update_favourite() functions implemented in other places are problematic because it requires the previous lat/lon to be accurately known before it changes it in the database.  This won't work if there are a update message gets missed.)

3. MQTT might be more a appropriate protocol for the purpose of sending and receiving live locations, instead of http polling.  This was noted in [this issue thread](https://github.com/traccar/traccar/issues/1582#issuecomment-343252080) for traccar.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/monitoring/LiveMonitoringHelper.java
+++ b/OsmAnd/src/net/osmand/plus/monitoring/LiveMonitoringHelper.java
@@ -1,6 +1,8 @@
 package net.osmand.plus.monitoring;
 
 import android.content.Context;
+import android.content.Intent;
+import android.net.Uri;
 import android.os.AsyncTask;
 
 import net.osmand.PlatformUtil;
@@ -9,9 +11,14 @@ import net.osmand.plus.OsmAndLocationProvider;
 import net.osmand.plus.OsmandApplication;
 import net.osmand.plus.OsmandPlugin;
 import net.osmand.plus.OsmandSettings;
+import net.osmand.plus.activities.MapActivity;
+import net.osmand.plus.helpers.ExternalApiHelper;
 import net.osmand.util.MapUtils;
 
 import org.apache.commons.logging.Log;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
 
 import java.io.BufferedReader;
 import java.io.InputStream;
@@ -33,9 +40,15 @@ public class LiveMonitoringHelper  {
 	private final static Log log = PlatformUtil.getLog(LiveMonitoringHelper.class);
 	private ConcurrentLinkedQueue<LiveMonitoringData> queue;
 	private boolean started = false;
+	private OsmandApplication app = null;
+
+	// set by OsmandMonitoringPlugin.registerLayers()
+	MapActivity activity = null;
+	ExternalApiHelper apiHelper = null;
 
 	public LiveMonitoringHelper(Context ctx){
 		this.ctx = ctx;
+		app = (OsmandApplication) ctx.getApplicationContext();
 		settings = ((OsmandApplication) ctx.getApplicationContext()).getSettings();
 		queue = new ConcurrentLinkedQueue<>();
 	}
@@ -198,6 +211,26 @@ public class LiveMonitoringHelper  {
 					is.close();
 				}
 				log.info("Monitor response (" + urlConnection.getHeaderField("Content-Type") + "): " + responseBody.toString());
+
+				try {
+					JSONObject obj = new JSONObject(responseBody.toString());
+					JSONArray osmandapicmds = obj.getJSONArray("osmandapicmds");
+					for (int i = 0; i < osmandapicmds.length(); i++) {
+						String osmandapicmd = osmandapicmds.getString(i);
+						//String osmandapicmd = "osmand.api://add_map_marker?lat=54.6622958&lon=-5.7666790&name=thing2";
+						log.info("Monitor received"+i+": "+osmandapicmd);
+						if (osmandapicmd.startsWith("osmand.api://")) {
+							Intent intent = new Intent(activity, app.getAppCustomization().getFavoritesActivity());
+							Uri netdata = Uri.parse(osmandapicmd);
+							intent.setData(netdata);
+							Intent result = apiHelper.processApiRequest(intent);
+						}
+					}
+				} catch (JSONException e) {
+					String message = "JSON parsing error: " + (e.getMessage() == null ? "unknown" : e.getMessage());
+					//app.showShortToastMessage(MessageFormat.format(activity.getString(R.string.error_message_pattern), message));
+					log.error(message);
+				}
 			}
 
 			urlConnection.disconnect();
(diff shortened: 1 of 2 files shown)


vshcherb:
1. First of all thanks for pull request.
2. It definitely needs  a UI setting which will be switched of by default cause it is pretty security sensitive feature

OsmAnd/src/net/osmand/plus/monitoring/LiveMonitoringHelper.java
```diff
@@ -198,6 +211,26 @@ public void sendData(LiveMonitoringData data) {
...
+
+				try {
+					JSONObject obj = new JSONObject(responseBody.toString());
+					JSONArray osmandapicmds = obj.getJSONArray("osmandapicmds");
+					for (int i = 0; i < osmandapicmds.length(); i++) {
+						String osmandapicmd = osmandapicmds.getString(i);
+						//String osmandapicmd = "osmand.api://add_map_marker?lat=54.6622958&lon=-5.7666790&name=thing2";
+						log.info("Monitor received"+i+": "+osmandapicmd);
+						if (osmandapicmd.startsWith("osmand.api://")) {
+							Intent intent = new Intent(activity, app.getAppCustomization().getFavoritesActivity());
```
vshcherb: app.getAppCustomization().getFavoritesActivity()
it looks very weird...

OsmAnd/src/net/osmand/plus/monitoring/OsmandMonitoringPlugin.java
```diff
@@ -106,6 +107,8 @@ public String getHelpFileName() {
 
 	@Override
 	public void registerLayers(MapActivity activity) {
+		liveMonitoringHelper.activity = activity;
```
vshcherb: this is definitely could cause "UI Activity leak" and it is bad

vshcherb:
I don't know where to take UI activity actually and I think it should be propagated through "sendData" method though it also could be running in the background service so there will be no activity at all. So probably it is better to get it working without activity in some sense.

@Chumva please add a comment if you have

Chumva:
There are several commands which change UI or show some dialogues, which need mapActivity. Probably we could add broadcast receivers as in OsmandAidlApi for these commands and remove mapActivity as a global variable.

goatchurchprime:
I guess couldn't sneak that one past you!

Here's the problem.

There are two different APIs in Osmand.  The first uses this AIDL (Android Interface Definition Language) with function call-backs, registering receivers and all sorts of really complicated stuff I don't understand, and the other is implemented in the `processApiRequest()` function and takes strings of the form "osmand.api://add_map_marker..." that are parsed using the standard `Uri.parse()` function.  

I obviously wanted to use the string-based API because the `LiveMonitoringHelper` receives response strings from the server, and I wouldn't need to invent a new format to interpret those strings into corresponding AIDL function calls-- or have to learn about AIDL.  

So that's two good reasons I wanted to use the string-based API as implemented in the `processApiRequest()` function, even though the implementation is incomplete.  

Now, most of the [apiHelper.processApiRequest()](https://github.com/osmandapp/Osmand/blob/21ac829a3da5d2b8ebeee6dd67c08a90a604c824/OsmAnd/src/net/osmand/plus/helpers/ExternalApiHelper.java#L174) function works with just a `Uri` object, apart from the `loadGPXFile()` feature which needs some extra data that is conveniently packaged in the Intent object.  

So if the `processApiRequest()` function were split into two functions `processApiRequest_ShowGPX()` and `processApiRequest_EverythingElse()`, then the second one could be called with just a `Uri` object without having to package the data up into an unnecessary `Intent` object.  

(The other schemes, "file", "content" and "google.navigation" that are also decoded in the `onResume()` function in addition to "osmand.api", don't pollute their implementations with these `Intent` objects.)

I obviously made this feature with the minimum of code-changes, so it's hacky.  If I knew the policy for the text-string-based API (as is used only by the `onResume()` function), then I'd know what direction to push it in.

The implementations of each of the API features are in various `SomethingHelper` classes.  For example, the implementation `ExternalApiHelper.searchAndNavigate()` is called from `ExternalApiHelper.processApiRequest()` and from a local function defined in `OsmandAidlApi.registerNavigateSearchReceiver()`.  

And the implementation `FavouritesDbHelper.addFavourite()` is also called from `ExternalApiHelper.processApiRequest()`, but its AIDL-API gets to it through `OsmandAidlApi.addFavorite()` which is actually called by `OsmandAidlServiceV2.addFavorite()`.  (I know that I do not understand how Services work.)

I think that the cleanest place for a URI-formatted string-driven API to use as the entry point has got to be the `processApiRequest_EverythingElse()` function proposed above, but without the `Intent result` return value.  

However, I don't know if we are sure that a string-driven API based on the format "osmand.api://..." is desirable or in the right place.  Just because I found a half-working API being used from the `onResume()` function doesn't mean it's a good thing to encourage.  It's is extremely useful to have, but when it's there you can get lazy with internal functions packaging up strings just to call it because it's so powerful and easy.

The next thing I'll need are implementations for "osmand.api://update_favorite" and "osmand.api://delete_favorite", and then I'll have the minimum required to replicate live tracking without Telegram.

PR #8242, opened 2020-01-16
Import Profile

nazar-kutz:
Add UI and behavior for profile and routing file importing


OsmAnd/src/net/osmand/plus/SettingsHelper.java
```diff
@@ -83,7 +83,7 @@
 	private ImportAsyncTask importTask;
 
 	public interface SettingsImportListener {
-		void onSettingsImportFinished(boolean succeed, boolean empty);
+		void onSettingsImportFinished(boolean succeed, boolean empty, List<SettingsItem> items);
```
max-klaus: Needs annotation: List<SettingsItem> items

PR #8258, opened 2020-01-18
Fix typo in IT strings regarding first wizard

ema987:


comradekingu:
@ema987 Fixed in https://hosted.weblate.org/translate/osmand/main/it/?checksum=c03b653f2b63b23f

PR #8309, opened 2020-01-24
New strings reworked

comradekingu:
GPLv3+ licensed.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -11,21 +11,21 @@
 	Thx - Hardy
 
 -->
-    <string name="permission_is_required">Permission is required to use this option.</string>
-    <string name="logcat_buffer_descr">Check and share detailed logs of the application</string>
+    <string name="permission_is_required">You must grant the required permission to use this.</string>
+    <string name="logcat_buffer_descr">Check and share detailed logs of the app</string>
     <string name="file_does_not_contain_routing_rules">No routing rules in \'%1$s\'. Please choose another file.</string>
     <string name="not_support_file_type_with_ext">Select a supported %1$s extension file instead.</string>
-    <string name="monitoring_min_speed_descr">This is a low-speed cut-off filter to not record points below a certain speed. This may make recorded tracks look "smoother" when viewed on the map.</string>
-    <string name="monitoring_min_speed_descr_side_effect">Side effect: Your track will be missing all sections where the minimum speed criterion was not met (e.g. where you push your bike up a steep hill). Also, there will be no information about periods at rest, like breaks. This has effects on any analysis or post-processing, like when trying to determine the total length of your trip, time in motion, or your average speed.</string>
-    <string name="monitoring_min_speed_descr_recommendation">Recommendation: Try using the motion detection via the "logging minimum displacement" filter (B) first, it may produce better results, and you will lose less data. If your tracks remain noisy at low speeds, try non-zero values here. Please note that some measurements may not report any speed value at all (some network-based methods), in which case you would not record anything.</string>
-    <string name="monitoring_min_speed_descr_remark">Remark: "speed > 0" check: Most GPS chipsets report a speed value only if the algorithm determines you are in motion, and none if you are not. Hence using the "> 0" setting in this filter in a sense uses the motion detection of the GPS chipset. But even if not filtered here at recording time, we still use this feature in our GPX analysis to determine the "Distance corrected", i.e. the value displayed in that field is the "distance recorded while in motion".</string>
-    <string name="monitoring_min_accuracy_descr">This will record only points measured with a minimum "accuracy" indication (in meters/feet, as reported by Android for your chipset). Accuracy refers to the scatter of repeated measurements, and is not directly related to precision, which defines how close your measurements are to your "true" position.</string>
-    <string name="monitoring_min_accuracy_descr_side_effect">Side effect: As a result of filtering by accuracy, points may be entirely missing for e.g. below bridges, under trees, between high buildings, or with certain weather conditions.</string>
-    <string name="monitoring_min_accuracy_descr_recommendation">Recommendation: It is hard to predict what will be recorded and what not, it may be best to turn this filter off.</string>
-    <string name="monitoring_min_accuracy_descr_remark">Remark: If GPS had been off immediately before a recording, the first point measured may have a decreased accuracy,so in our code we may want to wait a second or so before recording a point (or record the best of 3 consecutive points, etc.), but this is not yet implemented.</string>
-    <string name="monitoring_min_distance_descr">This filter avoids duplicate points being recorded where too little actual motion may have occurred, makes a nicer spatial appearance of tracks not post-processed later.</string>
-    <string name="monitoring_min_distance_descr_side_effect">Side effects: Periods at rest are not recorded at all or by just one point each. Small (real world) movements (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less information for post-processing, and has worse stats by filtering out obviously redundant points at recording time, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
-    <string name="monitoring_min_distance_descr_recommendation">Recommendation: A setting of 5 meters may work well for you if you do not require to capture details finer than that, and do not want to explicitly capture data while at rest.</string>
+    <string name="monitoring_min_speed_descr">This is a low-speed cut-off filter smoothing recorded tracks discarded track points below a certain speed.</string>
+    <string name="monitoring_min_speed_descr_side_effect">Whole sections below minimum speed will be discarded, like the steep hill you pushed your bike up. This includes stationary periods, like breaks. Total length, time in motion, and your average speed can still be determined.</string>
+    <string name="monitoring_min_speed_descr_recommendation">Try using motion detection via the "Minimum logging displacement" filter (B) first. It may produce better results, and you will lose less data. If your tracks remain noisy at low speeds, try non-zero values here. Please note, some measurements may not report any speed value at all (some network-based methods), in which case you would not record anything.</string>
+    <string name="monitoring_min_speed_descr_remark">"Speed > 0" check: Most GPS chipsets only report a speed if determining you are in motion. Hence using the "> 0" setting in this filter in a sense uses the motion detection of the GPS chipset. Even if not filtered in OsmAnd at recording time, it still uses this feature GPX analysis to determine the \"Distance corrected\", i.e. the value displayed in that field is the "distance recorded while in motion".</string>
+    <string name="monitoring_min_accuracy_descr">This will record only points measured with a minimum "accuracy" indication (in meters/feet, as reported by Android for your chipset). \"Accuracy\" refers to the scatter of repeated measurements, and is not directly related to precision, which defines how close your measurements are to your \"true\" position.</string>
+    <string name="monitoring_min_accuracy_descr_side_effect">As a result of filtering by accuracy, points may not be recorded below bridges, under trees, between high buildings, or with certain weather conditions.</string>
+    <string name="monitoring_min_accuracy_descr_recommendation">It is hard to predict what will be recorded, it may be best to turn this filter off.</string>
+    <string name="monitoring_min_accuracy_descr_remark">If GPS is only on right before a recording, first point accuracy may decrease, so the OsmAnd strategy may be to wait a second or so before recording a point (or take the median of 3 consecutive points, etc.) in the future.</string>
+    <string name="monitoring_min_distance_descr">This filter avoids duplicate point recording for insufficient actual motion, for a nicer spatial appearance of tracks not post-processed later.</string>
+    <string name="monitoring_min_distance_descr_side_effect">Periods at rest are not recorded, or one point each. Small (real world) movement (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less info for post-processing, and has worse stats by filtering out superfluous points when recording, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
+    <string name="monitoring_min_distance_descr_recommendation">A setting of 5 meters may work well if you don\'t require any finer details, and don\'t want accuracy in determined stops.</string>
     <string name="live_monitoring_time_buffer">Time buffer</string>
     <string name="live_monitoring_tracking_interval">Tracking interval</string>
     <string name="live_monitoring_adress">Web address</string>
@@ -34,14 +34,14 @@
     <string name="monitoring_min_speed">Minimum speed</string>
     <string name="monitoring_min_accuracy">Minimum accuracy</string>
     <string name="monitoring_min_distance">Minimum displacement</string>
-    <string name="tracks_view_path">Menu — My places — Tracks</string>
-    <string name="multimedia_notes_view_path">Menu — My places — Notes</string>
-    <string name="osm_edits_view_path">Menu — My places — OSM Edits</string>
-    <string name="reset_plugin_to_default">Reset plugin settings to default</string>
+    <string name="tracks_view_path">Menu → My places → Tracks</string>
+    <string name="multimedia_notes_view_path">Menu → My places → Notes</string>
+    <string name="osm_edits_view_path">Menu → My places → OSM edits</string>
+    <string name="reset_plugin_to_default">Reset plugin settings to defaults</string>
     <string name="multimedia_rec_split_title">Recorder split</string>
     <string name="multimedia_use_system_camera">Use system app</string>
     <string name="multimedia_photo_play_sound">Camera shutter sound</string>
-    <string name="osm_authorization_success">Authorization is successful</string>
+    <string name="osm_authorization_success">Authorized</string>
     <string name="import_from_file">Import from file</string>
     <string name="import_routing_file">Import routing file</string>
     <string name="import_profile">Import profile</string>
@@ -125,7 +125,7 @@
     <string name="quick_action_contour_lines_show">Show contour lines</string>
     <string name="quick_action_contour_lines_hide">Hide contour lines</string>
     <string name="quick_action_show_hide_contour_lines">Show/hide contour lines</string>
-    <string name="quick_action_hillshade_descr">A button to show or hide hillshades on the map.</string>
+    <string name="quick_action_hillshade_descr">A button to show or hide map hillshading.</string>
     <string name="quick_action_hillshade_show">Show hillshade</string>
     <string name="quick_action_hillshade_hide">Hide hillshade</string>
     <string name="quick_action_show_hide_hillshade">Show/hide hillshade</string>
@@ -1112,11 +1112,11 @@
 	and more…
     </string -->
     <string name="rendering_attr_contourColorScheme_description">Contour lines color scheme</string>
-    <string name="save_track_min_speed">Logging minimum speed</string>
+    <string name="save_track_min_speed">Minimum logging speed</string>
     <string name="save_track_min_speed_descr">Filter: No logging of points below this speed.</string>
-    <string name="save_track_min_distance">Logging minimum displacement</string>
+    <string name="save_track_min_distance">Minimum logging displacement</string>
     <string name="save_track_min_distance_descr">Filter: Set minimum distance from a point to log a new one.</string>
-    <string name="save_track_precision">Logging minimum accuracy</string>
+    <string name="save_track_precision">Minimum logging accuracy</string>
     <string name="save_track_precision_descr">Filter: No logging unless this accuracy is reached.</string>
     <string name="christmas_poi">Christmas POI</string>
     <string name="christmas_desc">Anticipating Christmas and New Year holidays, you can choose to display associated POIs like Christmas trees and markets, etc.</string>


OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_distance_descr">This filter avoids duplicate points being recorded where too little actual motion may have occurred, makes a nicer spatial appearance of tracks not post-processed later.</string>
-    <string name="monitoring_min_distance_descr_side_effect">Side effects: Periods at rest are not recorded at all or by just one point each. Small (real world) movements (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less information for post-processing, and has worse stats by filtering out obviously redundant points at recording time, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
-    <string name="monitoring_min_distance_descr_recommendation">Recommendation: A setting of 5 meters may work well for you if you do not require to capture details finer than that, and do not want to explicitly capture data while at rest.</string>
```
comradekingu: Not 100% sure here. What does "explicitly" signify other than it being an active and visible choice to do so?

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
 	Thx - Hardy
 
 -->
-    <string name="permission_is_required">Permission is required to use this option.</string>
-    <string name="logcat_buffer_descr">Check and share detailed logs of the application</string>
+    <string name="permission_is_required">You must grant the required permission to use this.</string>
```
sonora: Not better, but longer than the original?

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="logcat_buffer_descr">Check and share detailed logs of the app</string>
```
sonora: Fine by me.

OsmAnd/res/values/strings.xml
```diff
@@ -1112,11 +1112,11 @@
 	and more…
     </string -->
     <string name="rendering_attr_contourColorScheme_description">Contour lines color scheme</string>
-    <string name="save_track_min_speed">Logging minimum speed</string>
+    <string name="save_track_min_speed">Minimum logging speed</string>
```
sonora: No. ;)

"Logging minimum speed" has a different meaning from "Minimum logging speed". (There is no such thing as "logging speed" similarly in your changes below.)

OsmAnd/res/values/strings.xml
```diff
@@ -125,7 +125,7 @@
     <string name="quick_action_contour_lines_show">Show contour lines</string>
     <string name="quick_action_contour_lines_hide">Hide contour lines</string>
     <string name="quick_action_show_hide_contour_lines">Show/hide contour lines</string>
-    <string name="quick_action_hillshade_descr">A button to show or hide hillshades on the map.</string>
+    <string name="quick_action_hillshade_descr">A button to show or hide map hillshading.</string>
```
sonora: We call them hillshades everywhere.

OsmAnd/res/values/strings.xml
```diff
@@ -34,14 +34,14 @@
...
-    <string name="reset_plugin_to_default">Reset plugin settings to default</string>
+    <string name="tracks_view_path">Menu → My places → Tracks</string>
+    <string name="multimedia_notes_view_path">Menu → My places → Notes</string>
+    <string name="osm_edits_view_path">Menu → My places → OSM edits</string>
+    <string name="reset_plugin_to_default">Reset plugin settings to defaults</string>
     <string name="multimedia_rec_split_title">Recorder split</string>
     <string name="multimedia_use_system_camera">Use system app</string>
     <string name="multimedia_photo_play_sound">Camera shutter sound</string>
-    <string name="osm_authorization_success">Authorization is successful</string>
+    <string name="osm_authorization_success">Authorized</string>
```
sonora: yes, fine.

OsmAnd/res/values/strings.xml
```diff
@@ -34,14 +34,14 @@
...
     <string name="monitoring_min_accuracy">Minimum accuracy</string>
     <string name="monitoring_min_distance">Minimum displacement</string>
-    <string name="tracks_view_path">Menu — My places — Tracks</string>
-    <string name="multimedia_notes_view_path">Menu — My places — Notes</string>
-    <string name="osm_edits_view_path">Menu — My places — OSM Edits</string>
-    <string name="reset_plugin_to_default">Reset plugin settings to default</string>
+    <string name="tracks_view_path">Menu → My places → Tracks</string>
+    <string name="multimedia_notes_view_path">Menu → My places → Notes</string>
+    <string name="osm_edits_view_path">Menu → My places → OSM edits</string>
+    <string name="reset_plugin_to_default">Reset plugin settings to defaults</string>
```
sonora: I welcome implementing a convention how we state menu paths. Arrows one possible way. Maybe open an issue to find a consensus, and then we can implement in one change to harmonize overall?

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_accuracy_descr_remark">If GPS is only on right before a recording, first point accuracy may decrease, so the OsmAnd strategy may be to wait a second or so before recording a point (or take the median of 3 consecutive points, etc.) in the future.</string>
+    <string name="monitoring_min_distance_descr">This filter avoids duplicate point recording for insufficient actual motion, for a nicer spatial appearance of tracks not post-processed later.</string>
+    <string name="monitoring_min_distance_descr_side_effect">Periods at rest are not recorded, or one point each. Small (real world) movement (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less info for post-processing, and has worse stats by filtering out superfluous points when recording, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
+    <string name="monitoring_min_distance_descr_recommendation">A setting of 5 meters may work well if you don\'t require any finer details, and don\'t want accuracy in determined stops.</string>
```
sonora: Honestly, I prefer the original description. It is less technical, but precisely describes what will happen and the side effects.

Also: There are logical weaknesses in your suggestion, like "This is a low-speed cut-off filter smoothing recorded tracks discarded track points..." is not accurate. Yes, the track smoothing is the result, but what will actually happen is simply that no points are being recorded below the minimum speed, and that's what the original description state well enough. Why change it`?

(I think there is also a word missing in line 18.)

OsmAnd/res/values/strings.xml
comradekingu: My problem there is I somehow think it should be "show or hide hillshade on the map", without the s. Hills only cast one shade. Hillshades is otherwise the plural of many hills doing it, so that is fine.
As it is a layer, "throw/s hillshade on the map" just sounds functionally like what it does to me.
However, as an explanation, saying the same word twice doesn't do anything, so I thought this to be the better compromise.

OsmAnd/res/values/strings.xml
comradekingu: This is what was used previously in other strings, leaving it now consistent. If so they could all be changed and one would only have to search for the arrow.

OsmAnd/res/values/strings.xml
comradekingu: It has me confused. In my mind these are the options.

**Logging minimum speed**: Starts with something that could be felling trees. Means something is logging the minimum speed, as an action.
**Minimum logging speed**: The slowest speed that is logged (and I thought this to be the intention)
**Logging speed minimum**: The slowest speed logging can be done at, as a convoluted way of saying option 2.

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_distance_descr">This filter avoids duplicate points being recorded where too little actual motion may have occurred, makes a nicer spatial appearance of tracks not post-processed later.</string>
-    <string name="monitoring_min_distance_descr_side_effect">Side effects: Periods at rest are not recorded at all or by just one point each. Small (real world) movements (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less information for post-processing, and has worse stats by filtering out obviously redundant points at recording time, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
-    <string name="monitoring_min_distance_descr_recommendation">Recommendation: A setting of 5 meters may work well for you if you do not require to capture details finer than that, and do not want to explicitly capture data while at rest.</string>
+    <string name="monitoring_min_speed_descr">This is a low-speed cut-off filter smoothing recorded tracks discarded track points below a certain speed.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_speed_descr">A low-speed cut-off filter smoothing recorded tracks by discarding track points below a certain speed.</string>
```

OsmAnd/res/values/strings.xml
comradekingu: Fixed 18. I went for shorter, less technical and complex. It really is a pain trying to understand and translate the originals with those compounding problems. What happens is the worst strings end up untranslated (or erroneous like my attempt here), which is a disaster for users that rely on translations.

"Discarded" is what signifies no points recording below threshold. The rewritten attempt is very much clearer following your suggestion.

For someone not well versed in GPS, telling what happens, and then what it does to do that is a lot more clear than the technical info up front.

At first I thought code annotation had been imported into the strings by accident.

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_distance_descr_side_effect">Side effects: Periods at rest are not recorded at all or by just one point each. Small (real world) movements (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less information for post-processing, and has worse stats by filtering out obviously redundant points at recording time, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
-    <string name="monitoring_min_distance_descr_recommendation">Recommendation: A setting of 5 meters may work well for you if you do not require to capture details finer than that, and do not want to explicitly capture data while at rest.</string>
+    <string name="monitoring_min_speed_descr">A low-speed cut-off filter smoothing recorded tracks by discarding track points below a certain speed.</string>
+    <string name="monitoring_min_speed_descr_side_effect">Whole sections below minimum speed will be discarded, like the steep hill you pushed your bike up. This includes stationary periods, like breaks. Total length, time in motion, and your average speed can still be determined.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_speed_descr_side_effect">Sections (and breaks) below minimum speed are not recorded, like the steep hill you pushed your bike up. Total length, time in motion, and average speed can still be determined.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_distance_descr_recommendation">Recommendation: A setting of 5 meters may work well for you if you do not require to capture details finer than that, and do not want to explicitly capture data while at rest.</string>
+    <string name="monitoring_min_speed_descr">A low-speed cut-off filter smoothing recorded tracks by discarding track points below a certain speed.</string>
+    <string name="monitoring_min_speed_descr_side_effect">Sections (and breaks) below minimum speed are not recorded, like the steep hill you pushed your bike up. Total length, time in motion, and average speed can still be determined.</string>
+    <string name="monitoring_min_speed_descr_recommendation">Try using motion detection via the "Minimum logging displacement" filter (B) first. It may produce better results, and you will lose less data. If your tracks remain noisy at low speeds, try non-zero values here. Please note, some measurements may not report any speed value at all (some network-based methods), in which case you would not record anything.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_speed_descr_recommendation">Try increasing the value required to detect motion in the \"Minimum logging displacement\" filter (B) first. This could end up recording nothing with some network-based methods, because there is no reported speed value.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_speed_descr">A low-speed cut-off filter smoothing recorded tracks by discarding track points below a certain speed.</string>
+    <string name="monitoring_min_speed_descr_side_effect">Sections (and breaks) below minimum speed are not recorded, like the steep hill you pushed your bike up. Total length, time in motion, and average speed can still be determined.</string>
+    <string name="monitoring_min_speed_descr_recommendation">Try increasing the value required to detect motion in the \"Minimum logging displacement\" filter (B) first. This could end up recording nothing with some network-based methods, because there is no reported speed value.</string>
+    <string name="monitoring_min_speed_descr_remark">"Speed > 0" check: Most GPS chipsets only report a speed if determining you are in motion. Hence using the "> 0" setting in this filter in a sense uses the motion detection of the GPS chipset. Even if not filtered in OsmAnd at recording time, it still uses this feature GPX analysis to determine the \"Distance corrected\", i.e. the value displayed in that field is the "distance recorded while in motion".</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_speed_descr_remark">\"Speed > 0\" check: Most GPS chipsets only report a speed when registering motion. Filtering \"> 0\" relies on reported GPS chipset speed. Even if not recorded into the GPX track, the filtering method is used to analyze it in \"Distance corrected\" to get "distance recorded while in motion".</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_speed_descr">A low-speed cut-off filter smoothing recorded tracks by discarding track points below a certain speed.</string>
+    <string name="monitoring_min_speed_descr_side_effect">Sections (and breaks) below minimum speed are not recorded, like the steep hill you pushed your bike up. Total length, time in motion, and average speed can still be determined.</string>
+    <string name="monitoring_min_speed_descr_recommendation">Try increasing the value required to detect motion in the \"Minimum logging displacement\" filter (B) first. This could end up recording nothing with some network-based methods, because there is no reported speed value.</string>
+    <string name="monitoring_min_speed_descr_remark">\"Speed > 0\" check: Most GPS chipsets only report a speed when registering motion. Filtering \"> 0\" relies on reported GPS chipset speed. Even if not recorded into the GPX track, the filtering method is used to analyze it in \"Distance corrected\" to get "distance recorded while in motion".</string>
```
comradekingu: I am not sure this is correct, as I forgot what "Distance corrected" is.

comradekingu:
```suggestion
    <string name="monitoring_min_speed_descr_remark">\"Speed > 0\" check: Most GPS chipsets only report a speed when registering motion. Filtering \"> 0\" relies on reported GPS chipset speed. Even if not recorded into the GPX track, the filtering method is used to analyze it in \"Distance corrected\" to get \"distance recorded while in motion\".</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_accuracy_descr_side_effect">As a result of filtering by accuracy, points may not be recorded below bridges, under trees, between high buildings, or with certain weather conditions.</string>
+    <string name="monitoring_min_accuracy_descr_recommendation">It is hard to predict what will be recorded, it may be best to turn this filter off.</string>
+    <string name="monitoring_min_accuracy_descr_remark">If GPS is only on right before a recording, first point accuracy may decrease, so the OsmAnd strategy may be to wait a second or so before recording a point (or take the median of 3 consecutive points, etc.) in the future.</string>
+    <string name="monitoring_min_distance_descr">This filter avoids duplicate point recording for insufficient actual motion, for a nicer spatial appearance of tracks not post-processed later.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_distance_descr">This lossy filter avoids duplicate point recording for insufficient actual motion, to get nice tracks without having to do post-processing.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_accuracy_descr_side_effect">Side effect: As a result of filtering by accuracy, points may be entirely missing for e.g. below bridges, under trees, between high buildings, or with certain weather conditions.</string>
-    <string name="monitoring_min_accuracy_descr_recommendation">Recommendation: It is hard to predict what will be recorded and what not, it may be best to turn this filter off.</string>
-    <string name="monitoring_min_accuracy_descr_remark">Remark: If GPS had been off immediately before a recording, the first point measured may have a decreased accuracy,so in our code we may want to wait a second or so before recording a point (or record the best of 3 consecutive points, etc.), but this is not yet implemented.</string>
-    <string name="monitoring_min_distance_descr">This filter avoids duplicate points being recorded where too little actual motion may have occurred, makes a nicer spatial appearance of tracks not post-processed later.</string>
```
comradekingu: Did "spatial appearance" do anything meaningful here?

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_accuracy_descr_recommendation">It is hard to predict what will be recorded, it may be best to turn this filter off.</string>
+    <string name="monitoring_min_accuracy_descr_remark">If GPS is only on right before a recording, first point accuracy may decrease, so the OsmAnd strategy may be to wait a second or so before recording a point (or take the median of 3 consecutive points, etc.) in the future.</string>
+    <string name="monitoring_min_distance_descr">This lossy filter avoids duplicate point recording for insufficient actual motion, to get nice tracks without having to do post-processing.</string>
+    <string name="monitoring_min_distance_descr_side_effect">Periods at rest are not recorded, or one point each. Small (real world) movement (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less info for post-processing, and has worse stats by filtering out superfluous points when recording, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_distance_descr_side_effect">Small sideways motion and periods at rest are either not recorded, or by one point each. The resulting GPX file has fewer tracking points for post-processing, loosing some accuracy by filtering out superfluous ones, and potentially relying on ones caused by bad reception or GPS chipset effects.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_speed_descr_recommendation">Try increasing the value required to detect motion in the \"Minimum logging displacement\" filter (B) first. This could end up recording nothing with some network-based methods, because there is no reported speed value.</string>
+    <string name="monitoring_min_speed_descr_remark">\"Speed > 0\" check: Most GPS chipsets only report a speed when registering motion. Filtering \"> 0\" relies on reported GPS chipset speed. Even if not recorded into the GPX track, the filtering method is used to analyze it in \"Distance corrected\" to get \"distance recorded while in motion\".</string>
+    <string name="monitoring_min_accuracy_descr">This will record only points measured with a minimum "accuracy" indication (in meters/feet, as reported by Android for your chipset). \"Accuracy\" refers to the scatter of repeated measurements, and is not directly related to precision, which defines how close your measurements are to your \"true\" position.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_accuracy_descr">Records only points within a minimum \"accuracy\" indication of how scattered repeated measurements are (in m / ft), as reported by Android for your chipset). From that one can work out precision (how close they are to your \"true\" position).</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_speed_descr_recommendation">Try increasing the value required to detect motion in the \"Minimum logging displacement\" filter (B) first. This could end up recording nothing with some network-based methods, because there is no reported speed value.</string>
+    <string name="monitoring_min_speed_descr_remark">\"Speed > 0\" check: Most GPS chipsets only report a speed when registering motion. Filtering \"> 0\" relies on reported GPS chipset speed. Even if not recorded into the GPX track, the filtering method is used to analyze it in \"Distance corrected\" to get \"distance recorded while in motion\".</string>
+    <string name="monitoring_min_accuracy_descr">Records only points within a minimum \"accuracy\" indication of how scattered repeated measurements are (in m / ft), as reported by Android for your chipset). From that one can work out precision (how close they are to your \"true\" position).</string>
+    <string name="monitoring_min_accuracy_descr_side_effect">As a result of filtering by accuracy, points may not be recorded below bridges, under trees, between high buildings, or with certain weather conditions.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_accuracy_descr_side_effect">Filtering by accuracy may fail to record points under bridges, trees, between high buildings, or in some weather conditions.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_speed_descr_remark">\"Speed > 0\" check: Most GPS chipsets only report a speed when registering motion. Filtering \"> 0\" relies on reported GPS chipset speed. Even if not recorded into the GPX track, the filtering method is used to analyze it in \"Distance corrected\" to get \"distance recorded while in motion\".</string>
+    <string name="monitoring_min_accuracy_descr">Records only points within a minimum \"accuracy\" indication of how scattered repeated measurements are (in m / ft), as reported by Android for your chipset). From that one can work out precision (how close they are to your \"true\" position).</string>
+    <string name="monitoring_min_accuracy_descr_side_effect">Filtering by accuracy may fail to record points under bridges, trees, between high buildings, or in some weather conditions.</string>
+    <string name="monitoring_min_accuracy_descr_recommendation">It is hard to predict what will be recorded, it may be best to turn this filter off.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_accuracy_descr_recommendation">This is a hard cutoff for unreliable data, making good data inevitable, and recording anything more unpredictable.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_accuracy_descr_remark">If GPS is only on right before a recording, first point accuracy may decrease, so the OsmAnd strategy may be to wait a second or so before recording a point (or take the median of 3 consecutive points, etc.) in the future.</string>
+    <string name="monitoring_min_distance_descr">This lossy filter avoids duplicate point recording for insufficient actual motion, to get nice tracks without having to do post-processing.</string>
+    <string name="monitoring_min_distance_descr_side_effect">Small sideways motion and periods at rest are either not recorded, or by one point each. The resulting GPX file has fewer tracking points for post-processing, loosing some accuracy by filtering out superfluous ones, and potentially relying on ones caused by bad reception or GPS chipset effects.</string>
+    <string name="monitoring_min_distance_descr_recommendation">A setting of 5 meters may work well if you don\'t require any finer details, and don\'t want accuracy in determined stops.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_distance_descr_recommendation">5 meters may work well if you don\'t require any finer detail, giving away as much in determining stops.</string>
```

OsmAnd/res/values/strings.xml
comradekingu: The convention is to instruct directly. It is easier to understand because it has less abstraction. I would go "to make use of", but that is more British-English.  "Permission is required" is un/referenced noun uncanny wally, should have been "permission required" or "A permission is required". If having to add "to use this option", might as-well lower all barriers and directly instruct the user in what to do.

OsmAnd/res/values/strings.xml
comradekingu: It is however a bit cold.
Can the already present
<string name="logged_as">Logged in as %1$s</string>
be used?

OsmAnd/res/values/strings.xml
sonora: Well spotted. I think 'explicitly' is obsolete.

OsmAnd/res/values/strings.xml
sonora: Yes, being specific would be best here. Have not checked the context and what is known there. How about 'Please garnt the required permission to use this'? But then, have you checked when this occurs and how the user is even notified which permission it is?

OsmAnd/res/values/strings.xml
sonora: 'Cold' is well stated. ;) Would need to research where this is used, i.e. if logged in is correct, and if we have a variable there to fill % $1s?

OsmAnd/res/values/strings.xml
sonora: We could work with 'threshold', but asking my colleagues here that would lead to lengthy translations in some languages, so I'd like to stick to 'min something'.
How about:
Minimum speed for logging
Minimum accuracy for logging
Minimum displacement for logging

OsmAnd/res/values/strings.xml
sonora: I guess it means to say 'smoother' ;)

OsmAnd/res/values/strings.xml
sonora: 'Shades' (as in shades of color, or Shades of Grey') should be perfectly fine, I see no issues.

OsmAnd/res/values/strings.xml
sonora: Apologies, I had lost track of that issue, and it looks like noone else commentrd there. Will review at some point and then we try to resolve.

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_distance_descr_side_effect">Side effects: Periods at rest are not recorded at all or by just one point each. Small (real world) movements (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less information for post-processing, and has worse stats by filtering out obviously redundant points at recording time, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
-    <string name="monitoring_min_distance_descr_recommendation">Recommendation: A setting of 5 meters may work well for you if you do not require to capture details finer than that, and do not want to explicitly capture data while at rest.</string>
+    <string name="monitoring_min_speed_descr">A low-speed cut-off filter smoothing recorded tracks by discarding track points below a certain speed.</string>
+    <string name="monitoring_min_speed_descr_side_effect">Sections (and breaks) below minimum speed are not recorded, like the steep hill you pushed your bike up. Total length, time in motion, and average speed can still be determined.</string>
```
sonora: No! The original is better. Is explains what this does, why you would use it, and explains the negative side effects. Portions of your statement are misleading, like "Total length, time in motion, and average speed can still be determined.". Of course it can, but it will be wrong. :)

OsmAnd/res/values/strings.xml
comradekingu: I think I know why I don't like it now. "Shades" has lower resolution, effectively explaining dithering, rather than the organic concept of something casting a shade.
Map *shades sounds like the map is casting the shade, rather than the shade being cast onto the map.
They also are not called hillshades everywhere. (I did fix the ones that weren't).

```
<string name="quick_action_hillshade_descr">A button to show or hide map hillshading.</string>
<string name="quick_action_hillshade_show">Show hillshade</string>
<string name="quick_action_hillshade_hide">Hide hillshade</string>
<string name="quick_action_show_hide_hillshade">Show/hide hillshade</string>
<string name="contour_lines_and_hillshade">Contour lines and hillshade</string>
<string name="contour_lines_hillshade_maps">Contour lines and hillshade maps</string>
<string name="hillshade_menu_download_descr">Download the \'Hillshade Overlay\' map to show vertical shading.</string>
<string name="hillshade_purchase_header">Install the \'Contour lines\' plugin to show graded vertical areas.</string>
<string name="hillshade_layer_disabled">Hillshade layer disabled</string>
<string name="osmand_rastermaps_plugin_description">Access many types of online (so called tile or raster) maps, from predefined OSM tiles (like Mapnik) to satellite images, and special purpose layers like weather maps, climate maps, geological maps, hillshade layers, etc.\n\n
    Any of these maps can either be used as the main (base) map to be displayed, or as an overlay or underlay to another base map (like OsmAnd\'s standard offline maps). Certain elements of the OsmAnd vector maps can be hidden via the \'Configure map\' menu to make any underlay map more visible.\n\n
    Download tile maps directly online, or prepared them for offline use (manually copied to OsmAnd\'s data folder) as an SQLite database which can be produced by a variety of 3rd party map preparation tools.</string>
<string name="osmand_srtm_long_description_1000_chars">This plugin provides both a contour line overlay and a (relief) hillshade layer to be displayed on top of OsmAnd\'s standard maps. This functionality will be much appreciated by athletes, hikers, trekkers, and anybody interested in the relief structure of a landscape.\n\n
    The global data (between 70 ° north and 70 ° south) is based on measurements by SRTM (Shuttle Radar Topography Mission) and ASTER (Advanced Spaceborne Thermal Emission and Reflection Radiometer), an imaging instrument onboard Terra, the flagship satellite of NASA\'s Earth Observing System. ASTER is a cooperative effort between NASA, Japan\'s Ministry of Economy, Trade and Industry (METI), and Japan Space Systems (J-spacesystems).
</string>
<string name="srtm_plugin_description">This plugin provides both a contour line overlay and a (relief) hillshade layer to be displayed on top of OsmAnd\'s standard maps. This functionality will be much appreciated by athletes, hikers, trekkers, and anybody interested in the relief structure of a landscape. (Please note that contour line and/or relief data are separate, additional downloads available after activating the plugin.)
		\n\nThe global data (between 70 ° north and 70 ° south) is based on measurements by SRTM (Shuttle Radar Topography Mission) and ASTER (Advanced Spaceborne Thermal Emission and Reflection Radiometer), an imaging instrument onboard Terra, the flagship satellite of NASA\'s Earth Observing System. ASTER is a cooperative effort between NASA, Japan\'s Ministry of Economy, Trade and Industry (METI), and Japan Space Systems (J-spacesystems).
</string>
<string name="layer_hillshade">Hillshade layer</string>
<string name="download_hillshade_maps">Hillshade</string>
<string name="osmand_extended_description_part5">
        Cycling\n
        • Find cycle paths on the map\n
        • GPS navigation in cycling mode builds your route using cycle paths\n
        • See your speed and altitude\n
        • GPX recording option enables you to record your trip and share it\n
        • Via an additional plugin you can enable contour lines and hillshading
</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_distance_descr">This filter avoids duplicate points being recorded where too little actual motion may have occurred, makes a nicer spatial appearance of tracks not post-processed later.</string>
-    <string name="monitoring_min_distance_descr_side_effect">Side effects: Periods at rest are not recorded at all or by just one point each. Small (real world) movements (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less information for post-processing, and has worse stats by filtering out obviously redundant points at recording time, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
-    <string name="monitoring_min_distance_descr_recommendation">Recommendation: A setting of 5 meters may work well for you if you do not require to capture details finer than that, and do not want to explicitly capture data while at rest.</string>
+    <string name="monitoring_min_speed_descr">A low-speed cut-off filter smoothing recorded tracks by discarding track points below a certain speed.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_speed_descr">Track smoothing by not recording points below a certain speed.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_speed_descr_side_effect">Sections (and breaks) below minimum speed are not recorded, like the steep hill you pushed your bike up. Total length, time in motion, and average speed will not be exact.</string>
+    <string name="monitoring_min_speed_descr_recommendation">Try increasing the value required to detect motion in the \"Minimum logging displacement\" filter (B) first. This could end up recording nothing with some network-based methods, because there is no reported speed value.</string>
+    <string name="monitoring_min_speed_descr_remark">\"Speed > 0\" check: Most GPS chipsets only report a speed when registering motion. Filtering \"> 0\" relies on reported GPS chipset speed. Even if not recorded into the GPX track, the filtering method is used to analyze it in \"Distance corrected\" to get \"distance recorded while in motion\".</string>
+    <string name="monitoring_min_accuracy_descr">Records only points within a minimum \"accuracy\" indication of how scattered repeated measurements are (in m / ft), as reported by Android for your chipset). From that one can work out precision (how close they are to your \"true\" position).</string>
```
comradekingu:
```suggestion
    True position precision as calculated by discarding sequential points scattered outside of a minimum.
```

comradekingu:
```suggestion
    <string name="monitoring_min_accuracy_descr">True position precision as calculated by discarding sequential points scattered outside of a minimum.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_distance_descr">This filter avoids duplicate points being recorded where too little actual motion may have occurred, makes a nicer spatial appearance of tracks not post-processed later.</string>
-    <string name="monitoring_min_distance_descr_side_effect">Side effects: Periods at rest are not recorded at all or by just one point each. Small (real world) movements (e.g. sideways, to mark a possile turnoff on your trip) may be filtered out. Your file contains less information for post-processing, and has worse stats by filtering out obviously redundant points at recording time, while potentially keeping artifacts caused by bad reception or GPS chipset effects.</string>
-    <string name="monitoring_min_distance_descr_recommendation">Recommendation: A setting of 5 meters may work well for you if you do not require to capture details finer than that, and do not want to explicitly capture data while at rest.</string>
+    <string name="monitoring_min_speed_descr">Track smoothing by not recording points below a certain speed.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_speed_descr">Smoothed GPX tracking by not recording points below a certain speed.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_speed_descr">This is a low-speed cut-off filter to not record points below a certain speed. This may make recorded tracks look "smoother" when viewed on the map.</string>
-    <string name="monitoring_min_speed_descr_side_effect">Side effect: Your track will be missing all sections where the minimum speed criterion was not met (e.g. where you push your bike up a steep hill). Also, there will be no information about periods at rest, like breaks. This has effects on any analysis or post-processing, like when trying to determine the total length of your trip, time in motion, or your average speed.</string>
-    <string name="monitoring_min_speed_descr_recommendation">Recommendation: Try using the motion detection via the "logging minimum displacement" filter (B) first, it may produce better results, and you will lose less data. If your tracks remain noisy at low speeds, try non-zero values here. Please note that some measurements may not report any speed value at all (some network-based methods), in which case you would not record anything.</string>
```
comradekingu: 1. Why would (only?) using "minimum logging displacement" be better. "Lose less data" where, in capturing or saving?
2. What is "here"?
3. Some measurements may not report any speed value at all in min speed, or min displacement?

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
-    <string name="monitoring_min_speed_descr_remark">Remark: "speed > 0" check: Most GPS chipsets report a speed value only if the algorithm determines you are in motion, and none if you are not. Hence using the "> 0" setting in this filter in a sense uses the motion detection of the GPS chipset. But even if not filtered here at recording time, we still use this feature in our GPX analysis to determine the "Distance corrected", i.e. the value displayed in that field is the "distance recorded while in motion".</string>
```
comradekingu: I can't find "Distance corrected" or "Distance recorded while in motion" anywhere…

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
...
+    <string name="monitoring_min_speed_descr_remark">\"Speed > 0\" check: Most GPS chipsets only report a speed when registering motion. Filtering \"> 0\" relies on reported GPS chipset speed. Even if not recorded into the GPX track, the filtering method is used to analyze it in \"Distance corrected\" to get \"distance recorded while in motion\".</string>
+    <string name="monitoring_min_accuracy_descr">True position precision as calculated by discarding sequential points scattered outside of a minimum.</string>
+    <string name="monitoring_min_accuracy_descr_side_effect">Filtering by accuracy may fail to record points under bridges, trees, between high buildings, or in some weather conditions.</string>
+    <string name="monitoring_min_accuracy_descr_recommendation">This is a hard cutoff for unreliable data, making good data inevitable, and recording anything more unpredictable.</string>
+    <string name="monitoring_min_accuracy_descr_remark">If GPS is only on right before a recording, first point accuracy may decrease, so the OsmAnd strategy may be to wait a second or so before recording a point (or take the median of 3 consecutive points, etc.) in the future.</string>
```
comradekingu:
```suggestion
    <string name="monitoring_min_accuracy_descr_remark">The first recorded point is less accurate if GPS is only turned on right before. A one second delay or median of 3 consecutive points, etc. may be used in the future.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -11,21 +11,21 @@
 	Thx - Hardy
 
 -->
-    <string name="permission_is_required">Permission is required to use this option.</string>
-    <string name="logcat_buffer_descr">Check and share detailed logs of the application</string>
+    <string name="permission_is_required">Please grant the required permission to use this option.</string>
+    <string name="logcat_buffer_descr">Check and share detailed logs of the app</string>
```
comradekingu:
```suggestion
    <string name="logcat_buffer_descr">Check and share detailed logs of the app.</string>
```

comradekingu: Spelling: Missing dot for description

comradekingu:
@sonora Is this good now, or do you want me to move the "monitoring" strings to a new PR?

PR #8473, opened 2020-02-15
Fix 7763 selected gpx status

Dima-1:


OsmAnd/src/net/osmand/plus/GpxSelectionHelper.java
```diff
@@ -25,6 +25,7 @@
 import net.osmand.plus.helpers.GpxUiHelper;
 import net.osmand.plus.helpers.GpxUiHelper.GPXDataSetAxisType;
 import net.osmand.plus.helpers.GpxUiHelper.GPXDataSetType;
+import net.osmand.plus.myplaces.AvailableGPXFragment;
```
vshcherb: This import is incorrect

PR #8477, opened 2020-02-16
Fix #8476 Missing some street names in the detailed route

### LINKED ISSUE #8476: Missing some street names in the detailed route
# Title

Missing some street names in the detailed route. The empty street names are also displayed in the top of the screen during navigation.

# Problem

The detailed route doesn’t display street names that follow roundabout exits. See turns at **1.59 km** and at **2.11 km** in image below.
Route is attached as file [2020-02-16.gpx.zip](https://github.com/osmandapp/Osmand/files/4210528/2020-02-16.gpx.zip)

2020-02-16 Problem

# Reason

Let’s have a look at the roundabout at **1.59 km** from start point. In the below screenshot we can see the roundabout exits to one way road.

2020-02-16 Roundabout

I opened **OpenStreetMap** site (https://www.openstreetmap.org/edit#map=18/55.74180/37.86592) and selected roundabout exit to view its attributes. The selected object (https://www.openstreetmap.org/way/24686799#map=19/55.74197/37.86569) has no name (name is indicated using red box). 

2020-02-16 Openstreetmap

# Resolution

Street name is evaluated in net.osmand.plus.routing.RouteCalculationResult#convertVectorResult, by checking the segment that follows a turn (in our case the roundabout exit). 

Why not to try to detect street name by checking several segments after turn? This approach is implemented in my fork of repo. I'll provide link after publishing changes. 

After applying the new approach I captured screenshot with detailed route once again.

2020-02-16 Resolution

alrega:
When evaluating direction from segments, algorithm attempts to detect name of the street that follows turn by examining several segments.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
+++ b/OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
@@ -33,6 +33,15 @@ import static net.osmand.binary.RouteDataObject.HEIGHT_UNDEFINED;
 public class RouteCalculationResult {
 	private final static Log log = PlatformUtil.getLog(RouteCalculationResult.class);
 
+	/**
+	 * When generating list of directions from segments, algorithm attempts to evaluate street names
+	 * that follow the route after turn. It's quite often segments follow roundabout exit refer to
+	 * unnamed road, so algorithm scans several segments following the exit to find out the street
+	 * name.
+	 * Maximum distance to scan in meter is specified by this parameter.
+	 */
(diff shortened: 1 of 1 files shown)


vshcherb:
This could be erroneous in some cases, probably OSM has missing street names only for short distance segments like 50-100 m and algorithms doesn't have checks for distance, so it is better to check the data cases and limit algorithm as much as possible

alrega:
I'll add distance checking on weekend.

alrega:
@vshcherb , I added distance check to the algorithm. Limit is set to 150 meters.

PR #8496, opened 2020-02-18
Fix #7763 selected gpx file status

### LINKED ISSUE #7763: Track screen's folder tick marks do not refllect selected gpx file status
How to reproduce:
1. Orgamize some of your gpx files in folders (under OsmAnd's 'tracks' folder)
2. On the My places / tracks screen, then tap the 'Show on map' button near the bottom center of the screen, this will produce selection tick marks in front of all gpx files and folders.
3. Select the tick mark in front of one of the subfoldets. This will correctly select and display all tracks contained.
4. Now leave the screen and later return to it: The tickmark of the selected folde remains empty, although all contained files are corrently selected.

 The folder tick mark must correctly dhow the display status of selected gpx tracks, preferably even indicate by perhaps gray vs. lack checkmark color if all or just some contained files are selected to be displayed.visible.

Dima-1:


OsmAnd/src/net/osmand/plus/GpxSelectionHelper.java
```diff
@@ -689,6 +727,15 @@ private void syncGpxWithMarkers(GPXFile gpxFile) {
 		private boolean joinSegments;
 		private boolean showCurrentTrack;
 		private boolean splitProcessed = false;
+		private String loadedName = "";
```
vshcherb: This field is not needed, GPXFile is enough

OsmAnd/src/net/osmand/plus/GpxSelectionHelper.java
```diff
@@ -440,10 +440,20 @@ private static String formatSplitName(double metricEnd, GpxDisplayGroup group, O
 		}
 	}
 
+	private SelectedGpxFile getSelectedFileByLoadedFileName(String path) {
+		List<SelectedGpxFile> newList = new ArrayList<>(selectedGPXFiles);
```
vshcherb: This method looks strange

PR #8511, opened 2020-02-19
Fix #7763 gpx track selection

### LINKED ISSUE #7763: Track screen's folder tick marks do not refllect selected gpx file status

Dima-1:


OsmAnd/src/net/osmand/plus/myplaces/AvailableGPXFragment.java
```diff
@@ -605,19 +625,22 @@ public boolean onPrepareActionMode(ActionMode mode, Menu menu) {
 
 			@Override
 			public boolean onActionItemClicked(ActionMode mode, MenuItem item) {
-				runSelection(false);
+				HashMap<String, Boolean> selectedItemsFileNames = new HashMap<>();
+				originalSelectedItems.addAll(selectedItems);
+				for (GpxInfo gpxInfo : originalSelectedItems) {
+					if (gpxInfo.currentlyRecordingTrack) {
+						selectedItemsFileNames.put(CURRENT_TRACK, true);
```
Chumva: Why always true for the current track?

OsmAnd/src/net/osmand/plus/GpxSelectionHelper.java
```diff
@@ -54,7 +56,12 @@
 	private Map<GPXFile, Long> selectedGpxFilesBackUp = new java.util.HashMap<>();
 	private SavingTrackHelper savingTrackHelper;
 	private final static Log LOG = PlatformUtil.getLog(GpxSelectionHelper.class);
+	private SelectGpxTask selectGpxTask;
+	private SelectGpxTaskListener gpxTaskListener;
 
+	public void setGpxTaskListener(SelectGpxTaskListener gpxTaskListener) {
```
vshcherb: No field is needed ! Fields should be created carefully and when it is not needed better to use parameter for method

OsmAnd/src/net/osmand/plus/GpxSelectionHelper.java
```diff
@@ -698,6 +715,10 @@ public void setGpxFile(GPXFile gpxFile, OsmandApplication app) {
 			processPoints(app);
 		}
 
+		public boolean isLoaded() {
+			return gpxFile.modifiedTime != 0;
```
vshcherb: better to use -1 if it is going to be used. Sometimes 0 is system time

OsmAnd/src/net/osmand/plus/GpxSelectionHelper.java
```diff
@@ -54,7 +56,12 @@
 	private Map<GPXFile, Long> selectedGpxFilesBackUp = new java.util.HashMap<>();
 	private SavingTrackHelper savingTrackHelper;
 	private final static Log LOG = PlatformUtil.getLog(GpxSelectionHelper.class);
+	private SelectGpxTask selectGpxTask;
```
vshcherb: Field is not used anywhere

Chumva: This field is used for checking running SelectGpxTask in runSelection

OsmAnd/src/net/osmand/plus/GpxSelectionHelper.java
Chumva: I know, but with this listener, we can get updates for necessary screens if update working in the background and it doesn`t update new screen.

vshcherb:
Fix #7763

PR #8669, opened 2020-03-13
Change Settings UI -> Distance to recalculate

nazar-kutz:


OsmAnd/src/net/osmand/plus/OsmandSettings.java
```diff
@@ -3467,6 +3467,7 @@ protected boolean setValue(Object prefs, String val) {
 	}
 
 	public final CommonPreference<Float> ROUTE_RECALCULATION_DISTANCE = new FloatPreference("routing_recalc_distance", 0.f).makeProfile();
+	public final CommonPreference<Boolean> USE_ROUTE_RECALCULATION_DISTANCE = new BooleanPreference("use_routing_recalc_distance", false).makeProfile();
```
vshcherb: USE_ROUTE_RECALCULATION_DISTANCE it should be reverted, new setting is not needed same as changes to RoutingHelper it only possibly breaks existing functionality.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
```diff
@@ -442,10 +442,12 @@ private Location setCurrentLocation(Location currentLocation, boolean returnUpda
 				if (allowableDeviation == 0) {
 					allowableDeviation = getDefaultAllowedDeviation(settings, route.getAppMode(), posTolerance);
 				}
+				boolean isRecalculationActivated = allowableDeviation > 0 &&
```
vshcherb: USE_ROUTE_RECALCULATION_DISTANCE it should be reverted, new setting is not needed same as changes to RoutingHelper it only possibly breaks existing functionality.

sonora:
What are we doing here? We already have a profile-dependent setting for this.

Is this meant to be a bug fix for something?

Rather than introducing yet another user-accessible option, as if this was a user decision and  'matter of taste', I think our task is to provide an algorithm to determine the value in the background, that's what makes a dumb app a progressive app :)

That 'algorithm' can simply be a constant each for each profile. Or it could be something more elaborate, but then we would have to think about which scenarios or deficiencies we are trying to improve, and which parameters an information we have for this.

nazar-kutz:
In this case, this is just an UI improvement.
Additional variable was added to separate the setting state (on / off) and its numerical value.

Issue #8704, opened 2020-03-23
Daily folders option for recoded gpx tracks vanished


jensMF:
In the version that is distributed on google play store at the moment, the option to save recorded tracks in daily folders is gone. In master the code is just commented.
Does it only needs to be uncommented?

vshcherb:
We've found that's it is quite intrusive and it will create extra problems such fragmentation. In case of 1000 folders it will be a bad solution, so we will continue to work on GPX collection in Q3-Q4 and meanwhile you can use previous or custom builds

PR #8705, opened 2020-03-23
fix for #8704

jensMF:
Well there was the DAILY_DIRECTORY constant missing in `src/net/osmand/plus/monitoring/SettingsMonitoringActivity.java`


vshcherb:
https://github.com/osmandapp/Osmand/issues/8704

Issue #8706, opened 2020-03-23
Audio Delay in all modes.


ilyaguy:
I've installed OsmAnd on android car radio, and there's issue:
Car radio cut beginning of audio advice.
OsmAnd has option to change delay before start output, but it works only in "Phone Call Channel" and it seems that my car radio doesn't use this channel at all.
Is it possible to allow this delay works in every audio cahnnel?

sonora:
Can you please describe your setup, both in terms of hardware and the configuration/settings on all systems?

I do not currently understand the request, because from all setups I have come across, for the media (i.e. music) and notification (I.e. ringtone) channels to be auditable on the car stereo at all, the stereo needs to be set to Aux/bluetooth input, meaning you would **always** (exclusively and permanently) hear only the external device, and it would be an 'always on' situation (not the interruption situation I implement via the phone call channel). Hence it is unclear why there should be a cut-off at all.

ilyaguy:
2DIN Android car radio CT-0011 (Android 6.0, touch screen 1024x600).
Installed latest OsmAnd+ from playmarket on it. So, there's only one device.
I've configure voice guidance. It works mostly fine, except one thing: osmand or device cut about 0.5 seconds beginning of phrase. 

I've opened development plugin and go to "Test Voice" screen. When I tapped on phrase there's no first word for all of them. Or, if word is long enough, I hear end of it: "...га" instead of "Увага", etc.

In development plugin there's option to change delay before audio, but it works only if "Phone Call Channel" selected to use voice notifications. But if I selected it, there's no sound at all. (This is a problem of device, I think). 

My request is about enabling this delay for all audio output mode: "media", "notifications" and "phone call".

I can shoot and upload video if it could help.

sonora:
Ok, I think I understand the issue now, let me think about it as time allows. I have to find a solution not complicating the world for the 99% of users not needing this. If I expanded the timeout to just the notification channel, would that work for you?

ilyaguy:
I think it will help. Because when I tap two or more times on the same phrase I heard second and following plays completely and without any cuts.

sonora:
@vshcherb 
[...]

PR #8727, opened 2020-03-29
Voice prompt delay for all audio manager streams

sonora:


sonora:
See https://github.com/osmandapp/Osmand/issues/8706#issuecomment-605637809 for background.

PR #8787, opened 2020-04-12
Fix #6838 Announce the change of street names

### LINKED ISSUE #6838: Feature request: announce change of street names
When one drives on a straight road, OsmAnd does not announce the change of streets. Example: 5.83111 / - 55.18509.
When the direction crosses the Kernkampweg, the street name changes from Gravenberchstraat to Johannes Mungrastraat. Although it is shown in the display, OsmAnd does not announce it.

alrega:
One road may be result of several streets changing one into another. In that case OsmAnd didn't indicate that street name would change, because thare was no turn left/right in the route.

Commit adds the new feature. When two segments of stright part route have different street names, a new 'stright' direction with the new name will be inserted into the list of directions.

This is not applied to roads that have international/national refs, like 'E 101'

### DIFF
--- a/OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
+++ b/OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
@@ -362,7 +362,39 @@ public class RouteCalculationResult {
 					i--;
 				}
 			}
+
+			String streetName = null;
 			TurnType turn = s.getTurnType();
+
+			// Try to detect the case when a street changes is name to another (typically after road crossing)
+			// Then add RouteDirectionInfo of TurnType.straight()
+			if (turn == null && routeInd + 1 < list.size()) {
+				String curStreetName = s.getObject().getName(ctx.getSettings().MAP_PREFERRED_LOCALE.get(),
+						ctx.getSettings().MAP_TRANSLITERATE_NAMES.get());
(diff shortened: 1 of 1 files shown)


alrega:
Hello @vshcherb, I can't assign a reviewer to this PR. Could you help?
Thanks in advance.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
```diff
@@ -362,8 +362,40 @@ public double getRouteVisibleAngle() {
...
+			String streetName = null;
 			TurnType turn = s.getTurnType();
 
+			// Try to detect the case when a street changes is name to another (typically after road crossing)
+			// Then add RouteDirectionInfo of TurnType.straight()
+			if (turn == null && routeInd + 1 < list.size()) {
+				String curStreetName = s.getObject().getName(ctx.getSettings().MAP_PREFERRED_LOCALE.get(),
+						ctx.getSettings().MAP_TRANSLITERATE_NAMES.get());
+				String curIntRef = s.getObject().getValue("int_ref");
+				String curRoadRef2 = s.getObject().getValue("road_ref_2");
```
vshcherb, line 375: road_ref_2, int_ref looks strange cause they are specific to some specific map area and it doesn't look as a general solution

PR #8830, opened 2020-04-22
Discount badge

Dima-1:


OsmAnd/res/values/strings.xml
```diff
@@ -716,7 +718,7 @@
     <string name="osm_live_payment_annual_title">Annually</string>
     <string name="osm_live_payment_month_cost_descr">%1$s / month</string>
     <string name="osm_live_payment_month_cost_descr_ex">%1$.2f %2$s / month</string>
-    <string name="osm_live_payment_discount_descr">Save %1$s.</string>
+    <string name="osm_live_payment_discount_descr">Save %1$s</string>
```
max-klaus, line 721: Needs to be changed in all translations

OsmAnd/src/net/osmand/plus/chooseplan/ChoosePlanDialogFragment.java
```diff
@@ -352,18 +355,21 @@ private void setupOsmLiveCardButtons(boolean progress) {
...
 			for (final InAppSubscription s : visibleSubscriptions) {
-				if (s.isPurchased() || s.getIntroductoryInfo() != null) {
-					anyPurchasedOrIntroducted = true;
+				if (s.isPurchased()) {
+					anyPurchased = true;
 					break;
 				}
 			}
+			InAppSubscription subscriptionMaxDiscount = purchaseHelper.getLiveUpdates()
+					.getSubscriptionWithMaxDiscount(purchaseHelper.getMonthlyLiveUpdates());
```
max-klaus, line 366: Remove getSubscriptionWithMaxDiscount. First - it is wrong since it iterates through all and not visible. Second - it is redundant, use for loop above.

Issue #8835, opened 2020-04-24
Rendering issue (e.g. in measurement tool) caused by bug in calculatePath()


febeslmeisl:
While developing a new feature in my local branch I think I have spotted an error in existing master branch code, that is also causing a (minor) rendering issue in current measurement tool functionality, so reporting here:

Steps to reproduce (checked on official OsmAnd~ 3.6.3):
- Put your device in  Landscape orientation.
- Set map to north is on top orientation
- Start measurement tool and add a start way point on your current location
- Add a new way point far in north
- Go back to start position and zoom in that the northern way point is **far** out of screen
- Rotate landscape that east is around top
- You see the dashed line of the measurement tool starting in "the middle of nowhere" going to center. Please also refer to [this Screenshot below](https://user-images.githubusercontent.com/34205180/80488186-3e195480-895e-11ea-8ccd-cd19f7cf63d6.png).

![Screenshot_20200424-153037](https://user-images.githubusercontent.com/34205180/80488186-3e195480-895e-11ea-8ccd-cd19f7cf63d6.png)

My root cause analysis so far:
- Method `calculatePath(RotatedTileBox tb, List<Float> xs, List<Float> ys, List<GeometryWayStyle> styles, List<Pair<Path, GeometryWayStyle>> paths)` of net.osmand.plus.views.OsmandMapLayer causing rendering issues if map in not in "north is on top" orientation. Because the view rect that is checked for paths going in or out is not rotated along with map. 
- Also `calculateIntersection` of net.osmand.util.MapAlgorithms in my opinion has a special case not handled correctly also contributing to this issue: The cases where `(outx == rightX || outx == leftX || outy == topY || outy == bottomY)` here currently the return value is comment out (maybe for good reason?).

What to do:
I'm very new to OsmAnd and not so familiar with its architecture, therefore I'm not sure about best/most performant way to fix it. But here is what is working for me (with minimal changes in existing code) as a orientation:

- For calculatePath in `net.osmand.plus.views.OsmandMapLayer`: rotating all tx,ty points before calculating view rect intersection by RotatedTileBox's angle and rotating back before drawing (with `mtx` and `inv_mtx` Matrix):
```java
		boolean segmentStarted = false;

		mtx.setSinCos((float)tb.getRotateSin(), (float)tb.getRotateCos(), tb.getCenterPixelX(), tb.getCenterPixelY());
[...]

PR #8892, opened 2020-05-05
Fix issue 8835

febeslmeisl:
As discussed in #8835 a PR for easier review


vshcherb:
The issue is confirmed though it is unlikely will be fixed in that way. Most likely we're going to change it similar to RouteLayer, where canvas is rotated before drawing / calculatePath

vshcherb:
Was fixed differently

PR #8945, opened 2020-05-13
Fix cs/de/es/fr/it/pl/sk translations of pause/resume

eumiro:
Fix a few translations of modified strings `gpx_monitoring_start` and `gpx_monitoring_stop` changed in 
https://github.com/osmandapp/Osmand/commit/75b8b0de3a69e429f850aa0e60be17192c909eba


sonora:
Thanks! It would be good if we could have an equally quick fix also for string `shared_string_resume`, which I had to introduce, because we are nearing a release and I would hate to have it without that string in too many languages. Need not be a pull request, could also be done via https://hosted.weblate.org/projects/osmand/.

PR #8967, opened 2020-05-14
Fix default color favorites

Dima-1:


OsmAnd/src/net/osmand/plus/AppInitializer.java
```diff
@@ -228,8 +228,20 @@ public void onFinish(AppInitializer init) {
 				app.getSettings().migratePreferences();
 				startPrefs.edit().putInt(VERSION_INSTALLED_NUMBER, VERSION_3_6).commit();
 			}
-			if (prevAppVersion < VERSION_3_7) {
+			if (prevAppVersion < VERSION_3_7 || Version.getAppVersion(app).equals("3.7.1")) {
```
vshcherb, line 231: Looks bad
Why it is not simply prevAppVersion < VERSION_3_7_1

vshcherb: // Make 4 digits version  from now
	public static final int VERSION_3_7_1 = 3710;

vshcherb: Move all migration code to Another class AppVersionUpgradeOnInit

OsmAnd/src/net/osmand/plus/FavouritesDbHelper.java
```diff
@@ -183,6 +183,28 @@ public void run() {
...
+		for (FavouritePoint fp : cachedFavoritePoints) {
+			if (fp.getColor() == 0xFF000000) {
+				fp.setColor(0);
+			}
+			FavoriteGroup group = getOrCreateGroup(fp, 0);
+			group.points.add(fp);
+		}
+		sortAll();
+		saveCurrentPointsIntoFile();
+		context.runInUIThread(new Runnable() {
```
vshcherb, line 198: Is there a ready method for updating all listeners?

PR #8978, opened 2020-05-15
Route merge

MadWasp79:


OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapTransportReaderAdapter.java
```diff
@@ -43,6 +44,8 @@ private int readInt() throws IOException {
 
 		int stopsFileOffset = 0;
 		int stopsFileLength = 0;
+		int incompleteRoutesOffset = 0;
```
vshcherb, line 47: This is not part of cachedosmandindexes so it won't work correctly with cache

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -378,7 +382,7 @@ public QuadRect getSegmentRect() {
 
 		public List<Way> getGeometry() {
 			List<Way> list = new ArrayList<>();
-			route.mergeForwardWays();
+			route.mergeForwardWays(); //TODO merge ways of all Route parts
```
vshcherb, line 385: ???

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -903,7 +923,229 @@ public TransportRoutingContext(TransportRoutingConfiguration cfg, NativeLibrary
...
+					segs.add(newSeg);
+				}
+			}
+			return segs;
+		}
+		
+		private List<TransportRoute> findIncompleteRouteParts(TransportRoute baseRoute) throws IOException {
+			List<TransportRoute> allRoutes = null;
+			// TODO completely irrelevant always reiteration over all maps (especially not in bbox of the route probabl)
+			for (BinaryMapIndexReader bmir : routeMap.keySet()) {
```
vshcherb, line 1130: TODO

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -903,7 +923,229 @@ public TransportRoutingContext(TransportRoutingConfiguration cfg, NativeLibrary
...
+			}
+			return merged;
+		}
+
+		
+		
+		private LinkedList<List<TransportStop>> parseRoutePartsToSegments(List<TransportRoute> routeParts) {
+			LinkedList<List<TransportStop>> segs = new LinkedList<List<TransportStop>>();
+			// here we assume that missing stops come in pairs <A, B, C, MISSING, MISSING, D, E...>
+			// TODO check generation that are doubles
```
vshcherb, line 1108: TODO

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -785,18 +794,23 @@ public TransportRoutingContext(TransportRoutingConfiguration cfg, NativeLibrary
...
+					
 				for (TransportStop stop : stops) {
+					// skip missing stops
+					if (stop.isMissingStop()) {
+						continue;
+					}
 					long stopId = stop.getId();
 					TransportStop multifileStop = loadedTransportStops.get(stopId);
 					int[] rrs = stop.getReferencesToRoutes();
+					// TODO what is this?
```
vshcherb, line 813: TODO why do we skip missing stop?

vshcherb: TODO why duplicate same code?

MadWasp79: I thought that we need to skip missing stops so that they were not used during route calculation as valid start/finish/change stops. But I could be wrong with the exact place of such skip. Will look into this.

PR #8982, opened 2020-05-15
Fix #8901

### LINKED ISSUE #8901: [Feature request] Intent to open polyline
Lots of [cycle.travel](https://cycle.travel/map) users like to view their planned routes in OsmAnd, but [sometimes find it complex](https://forum.cyclinguk.org/viewtopic.php?f=1&t=137308&sid=e61e1d864bf305304eeca3b0dffe2812). I'd like to make this easier.

Could you consider adding an intent to open a polyline route in OsmAnd, so that websites like cycle.travel could offer an "Open in OsmAnd" button?

This could potentially be implemented by:

- an intent which takes a Google-encoded polyline and parses that similarly to a GPX track;
- or an intent which takes a URL of a GPX track; on receiving the intent, OsmAnd fetches the file from the given URL, and then opens it.

Chumva:


OsmAnd/src/net/osmand/plus/activities/MapActivity.java
```diff
@@ -1820,6 +1826,46 @@ private boolean parseTileSourceIntent() {
 		return false;
 	}
 
+	private boolean parseOpenGpxIntent() {
```
vshcherb, line 1829: Stop bloating MapActivity - move Intent base parsing to separate class

PR #9019, opened 2020-05-20
Fix ConcurrentModificationException

nazar-kutz:


OsmAnd-java/src/main/java/net/osmand/osm/MapPoiTypes.java
```diff
@@ -375,7 +380,7 @@ public void initFromInputStream(InputStream is) {
 							lastCategory.addExcludedPoiAdditionalCategories(parser.getAttributeValue("", "excluded_poi_additional_category").split(","));
 							lastCategoryPoiAdditionalsCategories.removeAll(lastCategory.getExcludedPoiAdditionalCategories());
 						}
-						categories.add(lastCategory);
```
vshcherb, line 378: Why do we copy each iteration?

PR #9054, opened 2020-05-22
Fix tablet sizes

Dima-1:


vshcherb:
1. Delete map_* icons and use only ic_*
2. Make all icons consistent Grey (configurable probably will change to white) in the night mode
3. Convert all icons to vector

PR #9210, opened 2020-06-11
Add vessel's height for Boat navigation

Dima-1:


noci2012:
Boats are defined by 4 values.

Length, Beam (width), Depth (the part into the water),  Height (the part above the water).
There are height restrictions like bridges, power lines on pylons etc.
There are depth restictions (waterways are deeper than banks, gates, under water powerlines, pipelines etc. ) 
There are width restrictions (gates, bridges, sometimes waterways like canals.
There are length restrictions like gates or tight bends in waterways.

So just adding height is only the 3rd of the 4 key values for routing a ship.

PR #9269, opened 2020-06-18
Fix remove Add action from add action menu

Dima-1:


vshcherb:
is it correct??? 
all places checked?

PR #9295, opened 2020-06-23
process copied text in osmand

veliymolfar:


vshcherb:
Опять MapActivity? Уже говорили, что этот класс надо уменьшать, так что на доработку

decadent:
If I'm not mistaken, ‘PROCESS_TEXT’ means the overflow menu on text selection (i.e. ‘...’). IMO that menu is better kept for a few most-often-used actions, like web search, and not for all apps on the device. I think the ‘share’ menu is better for this action.

PR #9310, opened 2020-06-24
FIX: rescale arrow image using the stroke.

ehermann:
It tries to compute width of the stroke at the bottom of the image (where it's "attached" to the stroke).
To keep it safe for future changes, if it fails to compute a width it will fall back to the current de facto ratio in the image (2.25)


vshcherb:
This code looks very strange cause it browses the picture by getting pixels, it's weird.
Simple code :
1. should have coefficient 1.0 by default scale (1.0 should be skipped for postScale)
2. there should be a quick check or cache when it should be scaled base on default line width.

For now this code is quite  weird, I leave the original issue open so we can refactor it later.

ehermann:
You are completely right the "weird" part I did after I decided to make a pull request, since I wanted to have a solution that was independent of the arrow image you have right now. My first version was much simpler and I'll share with you in another pull request just in case it interest you.
Best regards

PR #9317, opened 2020-06-25
handle share text intent

veliymolfar:


decadent:
Thanks @veliymolfar and @vshcherb!

PR #9321, opened 2020-06-25
Option to use volume buttons as zoom

Dima-1:


OsmAnd/src/net/osmand/plus/activities/MapActivity.java
```diff
@@ -1546,6 +1546,14 @@ public void run() {
 				uiHandler.sendMessageDelayed(msg, LONG_KEYPRESS_DELAY);
 			}
 			return true;
+		} else if (settings.USE_VOLUME_BUTTONS_AS_ZOOM.get()) {
```
vshcherb, line 1549: Again code in map activity should be moved out

Issue #9307, opened 2020-06-24
Wrong arrow width on route action drawing


ehermann:
Hello,
When choosing a large width route style (for better outdoor visibility for instance) the arrow tip is of the wrong size.
Example using strokeWidth="20"

![image](https://user-images.githubusercontent.com/467129/85527566-4662d880-b60b-11ea-9ea9-10956551c9fd.png)
Expected result:
![image](https://user-images.githubusercontent.com/467129/85528273-f9333680-b60b-11ea-8de7-7118aaf31d74.png)
P.S. I already have the correction for this, I'll make a pull request asap.

vshcherb:
Looking forward with PR

PR #9339, opened 2020-06-27
FIX: rescale arrow if stroke is too large

ehermann:
2.25 is the ratio of the stroke size compared to the image size.


ehermann:
#9307

PR #9362, opened 2020-07-01
Fix vehicle parameters numeric dialog

Dima-1:


OsmAnd/src/net/osmand/plus/settings/preferences/SizePreference.java
```diff
@@ -85,7 +89,8 @@ public CharSequence getSummary() {
 		String persistedString = getValue();
 		if (!persistedString.equals(defaultValue)) {
 			try {
-				persistedString = String.valueOf(Float.parseFloat(persistedString) + 0.01f);
+				final DecimalFormat df = new DecimalFormat("#.####", new DecimalFormatSymbols(Locale.US));
```
vshcherb, line 92: Strange part.... didn't understand why so complicated

Dima-1: Yes, String.valueOf() is redundant

PR #9466, opened 2020-07-16
Suggest better wording for the 3 route-by-GPX options

sonora:


ehermann:
I'm not sure that "Route can extend before and after the track" is much better. But I agree that "Calculate OsmAnd route for first and last route segment" is not clear enough, By the way I had to look at the code to understand what it does.
"Calculate route to/from the track", maybe?

vshcherb:
Will be changed soon by different design, so no need

vshcherb:
The idea of new UI:

Route from my position to track:
- Nearest point VS - Start point
- Straight line VS - Route

PR #9475, opened 2020-07-18
Add rotation to setMapLocation API

ehermann:
Added a rotation parameter to setMapLocation.
I think it's better to have it before animated parameter, but if you prefer to put it at the end I can do it.
I also added a  SetMapLocationParams constructor without rotation for backward compatibility

### DIFF
--- a/OsmAnd-api/src/net/osmand/aidlapi/map/SetMapLocationParams.java
+++ b/OsmAnd-api/src/net/osmand/aidlapi/map/SetMapLocationParams.java
@@ -10,12 +10,18 @@ public class SetMapLocationParams extends AidlParams {
 	private double latitude;
 	private double longitude;
 	private int zoom;
+	private float rotation;
 	private boolean animated;
 
 	public SetMapLocationParams(double latitude, double longitude, int zoom, boolean animated) {
+		this(latitude, longitude, zoom, Float.NaN, animated);
+	}
+	
+	public SetMapLocationParams(double latitude, double longitude, int zoom, float rotation, boolean animated) {
 		this.latitude = latitude;
 		this.longitude = longitude;
 		this.zoom = zoom;
+		this.rotation = rotation;
 		this.animated = animated;
 	}
 
@@ -47,6 +53,10 @@ public class SetMapLocationParams extends AidlParams {
 		return zoom;
 	}
 
+	public float getRotation() {
+		return rotation;
+	}
+	
 	public boolean isAnimated() {
 		return animated;
 	}
@@ -56,6 +66,7 @@ public class SetMapLocationParams extends AidlParams {
 		bundle.putDouble("latitude", latitude);
 		bundle.putDouble("longitude", longitude);
 		bundle.putInt("zoom", zoom);
+		bundle.putFloat("rotation",rotation);
 		bundle.putBoolean("animated", animated);
 	}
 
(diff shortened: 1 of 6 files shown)


OsmAnd-api/src/net/osmand/aidlapi/map/SetMapLocationParams.java
```diff
@@ -56,6 +66,7 @@ public void writeToBundle(Bundle bundle) {
 		bundle.putDouble("latitude", latitude);
 		bundle.putDouble("longitude", longitude);
 		bundle.putInt("zoom", zoom);
```
vshcherb, line 68: new fields always should be last parameter

cc @Chumva

Chumva: It is ok to add new fields in Bundle in a different order, but preferably add them at the end for better understanding differences among versions. But in parcelable order of written and read values is important and changes in it will break reading values.

OsmAnd-telegram/src/net/osmand/aidl/map/SetMapLocationParams.java
```diff
@@ -52,13 +63,15 @@ public void writeToParcel(Parcel out, int flags) {
 		out.writeDouble(latitude);
 		out.writeDouble(longitude);
 		out.writeInt(zoom);
+		out.writeFloat(rotation);
```
Chumva, line 66: This one should be written last.

OsmAnd-telegram/src/net/osmand/aidl/map/SetMapLocationParams.java
```diff
@@ -52,13 +63,15 @@ public void writeToParcel(Parcel out, int flags) {
...
 		out.writeByte((byte) (animated ? 1 : 0));
 	}
 
 	private void readFromParcel(Parcel in) {
 		latitude = in.readDouble();
 		longitude = in.readDouble();
 		zoom = in.readInt();
+		rotation = in.readFloat();
```
Chumva, line 74: Also, read the new value in the end.

OsmAnd/src/net/osmand/aidl/OsmandAidlApi.java
```diff
@@ -300,6 +302,9 @@ public void onReceive(Context context, Intent intent) {
 							zoom = zoom > mapView.getMaxZoom() ? mapView.getMaxZoom() : zoom;
 							zoom = zoom < mapView.getMinZoom() ? mapView.getMinZoom() : zoom;
 						}
+						if(rotation != Float.NaN) {
```
Chumva, line 305: Probably it is better to use !Float.isNaN().

OsmAnd/src/net/osmand/aidl/map/SetMapLocationParams.java
```diff
@@ -8,12 +8,18 @@
 	private double latitude;
 	private double longitude;
 	private int zoom;
+	private float rotation;
 	private boolean animated;
 
 	public SetMapLocationParams(double latitude, double longitude, int zoom, boolean animated) {
+		this(latitude, longitude, zoom, Float.NaN, animated);
```
Chumva, line 15: The second constructor is not necessary for backwards compatibility, the only order of written and read values is important, except for some situations with lists of items.

OsmAnd-api/src/net/osmand/aidlapi/map/SetMapLocationParams.java
ehermann: Do you think it's better to reorder everything even the parameters order in the constructor, or just the putFloat?

Chumva: `putFloat` and `getFloat` will be enough

ehermann:
Just pushed a new version with the suggested changes.

PR #9477, opened 2020-07-19
Add aidl interface to lock and unlock the screen.

ehermann:
For instance it can allow a service to wake up the navigation using media buttons

### DIFF
--- a/OsmAnd/src/net/osmand/aidl/IOsmAndAidlInterface.aidl
+++ b/OsmAnd/src/net/osmand/aidl/IOsmAndAidlInterface.aidl
@@ -96,6 +96,7 @@ import net.osmand.aidl.mapmarker.RemoveMapMarkersParams;
 
 import net.osmand.aidl.quickaction.QuickActionParams;
 import net.osmand.aidl.quickaction.QuickActionInfoParams;
+import net.osmand.aidlapi.lock.SetLockStateParams;
 
 // NOTE: Add new methods at the end of file!!!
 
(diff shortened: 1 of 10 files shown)


OsmAnd/src/net/osmand/aidl/IOsmAndAidlInterface.aidl
```diff
@@ -96,6 +96,7 @@ import net.osmand.aidl.mapmarker.RemoveMapMarkersParams;
 
 import net.osmand.aidl.quickaction.QuickActionParams;
 import net.osmand.aidl.quickaction.QuickActionInfoParams;
+import net.osmand.aidlapi.lock.SetLockStateParams;
```
Chumva, line 99: This will bring some problems for apps which try to update to new API version because `aidlapi` is another package (newer API) and it might be unimported in some apps.

ehermann: I must confess that I didn't completely understand the concept behind OsmandAidlServiceV2 and OsmandAidlService, since there are a lot of code duplication. 
Is V2 intended to replace the old?

Chumva: Yes, V2 intended to replace old, because there were issues with backwards compatibility for new fields. V2 has a similar structure but uses bundle as the main storage element and it solves some issues with write/read fields order and backwards compatibility for items inside lists.  It is not necessary to add new methods to old api.

ehermann: Ok, now it's clear! Thanks!
So, maybe I'll make a commit removing my adds in the old api, It would be easier to maintain with less code duplication. What do you think?

Chumva: I don't think that removing these changes is necessary, there is much code duplication and one more method will not change the situation. It will be better to test new methods thoroughly instead of removing this adds.

ehermann:
Is there already a place for aidlapi  tests?

Chumva:
Unfortunately, there are no tests for the aidl API implemented.

PR #9509, opened 2020-07-23
Track point menu fixes

Chumva:


OsmAnd/res/values/strings.xml
```diff
@@ -11,6 +11,7 @@
 	Thx - Hardy
 
 -->
+    <string name="overwrite_gpx">Overwrite GPX</string>
```
vshcherb, line 14: overwrite track

Chumva: Done

PR #9567, opened 2020-08-03
Fix external device keypresses

kconger:
Looks like a recent change to handle onKeyUp and onKeyDown produces duplicate input with External Input Device.  This is a quick fix.


nazar-kutz:
@kconger Thanks for your fix!

cngripper:
Hi,
I dont think this fix is correct, now it its impossible to scroll, for example with an GENERIC_EXTERNAL_DEVICE?

vshcherb:
Please create an issue if it won't be fixed

PR #9575, opened 2020-08-06
Update grammar and links in multiple files

TacoTheDank:
Fixed up a bunch of old links, grammar, and formatting.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,26 +1,32 @@
 Before entering an issue, please read the information below.
 
-```GitHub is a main development tool for developers. There are hundreds requests a month and there are only few developers. So by opening an issue, please know that your issue will be sent out to all developers and acknowledge that it could be closed without explanation or with a brief message. Comments on the closed issues are also sent to all developers, so you will definitely will be heard. However, there is no guarantee that a developer will pick up the issue to work on it.```
+```GitHub is our main development tool for our developers. There are hundreds of requests a month and there are relatively few developers.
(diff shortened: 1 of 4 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -1,26 +1,32 @@
 Before entering an issue, please read the information below.
 
-```GitHub is a main development tool for developers. There are hundreds requests a month and there are only few developers. So by opening an issue, please know that your issue will be sent out to all developers and acknowledge that it could be closed without explanation or with a brief message. Comments on the closed issues are also sent to all developers, so you will definitely will be heard. However, there is no guarantee that a developer will pick up the issue to work on it.```
+```GitHub is a massive development tool for developers. There are hundreds of requests a month and there are relatively few developers.
```
njohnston, line 3: I don't think "massive" is what was intended in the original text. I think the intention was probably "GitHub is mostly a tool for developers" or similar.

vshcherb: I agree that should be a separate request better, so we don't need to freeze this one with discussion about this paragraph

TacoTheDank: I couldn't quite understand what the original sentence meant to say, either, since it's kinda broken English.

vshcherb: Well 2 deleted sentences certainly have meaning. They describe about comments  or about new issues and explain something to avoid frustration from some people

TacoTheDank: ? I don't think I deleted any sentences...

I updated the commit to better reflect what you wanted it to say (I think).

PR #9656, opened 2020-08-17
Allow custom "user-agent" for tile download

ParksProjets:
This PR allows you to use a custom "User-Agent" for tile request, instead of the default one. Like "Referer", this parameter is optional in both SQLite and MetaInfo formats. Btw, this PR also loads `referer` parameter from MetaInfo.


vshcherb:
Looks good, hopefully you can test it and give feedback

PR #9666, opened 2020-08-19
Fix #9533

### LINKED ISSUE #9533: [OsmAnd Edit Plugin] Configuring context menu is not persistent
When configuring the actions "create/edit note" and "create/edit POI" to be always visible as 1 of the four actions in the context menu (for quick access), they are getting kicked out after certain workflow.

As far as I only keep adding POIs / notes everything works fine.
However when I go and select an edited note / POI, the entry disappears from the main actions context menu.
Instead now an item appears at last position of the additional actions menu: "edit POI" or "edit note".
The main context menu now has only 3 entries. 
If I now move the edit action into it, it continues to work for edit, however when clicking somewhere else on the map (where no POI/note can be edited), the edit is gone again and "create ...." is added again as last item in the additional actions menu.

**Root cause** looks like actions are dynamically added and removed and thereby breaking the functionality of adding them consistently to the main context menu (or anywhere above last entry in additional actions menu).

**Solution**:
- Do not create new actions, but edit the existing ones (update the title and action)
- or always keep the title "create/edit" and only switch the action
--- if possible

v. 3.7.4

androiddevkotlin:


Chumva:
This was already fixed here - https://github.com/osmandapp/OsmAnd/pull/9646

PR #9711, opened 2020-08-28
Plan route save as new track

Dima-1:


OsmAnd/res/values/strings.xml
```diff
@@ -24,6 +26,9 @@
     <string name="disable_recording_once_app_killed_descrp">Will pause track logging when the app is killed (via recent apps). (OsmAnd background indication disappears from the Android notification bar.)</string>
     <string name="monitoring_control_start">REC</string>
     <string name="number_of_gpx_files_selected_pattern">%s track files selected</string>
+    <string name="shared_string_file_name">File name</string>
+    <string name="simplified_track_description">Only the route line will be saved, the waypoints will be deleted.</string>
```
sonora, line 30: Is this tchnicallly correct? Or should it say something like "Only turning points will be saved, all other track points will be deleted."

vshcherb:
It's correct,
Simplified track - only geometry points will be saved
Full track - Geometry points + Route points with extended information about turns.
Some software doesn't work correctly with <rtept> so we would like to keep simple track option available as well

PR #9731, opened 2020-09-02
ui fixes started

vshcherb:


OsmAnd/res/values/dimens.xml
```diff
@@ -16,5 +16,6 @@
 
     <dimen name="line_spacing_multiplier_description">1.2</dimen>
     <dimen name="line_spacing_extra_description">5dp</dimen>
+    <dimen name="point_editor_fragment_expand_button_height">42dp</dimen>
```
vshcherb, line 19: doesn't look correct, consult @dmpr0 @Chumva

PR #9751, opened 2020-09-04
Profile import issues fix

Dima-1:


OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java
```diff
@@ -3804,6 +3808,13 @@ protected boolean setValue(Object prefs, String val) {
 		return customRoutingProps.get(attrName);
 	}
 
+	{
```
vshcherb, line 3811: Bad fix, parameters could be dynamic, so they shouldnot be set here.

getCustomRoutingProperty - doesn't make any sense cause it shouldn't change any state!

PR #9798, opened 2020-09-13
Save dialog in route details

Dima-1:


OsmAnd/src/net/osmand/FileUtils.java
```diff
@@ -165,6 +171,22 @@ public static File checkRenamePossibility(OsmandApplication ctx, File source, St
 		return dest;
 	}
 
+	public static String createName(OsmandApplication app) {
```
vshcherb, line 174: Give proper file name
1. Parameter Name Format 
2. Parameter  Folder name

vshcherb: Method name - createUniqueFileName

PR #9820, opened 2020-09-16
Plan a route executor multiple calculations bug fix

angry010101:
When a user starts "route between points"  and switches profiles in "plan a route" the calculations might not stop.


scaidermern:
Why does the calculation even start before the user selected a profile?

vshcherb:
Profile could be selected on the step before, for example in follow track.

PR #9835, opened 2020-09-17
Length limit on the "Vehicle parameters" screen shows 0.01 m

androiddevkotlin:


OsmAnd/src/net/osmand/plus/settings/preferences/SizePreference.java
```diff
@@ -100,7 +106,11 @@ public CharSequence getSummary() {
 		return summary;
 	}
 
-	public String getValue() {
+	private boolean isPersistedStringEqualsZero(String persistedString) {
```
vshcherb, line 109: 1. Doube.parseDouble() = 0.0d is the same and even simpler
2. Just FYI BigDecimal.ZERO

Chumva:
This is a copy of #9864

PR #9847, opened 2020-09-18
Issues 158 no ele after attach

angry010101:
https://github.com/osmandapp/OsmAnd-Issues/issues/158

Strange point in the ocean after manual attach thru "Plan a route". - !!! CHECK FIXME !!!


OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
```diff
@@ -549,8 +549,10 @@ private boolean initRoutingPoint(GpxPoint start, GpxRouteApproximation gctx, dou
 		if (start != null && start.pnt == null) {
 			gctx.routePointsSearched++;
 			RouteSegmentPoint rsp = findRouteSegment(start.loc.getLatitude(), start.loc.getLongitude(), gctx.ctx, null, false);
-			if (MapUtils.getDistance(rsp.getPreciseLatLon(), start.loc) < distThreshold) {
-				start.pnt = rsp;
+			if (rsp != null) {
```
vshcherb, line 552: could be online rsp != null &&

PR #9871, opened 2020-09-22
Fix #9623

### LINKED ISSUE #9623: Map style is sometimes not switched as expected with profile switch
Short description: Use different map styles on Walking and Driving profiles. Switch from Walking to Driving. It keeps showing the Walking map style.

I have seen this phenomenon a couple of times, but I can not reproduce it "on demand".

I did this [see text underneath figures]: 
Select the _Walking_ profile with map style Touring view, and Underlay map (online). (It sometimes is working a bit hard rendering a lot of map details). 

Map-style-problem-1
Then I switch to _Driving_ profile, where map style _OsmAnd_ was selected at earlier session. (Btw, Driving is not using underlay map).

Map-style-problem-2
Map style OsmAnd was expected here (and expected used on map).
Problem: it uses Touring view, both text and map. (As if it hangs from previous profile(?))

Map-style-problem-3
I tap on Map style, the radio-button list is actually correctly preselected on OsmAnd.
When I tap Apply, then it starts using OsmAnd map style.

(OsmAnd+ 3.7.4 on android. Using offline maps + online underlay map.)

nazar-kutz:


vshcherb:
Too complicated. Better to avoid listeners and call refresh context menu after profile switch or something like that. 
Listeners have leaking problems etc... So it's better to invalidate cache directly if it's persisted or invalidate cache immediately on opening configure map menu

OsmAnd/src/net/osmand/plus/dashboard/DashboardOnMap.java
```diff
@@ -1032,6 +1042,12 @@ void onDetach(DashBaseFragment dashBaseFragment) {
 		}
 	}
 
+	public void onMapSettingsUpdated() {
+		if (configureMapMenu != null) {
+			configureMapMenu.updateMenuItem(MAP_STYLE_ID);
```
vshcherb, line 1047: Here we should not use MAP_STYLE_ID. We don't know which item will be updated and we don't know how many items their ids, etc...

OsmAnd/src/net/osmand/plus/dashboard/DashboardOnMap.java
```diff
@@ -656,10 +659,16 @@ public void setDashboardVisibility(boolean visible, DashboardType type, Dashboar
 				fragment.show(mapActivity.getSupportFragmentManager(), MapillaryFirstDialogFragment.TAG);
 				settings.MAPILLARY_FIRST_DIALOG_SHOWN.set(true);
 			}
+
+			deleteTmpReferences();
 		}
 		mapActivity.updateStatusBarColor();
 	}
 
+	private void deleteTmpReferences() {
```
vshcherb, line 668: Looks bad

OsmAnd/src/net/osmand/plus/dashboard/DashboardOnMap.java
vshcherb: Here should be code like: cm.updateMenuItem(). configureMapMenu shouldn't be stored, it's 1 method

Issue #9621, opened 2020-08-12
OsmAnd crashes when faulty Online tracking web address is given


rrooggiieerr:
I'm trying to log my location to an online webservice. I made an error in the url and OsmAnd ket crashing until I manages to disable GPX logging

This is what the wrong url looks like:

http://example.com/log?lat={0}&lom={1}&hd={3}&al={4}&sp={5

Notice the missing } at the end. Then the } is added OsmAnd is no longer crashing

So I think the app should be made a bit more robust by not crashing on a faulty Online tracking web address

AlexSyt:
@rrooggiieerr , should be fixed now :)

PR #9974, opened 2020-10-07
#9621 Add validation for the online tracking link

AlexSyt:
@Chumva , and one more to go)


Chumva:
@AlexSyt This should fix that issue, but there are two places in UI with this preference. The one you added is legacy one and it will be removed in the near future. New settings UI with this pref is `LiveMonitoringFragment`. Can you add or move your check there?)

AlexSyt:
@Chumva , ok, I'll do it, thanks)

AlexSyt:
@Chumva , done :)

PR #10019, opened 2020-10-13
Adding Osmand Diagram

deborafb:
This document describes part of the architecture of the OsmAnd project, It is important to highlight that the entire architecture is not described.This description was based mainly on the C4 model.


vshcherb:
We are looking forward to new documentation here https://docs.osmand.net/en
and https://github.com/osmandapp/docs

PR #10057, opened 2020-10-21
Favorites: Allow to share the same name if favorites belong to differ…

androiddevkotlin:
…ent groups


vshcherb:
И все? Все остальное работает?

Chumva:
There is support for duplicate names in different categories based on keys in FavouritesDbHelper https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/FavouritesDbHelper.java#L588.

PR #10071, opened 2020-10-23
Open Places Images

angry010101:
- add sign in via webview
- images download from openplacereviews
- images upload to blockchain

### DIFF
--- a/OsmAnd/res/values-fa/phrases.xml
+++ b/OsmAnd/res/values-fa/phrases.xml
@@ -226,14 +226,14 @@
     <string name="poi_pension_fund">صندوق بازنشستگی</string>
     <string name="poi_migration">سازمان مهاجرت</string>
     <string name="poi_tax_inspection">اداره مالیات</string>
-    <string name="poi_office_administrative">دفتر اجرایی</string>
+    <string name="poi_office_administrative">اداره حکومتی</string>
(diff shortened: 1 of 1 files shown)


OsmAnd/AndroidManifest.xml
```diff
@@ -66,6 +66,7 @@
 		<meta-data android:name="com.sec.minimode.icon.landscape.normal" android:resource="@mipmap/icon" android:value="" />
         <activity android:name="net.osmand.plus.activities.HelpActivity" />
         <activity android:name="net.osmand.plus.activities.ExitActivity" />
+        <activity android:name=".osmedit.opr.OPRWebviewActivity" />
```
vshcherb, line 69: net.osmand.plus.

vshcherb:
All comments  from previous review are still valid

OsmAnd/build.gradle
```diff
@@ -227,11 +227,15 @@ android {
 
 	buildTypes {
 		debug {
+			buildConfigField "String", "OPR_BASE_URL", "\"https://test.openplacereviews.org/\""
```
vshcherb, line 230: OPR_BASE_URL - should be encoded inside app, and now should be always test, later will be changed always https://openplacereviews.org/

PR #10079, opened 2020-10-26
Add favorites: Add names for icon groups to strings

androiddevkotlin:


OsmAnd/res/values/strings.xml
```diff
@@ -11,6 +11,14 @@
 	Thx - Hardy
 
 -->
+    <string name="category_group_travel_icon">Travel</string>
```
vshcherb, line 14: better icon_group_travel

PR #10107, opened 2020-10-29
Mgrs coordinates

jBackhaus:
Hey everyone!
I added support for the [MGRS coordinate standard](https://en.wikipedia.org/wiki/Military_Grid_Reference_System). It is sometimes called "NATO UTM Standard".

This standard was already included in jwetherell's openmap.common-package that OsmAnd includes. But it was not used by now.

This coordinate standard is used intensively by all the public emergency services – like ambulances, firefighters, etc. – at least in Germany. That's why I added it.

The second commit extends jwetherell's package itself by an option of flavoring the coordinate string. This adds spaces between the coordinates parts what makes the coordinates easier to read and –much more important – allows to search for them even with spaces included (which is quite common). 

I also included an option for modifying the accuracy of the coordinate (length of the numbers) at this point, which still keeps the highest accuracy (1m) for calculation. This is different than jwetherell's implementation, which would decrease the overall accuracy at a point's creation. Currently, this is hardcoded in OsmAnd to keep the highest resolution but one might allow the user to modify the accuracy in the profile settings later.

Currently, there are only English and German translations of the settings descriptions. One might add different languages later, but the name 'MGRS' itself is internationally common and needs no translation.

### DIFF
--- a/OsmAnd-java/src/main/java/com/jwetherell/openmap/common/MGRSPoint.java
+++ b/OsmAnd-java/src/main/java/com/jwetherell/openmap/common/MGRSPoint.java
@@ -14,6 +14,10 @@
 
 package com.jwetherell.openmap.common;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 public class MGRSPoint extends ZonedUTMPoint {
 
     /**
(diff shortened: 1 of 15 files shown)


dmpr0:
Hello! Thank you for your contribution.
I have tested it, all works fine: coordinates widget, context menu, settings, except for the search in "Search – Address – Coordinates", regular search show results:
coordinates

jBackhaus:
Fixed this.
I moved the removing of possible spaces in the MGRS string into the decoding inside the MRGSPoint class. Now it catches all the occurrences.

PR #10145, opened 2020-11-06
Fix #10021 Favorite size should not be arbitrary

### LINKED ISSUE #10021: Favorite size should not be arbitrary
### Description

Favorites are currently displayed in two sizes, large and small. For a given favorite, it is arbitrary whether it gets the small icon or the large one. An algorithm is deciding, apparently to make the map more beautiful.

### The issue

This is a UX problem. In practice, on many maps the mini-favorites are very difficult to see. Check example capture: the light green favorite on the left might be a park, and good luck even seeing the yellow favorite by the bridge. Feature request #9104 will be useless with these small icons, for example. **What if the user wants to be able to see _all_ favorites, even if there are a lot of them? Different favorite sizes are great but it should not be done arbitrarily!**

Please at least provide an option to turn off the sizing algorithm so that all favorites are equal. 

### How to reproduce?

1. Add some favorites somewhat near to each other.
2. Zoom out.

### Your Environment
OsmAnd Version: 3.8.3
Android/iOS version: 8

Dima-1:


vshcherb:
Any constant is possible ?

vshcherb:
in common class for example instead of hardcode in many places

PR #10184, opened 2020-11-14
Fix typo in strings.xml

aceman444:


sonora:
Ah, yeah, had also hust seen and fixed. Thanks!

PR #10198, opened 2020-11-16
Export UI first screen

Chumva:


OsmAnd/res/values/strings.xml
```diff
@@ -11,6 +11,12 @@
 	Thx - Hardy
 
 -->
+    <string name="export_not_enough_space">There is not enough space</string>
+    <string name="export_not_enough_space_descr">OsmAnd is needed %1$s to export a file. Your device only has %2$s free. You can free up space on the device, or skip some items from export.</string>
```
alex-osm, line 15: Fix string

alex-osm (changes requested):
Fix string

PR #10203, opened 2020-11-16
Login strings reworked

comradekingu:


OsmAnd/res/values/strings.xml
```diff
@@ -14,10 +14,10 @@
     <string name="shared_string_search_history">Search history</string>
     <string name="register_opr_have_account">I already have an account</string>
     <string name="register_opr_create_new_account">Create new account</string>
-    <string name="register_on_openplacereviews_desc">Photos are provided by open data project OpenPlaceReviews.org. In order to upload your photos you need to sign up on website.</string>
+    <string name="register_on_openplacereviews_desc">Log in on the open data project website OpenPlaceReviews.org to upload even more photos.</string>
```
vshcherb, line 17: This is not a login cause it is method for "register" or "sign up"

comradekingu: I didn't understand that.
What I tried to do was say logging in allows uploading more images than the user sees.

vshcherb: Hm... I need to check where this displayed probably it has 2 alternatives

PR #10276, opened 2020-11-26
Fix_10257

Chumva:


vshcherb:
There is no encode on backend so it's not clear how did it gets encoded? Is it encoded already in OSM?

Chumva:
Yes, it looks like there are some image filenames with encoding in OSM.

PR #10283, opened 2020-11-26
Send GPX Bottom Sheet

androiddevkotlin:
Bottom sheet should scroll outside the screen
Check keyboard doesn't show after open


androiddevkotlin:
Doesn't work as intended

PR #10326, opened 2020-12-03
Remove default GPX color

Chumva:


vdcrim:
Sorry if I'm jumping the gun, but does this implies that OsmAnd GPX extensions will no longer be read when importing a GPX file? That would be a great loss. It's very useful to add already formatted tracks. For instance, when you have a hundred way-points and want them to have different icons.

Chumva:
@vdcrim These extensions will be read as usual, but gpx appearance tags will be saved to internal SQLite DB, instead of gpx tags. And during gpx file sharing, they will be added to the gpx file.

jensMF:
@Chumva adding the appearance tags to the gpx file during sharing only works when I share the track by tapping on it and choose share.
When I tap on open track or navigate to the track in My places an choose sharing there, the appearance tags are not added. When I choose the three-dot menu behind the filename in My places it dosn't work either.

PR #10355, opened 2020-12-07
Fix profiles export

Chumva:


prallo:
Still not fixed, cannot export profile ! 
Screenshot_20230128_103602_OsmAnd+
Screenshot_20230128_103913_OsmAnd+

PR #10357, opened 2020-12-08
Fix recalculate all segments with the selected profile

Dima-1:


OsmAnd/src/net/osmand/plus/helpers/AvoidSpecificRoads.java
```diff
@@ -163,11 +164,17 @@ public String getRoadName(@Nullable RouteDataObject obj) {
 		return app.getString(R.string.shared_string_road);
 	}
 
-	private void recalculateRoute() {
+	private void recalculateRoute(MapActivity mapActivity) {
 		RoutingHelper rh = app.getRoutingHelper();
 		if (rh.isRouteCalculated() || rh.isRouteBeingCalculated()) {
 			rh.recalculateRouteDueToSettingsChange();
 		}
+		if (mapActivity != null) {
```
vshcherb, line 172: Make similar to RoutingHelper via app -> remove dependency on UI MeasurementToolFragment

PR #10378, opened 2020-12-10
Add "Truck" and "Motorcycle" profile by default

androiddevkotlin:
Custom profile (deleteable) profile

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -433,7 +433,7 @@
     <string name="hillshade_description">Hillshade map using dark shades to show slopes, peaks and lowlands.</string>
     <string name="shared_string_terrain">Terrain</string>
     <string name="import_rendering_file">Import rendering file</string>
-    <string name="profile_type_custom_string">Custom profile</string>
+    <string name="profile_type_custom_string">User profile</string>
     <string name="shared_string_angle_param">Angle: %s°</string>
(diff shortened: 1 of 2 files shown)


OsmAnd/res/values/strings.xml
```diff
@@ -433,7 +433,7 @@
     <string name="hillshade_description">Hillshade map using dark shades to show slopes, peaks and lowlands.</string>
     <string name="shared_string_terrain">Terrain</string>
     <string name="import_rendering_file">Import rendering file</string>
-    <string name="profile_type_custom_string">Custom profile</string>
```
vshcherb, line 436: probably new string?

OsmAnd/src/net/osmand/plus/AppVersionUpgradeOnInit.java
```diff
@@ -175,15 +175,14 @@ public void resetFirstTimeRun(SharedPreferences startPrefs) {
 	}
 
 	private void addCustomProfile() {
```
vshcherb, line 177: This is incorrect approach cause profiles will be added once during upgrade and they should not be deletable actually

Chumva: Okay, I assumed that they should behave as other custom profiles which user can add with the ability to delete them.

Issue #10230, opened 2020-11-19
routing takes unnecessary u-turns - should auto-report those


KlavsKlavsen:
### Routing engine

- [x ] OsmAnd's in-app offline routing

Car, fastest route.

### Start and end points
I asked for a route from "hædersdalvej, hvidovre" https://www.openstreetmap.org/way/10636079
to "vestergade, roskilde" https://www.openstreetmap.org/way/145394545

### Actual and expected routes

![image](https://user-images.githubusercontent.com/2737885/99667180-4d7ffc80-2a6c-11eb-9b3e-6fedb60a37d8.png)

Problem was that osmand told me to turn right (right after getting off freeway) - and then directly after, do a u-turn (which I actually could not until a few km further down).

I could have just turned left.

### Is this a regression?

I don't know. I think its a map issue?

## 🌍 Your Environment

osmand 3.8.3

Android 8.1 (lineageOS 15.1)

**Maps used (online or offline):**
Offline - Denmark map.

- [x] Offline maps offered within the OsmAnd app for download.

osmgit:
I've fixed this problem. I can confirm this was a problem with the turn restrictions at that node on the map. In this case Left Turn was set to allowed.

Do you have an openstreetmap.org account? If not then you just go here https://www.openstreetmap.org and sign up. You can then log in and edit or add to the map!

vshcherb:
Yes we have this bug reported often and we will need to create a solution for it, we will need somehow to introduce a right penalty for such turns that will penalize enough to avoid them

KlavsKlavsen:
@osmgit I do have an account on openstreetmap.org and I have added roadnames I found weren't in openstreetmap etc. - but I have yet to figure out how to figure what goes wrong (f.ex. here) - with the routing, and as such - how to fix this in openstreetmap. It seems rather difficult to debug, if you're not very well versed in the subject (and I'm really not :)
If you wrote down "incident reports" and corrections - a small.. Had this bug: "image of routing error" - and described how you fixed it (in absense og a git log I could view and repeat for openstreetmap.org ) - then everyone could learn from how you fixed those - and thus hopefully more easily, figure out how to fix the bug they found ?

KlavsKlavsen:
Is the route calculations documented somewhere @vshcherb ? then I'd gladly read that and see if I have any bright ideas :)
[...]

PR #10481, opened 2020-12-28
#10230 Change turnLeft and turnRight, calculateTurnTime()

alisa911:


OsmAnd-java/src/main/java/net/osmand/router/GeneralRouter.java
```diff
@@ -222,10 +222,10 @@ public void addAttribute(String k, String v) {
 		attributes.put(k, v);
 		if(k.equals("restrictionsAware")) {
 			restrictionsAware = parseSilentBoolean(v, restrictionsAware);
-		} else if(k.equals("leftTurn")) {
-			leftTurn = parseSilentFloat(v, leftTurn);
-		} else if(k.equals("rightTurn")) {
-			rightTurn = parseSilentFloat(v, rightTurn);
+		} else if(k.equals("sharpTurn")) {
+			sharpTurn = parseSilentFloat(v, sharpTurn);
+		} else if(k.equals("slightTurn")) {
```
vshcherb, line 227: same as C++ Backward compaitibility

OsmAnd-java/src/main/java/net/osmand/router/GeneralRouter.java
```diff
@@ -605,11 +605,11 @@ public float getMaxSpeed() {
 	}
 
 	public double getLeftTurn() {
```
vshcherb, line 607: functions should be renamed

OsmAnd-java/src/main/java/net/osmand/router/GeneralRouter.java
```diff
@@ -654,6 +652,11 @@ public double calculateTurnTime(RouteSegment segment, int segmentEnd, RouteSegme
 			}
 			return 0;
 		}
```
vshcherb, line 654: same as in C++, penalty should be summed up

vshcherb:
It would be good to add test to routing!

PR #10493, opened 2020-12-30
Fix #10395 Add support of incomplete Social Media URLs

### LINKED ISSUE #10395: Support account names in contact:* Now it support only full links.
# 🚀 feature request

### Description

https://wiki.openstreetmap.org/wiki/Key:contact , section Social Media / Messengers / Web Services describes value as account name or full URL. Osmand works correctly only if full URL. If account name then osmand only shows it as text but does not create hyperlink to open contact.

Each social network should have predefined url template in code and Osmand should use it if value of tag does not contain **http(s)://**

For example 
contact:instagram - https://www.instagram.com/{value}/
contact:vk - https://vk.com/{value}
contact:facebook - https://www.facebook.com/{value}
contact:twitter - https://twitter.com/{value}
contact:youtube - https://www.youtube.com/c/{value}
contact:ok - https://ok.ru/{value}
contact:telegram - https://tttttt.me/{value}

cepprice:
[Demo](https://youtu.be/UjO2DE9Z3Qo)

### DIFF
--- a/OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityMenuBuilder.java
+++ b/OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityMenuBuilder.java
@@ -15,10 +15,6 @@ import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
-import androidx.annotation.NonNull;
-import androidx.appcompat.view.ContextThemeWrapper;
-import androidx.core.content.ContextCompat;
-
 import net.osmand.AndroidUtils;
 import net.osmand.PlatformUtil;
 import net.osmand.data.Amenity;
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityMenuBuilder.java
```diff
@@ -392,6 +403,80 @@ public void buildInternal(View view) {
 				isUrl = true;
 			}
 
+			if (!isUrl) {
```
vshcherb, line 406: Too long block, it should be a separate method

vshcherb:
Checks are too complicated and there is too much code.
Here could be one map:
mp = {"vk", "https://vk.com/$1"}, {"ok", "https://ok.ru/$1" }
And check could be as easy as

```
if(mp.containsKey(key) &&  !value.contains("/")) { value = String.format(mp.get(key), value); }
```
Simple code - better code!

PR #10494, opened 2020-12-30
Fix opening links to maps "geo:0,0"

Skalii:


OsmAnd-java/src/main/java/net/osmand/util/GeoPointParserUtil.java
```diff
@@ -558,6 +558,10 @@ public static GeoParsedPoint parse(final String uriString) {
 				if (lat == 0.0 && lon == 0.0 && positionInSearchRequestMatcher.find()) {
 					lat = Double.valueOf(positionInSearchRequestMatcher.group(1));
 					lon = Double.valueOf(positionInSearchRequestMatcher.group(2));
```
vshcherb, line 560: here should be a comment why do we set lat / lon to 0? 
I'm not sure if it's correct or not (probably it is)

Skalii: because from the line "geo:0,0?q=Van+Bleiswijkstraat+66,+2582+LG+The+Hague" in any case 66 and 2582 will be taken as latitude and longitude, which is wrong, because this is an address and there are a lot of such addresses. I decided to compare the minimum and maximum values ​​for coordinates in order to minimize the creation of incorrect coordinates.  maybe this is a temporary solution, maybe there are addresses that are in the min/max range, but nevertheless, similar addresses as from the issue are now processed normally, because further in the method there is a check for equality to 0.0 and the creation of a GeoParsedPoint from the searchRequest string

PR #10516, opened 2021-01-03
Language packages are updated every day

androiddevkotlin:


OsmAnd/src/net/osmand/plus/download/DownloadOsmandIndexesHelper.java
```diff
@@ -165,7 +165,7 @@ private static void listVoiceAssets(IndexFileList result, AssetManager amanager,
 					File destFile = new File(voicePath, target.substring(IndexConstants.VOICE_INDEX_DIR.length(),
 							target.indexOf("/", IndexConstants.VOICE_INDEX_DIR.length())) + "/" + lang + "_tts.js");
```
vshcherb, line 166: looks weird asset.source is not used at all

Issue #10209, opened 2020-11-16
Make OsmANd GPX compatible with BaseCamp


AnastasiaTiutina:
### Description

Faulty import file for Garmin Basecamp 

![gcflmekafoacpinl](https://user-images.githubusercontent.com/72389175/99291205-1af5ba00-2848-11eb-893c-1fd1425b444a.png)
![gcflmekafoacpinl](https://user-images.githubusercontent.com/72389175/99291224-20530480-2848-11eb-9e9d-20e28970a79a.png)
![icikfldmlnamcaoc](https://user-images.githubusercontent.com/72389175/99291229-22b55e80-2848-11eb-888a-2e5757ae1406.png)
![noemjjdiaohhdlac](https://user-images.githubusercontent.com/72389175/99291233-23e68b80-2848-11eb-8c35-a7653a7e3b7c.png)

### How to reproduce?

In your guide on the web you describe how to make GPX files from Osmand 
compatible to Garmin-BaseCamp / Mapsource using GPS-Babel.

I converted the file with these parameters with GPS-Babel:

The file format does not fit:

The files can be found in the attachment.

From my point of view it would be much better if the XML format of 
OsmAnd was basically compatible, so there would be no conversion at all.

One more hint: Planning GPS tracks for Garmin devices works without any 
problems. I like to show in my GPS trainings how easy it is to create a 
GPX track for an Edge or Fenix device with OsmAnd and transfer it to the 
device with Garmin-Connect. This can be an alternative for Komoot for 
many people. I have subsripted the OsmAnd version for iOS and Android.

jensMF:
related to #1326

jensMF:
@AnastasiaTiutina can you open this file in BaseCamp:

[2021-01-02_20-28_Sat_test.txt](https://github.com/osmandapp/OsmAnd/files/5760984/2021-01-02_20-28_Sat_test.txt)
(You need to rename it from .txt to .gpx)

?

vshcherb:
Looks like it's fixed with merge #1326

PR #10526, opened 2021-01-04
fix #1326 and #10209 by adding the XML namespace 'osmand' and osmand.xsd

### LINKED ISSUE #1326: OsmAnd creates invalid GPX v1.1 files
GPX v1.1 files generated by Osmand are invalid due to _speed_ (in _trkpt/extensions_) belonging to the default/gpx namespace. GPX v1.1 schema requires that [extensions belong to a "other" namespace](http://www.topografix.com/gpx/1/1/#type_extensionsType).

jensMF:
The location of osmand.xsd should be discussed and the link to it updated in the code when a decision is made.

fixes #1326 and #10209 and conflicts with #10542

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
+++ b/OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
@@ -1,4 +1,3 @@
-
 package net.osmand;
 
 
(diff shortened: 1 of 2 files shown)


vshcherb:
It is probably not bad idea but it would better I think to update documentation - if needed -https://docs.osmand.net/en/main@latest/development/osmand-file-formats/osmand-gpx. 

Even though XML used to be a strict format overtime it became less strict similar to JSON. In OSM we often deal with XML's that don't have such limitation.

Obviously better place for schema would be somewhere at https://github.com/osmandapp/Osmand-resources or at OsmAnd-java. 


So summarizing to be honest I don't see much sense of this complication

jensMF:
@vshcherb 
I opened another pull request with the same result but less changes. Maybe that one would be ok? I thought the other solution is a little bit like a hack, but otherwise its simple ...
see #10542

vshcherb:
Feel free to create PR to enhance documentation https://docs.osmand.net/en/main@latest/development/osmand-file-formats/osmand-gpx which we are going to support anyway

PR #10542, opened 2021-01-07
fix #1326 and #10209 by adding the XML namespace 'osmand' to the header and the writeExtensions function

### LINKED ISSUE #1326: OsmAnd creates invalid GPX v1.1 files

jensMF:
this fixes #1326 and #10209 and conflicts with #10526

I could also import a file created with this changes to BaseCamp without modification and I could verify correct XML syntax with `SAXCount`.


vshcherb:
Basecamp fix is a great one!

PR #10564, opened 2021-01-12
Add condition for BuildingInterpolation.ALPHABETIC

alisa911:


OsmAnd-java/src/main/java/net/osmand/data/Building.java
```diff
@@ -113,15 +113,20 @@ public float interpolation(String hno) {
...
-					char chF = fname.charAt(fname.length() - 1);
-					if (chF < ch) {
-						return -1;
-					}
+			String sname = getName2();
+
+			if (getInterpolationType() == BuildingInterpolation.ALPHABETIC && num == numB) {
+				char ch = hno.charAt(hno.length() - 1);
+				char chF = fname.charAt(fname.length() - 1);
+				char chS = sname.charAt(sname.length() - 1);
```
vshcherb, line 121: There is no interpolation between 2 location  like
```
if (numT > numB) {
				return ((float) num - numB) / (((float) numT - numB));					return ((float) num - numB) / (((float) numT - numB));
			}				}
```

PR #10587, opened 2021-01-16
Add BaseCamp compatibility for planned Tracks and claculated routes exported to gpx

jensMF:
I had to change the order in which wpt, rte and trk was written into the gpx files to comply to the gpx 1.1 standard.

I did test examples of both type of gpx files with BaseCamp and could successfully import them.


jensMF:
Probably this is necessary to fully fix #10209 as there is not stated, what kind of gpx file is meant: recorded tracks, planned routes or navigation routes shared via the details page of navigation.

PR #10595, opened 2021-01-16
Add an option for always adding track waypoints to the current track

njohnston:
When recording a track (and displaying it), and displaying one or more
tracks at the same time, OsmAnd asks which track to add any track waypoints
to.

This can become tedious when adding many track waypoints (like I do when
using OsmAnd to collect addresses). There's a possibility that you might
tap the wrong track by mistake, particularly if your phone screen is wet or
you're wearing gloves.

This change introduces a new setting for the trip recording plugin. When
enabled (it's disabled by default), track waypoints will always be added to
the current track (OsmAnd will not prompt ask which track to use).

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -12,6 +12,7 @@
 
 -->
 
+    <string name="always_store_waypoint_in_current_track_title">Always add track waypoints to the currently recording track</string>
     <string name="change_folder">Change folder</string>
     <string name="rename_track">Rename track</string>
     <string name="edit_track">Edit track</string>
--- a/OsmAnd/res/xml/monitoring_settings.xml
+++ b/OsmAnd/res/xml/monitoring_settings.xml
@@ -116,6 +116,17 @@
 		android:layout="@layout/simple_divider_item"
 		android:selectable="false" />
 
+	<net.osmand.plus.settings.preferences.SwitchPreferenceEx
+		android:key="always_store_waypoint_in_current_track"
+		android:layout="@layout/preference_with_descr_dialog_and_switch"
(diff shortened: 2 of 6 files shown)


OsmAnd/res/values/strings.xml
```diff
@@ -12,6 +12,7 @@
 
 -->
 
+    <string name="always_store_waypoint_in_current_track_title">Always add track waypoints to the currently recording track</string>
```
comradekingu, line 15:
```suggestion
    <string name="always_store_waypoint_in_current_track_title">Always add waypoints to current track</string>
```
?

njohnston:
@dmpr0 : thanks for the comment. I haven't had a chance to look at a nightly build, but I'm not sure if the changes will meet my requirements. When I tried the quick action tool before, it only allows creating a waypoint at the current location. I need the ability to create waypoints at arbitrary locations. Also, I don't want to have to specify what track the waypoint should be added to. I want the waypoint to be automatically added to the currently recording track.

What I want is to be able to long-press the map, tap "Add waypoint", enter a description, and then save. I don't want to be prompted for what track the waypoint should be added to.

(Like I said initially, I wanted this feature for adding many house numbers. Having to select which track to add waypoints to becomes annoying when you're adding hundreds of waypoints in a day, for example.)

vshcherb:
>  Also, I don't want to have to specify what track the waypoint should be added to. I want the waypoint to be automatically added to the currently recording track.
It's done.

vshcherb:
> (Like I said initially, I wanted this feature for adding many house numbers. Having to select which track to add waypoints to becomes annoying when you're adding hundreds of waypoints in a day, for example.)

You better use OSM Edit template which you can export as OSC file.

PR #10600, opened 2021-01-17
Fix #10517

### LINKED ISSUE #10517: Inconsistency for very small coordinates
### Description
As a grateful user, I mention a possible issue.

Consider a location such as 52.00000, -0.00090 saved as a favourite.

In the "Details" the coordinates are available to copy in multiple formats and all is well.

Choosing "Share" there are the options: Send, copy, copy location
, copy coordinates, geo, QR code. For send, copy, geo, and QR code the small coordinate is given in exponential notation: -9.0E-4. Sharing a location from one device to another, on the receiving device Osmand doesn't parse the exponential notation correctly

### Your Environment
OsmAnd Version: +3.8.4
Android/iOS version: 10

cepprice:


OsmAnd-java/src/main/java/net/osmand/util/MapUtils.java
```diff
@@ -333,8 +333,8 @@ public int compare(MapObject o1, MapObject o2) {
 		});
 	}
 
-	public static String buildGeoUrl(double latitude, double longitude, int zoom) {
-		return "geo:" + ((float) latitude) + "," + ((float) longitude) + "?z=" + zoom;
+	public static String buildGeoUrl(String latitude, String longitude, int zoom) {
```
vshcherb, line 336: previous method declaration with double as parameters is better!

PR #10601, opened 2021-01-17
Lefthand

androiddevkotlin:


OsmAnd/src/net/osmand/plus/views/mapwidgets/LanesControl.java
```diff
@@ -49,11 +51,18 @@ public LanesControl(MapActivity mapActivity, OsmandMapTileView view) {
 		lanesDrawable = new LanesDrawable(mapActivity, mapActivity.getMapView().getScaleCoefficient());
 		lanesView.setImageDrawable(lanesDrawable);
 		trackingUtilities = mapActivity.getMapViewTrackingUtilities();
-		locationProvider = mapActivity.getMyApplication().getLocationProvider();
-		settings = mapActivity.getMyApplication().getSettings();
 		mapRouteInfoMenu = mapActivity.getMapRouteInfoMenu();
-		rh = mapActivity.getMyApplication().getRoutingHelper();
```
vshcherb, line 55: ui (drawable / controls) should not have listener to settings!

vshcherb:
It was discussed many times that route will be recalculated if left / right changes , it's not UI issue! It's backend issue.

PR #10602, opened 2021-01-17
Fix #10464

### LINKED ISSUE #10464: Language packages are updated every day

cepprice:


vshcherb:
what the fix here?

Chumva:
This is only part of the fix, for "net.osmand" and "net.osmand.dev", because there was check only for "net.osmand.plus".

PR #10617, opened 2021-01-19
Fix UI trip recording

androiddevkotlin:


androiddevkotlin:
[Update trip recording dialog(Review)](https://github.com/osmandapp/OsmAnd-Issues/issues/413)

PR #10695, opened 2021-01-28
Add silent setting to import method of AIDL API 

cepprice:


OsmAnd-api/src/net/osmand/aidlapi/customization/ProfileSettingsParams.java
```diff
@@ -15,23 +15,26 @@
...
 	public static final String PROFILE_SETTINGS_URI_KEY = "profileSettingsUri";
 	private Uri profileSettingsUri;
 	private String latestChanges;
 	private int version;
 	private ArrayList<String> settingsTypeKeyList = new ArrayList<>();
-	boolean replace;
+	private boolean silent;
+	private boolean replace;
 
 	public ProfileSettingsParams(Uri profileSettingsUri, ArrayList<AExportSettingsType> settingsTypeList, boolean replace,
```
vshcherb, line 28: why do we pass ArrayList as parameter? it should be List.

PR #10739, opened 2021-02-02
Fix speed unit setting for "min/km" to show seconds 

Skalii:


OsmAnd/src/net/osmand/plus/OsmAndFormatter.java
```diff
@@ -557,11 +557,11 @@ private static String formatCoordinate(double coordinate, int outputType) {
 			sb.append(degDf.format(coordinate)).append(DELIMITER_DEGREES);
 		} else if (outputType == FORMAT_MINUTES) {
 			sb.append(minDf.format(formatCoordinate(coordinate, sb, DELIMITER_DEGREES)))
```
vshcherb, line 559: is it legit change for formatting?

PR #10750, opened 2021-02-03
Fix #5970 Make menu button active at once

### LINKED ISSUE #5970: Allow to move not yet uploaded OSM notes and POIs
It is already possible to move favorites, markers, GPX waypoints etc. However moving of newly created, not yet uploaded OSM notes and new POIs is not possible, the corresponding action is greyed out.

cepprice:


Chumva:
This was added to 3.9 here - https://github.com/osmandapp/OsmAnd/pull/10791

PR #10785, opened 2021-02-05
Fix bookmark after file update

Dima-1:


vshcherb:
Algorithm is not correct! Looks generic get article by id.
Proper algorithm:
1. If file timestamp & file name is the same, search by "route_id" inside OBF
2. If file with same name is present but timestamp is different, search by exact "route_title", "route_lang" & rough coordinates (5km)in that file
3. If no file with same name is prsent, search in all files  by exact "route_title", "route_lang" & rough coordinates (30km)

Chumva:
This was fixed here - https://github.com/osmandapp/OsmAnd/pull/10845.

PR #10813, opened 2021-02-09
Download all button implementation

nazar-kutz:


vshcherb:
I think it should be master & PR has conflicts

PR #10848, opened 2021-02-12
Implement precision result from 26 zoom

ivanPyrohivskyi:
Implement precision of 26 zoom on Android

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
+++ b/OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
@@ -714,9 +714,7 @@ public class BinaryMapPoiReaderAdapter {
 		Amenity am = null;
 		int x = 0;
 		int y = 0;
-		int preciseX = 0;
-		int preciseY = 0;
-		boolean isPrecised = false;
+		int precisionXY = 0;
 		boolean hasLocation = false;
 		StringBuilder retValue = new StringBuilder();
 		PoiCategory amenityType = null;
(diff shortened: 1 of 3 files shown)


OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
```diff
@@ -740,6 +742,16 @@ private Amenity readPoiPoint(int left31, int right31, int top31, int bottom31,
...
+						int x31 = xy[0] << 5;
+						int y31 = xy[1] << 5;
+						am.setLocation(MapUtils.get31LatitudeY(y31), MapUtils.get31LongitudeX(x31));
+					} else {
+						am.setLocation(MapUtils.get31LatitudeY(y), MapUtils.get31LongitudeX(x));
+					}
+				}
 				return am;
 			case OsmandOdb.OsmAndPoiBoxDataAtom.DX_FIELD_NUMBER:
 				x = (codedIS.readSInt32() + (px << (24 - zoom))) << 7;
```
vshcherb, line 757: shifting here to 7 and then later shifting back to 7 again, unnecessary operation

OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
```diff
@@ -740,6 +742,16 @@ private Amenity readPoiPoint(int left31, int right31, int top31, int bottom31,
 						am.setRoutePoint(arp);
 					}
 				}
+				if (hasLocation) {
+					if (precisionXY != 0) {
+						int[] xy = MapUtils.calculateFinalXYFromBaseAndPrecisionXY(24, 26, precisionXY, x >> 7, y >> 7, true);
```
vshcherb, line 747: There should be somewhere a constant of 24 and constant of 26

OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
ivanPyrohivskyi: Then need to do shift in if statement if (left31 > x || right31 < x || top31 > y || bottom31 < y) because x and y will be in 24 zoom format

ivanPyrohivskyi:
Added constants for 24 and 26 zooms and shifts

PR #10943, opened 2021-02-19
Hide "Stop without saving" button (m)

Skalii:


sonora:
I welcome that we hide this button, much safer this way! ;) 

But now of course the caption 'All unsaved data will be lost' is obsolete.

pebogufi:
If the button "Stop without saving" is removed, then there is only one functional button "Save and stop recording" remaining. 
What about moving this function to the calling window, that means:
"Stop" rename to "Save and stop recording". 
So one window can be totally removed. I would love that.

PR #11105, opened 2021-03-10
Track coloring based on speed / altitude / slope

cepprice:


cepprice:
https://github.com/osmandapp/OsmAnd-Issues/issues/510

PR #11162, opened 2021-03-16
Update strings.xml

MaartenDeen:
Small typofix in the string "Voor wandelaars, trektochten en fietsenrs in de natuur.", changed to "Voor wandelaars, trektochten en fietsen in de natuur."


Chumva:
@MaartenDeen You can edit translations in weblate https://hosted.weblate.org/projects/osmand/main/

MaartenDeen:
Thanks for the pointer, wasn't aware of that.

PR #11176, opened 2021-03-17
Fix space after "Next time" in strings.xml

aceman444:
Add probably missing space after "Next time" that all translators have added into their localized strings.


Chumva:
@aceman444 You can edit translations in weblate https://hosted.weblate.org/projects/osmand/main/

aceman444:
@Chumva, of course I do for the translations, but this looked like a bug in the source English string. I don't think you can edit the en-US original string on weblate.

PR #11251, opened 2021-03-25
fix replace getNonConditionalTag

Dima-1:


OsmAnd-java/src/main/java/net/osmand/binary/RouteDataObject.java
```diff
@@ -583,8 +583,8 @@ public void processConditionalTags(long conditionalTime) {
 									String nonCondTag = rtr.getTag();
 									int ks;
 									for (ks = 0; ks < pointTypes[i].length; ks++) {
-										RouteTypeRule toReplace = region.quickGetEncodingRule(pointTypes[i][j]);
-										if (toReplace != null && toReplace.getTag().contentEquals(nonCondTag)) {
```
vshcherb, line 587: 1. j -> ks - OK
NOT OK: This line is correct - toReplace.getTag().contentEquals(nonCondTag) 

2. C++ change?

Dima-1: C++ no changes needed

PR #11266, opened 2021-03-26
Fix slider when RTL

cepprice:


Chumva:
It doesn`t work with the force RTL option in English. Needs to be fixed in a separate issue.

PR #11349, opened 2021-04-05
Purchases screen UI fixes (without two checkboxes for now)

cepprice:


cepprice:
https://github.com/osmandapp/OsmAnd-Issues/issues/496

PR #11456, opened 2021-04-18
New UI

androiddevkotlin:
Navigation settings > Voice prompts
0. New UI

    Rename the Voice guidance menu item to Language.
    Language dialog changes:
    • Add toggle between "TTS | Recorded"
    • Remove "Install more" item and download "Recorded" voice packages without leaving the dialog.
    • Delete the "Don't use" option, because currently, we have the main switch on the "Voice prompts" screen in Navigation settings.
    • Add option "Use system language", detect system language (we do similar in Wikipedia, where we get system languages and put them at the top of the list), and use it by default.


androiddevkotlin:
Fix two issue: [Redesign](https://github.com/osmandapp/OsmAnd-Issues/issues/130) & [Spillover from Voice Prompts](https://github.com/osmandapp/OsmAnd-Issues/issues/566)

PR #11465, opened 2021-04-19
Purchases screen review

cepprice:


OsmAnd/src/net/osmand/plus/settings/fragments/SubscriptionsCard.java
```diff
@@ -100,23 +100,23 @@ public void updateSubscriptionsListCard() {
 	}
 
 	private void setupSupportRegion() {
```
vshcherb, line 102: Don't comment code only delete

OsmAnd/res/layout/subscriptions_card.xml
```diff
@@ -21,69 +21,69 @@
 		android:layout_width="match_parent"
 		android:layout_height="wrap_content" />
 
-	<LinearLayout
```
vshcherb, line 24: code needs to be deleted

Issue #11398, opened 2021-04-11
Untranslated Travel guides (wikivoyage) file names


aceman444:
When you go into the "Travel guides (beta)" menuitem in main menu, you get offered a list of files to download. Those wikivoyage files are named after continents, like Europe, Australia-oceania etc. These names are NOT translated, while I think they could, as the same regions are already translated in the main Map download dialog, so the strings should be available.

OsmAnd 3.9.10

PR #11468, opened 2021-04-19
Untranslated Travel guides (wikivoyage) file names

androiddevkotlin:
https://github.com/osmandapp/OsmAnd/issues/11398


aceman444:
Thanks, I will test it in a stable release.

PR #11498, opened 2021-04-22
Fix 393

alisa911:


OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
```diff
@@ -52,13 +53,27 @@ public double getUnknownPhraseMatchWeight() {
...
-		int localWordsMatched = alternateName != null ? 
-				requiredSearchPhrase.countWords(alternateName)  : requiredSearchPhrase.countWords(localeName) ;
- 		boolean match = localWordsMatched <= getSelfWordCount();
-		double res = ObjectType.getTypeWeight(match ? objectType : null);
+		String name = alternateName != null ? alternateName : localeName;
+		boolean fullCompleteMatch = false;
+		double res = 1;
+		if (getSelfWordCount() >= 1) {
+			res = ObjectType.getTypeWeight(objectType);
+			String preparedName = StringUtils.stripAccents(name.toLowerCase());
```
vshcherb, line 64: wrong

OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
```diff
@@ -37,28 +38,63 @@
...
- 		boolean match = localWordsMatched <= getSelfWordCount();
-		double res = ObjectType.getTypeWeight(match ? objectType : null);
+		String name = alternateName != null ? alternateName : localeName;
+		List<String> localResultNames = new ArrayList<>();
+		List<String> searchPhraseNames = new ArrayList<>();
+		SearchPhrase.splitWords(name, localResultNames);
+
+		String fw = requiredSearchPhrase.getFirstUnknownSearchWord();
+		List<String> ow = requiredSearchPhrase.getUnknownSearchWords();
+		if (fw != null) {
```
vshcherb, line 73: No check if (firstUnknownWordMatches) { + how it could be null?

OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
```diff
@@ -37,28 +38,63 @@
...
+			searchPhraseNames.add(fw);
+		}
+		if (!ow.isEmpty()) {
```
vshcherb, line 76: ow should be check if it's not null! not is empty

PR #11528, opened 2021-04-26
fix search 393

alisa911:


OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
```diff
@@ -127,6 +134,17 @@ private int getSelfWordCount() {
 		return inc;
 	}
 
+	private boolean checkPhraseIsObjectType(String searchPhraseName) {
+		if (object instanceof Amenity) {
```
vshcherb, line 138: not correct implementation cause it doesn't cover all cases of search should be refatored

PR #11549, opened 2021-04-29
T10889

alisa911:


OsmAnd-java/src/main/java/net/osmand/binary/Abbreviations.java
```diff
@@ -0,0 +1,43 @@
...
+        abbreviations.put("n", "North");
+        abbreviations.put("sw", "Southwest");
+        abbreviations.put("se", "Southeast");
+        abbreviations.put("nw", "Northwest");
+        abbreviations.put("ne", "Northeast");
+        abbreviations.put("ln", "Lane");
+    }
+
+    public static String replace(String word) {
+        return abbreviations.containsKey(word.toLowerCase()) ?
```
vshcherb, line 31: ышьзду пуе шы утщгпр

OsmAnd-java/src/main/java/net/osmand/binary/Abbreviations.java
```diff
@@ -0,0 +1,43 @@
...
+                abbreviations.get(word.toLowerCase()) : word;
+    }
+
+    public static String replaceAll(String phrase) {
```
vshcherb, line 35: not used = not needed

OsmAnd-java/src/main/java/net/osmand/binary/Abbreviations.java
vshcherb: no need to call 2 times containsKey + get: 1 get is enough!

OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
```diff
@@ -219,15 +211,14 @@ private SearchPhrase createNewSearchPhrase(SearchSettings settings, String fullT
 				boolean lastAndIncomplete = i == ws.length - 1 && !sp.lastUnknownSearchWordComplete;
 				if (wd.length() > 0 && (!conjunction || lastAndIncomplete)) {
 					if (first) {
-						sp.firstUnknownSearchWord = wd;
+						sp.firstUnknownSearchWord = Abbreviations.replace(wd.toLowerCase());
```
vshcherb, line 214: double lowercase here and in abbrevations here is not needed

OsmAnd-java/src/main/java/net/osmand/binary/Abbreviations.java
```diff
@@ -0,0 +1,43 @@
...
+        String[] words = phrase.split(DELIMITER);
```
vshcherb, line 36: if it's not changed phrase should not be updated at all!

OsmAnd-java/src/main/java/net/osmand/binary/Abbreviations.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public static String replace(String word) {
+        String value = abbreviations.get(word.toLowerCase());
+        return value != null ? value : word;
+    }
+
+    public static String replaceAll(String phrase) {
+        String[] words = phrase.split(DELIMITER);
+        ArrayList<String> result = new ArrayList<>();
```
vshcherb, line 37: Variable declration List<String>

OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
```diff
@@ -42,14 +42,7 @@ public static int getCommonGeocoding(String name) {
 	
 	static {
 		addFrequentlyUsed("santa");
-		addFrequentlyUsed("west");
```
vshcherb, line 45: not sure it can be deleted, doesn't look correct

OsmAnd-java/src/main/java/net/osmand/binary/Abbreviations.java
```diff
@@ -0,0 +1,44 @@
...
+        for (String word : words) {
+            result.add(replace(word));
+        }
+        String resultPhrase = StringUtils.join(result, DELIMITER);
```
vshcherb, line 41: we don't use stringutils in backend

PR #11553, opened 2021-04-29
Fix #10393 navigation incorrect on imported gpx-files

### LINKED ISSUE #10393: navigation incorrect on imported gpx-files
Osmand reroutes the predefined route of an imported gpx file. If the hiking path isn't available in Open Street Map, Osmand reroutes via an alternative path. This meshed up the navigation completely.
see attached schreenshots in Osmand. 

The route in Komoot browser shows (including warnings) shows where Osmand recalculates the route. 

### How to reproduce?
Import a gpx file in Osmand (I use Google drive), selecting the gpx-file in Google Drives opens Osmand

Start navigation
click on setting
Follow track
select imported GPX file
Start routing
The navigation takes a different route as the gpx file (orange is the gpx-track)

### Your Environment
OsmAnd Version: 3.8.4
Android/iOS version: 8.0
Device model: Samsung Galaxy S7

**Maps used (online or offline):**   
offline
Screenshot_20201213-164113_OsmAnd
Screenshot_20201213-165121_OsmAnd
complete route in Komoot
warnings in komoot

Dima-1:


OsmAnd/src/net/osmand/plus/routing/GPXRouteParams.java
```diff
@@ -140,6 +142,7 @@ public GPXRouteParams prepareGPXFile(GPXRouteParamsBuilder builder) {
 		private boolean leftSide;
 		private boolean passWholeRoute;
 		private boolean calculateOsmAndRouteParts;
+		private boolean followTrack;
 		private int selectedSegment = -1;
```
vshcherb, line 146: doesn't look correct,  all params should be for followTrack

OsmAnd/src/net/osmand/plus/routing/RouteProvider.java
```diff
@@ -260,11 +260,11 @@ private RouteCalculationResult calculateGpxRoute(RouteCalculationParams routePar
 					routeParams.intermediates, routeParams.ctx, routeParams.leftSide, null, null, routeParams.mode, true);
 		}
 
-		if (routeParams.gpxRoute.useIntermediatePointsRTE) {
+		if (routeParams.gpxRoute.useIntermediatePointsRTE && !routeParams.gpxRoute.followTrack) {
```
vshcherb, line 263: doesn't look logical, rte should not be used when track is present itself but here there is no such condition at all

PR #11650, opened 2021-05-11
Add public transport tags

Dima-1:


OsmAnd/src/net/osmand/plus/routepreparationmenu/cards/PublicTransportCard.java
```diff
@@ -280,6 +294,11 @@ private SpannableString getSecondLineDescrSpan(List<TransportRouteResultSegment>
 		return secondLineDesc;
 	}
 
+	private String getIntervalDescr(List<TransportRouteResultSegment> segments) {
+		String interval = Algorithms.capitalizeFirstLetter(app.getString(R.string.shared_string_interval));
+		return interval + " " + segments.get(0).route.getInterval();
```
vshcherb, line 299: crash on get(0) ?

OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapTransportReaderAdapter.java
```diff
@@ -502,34 +519,51 @@ private void readTransportSchedule(TransportSchedule schedule) throws IOExceptio
...
+
+	private Map<String, String> initializeTags(TIntObjectHashMap<String> stringTable,
+	                                           net.osmand.data.TransportRoute dataObject) {
+		Map<String, String> newMap = new HashMap<>();
+		for (Map.Entry<String, String> entry : dataObject.getTags().entrySet()) {
+			String string = stringTable.get(entry.getKey().charAt(0));
+			if (entry.getValue().length() > 0) {
+				newMap.put(string, entry.getValue());
+			} else {
+				newMap.put(string.split("/")[0], string.split("/")[1]);
```
vshcherb, line 559: no checks nothing... crashes left & right!!!

vshcherb:
```
int s = string.indexOf('/');
if ( s > 0 ) {
   newMap.put(string.substring(0, s), string.substring(s + 1));
}
```

PR #11723, opened 2021-05-20
Fix huge slope

cepprice:


OsmAnd/src/net/osmand/plus/track/GradientCard.java
```diff
@@ -22,8 +22,8 @@
 	private final GPXTrackAnalysis gpxTrackAnalysis;
 	private GradientScaleType selectedScaleType;
 
-	private final int minSlope = 0;
-	private final int maxSlope = 60;
+	private static final float minSlope = 0;
```
alex-osm, line 25: Constant name should be MIN_SLOPE

PR #11748, opened 2021-05-24
Make obf tracks clickable

cepprice:


OsmAnd-java/src/main/java/net/osmand/IndexConstants.java
```diff
@@ -63,6 +63,7 @@
 	public static final String MAP_MARKERS_INDEX_DIR = "/map markers";
 	public static final String GPX_RECORDED_INDEX_DIR = GPX_INDEX_DIR + "rec/";
 	public static final String GPX_IMPORT_DIR = GPX_INDEX_DIR + "import/";
+	public static final String GPX_FROM_OBF_DIR = GPX_INDEX_DIR + "obf/";
```
vshcherb, line 66: bad name, doesn't mean anything for user. it's better to use "travel" cause they are part of it or "maps" if travel doesn't work

OsmAnd/src/net/osmand/plus/views/layers/GPXLayer.java
```diff
@@ -999,6 +1018,70 @@ private SelectedGpxPoint createSelectedGpxPoint(SelectedGpxFile selectedGpxFile,
 		return new SelectedGpxPoint(selectedGpxFile, projectionPoint, prevPoint, nextPoint, bearing);
 	}
 
+	private GPXFile createGpxFileFromRenderedObject(NativeLibrary.RenderedObject object) {
```
vshcherb, line 1021: Too much code in 1 class > 1000 lines, should be redesigned

OsmAnd/src/net/osmand/plus/views/layers/GPXLayer.java
```diff
@@ -999,6 +1018,70 @@ private SelectedGpxPoint createSelectedGpxPoint(SelectedGpxFile selectedGpxFile,
...
+		OsmandApplication app = view.getApplication();
+		GPXFile gpxFile = new GPXFile(Version.getFullVersion(app));
+		String fileName = object.getFileNameByExtension(IndexConstants.GPX_FILE_EXT);
+		gpxFile.path = app.getAppPath(IndexConstants.GPX_FROM_OBF_DIR).getPath() + "/" + fileName;
+		GPXUtilities.Track track = new GPXUtilities.Track();
+		String refValue = object.getTagValue(TravelObfHelper.REF_TAG);
+
+		List<BinaryMapDataObject> segmentsObjects = new ArrayList<>();
```
vshcherb, line 1029: same code was part of travelobfhelper, so it should be there probably and combined with existing code

Issue #11683, opened 2021-05-14
Cannot add waypoint via track menu


Nine-Tailed:
### Description
It is not possible to add a waypoint through the track menu if at least one waypoint has not been previously added to it in another way.
![image](https://user-images.githubusercontent.com/12549744/118321062-bf11a280-b505-11eb-924e-0d4ce2320c0e.png)

After adding a waypoint to the track via quick actions, it becomes possible to add waypoints.
![image](https://user-images.githubusercontent.com/12549744/118321629-86be9400-b506-11eb-86d2-efe63acf95e2.png)

### Your Environment
OsmAnd Version: 4.0.0#32000m
Android/iOS version: 7.0
Device model: xiaomi redmi note 4

PR #11755, opened 2021-05-25
Add waypoint track menu

androiddevkotlin:
https://github.com/osmandapp/OsmAnd/issues/11683


OsmAnd/src/net/osmand/plus/track/TrackPointsCard.java
```diff
@@ -270,6 +278,7 @@ public void onPointsDeletionStarted() {
 	@Override
 	public void onPointsDeleted() {
 		updateGroups();
+		update();
```
Chumva, line 281: It calls adapter.synchronizeGroups() two times, firstly in updateGroups secondly in updateContent(). You can change updateGroups to clearSelectedGroups

PR #11756, opened 2021-05-25
add skip abbreviations

alisa911:


OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
```diff
@@ -209,19 +212,56 @@ private SearchPhrase createNewSearchPhrase(SearchSettings settings, String fullT
...
+						sp.otherUnknownWords.add(decryptAbbreviations ? Abbreviations.replace(wd) : wd);
 					}
 				}
 			}
 		}
 		return sp;
 	}
-	
+
+	private List<WorldRegion> getWorldRegions() {
```
vshcherb, line 230: Doesn't look good won't work on Android, it works differently probably WorldRegion should be passed by Android

PR #11760, opened 2021-05-25
Remove limitation by location for street search

ivanPyrohivskyi:


OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -419,8 +419,9 @@ public boolean publish(MapObject object) {
 						int x = MapUtils.get31TileNumberX(object.getLocation().getLongitude());
 						List<City> closestCities = null;
 						if (object instanceof Street) {
-							if ((locSpecified && !streetBbox.contains(x, y, x, y))
-									|| !phrase.isSearchTypeAllowed(ObjectType.STREET)) {
+							// remove limitation by location
```
vshcherb, line 422: in the code we use single line comments (so it's easier to read and comment blocks)

PR #11969, opened 2021-06-21
Backup promo

Chumva:


OsmAnd/src/net/osmand/plus/backup/ui/AuthorizeFragment.java
```diff
@@ -363,6 +394,11 @@ public void onRegisterUser(int status, @Nullable String message, @Nullable Strin
 		});
 	}
 
+	private boolean promoCodeSupported() {
+		return dialogType == LoginDialogType.SIGN_UP
+				&& Version.isDeveloperVersion(app) || !Version.isProductionVersion(app);
```
vshcherb, line 399: no brackets...
It should be checked if there is no Google Play support here

PR #11993, opened 2021-06-24
Fix_11904

ivanPyrohivskyi:


OsmAnd/src/net/osmand/plus/measurementtool/MeasurementToolFragment.java
```diff
@@ -2173,6 +2173,12 @@ private void startTrackNavigation() {
 			if (editingCtx.hasRoute() || editingCtx.hasChanges()) {
 				String trackName = getSuggestedFileName();
 				GPXFile gpx = editingCtx.exportGpx(trackName);
+				// come back wpt point to gpx after approximation
```
vshcherb, line 2176: Probably it's not the only place or not correct cause Attach to Roads could be called from Plan Route (MeasureFramgent) & Navigation.

OsmAnd-java/src/main/java/net/osmand/router/RouteExporter.java
```diff
@@ -46,17 +46,12 @@ public GPXFile exportRoute() {
 		return gpx;
 	}
 
-	public static GPXFile exportRoute(String name, List<TrkSegment> trkSegments, List<WptPt> points) {
+	public static GPXFile exportRoute(String name, List<TrkSegment> trkSegments) {
```
vshcherb, line 49: it should belong here or points should stay

Issue #3289, opened 2016-11-22
TTS destination:ref should use toward


hakuchi:
`destination:ref` should use `toward` and not `onto`

example:
https://www.openstreetmap.org/way/145035982

turn slightly right **onto** A99 toward Nürnberg
~but it should be:~
~turn slightly right **toward** A99 toward Nürnberg~

turn right on way with destination = toward
turn right on way with name/ref = onto

Tested on OsmAnd 2.5.4 and Nightly 2.5.0#15419D with latest Bavaria map.

sonora:
Theoretically correct. But in practice: I happened to drive that section some time ago myself, and I found no unusual or misleading prompt there. The section to turn onto (like most which are labeled with a destination:ref) is the direct link to A99, and it is also labeled accordingly, I think there is hardly any room for confusion ?

In our code, from what I remember, we only pass 3 different parameters to the voice engine: One is geared at containing a ref, the second a destination, and the 3rd a street name. The grammar for these 3 types may differ significantly in several languages. and this is why we pass destination:ref, being type ref, to the first field, and not to the second. (Resulting in the grammatical treatment you observe.)

We only use destination:ref as a fallback for if there is no ref present, anyway. And with the current practice we automatically ensure that we do not result in awkwardly announcing 2 different refs. 

Looking at all this, I am afraid there is no easy improvement which would not potentially break several other pieces of long-tested logic and grammar....

hakuchi:
what about combine the two keys "destination: ref" and "destination"

from:
turn slightly right **_onto_** A99 **_toward_** Nürnberg
to:
turn slightly right **toward** A99, Nürnberg

Can this handle the map creator by converting `destination:ref=A99` `destination=Nürnberg` to `destination=A99; Nürnberg`?

So this works also on streets with "ref", "destination:ref" and "destination".

Examples:
http://www.openstreetmap.org/way/31957655#map=15/45.8734/10.9004

On this roundabout exit, the voice guidance will be:
take second exit **_onto_** SS240 **_toward_** A22, Mori, Rovereto

2th:
www.openstreetmap.org/way/370563044
On this roundabout exit, the voice guidance will be:
take second exit **_onto_** N 4 **_toward_** B 2, A 73, Würzburg, Suhl, Fürth, Centrum

sonora:
[...]

PR #12005, opened 2021-06-25
Add destination:lanes to navigation. Set street name from motorway_ju…

ivanPyrohivskyi:
…nction if destination was empty

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/binary/RouteDataObject.java
+++ b/OsmAnd-java/src/main/java/net/osmand/binary/RouteDataObject.java
@@ -368,7 +368,11 @@ public class RouteDataObject {
 				destinationTagLang = "destination:lang:" + lang;
 			}
 			String destinationTagDefault = "destination";
+			String destinationLanesTag = "destination:lanes";
+			String destinationLanesTagFB = (direction == true) ? destinationLanesTag + ":forward" : destinationLanesTag + ":backward";
 			String destinationDefault = null;
+			String destinationLanesDefault = null;
+			String destinationLanesFBDefault = null;
 
 			for (int i = 0; i < kt.length; i++) {
 				int k = kt[i];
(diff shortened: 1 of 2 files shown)


OsmAnd-java/src/main/java/net/osmand/binary/RouteDataObject.java
```diff
@@ -368,7 +368,11 @@ public String getDestinationName(String lang, boolean transliterate, boolean dir
 				destinationTagLang = "destination:lang:" + lang;
```
vshcherb, line 368: The code here looks strange and probably is wrong (destination:ref needs to be ref) otherwise shield is not correctly displayed probably:
		//Issue #3289: Treat destination:ref like a destination, not like a ref

OsmAnd/src/net/osmand/plus/routing/CurrentStreetName.java
```diff
@@ -54,6 +54,10 @@ public static CurrentStreetName getCurrentName(@NonNull RoutingHelper routingHel
 			if (n.directionInfo.getExitInfo() != null) {
 				// don't display name of exit street name
 				streetName.exitRef = n.directionInfo.getExitInfo().getRef();
+				if (!isSet && !Algorithms.isEmpty(n.directionInfo.getExitInfo().getExitStreetName())) {
```
vshcherb, line 57: this is not needed and probably not correct

OsmAnd-java/src/main/java/net/osmand/binary/RouteDataObject.java
```diff
@@ -368,7 +368,11 @@ public String getDestinationName(String lang, boolean transliterate, boolean dir
...
 			}
 			String destinationTagDefault = "destination";
+			String destinationLanesTag = "destination:lanes";
```
vshcherb, line 371: Original idea, not to use destination:lanes but to use only destination on next roads.  Check code here https://github.com/osmandapp/OsmAnd/blob/8ce37e4d8fcfad83b9127f5c24ab089d483e12e3/OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java#L404

PR #12006, opened 2021-06-25
Remove unnecessary continue straight

Dima-1:


OsmAnd/src/net/osmand/plus/routepreparationmenu/cards/RouteDirectionsCard.java
```diff
@@ -43,6 +45,7 @@ protected void updateContent() {
 
 	private void createRouteDirections(LinearLayout cardsContainer) {
 		List<RouteDirectionInfo> routeDirections = app.getRoutingHelper().getRouteDirections();
+		removeUnnecessaryGoAhead(routeDirections);
```
vshcherb, line 48: Place to fix is very awkward of course. Also the bug is about our routing i.e. Attach to roads, so we should not produce unnecessary go ahead in the first place!

PR #12058, opened 2021-07-05
New route coloring types test

vshcherb:


OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -317,8 +317,8 @@ private void calculateTimeSpeed(RoutingContext ctx, List<RouteSegmentResult> res
 
 			for (int j = rr.getStartPointIndex(); j != rr.getEndPointIndex(); j = next) {
 				next = plus ? j + 1 : j - 1;
-				double d = measuredDist(road.getPoint31XTile(j), road.getPoint31YTile(j), road.getPoint31XTile(next),
-						road.getPoint31YTile(next));
+				double d = MapUtils.measuredDist31(road.getPoint31XTile(j), road.getPoint31YTile(j),
```
vshcherb, line 320: need to be revert

PR #12059, opened 2021-07-05
Fix shield view

ivanPyrohivskyi:


OsmAnd/src/net/osmand/plus/views/mapwidgets/MapInfoWidgetsFactory.java
```diff
@@ -964,6 +964,7 @@ public boolean updateInfo(DrawSettings d) {
 				if (streetName.shieldObject != null && streetName.shieldObject.nameIds != null
 						&& setRoadShield(shieldIcon, streetName.shieldObject)) {
 					AndroidUiHelper.updateVisibility(shieldIcon, true);
+					streetName.text = streetName.text.substring(streetName.text.indexOf("»"));
```
vshcherb, line 967: -1 check could be good

ivanPyrohivskyi: Fixed

PR #12095, opened 2021-07-12
New strings reworked, old changelogs unlisted

comradekingu:
Changed all changelogs to the oldest format. Some in between were different.
Should ideally all be moved to a Fastlane file.


OsmAnd/res/values/strings.xml
```diff
@@ -776,7 +776,7 @@
     <string name="route_recalculation_dist_descr">The route will be recalculated if the distance to the route is longer than specified parameter</string>
     <string name="routing_profile_direct_to">Direct-to-point</string>
     <string name="clear_recorded_data">Clear recorded data</string>
-    <string name="release_3_6">
+    <-- !string name="release_3_6">
```
comradekingu:
```suggestion
    <! --string name="release_3_6">
```

OsmAnd/res/values/strings.xml
```diff
@@ -620,7 +620,7 @@
     <string name="parking_positions">Parking positions</string>
     <string name="create_edit_poi">Create or edit POI</string>
     <string name="quick_action_transport_descr">Button showing or hiding public transport on the map.</string>
-    <string name="release_3_7">
+    <!-- string name="release_3_7">
```
comradekingu, line 623:
```suggestion
    <! --string name="release_3_7">
```

OsmAnd/res/values/strings.xml
```diff
@@ -776,7 +776,7 @@
     <string name="route_recalculation_dist_descr">The route will be recalculated if the distance to the route is longer than specified parameter</string>
     <string name="routing_profile_direct_to">Direct-to-point</string>
     <string name="clear_recorded_data">Clear recorded data</string>
-    <string name="release_3_6">
+    <!-- string name="release_3_6">
```
comradekingu, line 779:
```suggestion
    <! -- string name="release_3_6">
```

sonora:
Linguisticallyf ine by me.

I will leave the paragraph/line space decision for the change log to @vshcherb :Without explicit paragraph spacing this will appear as a rather crammed list with less readability, we may not want this on a screen which offers enough space anyway.

comradekingu:
@sonora Typographically doing line-spacing like that is worse. I am not very for or against, just forwarding feedback mostly.

OsmAnd/res/values/strings.xml
```diff
@@ -14,35 +14,35 @@
...
+    <string name="online_maps_required_descr">Could not find a route suited to your settings.\n\nPerhaps some maps are not loaded. You can run online checking of the necessary maps.</string>
+    <string name="offline_maps_required_descr">More offline maps needed to calculate the route.</string>
     <string name="online_direct_line_maps_link">Check the necessary maps</string>
-    <string name="direct_line_maps_required_descr">Route calculation takes longer than usual. \n\nPerhaps some maps are not loaded.</string>
-    <string name="setting_supported_by_style">This setting is supported by map style \'%1$s\'</string>
+    <string name="direct_line_maps_required_descr">Route calculation is taking longer than usual. \n\nPerhaps some maps are not loaded.</string>
+    <string name="setting_supported_by_style">This setting is supported by the \'%1$s\' map style</string>
     <string name="travel_routes">Travel routes</string>
-    <string name="walking_route_osmc_description">Assign color for routes based on their individually locally assigned color (if present on OpenStreetMap)  and type (international, regional, local)</string>
```
comradekingu, line 37: "Based on" caught by @websites on Weblate.

comradekingu:
@sonora 
Can we agree that adhering to the changelog line-spacing that other projects use is a smaller issue than fixing all the errors?
(If not I can edit it out.)

sonora:
Hi @comradekingu, as I said on July 12: Reviewed by me, fine by me, I would merge everything you had in this PR until that day. Except the removal of the extra line feeds between the changelog bullets.
- Looks like you drew no favorable response on removing these line feeds.
- I would still be glad to merge the July 12 package if you remove your line feed suggestion for now.

For any changes you added to this package afterwards: Can you please open a new PR. It needs a new review, and that is easier in small portions, why blow this one up and necessitate the re-review.

comradekingu:
@sonora Because in the scope of time it took to gauge whether the changelogs are in fact too drawn out, another set of totally unreviewed changes made their way in.
For "New strings reworked" to apply, those have to be added.
Are we somehow pretending changes I make in so doing are in need of reviewing as a whole?
Spare 30 seconds to review as many strings.

sonora:
I have used most of your suggestions in https://github.com/osmandapp/OsmAnd/commit/5db056753eb6d4ddb9bc8636e7c58bc5df9040a6, Thanks!

- I have not touched the changelog strings, needs checking if they are still referenced in the code which would cause errors.

- Regarding `reset_to_original`: there is already a string `reset_to_default`, so setting them both to value "Default" seems counterproductive. Should investigate where the new one is used and if it could be replaced entirely by `reset_to_default`.

Issue #12097, opened 2021-07-12
Trip/Track recording: automatcally start new segment on power status change


blacklion:
# 🚀 feature request

### Description

It is very useful to have different segments of track record (GPX) when travel on vehicle & do multiple radial detours by feet. 

Now you must remember to click "Start new segment" each time you leave vehicle and return to it.

Device with OsmAnd is powered vy external source (USB charger) in vehicle and lives on its own battery when user is out of vehicle. 

It will be very nice to have option in Track Recorder plugin to automatically start new segment when device power status (Charging / On battery) is changing (in both directions). It could produce artificial segments sometimes, but it is better than forgotten segment boundaries, IMHO.

### Describe the solution you'd like

Monitor ACTION_POWER_CONNECTED / ACTION_POWER_DISCONNECTED notifications and optionally start new track segment on each of it.

blacklion:
BTW, "Nightly" build has new Recording control interface, which hide "Start new segment" after additional click, so this feature becomes more useful than on "Release" (Play Market) version.

vshcherb:
Unfortunately we don't see much demand for this option yet to make it mainstream. Probably you would be interested to have a 3rd party plugin and control new segment / new track from separate application.

blacklion:
> Unfortunately we don't see much demand for this option yet to make it mainstream.

What is the problem to add feature where there is pull-request?

I understand, that you can be not interesting to invest developer's time to implement this from scratch, as there are a lot of more urgent issues, it is understandable.

blacklion:
> Probably you would be interested to have a 3rd party plugin and control new segment / new track from separate application.

As far as I understand (after reading  #2173 and #5525) it is impossible to make full-featured 3rd party plugin for OsmAnd+ (I have subscription and want to use it in the future, I don't want to build my own custom build). Am I right?

Code in Osmand/plugins doen't look like plugins :)

vshcherb:
No you are not right. https://github.com/osmandapp/osmand-api-demo/tree/master/OsmAnd-api-sample (https://github.com/osmandapp/osmand-api-demo/tree/master/). Also OsmAnd Telegram is fully done via  plugin API. 

Target app conects via AIDL protocol and controls OsmAnd.

blacklion:
I can not see "Start new segment" command. Also, additional app is again about «not forget to start», and additional window in carousel :-(

Could you say, what is wring with my PR? :)

vshcherb:
[...]

PR #12108, opened 2021-07-14
Add setting to enable track splitting on power status change.

blacklion:
This is implementation of #12097.
This change doesn't have translations for two new strings.


blacklion:
I'm not sure about code style, sorry. I've tried to mimic existing one, for example I didn't use `final` fields as they are not used in "old" code.

Also, I'm not sure, should new non-translated strings be added as-is to non-English resources.

vshcherb:
See #12097

OsmAnd/res/values/strings.xml
```diff
@@ -1866,6 +1866,8 @@
 	</string -->
     <string name="auto_split_recording_title">Auto-split recordings after gap</string>
     <string name="auto_split_recording_descr">Start new segment after gap of 6 min, new track after gap of 2 h, or new file after a longer gap if the date has changed.</string>
+    <string name="powerchange_split_recording_title">Auto-split recordings on power status change.</string>
+    <string name="powerchange_split_recording_descr">Start new segment when power status changes between AC and battery and back.</string>
```
comradekingu, line 1870:
```suggestion
    <string name="powerchange_split_recording_descr">Start new segment upon charging, and discharging.</string>
```

OsmAnd/res/values/strings.xml
```diff
@@ -1866,6 +1866,8 @@
 	</string -->
     <string name="auto_split_recording_title">Auto-split recordings after gap</string>
     <string name="auto_split_recording_descr">Start new segment after gap of 6 min, new track after gap of 2 h, or new file after a longer gap if the date has changed.</string>
+    <string name="powerchange_split_recording_title">Auto-split recordings on power status change.</string>
```
comradekingu, line 1869:
```suggestion
    <string name="powerchange_split_recording_title">Auto-split recordings with power status.</string>
```

PR #12311, opened 2021-07-29
Fix long deviation for motorway exit

Dima-1:


OsmAnd-java/src/main/java/net/osmand/router/GeneralRouter.java
```diff
@@ -484,8 +484,13 @@ public float getPenaltyTransition(RouteDataObject road) {
 	public float defineRoutingSpeed(RouteDataObject road) {
 		Float definedSpd = getCache(RouteDataObjectAttribute.ROAD_SPEED, road);
 		if (definedSpd == null) {
-			float spd = getObjContext(RouteDataObjectAttribute.ROAD_SPEED).evaluateFloat(road, defaultSpeed);
-			definedSpd = Math.max(Math.min(spd, maxSpeed), minSpeed);
```
vshcherb, line 488: this is incorrect

PR #12712, opened 2021-09-07
New Routing: Horsebackriding

derfuessl:
Translation of the parameters into German


vshcherb:
@Chumva 
We also will need to make this routing  default to Horse riding profile

derfuessl:
Can I help in any way?

vshcherb:
Well new profile similar to motorbike should be created and routing should be assigned.

vshcherb:
+ We need to add new Base profile for Horse riding (disabled by default)

cepprice:
https://github.com/osmandapp/OsmAnd/pull/new/horse_riding

PR #12793, opened 2021-09-17
Draw Mapillary vector tiles on 14-16 zooms

cepprice:


OsmAnd/src/net/osmand/plus/resources/GeometryTilesCache.java
```diff
@@ -12,16 +12,24 @@
 
 public class GeometryTilesCache extends TilesCache<GeometryTile> {
 
-	public GeometryTilesCache(AsyncLoadingThread asyncLoadingThread) {
+	private final int tileZoom;
```
alex-osm, line 15: Why tileZoom introduced?

OsmAnd/src/net/osmand/plus/resources/ResourceManager.java
```diff
@@ -250,9 +253,12 @@ public ResourceManager(OsmandApplication context) {
 		this.renderer = new MapRenderRepositories(context);
 
 		bitmapTilesCache = new BitmapTilesCache(asyncLoadingThread);
-		geometryTilesCache = new GeometryTilesCache(asyncLoadingThread);
+		mapillarySequenceLayerTilesCache = new GeometryTilesCache(asyncLoadingThread, MAX_SEQUENCE_LAYER_ZOOM);
```
alex-osm, line 256: Looks like we need single cache for vector tiles like we have for raster.

PR #12850, opened 2021-09-24
Fix negative time in motion

cepprice:


sonora:
Do we undrstand when/why this could be negative? Surely, just removing the sign does not make this a valid solution?

cepprice:
@sonora In track, attached to issue, timestamp of previous point > timestamp of next point (for each pair of neighbour points). As if all trkpts of segment inside GPX-file were reversed

vshcherb:
Please attach the gpx file here if possible

cepprice:
@vshcherb [2021-03-21_14-40_Sun.1.gpx.zip](https://github.com/osmandapp/OsmAnd/files/7235337/2021-03-21_14-40_Sun.1.gpx.zip)

vshcherb:
Probably we shoud not fix it like that or probably should but original issue is different.

OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
```diff
@@ -848,7 +848,8 @@ public GPXTrackAnalysis prepareInformation(long filestamp, SplitSegment... split
 						totalDistance += calculations[0];
 						segmentDistance += calculations[0];
 						point.distance = segmentDistance;
-						timeDiff = (int)((point.time - prev.time) / 1000);
+						long timeDiffMillis = Math.abs(point.time - prev.time);
```
vshcherb, line 851: Probably here we shoud use Math.max(0, point.time - prev.time)

Otherwise we could get ridiculous numbers... 
Or probably even better to have prevTime variable equals prevTime = Math.max(point.time, prevTime);

sonora:
Yes, that solution sounds better.

But I am still baffled how this has occurred in the first place and cannot reproduce. The file says creator="OsmAndRouterV2" which should mean it is somehow an exported route and not a recording. But I remember we had some issues with the XML namespace in early 2021, maybe this is a relic of a bug long gone?

Maybe we should not fix unless we find mire occurrences or can reproduce and find the root cause?

vshcherb:
Well it's probably wrong reverse + manual dismangle using Plan Route functionality. And probably it's a combination of 2 tracks, so the story is weird...

Issue #8108, opened 2019-12-18
Maxspeed in speed_camera alert


RussianFox:
Pls, add maxspeed in speed_camera alert if exist

lucumon:
It would be useful to also get distance from the voice prompt, as apps like "Coyote" do.
e.g.: "Speed camera at 300m, speed limit 50km/h"
Also, If the user is approaching the speed camera exceeding the speed limit, osmand could warn him by beeping.

grravity:
+1 to maxspeed and voice prompt

martenjj:
+1 also to maxspeed in voice prompt

martenjj:
See also #3721 for maxspeed visual alert.

Issue #3721, opened 2017-04-28
UI - Show max speed sign next to speed camera icon


ignaciolep:
Hello,

It would be great if OsmAnd could show the max speed sign next to the speed camera icon, reading the speed limit value from the maxspeed tag in the enforcement=maxspeed relation or from the highway=speed_camera node (device).

It could also show a traffic light sign for the enforcement=traffic_signals relation.

I made two mockups:

![max speed sign - camera](https://cloud.githubusercontent.com/assets/13091755/25543631/b66f2dec-2c2d-11e7-97d7-1af12155ccb6.png)
![traffic light sign - camera](https://cloud.githubusercontent.com/assets/13091755/25543632/b6714f50-2c2d-11e7-8e60-59151a26f11c.png)

Thanks!
Ignacio

pebogufi:
Great Idea !
Peter

pl71:
👍

polarbearing:
I see several problems here:

- the allowed speed is already shown in a widget. Arguably it is a bit too small to be observed while driving. However, effectively the speed with the camera symbol is redundant. 
- the first mockup uses the same speed symbol in a red circle that OsmAnd uses when I exceed the speed, which is confusing if I am driving below the limit
- we have cameras here in areas with variable speed limit, thus the app would indicate a higher limit than it is allowed when the variable speed is posted lower on the street. 
- I am **absolutely against** warning of traffic light cameras. Exceeding the speed can happen gradually when the driver has more attention to the road than to his instruments. But driving over a red traffic light is a concious decision in full view of the traffic situation, which poses an immediate and deadly hazard on crossing traffic. Such a warning would encourage to drive on red over junctions without a camera.

ignaciolep:
Perhaps the max allowed speed widget could increase its size during the camera alert. The idea is to quickly identify the speed limit of the camera.

I didn't understand the problem with variable speed limit areas. Shouldn't the speed_camera have the lowest maxspeed limit?

About the traffic light camera alert, of course that nobody wants to encourage illegal driving. Sometimes the camera is a just a trap to collect fines, for example when the yellow light duration time is too short or the light changes and the car stops a few centimeters over the pedestrian crossing.

Think of this alert as if it were a sign in the street announcing the camera:

[...]

PR #12923, opened 2021-10-04
Attempt to implement #3721 and #8108 (android part)

AlexNcUa:
There is attempt to implement issues #3721 (except traffic light camera) and #8108.

_Tested in simulation mode for:_
- Speed camera with assigned max speed same as route segment max speed
- Speed camera with assigned max speed different from the route segment max speed  (speed camera max speed expected to be displayed / announced)
- Speed camera without assigned max speed (route max speed expected to be displayed / announced)

_Changes description:_

- RouteCalculationResult attachAlarmInfo method extended to add max speed to speed camera AlarmInfo taken either from speed camera point (preferred) or from RouteSegmentResult.
- WaypointHelper getMostImportantAlarm and announceVisibleLocations updated to attach actual distance to speed camera AlarmInfo.
- AlarmWidget extended with ability to show max speed for speed camera when available.
- VoiceRouter.announceAlarm extended with special handling for speed camera announcement to announce distance and max speed when available.
- CommandBuilder / JsCommandBuilder extended with new speedCameraAlarm method, which allows announce speed camera with distance and max speed.
- TestVoiceActivity extended to test speed camera with distance and max speed announcement.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
+++ b/OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
@@ -15,6 +15,7 @@ import net.osmand.data.QuadRect;
 import net.osmand.map.WorldRegion;
 import net.osmand.plus.OsmandApplication;
 import net.osmand.plus.R;
+import net.osmand.plus.helpers.enums.SpeedConstants;
 import net.osmand.plus.routing.AlarmInfo.AlarmInfoType;
 import net.osmand.plus.settings.backend.ApplicationMode;
 import net.osmand.router.ExitInfo;
@@ -260,10 +261,12 @@ public class RouteCalculationResult {
 				locations.get(currentLocation).getLatitude(), locations.get(currentLocation).getLongitude());
 	}
 
-	private static void attachAlarmInfo(List<AlarmInfo> alarms, RouteSegmentResult res, int intId, int locInd) {
+	private static void attachAlarmInfo(List<AlarmInfo> alarms, RouteSegmentResult res, int intId, int locInd, SpeedConstants sc) {
 		int[] pointTypes = res.getObject().getPointTypes(intId);
 		if (pointTypes != null) {
 			RouteRegion reg = res.getObject().region;
+			float maxSpeed = -1;
+			AlarmInfo speedCameraAlarmInfo = null;
 			for (int r = 0; r < pointTypes.length; r++) {
 				RouteTypeRule typeRule = reg.quickGetEncodingRule(pointTypes[r]);
 				int x31 = res.getObject().getPoint31XTile(intId);
@@ -272,9 +275,32 @@ public class RouteCalculationResult {
 				loc.setLatitude(MapUtils.get31LatitudeY(y31));
 				loc.setLongitude(MapUtils.get31LongitudeX(x31));
 				AlarmInfo info = AlarmInfo.createAlarmInfo(typeRule, locInd, loc);
+
 				// For STOP first check if it has directional info
 				if ((info != null) && !((info.getType() == AlarmInfoType.STOP) && !res.getObject().isStopApplicable(res.isForwardDirection(), intId, res.getStartPointIndex(), res.getEndPointIndex()))) {
 					alarms.add(info);
+
+					if (info.getType() == AlarmInfoType.SPEED_CAMERA) {
+						speedCameraAlarmInfo = info;
+					}
+				}
+
+				if (maxSpeed <= 0 || maxSpeed == RouteDataObject.NONE_MAX_SPEED) {
+					maxSpeed = typeRule.maxSpeed();
+				}
+			}
+
+			if (speedCameraAlarmInfo != null) {
+				if (maxSpeed <= 0 || maxSpeed == RouteDataObject.NONE_MAX_SPEED) {
+					maxSpeed = res.getObject().getMaximumSpeed(res.isForwardDirection());
+				}
+
+				if (maxSpeed > 0 && maxSpeed != RouteDataObject.NONE_MAX_SPEED) {
+					if (sc.imperial) {
+						speedCameraAlarmInfo.setIntValue(Math.round(maxSpeed * 3.6f / 1.6f));
+					} else {
+						speedCameraAlarmInfo.setIntValue(Math.round(maxSpeed * 3.6f));
+					}
 				}
 			}
 		}
@@ -378,7 +404,7 @@ public class RouteCalculationResult {
 					lastHeight = h;
 				}
 				locations.add(n);
-				attachAlarmInfo(alarms, s, i, locations.size());
+				attachAlarmInfo(alarms, s, i, locations.size(), ctx.getSettings().SPEED_SYSTEM.get());
 				segmentsToPopulate.add(s);
 				if (i == s.getEndPointIndex()) {
 					break;
--- a/OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/AlarmWidget.java
+++ b/OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/AlarmWidget.java
@@ -265,7 +265,14 @@ public class AlarmWidget {
 			}
 			text = alarm.getIntValue() + "";
 		} else if (alarm.getType() == AlarmInfo.AlarmInfoType.SPEED_CAMERA) {
-			locImgId = R.drawable.warnings_speed_camera;
+			int maxSpeed = alarm.getIntValue();
+			if (maxSpeed > 0) {
+				locImgId = R.drawable.warnings_speed_camera_w_limit;
+				text = maxSpeed + "";
+			}
+			else {
+				locImgId = R.drawable.warnings_speed_camera;
+			}
 		} else if (alarm.getType() == AlarmInfo.AlarmInfoType.BORDER_CONTROL) {
 			locImgId = R.drawable.warnings_border_control;
 		} else if (alarm.getType() == AlarmInfo.AlarmInfoType.HAZARD) {
--- a/OsmAnd/src/net/osmand/plus/voice/CommandBuilder.java
+++ b/OsmAnd/src/net/osmand/plus/voice/CommandBuilder.java
@@ -30,6 +30,7 @@ public abstract class CommandBuilder {
 	protected static final String C_REACHED_POI = "reached_poi";
 	protected static final String C_THEN = "then";
 	protected static final String C_SPEAD_ALARM = "speed_alarm";
+	protected static final String C_SPEED_CAMERA_ALARM = "speed_camera_alarm";
 	protected static final String C_ATTENTION = "attention";
 	protected static final String C_OFF_ROUTE = "off_route";
 	protected static final String C_BACK_ON_ROUTE = "back_on_route";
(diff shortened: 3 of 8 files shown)


AlexNcUa:
Screenshots of visual representation in day / night mode:
screenshot_day_mode
screenshot_night_mode

AlexNcUa:
PR was updated:

- Design of icons was changed to be less confusing with regular "current" speed limit alert, and to be more consistent with american signs (when US region selected).

- Visual distance indication added.

- Additionally fixed issue (not directly related to speed cameras) with resizing alert icon on bigger screens (tablets), which was leading to bottom text out of icon area (for example for tunnels).

New screenshots attached.

Screenshot_2021-10-30-01-03-45-398

Screenshot_2021-10-30-01-01-47-131

Screenshot_2021-10-30-01-13-09-502

Screenshot_2021-10-30-01-15-42-493

OsmAnd/src/net/osmand/plus/voice/CommandBuilder.java
```diff
@@ -30,6 +30,7 @@
 	protected static final String C_REACHED_POI = "reached_poi";
 	protected static final String C_THEN = "then";
 	protected static final String C_SPEAD_ALARM = "speed_alarm";
+	protected static final String C_SPEED_CAMERA_ALARM = "speed_camera_alarm";
```
vshcherb, line 33: This approach will require to change all voice configuration that could be too much. So incremental approach or check could be implemented probably.

AlexNcUa: I have added corresponding fallback in the voice configuration files as part of the related PR https://github.com/osmandapp/OsmAnd-resources/pull/724/files

For all languages that contains
`dictionary["distance"] + (tts ? " - " : " ") + distance(dist) ` as part of the `function route_recalc(dist, seconds)` i have added implementation, once this part of announcement will be same. As well as the part of announcement with the speed limit will be the same with corresponding announcement from "regular" speed limit alarm. For all other languages I have added fallback to original announcement without speed / distance.

AlexNcUa: Only one thing that I am not sure, it is how to force voice configuration files update before (or during) latest version installation. It seems that the voice files downloaded separately from the application itself. Probably you can suggest how to check if the latest voice files are available on device (or if voice file contains corresponding function) from the CommandBuilder / JsCommandBuilder, I have not found something similar....

vshcherb: Voice files are updated automatically checked each android version update (probably you have same android version number).

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/AlarmWidget.java
```diff
@@ -138,29 +143,60 @@ public boolean updateInfo(DrawSettings drawSettings, boolean drawBitmap) {
 							icon.setImageResource(info.locImgId);
 						}
 					}
-					if (!Algorithms.objectEquals(info.text, cachedText) || cachedRegion != info.region) {
+
+					boolean isRegionChanged = cachedRegion != info.region;
```
vshcherb, line 147: We follow previous code style cause it's shorter version and simpler.

OsmAnd/src/net/osmand/plus/voice/CommandBuilder.java
vshcherb: We've already implemented before backward compatible voice changes, so old voice packages could continue to work with  ```isJsCommandExists(C_TAKE_EXIT)```

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/AlarmWidget.java
AlexNcUa: @vshcherb Kindly note, that it is not only code style related.

Most likely the initial implementation has an issue with if statement condition in L#193 (previous L#156):

`if (!Algorithms.objectEquals(info.bottomText, cachedBottomText) || cachedRegion != info.region)`

because the cachedRegion could be already updated in L#144 (previous numbering) above, so the second if statement will not detect the region change.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
```diff
@@ -284,9 +287,32 @@ private static void attachAlarmInfo(List<AlarmInfo> alarms, RouteSegmentResult r
...
+				}
+			}
+
+			if (speedCameraAlarmInfo != null) {
+				if (maxSpeed <= 0 || maxSpeed == RouteDataObject.NONE_MAX_SPEED) {
+					maxSpeed = res.getObject().getMaximumSpeed(res.isForwardDirection());
+				}
+
+				if (maxSpeed > 0 && maxSpeed != RouteDataObject.NONE_MAX_SPEED) {
+					if (sc.imperial) {
```
vshcherb, line 311: Looks strange that here we store converted data, I guess it should happen at the end UI / Voice prompt cause here it's to early to do conversions

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/AlarmWidget.java
vshcherb: The codestyle and technique we use is very simple. 
1 - if statement that checks all cached variable
2- in case **any** state is changed the whole widget / icon is rerendered / recalculated, so pseudocode is always like

```
if( new_state_A != cached_state_A || new_state_B != cached_state_B || ... ) {
      cached_state_A = new_state_A;
      cached_state_B = new_state_B;
....
<CODE using new_state_A, new_state_B, ... >

}
```

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/AlarmWidget.java
```diff
@@ -138,29 +143,60 @@ public boolean updateInfo(DrawSettings drawSettings, boolean drawBitmap) {
...
+								widgetTextLayoutParams.gravity = Gravity.CENTER;
+
+								widgetText.setLayoutParams(widgetTextLayoutParams);
+								widgetText.setGravity(Gravity.NO_GRAVITY);
+								widgetText.setTextSize(COMPLEX_UNIT_PX, res.getDimensionPixelSize(R.dimen.map_alarm_text_size));
+							}
 						}
 					}
-					if (!Algorithms.objectEquals(info.bottomText, cachedBottomText) || cachedRegion != info.region) {
+					if (!Algorithms.objectEquals(info.bottomText, cachedBottomText) || isRegionChanged || isAlarmTypeChanged) {
```
vshcherb, line 193: If we talk about this code, then it's simple 2 blocks should be combined into 1 if, otherwise the pattern is broken and it's hard to maintain.

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/AlarmWidget.java
```diff
@@ -138,29 +143,60 @@ public boolean updateInfo(DrawSettings drawSettings, boolean drawBitmap) {
...
 							widgetText.setText(cachedText);
 							Resources res = layout.getContext().getResources();
-							if (alarm.getType() == AlarmInfo.AlarmInfoType.SPEED_LIMIT && info.americanType && !info.isCanadianRegion) {
+							if (cachedAlarmType == AlarmInfo.AlarmInfoType.SPEED_LIMIT && info.americanType && !info.isCanadianRegion) {
 								int topPadding = res.getDimensionPixelSize(R.dimen.map_alarm_text_top_padding);
 								widgetText.setPadding(0, topPadding, 0, 0);
 							} else {
 								widgetText.setPadding(0, 0, 0, 0);
 							}
+
```
vshcherb, line 169: this is already quite big chunk of code so, it makes sense I think to move if block inside to a separate method

OsmAnd/src/net/osmand/plus/voice/CommandBuilder.java
AlexNcUa: @vshcherb thank you for refer the `isJsCommandExists`, have missed it.

So, you recommend to add fallback in Java code even if the voice files will be updated as part of another already prepared PR?

The fallback also will require passing additional argument `String fallbackAttentionType` to the speedCameraAlarm method, to avoid dependency between JsCommandBuilder and AlarmInfoType (the attentionType should be `String.valueOf(AlarmInfoType.SPEED_CAMERA)`), which itself could be slightly "dirty".

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/AlarmWidget.java
AlexNcUa: Absolutely agree that single `if` will be simpler for understanding, will combine it. My initial intention was to reduce number of changes in PR.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
AlexNcUa: @vshcherb To be honest, I also had strong doubts when writing this code.

On the right hand it is not the best place for conversion, on the left hand the `AlarmInfo.intValue` also used for speed limit alarm, and in that case the field will contain converted value.

It could be very confusing if the same field will contain speed value in different units of measurement.

Probably, it would be better to introduce dedicated explicit field in AlarmInfo for meters per second speed and provide getter method to provide converted speed. But it will require re-factoring in multiple places, and also I guess there was signifact reason to use several common fields like "floatValue", "intValue" etc. instead of "maxSpeedInMetersPerSecond", "distance" etc...

AlexNcUa:
@vshcherb I've updated implementation, and have changed AlarmInfo to always keep speed in meters per seconds to avoid early conversion and to provide it according to requested speed constants settings. Could you please check if it is better from the general conception prospective? Also not sure about naming conventions, please correct me if I am wrong somewhere.

OsmAnd/src/net/osmand/plus/voice/CommandBuilder.java
vshcherb: You can never say if somebody doesn't have custom tts package. Secondly, good practice is to update only packages / languages that you are more or less sure. So in that case to not add commands that don't support it. 

Indeed you need to pass parameter that make sense only for speed camera, so probably making a new method make sense but it should be checked anyway that this method exists and fallback should be called.

vshcherb:
As we planned it for release 4.2, we've decided to postpone it till 4.3 and merge with https://github.com/osmandapp/OsmAnd/issues/13413. 

Main ideas behind it:
- We probably want to switch pictograms and make speed (exceeding speed) main pictogram and add speedcam pictogram as additional (3rd pictogram to Max speed / Current speed)
- To simplify view we probably don't need distance to speed camera cause it won't be readable during driving. We use for tunnels to mimic real traffic sign.

We plan soon to finalize UI but this issue likely won't be merged in 4.2 release.

alensiljak:
Hi, the current version is 4.5.9. Is this feature still on the roadmap?
I find the current speed camera announcements to come too late. Most of the time when camera is right next to the car or after. And this is at an area with the speed limit of 40km/h. This is a problem in countries/areas for which no speed limit exists in OSM (but there is a camera POI).

PR #12986, opened 2021-10-13
turn off keepLeft and keepRight for multi lane road

alisa911:


vshcherb:
Doesn't look good, very specific change and doesn't look self-explainable.

PR #13007, opened 2021-10-17
fix #10239 indicate already uploaded gpx tracks

### LINKED ISSUE #10239: [Feature request] OSM: Indicate whether a track has been uploaded or not
Currently all tracks look the same in the app, uploaded or not.
It is hard to remember which tracks I have uploaded already.

It would be great if the app could show what tracks have been uploaded already.

The only alternative currently is: going the OSM website to compare the list of filenames. It is very time-wasting and error-prone.

The app OSMTracker has that feature (but OsmAnd is much nicer to use). I suspect it only sets a flag locally, without actually checking on the OSM server.

koenigle:
Fix #10239
Indicate which gpx tracks have already been uploaded to OSM. The information is maintained by populating the already existing field `lastUploadedTime` in the GPX database on a successful upload. A track is already uploaded if `lastUploadedTime` >= `lastModifiedTime`.
The additional information is only visible when selecting the upload action in the My Places view.

Screenshot_1634426090

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -10,6 +10,7 @@
 	- For wording and consistency, please note https://docs.osmand.net/en/main@latest/development/contributing-to-osmand/translating-osmand
 	Thx - Hardy
 -->
+    <string name="track_already_uploaded">Already uploaded</string>
     <string name="android_auto_map_placeholder_title">The map is displayed on the vehicle screen.</string>
     <string name="android_auto_map_placeholder_descr">Disconnect your device from Android Auto to view the map in the app.</string>
     <string name="android_auto_purchase_request_title">Purchase OsmAnd Pro or Maps+ to use OsmAnd with Android Auto</string>


vshcherb:
This UI is questionable and need to be reviewed by the team

koenigle:
I also struggeled with the UI, thus i'm open to other recommendations to depict this information.

Options i experimented with:
1. Single icon without text, e.g. the icon which is used for the action itself. However thereby it seems not clear whether already uploaded tracks or tracks which need to be uploaded are indicated. Maybe a good icon can solve this or one has to use two different icons and mark all tracks.
Positioning:
    - behind track title
    - right beside track time
    - bigger icon on the right hand side where three dots and switches are shown in other views

2. Icon with text, thus its more clear what is indicated.
Positioning:
    - behind track title / right beside track time - problem on smaller devices
    - current implementation below other stuff - uploaded tracks and non-uploaded tracks are not of equal height

For all scenarios it seems reasonable to only show the additional information in the upload view.

PR #13144, opened 2021-11-10
Track waypoints UI

cepprice:


OsmAnd/src/net/osmand/plus/track/TrackPointsCard.java
```diff
@@ -305,22 +363,110 @@ public void filter(String text) {
...
+			addressNotFoundMsg = PointDescription.getAddressNotFoundStr(app);
+		}
+
 		public void synchronizeGroups(@NonNull List<GpxDisplayGroup> displayGroups) {
 			DisplayGroupsHolder displayGroupsHolder = DisplayPointsGroupsHelper.getGroups(app, displayGroups, filteredItems);
 			groups.clear();
 			itemGroups.clear();
 			groups.addAll(displayGroupsHolder.groups);
 			itemGroups.putAll(displayGroupsHolder.itemGroups);
+			lookupAddresses();
```
alex-osm, line 415: Do not lookup addresses. Use tag address if exists.

PR #13237, opened 2021-11-23
Fix Wikipedia Duplicates

nazar-kutz:


OsmAnd/src/net/osmand/plus/search/QuickSearchDialogFragment.java
```diff
@@ -1174,8 +1174,9 @@ public void onSearchListFragmentResume(QuickSearchListFragment searchListFragmen
 					searchEditText.setText(searchQuery);
 					searchEditText.setSelection(searchQuery.length());
 				}
-				if (getResultCollection() != null) {
-					updateSearchResult(getResultCollection(), false);
+				SearchResultCollection res = getResultCollection();
+				if (res != null) {
+					updateSearchResult(res, false);
```
alex-osm, line 1179: Why filter on UI?

PR #13257, opened 2021-11-26
GPX add <time> tag to <metadata>

nazar-kutz:


OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
```diff
@@ -1923,6 +1978,15 @@ private int getItemsToWriteSize() {
 			}
 			return size;
 		}
+
+		public void setupCreationTime() {
```
alex-osm, line 1982: Why time not setup at GPXFile constructors?

OsmAnd-java/src/main/java/net/osmand/router/RouteExporter.java
```diff
@@ -34,6 +34,7 @@ public RouteExporter(String name, List<RouteSegmentResult> route, List<Location>
 
 	public GPXFile exportRoute() {
 		GPXFile gpx = new GPXFile(OSMAND_ROUTER_V2);
+		gpx.setupCreationTime();
```
alex-osm, line 37: Did you analyse all places in the OsmAnd where gpx files creates?

OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
```diff
@@ -496,6 +496,30 @@ public boolean hasRoute() {
 		public Copyright copyright = null;
 		public Bounds bounds = null;
 
+		public Metadata() { }
+
+		public Metadata(Metadata source) {
```
alex-osm, line 501: Probably we will need refactor GPS Filter task at the 4.1 branch. Since changes of GPXUtilities.java affects GPS Filter, would be good to make them via separate commit.

PR #13394, opened 2021-12-18
New strings reworked

comradekingu:
Something close to this would be highly beneficial.


vshcherb:
There are conflicts but I don'w know how

sonora:
Meanwhile superseded - thanks.

PR #13414, opened 2021-12-21
added all osm-related settings to separate card

mmh-dev:


OsmAnd/src/net/osmand/plus/dialogs/ConfigureMapMenu.java
```diff
@@ -22,6 +22,7 @@
 import static net.osmand.aidlapi.OsmAndCustomizationConstants.TRANSPORT_ID;
 import static net.osmand.plus.ContextMenuAdapter.makeDeleteAction;
 import static net.osmand.plus.ContextMenuItem.INVALID_ID;
+import static net.osmand.plus.plugins.osmedit.OsmEditingPlugin.RENDERING_CATEGORY_OSM_ASSISTANT;
```
vshcherb, line 25: This Import is not correct, should be part of plugin @Chumva @mmh-dev

PR #13517, opened 2022-01-10
fixed bug when widgets in night mode become white after rotating the …

mmh-dev:
…screen


OsmAnd/src/net/osmand/plus/views/layers/MapInfoLayer.java
```diff
@@ -459,7 +456,6 @@ private TextState calculateTextState() {
 
 	@Override
 	public void onDraw(Canvas canvas, RotatedTileBox tileBox, DrawSettings drawSettings) {
-		this.drawSettings = drawSettings;
```
vshcherb, line 462: Doesn't look as a right fix, draw settings should be correctly set.

PR #13723, opened 2022-02-09
Enhance category editor

cepprice:


OsmAnd/res/values/strings.xml
```diff
@@ -15,6 +15,10 @@
     <string name="shared_string_shape">Shape</string>
     <string name="default_appearance_desc">Default color, icon and shape will apply to the added favorite points into the group.</string>
     <string name="default_appearance">Default appearance</string>
+    <string name="android_api">Android API</string>
```
vshcherb, line 18: <string name="android_api">Android API</string>
    <string name="google_play_services">Google Play Services</string>
should be no_translate

PR #13727, opened 2022-02-09
fix 13407,1104

alisa911:


OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -1568,6 +1574,22 @@ protected RoadSplitStructure calculateRoadSplitStructure(RouteSegmentResult prev
 		return rs;
 	}
 
+	private boolean hasTU(String turnLanesPrevSegm, boolean attachedOnTheRight) {
```
vshcherb, line 1577: yet another parsing turn:lanes

PR #13948, opened 2022-03-18
OpenGL rendering for RoutingLayer

ivanPyrohivskyi:


OsmAnd/src/net/osmand/plus/views/layers/RouteLayer.java
```diff
@@ -586,4 +610,148 @@ public boolean runExclusiveAction(@Nullable Object o, boolean unknownLocation) {
...
+				ll.get(index).add(loc);
+			} else if (i < locations.size() - 1) {
+				index++;
+				ll.add(new ArrayList<>());
+			}
+		}
+		return ll;
+	}
+
+	private FColorARGB getFColorARGB(int color) {
```
alex-osm, line 750: https://github.com/osmandapp/Osmand/blob/d9d3df0cacd1c7913c357588867004c053d6c53d/OsmAnd/src/net/osmand/plus/utils/NativeUtilities.java#L19 ?

OsmAnd/src/net/osmand/plus/views/layers/RouteLayer.java
```diff
@@ -150,6 +161,14 @@ public void onPrepareBufferImage(Canvas canvas, RotatedTileBox tileBox, DrawSett
 
 			final QuadRect latlonRect = cp.getLatLonBounds();
 			final QuadRect correctedQuadRect = getCorrectedQuadRect(latlonRect);
+
+			if (view.hasMapRenderer()) {
+				//OpenGL
+				drawRouteSegment(tileBox);
```
alex-osm, line 167: Looks like it needs optimisation. OpenGL should be called once per route line calculated / cancelled. Also it should not depend on tileBox.

Issue #13864, opened 2022-03-06
Display of hiking path signs (text on shield)


okievv:
I am having trouble with specifying the correct OSMC:symbol tag for e.g. route id 2539062.

Original sign is a red digit on a white background. But this doesn't reflect in OSMAND+.
In both cases, I can only see an empty white square, i.e. no digit '1'. 

I tried 2 syntaxes:

"osmc:symbol"="red:white:::1:red"
(according to https://wiki.openstreetmap.org/wiki/DE:Key:osmc:symbol which states:
osmc:symbol=wegfarbe:hintergrund:vordergrund:vordergrund2:text:textfarbe
)

and

"osmc:symbol"="red:white:1:red"
(according to  https://wiki.openstreetmap.org/wiki/Key:osmc:symbol which states:
osmc:symbol=waycolor:background[:foreground][[:foreground2]:text:textcolor]
)

Foreground colours can be skipped in case the path sign is a character.
 
Any advice?

vshcherb:
There are different colors for way and symbol - https://github.com/osmandapp/OsmAnd-tools/blob/8109d57fd347f4ad9dda5beaa3e5efa8ffe3a73d/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/osm/MapRenderingTypesEncoder.java#L1326.

As I understand your syntax ```"osmc:symbol"="red:white:::1:red"``` should work. Did you regenerate map with OsmAndMapCreator ? What do you see as output for Binary ./inspector.sh

okievv:
I did not regenerate. I was waiting on the result after the monthly map updates came in.

vshcherb:
ok, we will check

xmd5a2:
Currently we don't support rendering of 'text' subtag over background

vshcherb:
Implemented in 4.2

PR #14214, opened 2022-05-06
Add osmc text to route icons

Chumva:


michaelblyons:
Is this related to #13864?

PR #14247, opened 2022-05-10
fix routing time

alisa911:


OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
```diff
@@ -726,6 +726,11 @@ private boolean needRequestPrivateAccessRouting(RoutingContext ctx, List<LatLon>
 
 	public List<RouteSegmentResult> searchRoute(final RoutingContext ctx, LatLon start, LatLon end, List<LatLon> intermediates,
 	                                            PrecalculatedRouteDirection routeDirection) throws IOException, InterruptedException {
+		ctx.slat = start.getLatitude();
```
vshcherb, line 729: Unnecessary variables (startY, startX)

OsmAnd-java/src/main/java/net/osmand/router/RoutingContext.java
```diff
@@ -70,6 +70,22 @@ public class RoutingContext {
 	
 	public boolean publicTransport;
 	
+	public double slat;
+	public double slon;
+	public double elat;
+	public double elon;
+	
+	
+	public double startPosDist;
```
vshcherb, line 79: Too many variables, code pollution

PR #14272, opened 2022-05-12
Fix AA NPE

cepprice:


OsmAnd/src/net/osmand/plus/auto/TripHelper.java
```diff
@@ -193,7 +193,9 @@ public Trip buildTrip(float density) {
 			if (!deviatedFromRoute) {
 				nextDirInfo = routingHelper.getNextRouteDirectionInfo(calc, true);
 				CurrentStreetName currentName = routingHelper.getCurrentName(nextDirInfo);
-				tripBuilder.setCurrentRoad(currentName.text);
```
vshcherb, line 196: where is else ? 
```
					tripBuilder.setCurrentRoad(currentName.text == null ? "" : currentName.text);
```

cepprice: `currentRoad`  is not mandatory

PR #14283, opened 2022-05-13
OsmEdits layer

ivanPyrohivskyi:
Problem present here of resources initialization (solved) and text below icons (need to fix in OpenGL)


OsmAnd/src/net/osmand/plus/plugins/osmedit/OsmEditsLayer.java
```diff
@@ -322,4 +374,107 @@ public boolean isTextVisible() {
...
+	}
+
+	/** OpenGL */
+	private void setMovableObject(@NonNull OsmPoint objectInMotion) {
+		if (markersCollection == null) {
+			return;
+		}
+		int x = MapUtils.get31TileNumberX(objectInMotion.getLongitude());
+		int y = MapUtils.get31TileNumberY(objectInMotion.getLatitude());
+		if (movableObject != null && movableObject.getX() == x && movableObject.getY() == y) {
```
ivanPyrohivskyi, line 457: if (movableObject != null)
enough here

PR #14389, opened 2022-05-26
Add select hiking routes OpenGL

Dima-1:


OsmAnd-java/src/main/java/net/osmand/router/network/NetworkRouteSelector.java
```diff
@@ -835,8 +835,7 @@ public static List<RouteKey> getRouteKeys(RouteDataObject obj) {
 		}
 
 
-		public static List<RouteKey> getRouteStringKeys(RenderedObject o) {
```
alex-osm, line 838: Do not remove it!
Make similar for MapObject

OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
```diff
@@ -308,6 +309,11 @@ private void selectObjectsFromOpenGl(@NonNull MapSelectionResult result, @NonNul
 										amenity.getX().add(points31.get(k).getX());
 										amenity.getY().add(points31.get(k).getY());
 									}
+								} else {
```
alex-osm, line 312: is it correct? Where isRoute variable?

PR #14419, opened 2022-05-30
Fix live updates OpenGL

Dima-1:


OsmAnd/src/net/osmand/plus/views/corenative/NativeCoreContext.java
```diff
@@ -63,6 +63,7 @@ public static void init(OsmandApplication app) {
 				ObfsCollection obfsCollection = new ObfsCollection();
 				obfsCollection.addDirectory(directory.getAbsolutePath(), false);
 				obfsCollection.addDirectory(app.getAppPath(IndexConstants.ROADS_INDEX_DIR).getAbsolutePath(), false);
+				obfsCollection.addDirectory(app.getAppPath(IndexConstants.LIVE_INDEX_DIR).getAbsolutePath(), false);
```
vshcherb, line 66: There is a checkbox in development to support or not Live updates? Did you check the core-legacy alternative code?

PR #14549, opened 2022-06-15
Fix polygons in OpenGL 2.5D

ivanPyrohivskyi:


OsmAnd/src/net/osmand/plus/resources/ResourceManager.java
```diff
@@ -1357,6 +1368,9 @@ public void closeFile(String fileName) {
 		if (resource != null) {
 			resource.close();
 		}
+		for (ResourceListener l : resourceListeners) {
+			l.onMapClosed();
```
alex-osm, line 1372: Call l.onMapClosed with fileName

OsmAnd/src/net/osmand/plus/resources/ResourceManager.java
```diff
@@ -1023,6 +1032,8 @@ public List<String> indexingMaps(@Nullable IProgress progress, @NonNull List<Fil
 				log.error("Index file could not be written", e);
 			}
 		}
+		backupedFileNames.clear();
```
alex-osm, line 1035: Concurrency

PR #14625, opened 2022-06-22
Fix favorites, waypoints, markers in OpenGL

ivanPyrohivskyi:


OsmAnd/src/net/osmand/plus/helpers/LauncherShortcutsHelper.java
```diff
@@ -65,7 +65,7 @@ public void onFavoriteDataUpdated(@NonNull FavouritePoint point) {
 			}
 
 			@Override
-			public void onFavoritePropertiesUpdated() {
+			public void onFavoritePropertiesUpdated(boolean syncWithMarker) {
```
vshcherb, line 68: Looks overcomplicated probably we need to review it again.

PR #14631, opened 2022-06-23
New app strings reworked

comradekingu:
Closes #13394


sonora:
Looks superseded and mostly implemented - Thanks!

(Some remaining items look like they are not appropriate, containing misunderstandings.)

PR #14653, opened 2022-06-24
Fix lags in `Configure screen` 

cepprice:


OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/WidgetsSettingsHelper.java
```diff
@@ -0,0 +1,205 @@
...
+public class WidgetsSettingsHelper {
+
+	private final OsmandSettings settings;
+	private final MapActivity mapActivity;
+	private final ApplicationMode appMode;
+
+	private final MapWidgetRegistry widgetRegistry;
+	private final MapWidgetsFactory widgetsFactory;
+
+	public static void resetConfigureScreenSettings(@NonNull MapActivity mapActivity,
```
vshcherb, line 37: 1) Why do we need this class MapWidgetsFactory is very small class and much better naming with than helper
2) why so many static methods? They all could be part of constructor and normal methods, it's much better to not wrap around static methods

cepprice: 1. I don't understand question

PR #14738, opened 2022-07-02
Marker size in Auto mode [WIP]

landswellsong:
I noticed that a few of the map markers don't switch the bitmaps when switching to Auto Mode which results in images like this due to DPI mismatch:
osm bug

There  is probably a lot of them, but the most offending ones also set their size according to bitmap, so far those are the context menu orange marker and the Telegram user icon.

I'm in no way an Android / Java coder but I seem to have fixed one of them:

osm bug fixed

I'd love some feedback if this is the way to go and I should proceed with Telegram ones or there can be a more systemic solution to this problem.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/ContextMenuLayer.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/ContextMenuLayer.java
@@ -12,6 +12,7 @@ import android.graphics.PointF;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.os.Vibrator;
+import android.view.Display;
 import android.view.GestureDetector;
 import android.view.GestureDetector.SimpleOnGestureListener;
 import android.view.MotionEvent;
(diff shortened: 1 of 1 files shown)


vshcherb:
Yes we have this task in porgress actually.

alex-osm:
Thanks for the request. Fixed by https://github.com/osmandapp/OsmAnd/commit/20afd37e7ea042e5b232bade161f6f95f305d9f5

alex-osm:
Which ones?

landswellsong:
@alex-osm the user makers from Telegram integration. Here's a sample, the marker on the phone:
image

Same view more or less on the car:
image

This one is really annoying because the user isn't in full control of the markers (i.e. they appear when someone sends you the location) and they can occupy a real large portion of the screen this way.

On a lesser scale it also applies to the arrow showing own location, it's way too big due to the same issue as in, the resource is taken for a wrong DPI, but it's at least solveable because that arrow scales from Text Scale.

alex-osm:
@landswellsong Makers from Telegram integration will be fixed soon. The arrow of my location is fixed already.

PR #14770, opened 2022-07-06
Routing logs

Chumva:


OsmAnd/src/net/osmand/plus/dialogs/WhatsNewDialogFragment.java
```diff
@@ -84,7 +84,7 @@ public static boolean wasNotShown() {
 	}
 
 	public static boolean shouldShowDialog(@NonNull OsmandApplication app) {
-		return app.getAppInitializer().checkAppVersionChanged() && notShown;
```
vshcherb, line 87: // add comment for 4.2.7+ releases only !

OsmAnd/src/net/osmand/plus/base/FailSafeFunctions.java
```diff
@@ -167,6 +167,8 @@ protected void onPostExecute(GPXFile result) {
 	
 	public static void enterRoutingMode(@NonNull MapActivity mapActivity, @Nullable GPXRouteParamsBuilder gpxRoute) {
 		OsmandApplication app = mapActivity.getMyApplication();
+		app.logRoutingEvent("enterRoutingMode gpxRoute " + gpxRoute);
```
vshcherb, line 170: gpxRoute has good toString?

OsmAnd/src/net/osmand/plus/OsmandApplication.java
```diff
@@ -1060,23 +1060,31 @@ public String getUserAndroidId() {
...
 			analyticsHelper.addEvent(event, AnalyticsHelper.EVENT_TYPE_APP_USAGE);
 		} catch (Exception e) {
 			LOG.error(e);
 		}
 	}
 
-	public void logMapDownloadEvent(String event, IndexItem item) {
+	public void logRoutingEvent(@NonNull String event) {
+		try {
+			analyticsHelper.addEvent(event, AnalyticsHelper.EVENT_TYPE_ROUTING);
```
vshcherb, line 1073: Is here timestamp added?

Chumva: Yes, it is added in AnalyticsHelper.addEvent() - https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/helpers/AnalyticsHelper.java#L266

OsmAnd/src/net/osmand/plus/base/FailSafeFunctions.java
Chumva: It should be enough - https://github.com/osmandapp/OsmAnd/pull/14770/files#diff-b54a0d5f4cefb955f45bb9466c923393fd3ea81298c5caca0d8bf5afe3898460R248

PR #14906, opened 2022-07-21
Fix selecting long hiking route memory crash

Dima-1:


OsmAnd-java/src/main/java/net/osmand/router/network/NetworkRouteContext.java
```diff
@@ -26,17 +24,17 @@
 import net.osmand.util.MapUtils;
 
 public class NetworkRouteContext {
-	
+
 	private static final int ZOOM_TO_LOAD_TILES = 15;
-	
+
+	private final TLongObjectMap<NetworkRoutePoint> routePoints = new TLongObjectHashMap<>();
```
vshcherb, line 30: routePoints definitely should be in tiles

PR #14912, opened 2022-07-22
Fix_14491

Chumva:


Chumva:
There was a problem with tiles size calculation, for a file we used its length, but for SQLite was used bitmap.getByteCount() which returns uncompressed image size. Bitmaps were 4-5 times bigger than data stored in sqlite

PR #15008, opened 2022-08-04
Fix POI background size on big text scale

Dima-1:


OsmAnd/src/net/osmand/plus/views/PointImageDrawable.java
```diff
@@ -325,7 +325,7 @@ private int getBigMergedBitmapWidth() {
 			width = Math.max(width, background.getWidth());
 			width = Math.max(width, listIcon.getWidth());
 		}
-		return width;
+		return scale != 1.0f ? (int) (width * scale) : width;
```
ivanPyrohivskyi, line 328: just width * scale

Issue #15128, opened 2022-08-22
Android 6: Search not working in nightly builds


Prusaker:
### Description
Search not working in nightly builds. I've tested on nightly builds for at least the last month.
Bug reproduced at devices with Android 6 and 7. **Devices with Android 9 works correctly.**

### How to reproduce?
Just get device with Andrid 6 or 7, install OsmAnd and try to search anything.

### Your Environment
OsmAnd Version: OsmAnd nightly 4.3.0#34493m (or any nightly build at least the last month)
Android/iOS version: tested at Android 6, 7 and 9
Device model: Google Nexus 6, Lenovo Tab 2 A10-70, XIaomi Redmi 8A

**Maps used (online or offline):**   
Tested at offline map.

<table style="width:100%">
  <tr>
<th><img src="https://user-images.githubusercontent.com/32092837/185876332-fa9f19f6-788b-47e6-8673-1af48abba7a0.png"  width="300" height="auto"/> </th>
  </tr>
</table>  

Logcat.log from device with Android 7:
[logcat.log](https://github.com/osmandapp/OsmAnd/files/9392690/logcat.log)

PR #15201, opened 2022-09-01
Fix "Android 6: CopyOnWriteArrayList sort problems"

### LINKED ISSUE #15129: Android 6: Night build crashes when trying to add a favorite point.
### Description
Night build crashes when trying to add a favorite point. I've tested on nightly builds for at least the last month.
Bug reproduced at devices with Android 6 and 7. **Devices with Android 9 works correctly.**

### How to reproduce?
Just get device with Andrid 6 or 7, install OsmAnd and try to add favorite point.

### Your Environment
OsmAnd Version: OsmAnd nightly 4.3.0#34493m (or any nightly build at least the last month)
Android/iOS version: tested at Android 6, 7 and 9
Device model: Google Nexus 6, Lenovo Tab 2 A10-70, XIaomi Redmi 8A

**Maps used (online or offline):**   
Tested at offline map.

https://user-images.githubusercontent.com/32092837/185881270-71656852-8fed-4b3a-8fe9-da505f1ccfa7.mp4
[exception.log](https://github.com/osmandapp/OsmAnd/files/9392787/exception.log)

nazar-kutz:
We had the same problem with sorting CopyOnWriteArrayList that resulted in diferent issues on Android 6: #15128, #15129, #15202, #15203.
**Problem:** Collections.sort uses ListIterator.set, but CopyOnWriteArrayList's ListIterator does not support the "remove", "set" or "add" methods.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/myplaces/FavouritesHelper.java
+++ b/OsmAnd/src/net/osmand/plus/myplaces/FavouritesHelper.java
@@ -19,6 +19,7 @@ import net.osmand.plus.OsmandApplication;
 import net.osmand.plus.R;
 import net.osmand.plus.mapmarkers.MapMarkersGroup;
 import net.osmand.plus.mapmarkers.MapMarkersHelper;
+import net.osmand.plus.utils.AndroidUtils;
 import net.osmand.util.Algorithms;
 
 import org.apache.commons.logging.Log;
@@ -602,7 +603,7 @@ public class FavouritesHelper {
 		for (FavoriteGroup g : favoriteGroups) {
 			Collections.sort(g.getPoints(), favoritesComparator);
 		}
-		Collections.sort(cachedFavoritePoints, favoritesComparator);
+		AndroidUtils.sortCopyOnWriteList(cachedFavoritePoints, favoritesComparator);
 	}
 
 	public static Comparator<FavouritePoint> getComparator() {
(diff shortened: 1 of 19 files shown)


OsmAnd/src/net/osmand/plus/myplaces/FavouritesHelper.java
```diff
@@ -602,7 +603,7 @@ public void sortAll() {
 		for (FavoriteGroup g : favoriteGroups) {
 			Collections.sort(g.getPoints(), favoritesComparator);
 		}
-		Collections.sort(cachedFavoritePoints, favoritesComparator);
```
vshcherb, line 605: Collections should be always recreated (!) in case they are changed to avoid all possible modify exceptions if they accessed from multiple threads.

OsmAnd/src/net/osmand/plus/utils/AndroidUtils.java
```diff
@@ -1178,6 +1178,32 @@ private static String getStringForField(@NonNull Context ctx, @Nullable Field fi
 		return null;
 	}
 
+	public static <T> void reverseCopyOnWriteList(List<T> list) {
```
vshcherb, line 1181: Not needed:
Collections should be always recreated (!) in case they are changed to avoid all possible modify exceptions if they accessed from multiple threads.

vshcherb:
> 	mapMarkers = new CopyOnWriteArrayList<>(pair.second.values());
>		mapMarkersGroups = new CopyOnWriteArrayList<>(pair.first.values());
Why CopyOnWriteArrayList is needed at all? It's probably a design flaw cause modification should only happen in 1 thread otherwise there are other bugs.

Issue #14936, opened 2022-07-26
Remaining issues with tapping routes


sonora:
Carry over from #14698:

1. It still seems unclear where exactly you have to tap on a route to get it selected. It seems that maybe you have to hit a displayed route shield, or an actual route point? Or is there a timing issue, that maybe some indexing takes place after panning during which tapping a route is flaky or may have no effect?
2. There are issues with some routes still not showing, but this seems perhaps only under certain circumstances or devices. The effect is that tapping them briefly shows the "Loading data..." toast, but this aborts with no subsequent context menu shown. Examples are https://www.openstreetmap.org/relation/3762491("Relation: Goldsteig 4: Neuhaus - Letzau/Oberhöll") and https://www.openstreetmap.org/relation/7323772 ("Relation: Trausnitz - ?? 1"). 
3. Possibly related to (2.): Special characters in the OSM relation name may be the root cause of (2), see above. In any case, even on devices where (2) is not an issue, Relation "Goldsteig 4: Neuhaus - Letzau/Oberhöll"is in OsmAnd displayed truncated to just "Oberhöll". (see screenshot at https://github.com/osmandapp/OsmAnd/issues/14698#issuecomment-1194425825)

scaidermern:
I already reported 1) in #14241. Tapping on route shields also doesn't work reliably (even after the alleged fix).

There is also a 4th issue: If multiple routes overlap then only one of them will get selected (the one that gets rendered, I guess). It is not possible to select the other routes.

vshcherb:
Looks more like 3 independent issue or a discussion then
1. You need to hit on road shields
2-3. Strange, attaching screenshots cause it works in my case

<details>
    <summary>Screenshots</summary>

![telegram-cloud-photo-size-4-5980796128263781044-y](https://user-images.githubusercontent.com/1042025/181010376-dfabbedd-61d2-43b3-90be-1e4293a1dd39.jpg)

![telegram-cloud-photo-size-4-5980796128263781045-y](https://user-images.githubusercontent.com/1042025/181010391-c4422ec7-3b5e-4191-ac99-f5f73d325676.jpg)


</details>

4. All routes on the same way should be suggested in the list. I've tested on (https://www.openstreetmap.org/relation/7323772#map=16/49.5175/12.2728, https://www.openstreetmap.org/relation/3762491#map=13/49.7371/12.2458) works ok.

scaidermern:
[...]

PR #15205, opened 2022-09-01
Fix first segment is in another tile for route selection

Dima-1:


vshcherb:
https://github.com/osmandapp/OsmAnd/issues/14936

OsmAnd-java/src/main/java/net/osmand/router/network/NetworkRouteContext.java
```diff
@@ -149,35 +149,36 @@ public List<NetworkRouteSegment> loadRouteSegment(int x31, int y31) throws IOExc
...
-		int tileX = x31 >> zm;
-		int tileY = y31 >> zm;
-		int tileLeft = tileX << zm;
-		int tileTop = tileY << zm;
-		int tileRight = (tileX + 1) << zm;
-		int tileBottom = (tileY + 1) << zm;
 		if (routing) {
-			SearchRequest<RouteDataObject> req = BinaryMapIndexReader.buildSearchRouteRequest(tileLeft, tileRight,
-					tileTop, tileBottom, null);
+			SearchRequest<RouteDataObject> req = BinaryMapIndexReader.buildSearchRouteRequest(x15, y15);
```
vshcherb, line 173: is this method needed ? refactoring goal is unclear... It looks that code before was better cause zoom was centralized.

OsmAnd-java/src/main/java/net/osmand/router/network/NetworkRouteContext.java
```diff
@@ -149,35 +149,36 @@ public List<NetworkRouteSegment> loadRouteSegment(int x31, int y31) throws IOExc
...
 	}
-	
+
 	public boolean isRouting() {
 		return routing;
 	}
-	
 
-	private NetworkRoutesTile loadTile(int x31, int y31, long tileId) throws IOException {
+	private NetworkRoutesTile loadTile(int x15, int y15, long tileId) throws IOException {
```
vshcherb, line 170: 15 is not unversal number, zoom should be passed as parameter.

PR #15286, opened 2022-09-12
Increase order of my location

ivanPyrohivskyi:
https://github.com/osmandapp/OsmAnd-iOS/issues/2117


OsmAnd/src/net/osmand/plus/views/MapLayers.java
```diff
@@ -166,7 +166,7 @@ public void createLayers(@NonNull OsmandMapTileView mapView) {
 		// 5.95 all text labels
 		// 6. point location layer
```
vshcherb, line 167: comment?

PR #15337, opened 2022-09-19
Add support for online tile source template URLs with only bounding box

naf419:
Some online map data sources do not have native support for tiled web map
(slippy map) requests. A common way of operating without the tile zoom/x/y
triplet is to specify a lat/lon bounding box in the tile request.

Previously, as a workaround, osmand suggested pointing at a redirecting
proxy that would rewrite requests for zoom/x/y tiles with an equivalent request
for a bounding box (i.e. http://whoots.mapwarper.net/tms)

Instead of this workaround, implement this functionality natively by adding
a "{bbox}" parameter to templated tile URLs, which will be replaced at request
time with the tile's lat/lon bounding box.

Examples of services that use the templated bounding box method:
1) ArcGIS rest service:
https://gis.charttools.noaa.gov/arcgis/rest/services/MCS/NOAAChartDisplay/MapServer/exts/MaritimeChartService/MapServer/export?dpi=96&transparent=true&format=png24&layers=show:0,1,2,3,4,5,6,7&bbox={bbox}&bboxSR=4326&imageSR=3857&size=256,256&f=image

2) WMS:
https://encdirect.noaa.gov/arcgis/services/encdirect/enc_harbour/MapServer/WmsServer?bbox={bbox}&format=image/png&service=WMS&version=1.1.1&request=GetMap&srs=EPSG:4326&width=256&height=256&layers=0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29&map=&styles=

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/map/TileSourceManager.java
+++ b/OsmAnd-java/src/main/java/net/osmand/map/TileSourceManager.java
@@ -3,6 +3,7 @@ package net.osmand.map;
 import net.osmand.PlatformUtil;
 import net.osmand.osm.io.NetworkUtils;
 import net.osmand.util.Algorithms;
+import net.osmand.util.MapUtils;
 
 import org.apache.commons.logging.Log;
 import org.xmlpull.v1.XmlPullParser;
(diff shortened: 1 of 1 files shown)


vshcherb:
That doesn't implement Android / iOS support though

naf419:
> That doesn't implement Android / iOS support though

I believe it does implement Android support, at least according to my test build of OsmAnd-nightlyFree-opengl-fat-debug.apk
iOS I admittedly have no idea about...

naf419:
> > That doesn't implement Android / iOS support though
> 
> I believe it does implement Android support, at least according to my test build of OsmAnd-nightlyFree-opengl-fat-debug.apk iOS I admittedly have no idea about...

OK I tried to address iOS with https://github.com/osmandapp/OsmAnd-core/pull/384 (but I admittedly know no way to test it)

PR #15363, opened 2022-09-23
Configure map / Tracks: Redesign

0xRe1nk0:


sonora:
- [ ] Looks like there is still an issue with selecting and clearing the checkboxes: Go to any folder tab, tap on a contained track: The checkbox will become selected. But somehow it also seems to become disabled, because another tapv will not deselect it any more.

- [ ] Also the behavior of the "Apply" button still seems weird. Looks like it clears the set checkboxes, but still leaves them grayed out?

sonora:
Can I please note the following, chances are we may want to keep this in mind during the design phase already here:

For **"Favorites"**,
- we have only a simple on/off toggle under `Configure map > Favorites`
- while all detailed configuration is done under screen `MyPlaces > Favorites`

For **"Tracks**",
- it currently seems like we are expanding `Configure map > Tracks` to be a powerful configuration screen?
- Essentially rivaling (perhaps even replacing?) what we today have under `MyPlaces > Tracks` ?

Ideally, we would harmonize this
- either by e.g. having both the **"Favorites"** and the **"Tracks"** configuration screens under `MyPlaces`, and only have a simple on/off toggle each under `Map configuration`,
- or we would from both places link to the same powerful configuration screen (one for Tracks and one for Favorites) so that the user will end up on the same screen no matter if they came via `Configure map (> Favorites / Tracks)`, or via `MyPlaces (>Favorites/Tracks) `?

vshcherb:
@sonora yes exactly expanding cause we want to consolidate it with my places as discussed before. So it should be powerful.

Tracks on/off should stay unless we change paradigm to advanced users - switches are clear only to experienced users they know what to expect (switch on) for other users it's fine to have extra screen to have fine tuned.

Consolidation for Favorites also very important to have switches per group and colours as well.

sonora:
The new strings we use in "_Default track appearance_" are still a little unclear. If you explain the following, I can fine-tune them:

- (1) "_these settings will be applied to all created, imported or recorded tracks..._" Do you mean to say all which which will be created/imported/recorded after the settings are changed, while all  tracks already existing now keep their former appearance?
- (2) When you say "_without their settings_", what do you mean? Perhaps something like without appearance settings specified inside the GPX file itself?
- (3) Can you explain the second paragraph? What is meant by "_the Default option_", and what are "_all tracks with this default option_"?

sonora:
An additional functional question: If you e.g. use the 'Touring view' Map style, without having modified anything the button "Custom" under "Select width" is framed, which may be  correct. But below that the slider is set to the "1" position, which I think is incorrect: The default width of tracks in 'Touring view' is a lot thicker than 1?

sonora:
PS2: We have a button "Hide all", which seems to always act on the tracks displayed on the map. Would be cool if this button could be a toggle switching between actions "Select all" / "Clear all, acting on all checkboxes of the currently displayed folder tab.

Chumva:
I will check tomorrow issue with width slider. 

Appearance parameters are applied to all selected tracks on previous screen, also they are saved in preferences and used in other gpx (as default) if they don't have this parameters specified in extensions or GpxDbHelper. Probably this behaviour will have some tweaks next week.

Also tracks sorting should be available in the beginning of the next week.

sonora:
Great work!

I have taken a shot at the strings, see if you like https://github.com/osmandapp/OsmAnd/commit/5c9cfa35f1d829144ce3b2f3a69b84d6c1b5a70d.

String `default_tracks_appearance_description` is still maybe unnecessary clumsy: I guess we could omit the reference to `%2$s` which really just inserts a second time the string `shared_string_default`, not bringing any new information?

sonora:
Also: I guess we need to use an extra text line for the display of the folder name of each track on tabs ON MAP and ALL. The space reserved now is too narrow on many screens, so the folder name may be broken in 4 or more lines.

Screenshot_20230217-212604_OsmAnd~

PR #15410, opened 2022-10-03
Add "Send UUID" preference and UI

nazar-kutz:


verdulo:
Could you please stop reusing string? Now it is difficult to translate. On one screen the string " shared_string_privacy_policy" is used as part of sentence, here is used as list item.

vshcherb:
shared_string_privacy_policy means it couldn't be translated differently only as is "Privacy policy" so it should be consistent and has 1 meaning

verdulo:
I know, that strings should be consistent: when I translate I often use "Search" function on Weblate and compare which words I used in translations of OpenStreetMap website and iD editor.

Standalone phrases "Terms Of Use" and "Privacy Policy" should be translated into Esperanto as "**Kondiĉoj pri uzado**" and "**Privateca politiko**"; but the whole sentence  _terms_of_service_desc_ should be translated as "Elŝutante mapojn vi konsentas kun niaj **privateca politiko** kaj **kondiĉoj pri uzado**.".
Phrase with the string "privacy_and_security_change_descr" should be translated as "Frapetu “Permesi”, se vi akceptas nian **privatecan politikon**".

Similar in Polish (which translation is maintained by someone else): "**Warunki użytkowania**", "**Polityka prywatności**", "Pobierając mapy zgadzasz się z naszą **polityką prywatności** i **warunkami użytkowania**.", "Naciśnij „Pozwól”, jeśli zgadzasz się z naszą **polityką prywatności**".

In the above suggestions I used the same words, but with different inflection.

vshcherb:
shared_string_ should only be used as standalone phrase without any other names. Here is a bit tricky situation cause we need separate words to make them links and intentionally highlight. If you think that's gramatically in appropriate you can use the trick
```
By download you agree to policy (Terms Of Use) and agreements (Privacy Policy)
```

PR #15499, opened 2022-10-18
Fix bearing

ivanPyrohivskyi:
Fix bearing when enabled recalculate routing in case of reverse direction.
To issue https://github.com/osmandapp/OsmAnd-Issues/issues/1597


vshcherb:
@sonora - we've fixed long time ago introduced method cause it's quite buggy to be honest in modern situations with jumping current rotation and (!) points on the route connected to each other pretty close.

vshcherb:
image
this is on simulated track how current location to next route location looks as you see it generates quite a lot of errors around intersection

Issue #15457, opened 2022-10-11
[Bug] Map Tiles No More Transparent (Alpha Channel Flattened from 2022-09-20?)


Dollarius:
I am using Strava's heatmaps as Overlay maps (source https://anygis.ru/Web/Html/Download_en?app=OsmandSqlite)
The maps have an in-built transparent layer so that even with the "Overlay transparency = 100%" set in OsmAnd, it was possible to see the main map underneath:
<table style="width:100%">
  <tr>
<th><img src="https://user-images.githubusercontent.com/56419637/195055049-3092af2f-46a0-46fd-a297-08be2912e037.png"  width="300" height="auto"/> </th>
  </tr>
</table>  

The in-built transparency was correctly displayed up to OsmAnd nightly of 2022-09-16.
From the nightly of 2022-09-27, the overlay maps are always flattened, so that it is not possible to naturally see the base map underneath (unless the transparency slider is set below 100%, leading to a sub-optimal view of both the base and the overlay map)
<table style="width:100%">
  <tr>
<th><img src="https://user-images.githubusercontent.com/56419637/195056142-ef9b5563-d484-4095-8bd2-7df032718110.png"  width="300" height="auto"/> </th>
  </tr>
</table>  

There existed sort of a "transition time" between 2022-09-20 and 2022-09-26 in which:
1. the overlay map was non-transparent (i.e., flattened) upon starting the app
2. but if opening the settings and going on any menu (i.e., "Help")
3. then upon going "back" the transparency would be correctly rendered
<table style="width:100%">
  <tr>
<th><img src="https://user-images.githubusercontent.com/56419637/195057371-5ab4299c-db25-474b-98cc-76cb8c839efa.png"  width="300" height="auto"/> </th>
  </tr>
</table>  

This behaviour is identical both with "OpenGL rendering" enabled or disabled (from plugins/OsmAnd development).
I am using the nightly builds of the kind
OsmAnd-master-gnb-2022-xx-yy.apk
https://download.osmand.net/night-builds/?C=M;O=D
I am running Android 12

I hope this information will help you solve this bug.
Thank you very much for your work!

Zirochkabila:
Reproduced on OsmAnd~ 4.3.0#13869mqta, released: 2022-10-14 **without OpenGL activation**

Zirochkabila:
Reproduced on OsmAnd~ 4.3.0#13869mqta, released: 2022-10-14
Deactivate OpenGL rendering to see the result

[...]

PR #15523, opened 2022-10-20
Fix overlay transparency for legacy

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/15457

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/MapLayers.java
+++ b/OsmAnd/src/net/osmand/plus/views/MapLayers.java
@@ -208,10 +208,11 @@ public class MapLayers {
 		app.getSettings().MAP_TRANSPARENCY.addListener(transparencyListener);
 
 		overlayTransparencyListener = change -> app.runInUIThread(() -> {
-			if (useOpenGLRender) {
+			MapRendererView mapRenderer = mapView.getMapRenderer();
+			if (mapRenderer != null) {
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/views/MapLayers.java
```diff
@@ -257,12 +259,13 @@ public void updateLayers(@Nullable MapActivity mapActivity) {
...
 	public void updateMapSource(@NonNull OsmandMapTileView mapView, CommonPreference<String> settingsToWarnAboutMap) {
 		OsmandSettings settings = app.getSettings();
+		boolean useOpenGLRender = settings.USE_OPENGL_RENDER.get();
 
 		// update transparency
 		int mapTransparency = 255;
 		if (settings.MAP_UNDERLAY.get() != null) {
 			mapTransparency = settings.MAP_TRANSPARENCY.get();
-		} else if (settings.MAP_OVERLAY.get() != null) {
+		} else if (useOpenGLRender && settings.MAP_OVERLAY.get() != null) {
```
vshcherb, line 268: I don't think this code was changed since it was working in Legacy, looks strange

ivanPyrohivskyi: This code wasn't present in Legacy

PR #15543, opened 2022-10-23
Fix #14993 Custom POI disappears when screen turns off

### LINKED ISSUE #14993: Custom POI display disappears automatically when screen turns off (or switching apps)
### Description
If I have custom POI display enabled, when display turns off (or I switch to another app and back to OsmAnd), custom POI overlay is **_turned off automatically_**. I would expect the custom POIs to remain being shown, just like regular built-in POI overplay remains being shown. 
It is _very_ annoying and makes custom POIs practically unusable :cry: 

### How to reproduce?
- Create a custom POI search (i.e. `Search` / `Categories` / `Custom search`)
- Enable showing that custom POI in `Configure map` / `POI overlay`
- notice POIs are correctly displayed in orange circles
- wait for screen to turn off (or switch to another app and back to OsmAnd)
- when back in OsmAnd, notice that all orange circles have disappeared!

Video:

https://user-images.githubusercontent.com/156656/182373614-a077a24f-93b0-44f6-874c-a29687aa35d9.mp4

 
### Your Environment
OsmAnd Version: OsmAnd~ 4.2.6 (F-droid latest)
Android/iOS version: Android 10
Device model: Huawei P30 Pro

nazar-kutz:
We should not use a single comma sign as a delimiter when storing a list of POI selected to display.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/AppVersionUpgradeOnInit.java
+++ b/OsmAnd/src/net/osmand/plus/AppVersionUpgradeOnInit.java
@@ -69,11 +69,11 @@ import java.util.Set;
 
 class AppVersionUpgradeOnInit {
 
-	public static final String FIRST_TIME_APP_RUN = "FIRST_TIME_APP_RUN"; //$NON-NLS-1$
-	public static final String VERSION_INSTALLED_NUMBER = "VERSION_INSTALLED_NUMBER"; //$NON-NLS-1$
-	public static final String NUMBER_OF_STARTS = "NUMBER_OF_STARTS"; //$NON-NLS-1$
-	public static final String FIRST_INSTALLED = "FIRST_INSTALLED"; //$NON-NLS-1$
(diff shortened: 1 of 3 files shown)


OsmAnd/src/net/osmand/plus/AppVersionUpgradeOnInit.java
```diff
@@ -103,8 +107,10 @@ class AppVersionUpgradeOnInit {
 	public static final int VERSION_4_0_05 = 4005;
 	// 4006 - 4.0-06 (Merge widgets: Intermediate time to go and Intermediate arrival time, Time to go and Arrival time)
 	public static final int VERSION_4_0_06 = 4006;
+	// 4007 - 4.0-07 (Update type of "Selected POI" preference)
+	public static final int VERSION_4_0_07 = 4007;
```
vshcherb, line 111: It should be 4.3.01 - 4301

PR #15611, opened 2022-11-01
Suggest to switch to OpenGL

Chumva:


OsmAnd/src/net/osmand/plus/helpers/LogsHelper.java
```diff
@@ -0,0 +1,163 @@
...
+
+import java.io.BufferedWriter;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.PrintStream;
+import java.lang.Thread.UncaughtExceptionHandler;
+
+public class LogsHelper {
```
alex-osm, line 31: LogHelper or FeedbackHelper

PR #15735, opened 2022-11-15
Separate favs

Dima-1:


sonora:
Since we re-model this anyway: "favorites" would be the correct spelling,  particularly also for folder/file/group names.

"favourites" is the UK version. In OsmAnd we use American English by default and maintain "English (United Kingdom)" as an optional display and voice option.

sonora:
@alex-osm Any chance we still implement my request above on the occasion? Thx!

alex-osm:
Hello @sonora! I saw your message, but we did not change the name yet, since it may broke backward compatibility.

sonora:
Yes, thanks for reacting! Yes, I never changed it due to that, but I think this is the point in time to change for the future ... or else we will carry this on to now also new folder and new file names ...

vshcherb:
We have cloud names already and we have import relying on the file name since it's file name so it should be quite technical its probably low priority to fix cause it could break quite a lot. Or at least I couldn't guarantee all scenario fixes.

Please mind that import/export happens between different version and it needs to be checked in that way.

sonora:
Then we had a problematic commit in place for about 6 hours by merging https://github.com/osmandapp/OsmAnd/pull/15846. I only discovered at night, fixed it temporarily here https://github.com/osmandapp/OsmAnd/commit/c03acb128be90c5fe83966c7563d301b8da1a453 and clean here https://github.com/osmandapp/OsmAnd/commit/fe5d847b366e275c7ce8465942d39700b11a58db.

It's good now and I tested also between old and new versions. But who tried yesterday late afternoon's builds would have run into an issue.

vshcherb:
it's a bit confusing now but ok we needed separate constants anyway

sonora:
Yes, separating the constants drove me in the first place. Then I realized that's also the transparent way to make a clean transition. I think reading the code now will be more transparent in the future, because we explicitly separate between the old and the new convention.

PR #15737, opened 2022-11-15
Fix drawing route line (non-OpenGL)

cepprice:


OsmAnd/src/net/osmand/plus/views/layers/RouteLayer.java
```diff
@@ -459,7 +459,7 @@ public void drawLocations(RotatedTileBox tb, Canvas canvas, double topLatitude,
 							currentLocation.setLongitude(lastMarkerLocation.getLongitude());
 						}
 					}
-				} else if (useMapCenter) {
+				} else if (useMapCenter && app.getSettings().ROTATE_MAP.get() == OsmandSettings.ROTATE_MAP_BEARING) {
```
ivanPyrohivskyi, line 462: @alex-osm perhaps better just to delete this "if" ? In original legacy code these was no present.

PR #15761, opened 2022-11-17
Show Sunrise/Sunset on widget for map center location

nazar-kutz:


OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/SunriseSunsetWidget.java
```diff
@@ -90,6 +99,15 @@ public boolean isShowTimeLeft() {
 		return getPreference().get();
 	}
 
+	private void updateCachedLocation() {
+		RotatedTileBox tileBox = mapView.getCurrentRotatedTileBox();
+		LatLon newCenterLatLon = tileBox.getCenterLatLon();
+		if (!Algorithms.objectEquals(cachedCenterLatLon, newCenterLatLon)) {
```
vshcherb, line 105: equals is a bad method, for sunrise / sunset - 1000 m is enough distance to check. Or if quick method is needed than it could ``` Math.abs(lat - newlat)  > 0.001 ```

PR #15827, opened 2022-11-25
improved fix for delay calculation

sonora:


vshcherb:
It looks like a hack the issue is unclear, so it needs to be fixed on another leve.

sonora:
Sure. Looks like `AnnounceTimeDistances(ApplicationMode appMode, OsmandApplication app)` is only called from two places:

(1) In `VoiceRouter.java` there is no issue, we have always passed the appropriate appMode correctly:
https://github.com/osmandapp/OsmAnd/blob/291fd118eb7ccc91f5e18a7ca0260ad2f49d6c87/OsmAnd/src/net/osmand/plus/routing/VoiceRouter.java#L186-L188


(2) I am uncertain about `AnnouncementTimeBottomSheet.java`, if we do not need it analog there, i.e. pass the routingHelper's appMode (if not null):
https://github.com/osmandapp/OsmAnd/blob/291fd118eb7ccc91f5e18a7ca0260ad2f49d6c87/OsmAnd/src/net/osmand/plus/settings/bottomsheets/AnnouncementTimeBottomSheet.java#L53-L55

vshcherb:
2) I am uncertain about AnnouncementTimeBottomSheet.java, if we do not need it analog there, i.e. pass the routingHelper's appMode (if not null):

This is pure UI so it's used only to display settings and not to clacluate - so in the end I don't see any issue except some weird multithread possiblity but how

PR #15845, opened 2022-11-28
fix restore gpx from web

alisa911:


OsmAnd/src/net/osmand/plus/settings/backend/backup/items/GpxSettingsItem.java
```diff
@@ -32,6 +32,7 @@ public GpxSettingsItem(@NonNull OsmandApplication app, @NonNull File file) throw
 
 	public GpxSettingsItem(@NonNull OsmandApplication app, @NonNull JSONObject json) throws JSONException {
 		super(app, json);
+		updateFile();
```
vshcherb, line 35: the code is weird and different from original code whcih should be as simple as 
```
this.file = new File(app.getAppPath(FileSubtype.GPX), fileName);
```

Chumva: Logic is similar to previous code, it was simply moved to subtype class of FileSettingsItem. 

Simple use of new File(app.getAppPath(FileSubtype.GPX), fileName); will lead to wrong path with duplicate tracks folder

PR #15896, opened 2022-12-05
Fix differen parking icons

Dima-1:


OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
```diff
@@ -324,7 +324,16 @@ private void selectObjectsFromOpenGl(@NonNull MapSelectionResult result, @NonNul
 								}
 								if (onPathMapSymbol == null) {
 									amenity = getAmenity(result.objectLatLon, obfMapObject);
-									if (amenity == null) {
+									if (amenity != null) {
+										RasterMapSymbol rasterMapSymbol = null;
+										try {
+											rasterMapSymbol = RasterMapSymbol.dynamic_pointer_cast(symbolInfo.getMapSymbol());
+										} catch (Exception ignore) {
```
vshcherb, line 331: log.error or log.warn is must have

ivanPyrohivskyi: No need log.error here. It is just trying the cast value (analog instanceof in Java)

OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
```diff
@@ -324,7 +324,16 @@ private void selectObjectsFromOpenGl(@NonNull MapSelectionResult result, @NonNul
...
+										}
+										if (rasterMapSymbol != null) {
```
alex-osm, line 333: Where `if (rasterMapSymbol.getContentClass() == MapSymbol.ContentClass.Icon/)` ?

PR #15939, opened 2022-12-11
Fix fav import from profile osf

sonora:


alex-osm:
@sonora Thanks for fixing!

PR #15943, opened 2022-12-12
Fix map rotation behavior (fixes #15836)

### LINKED ISSUE #15836: Map orientation inconsistencies when switching profiles
I guess the following behaviors are bugs?

(1) Have your 'Browse map'  map view profile set to _'Fixed (north is up)'_, and your Pedestrian profile to _'Compass orientation'_. Now from Browese map, use the map screen's top left profile selector to switch to Pedestrian, the map orientation will adjust correctly. But now go back to Browse map, and it seems the map there will now be manually rotated  to the last bearing of the conpass orientation?

(2) A similar but different inconsistency can be observed if a profile change is triggered via the route config dialog. There we currently change the map view profile as soon as that dialog is entered (#15833), but it seems we apply the new profile's (e.g. Pedestrian's) **map orientation** only when and after the user leaves that dialog by tapping the turn-by-turn Start button. I.e. **not** when they enter the dialog, **and not** when they leave the dialog (after the route is calculated) via the back button to just view the route?

EDIT: It looks like both (1) and (2) are reproducible only under unknown pre-conditions, sometimes I see only (1) or (2), sometimes both.

sonora:


sonora:
This is tested and fixes #15836, here is the test plan: https://github.com/osmandapp/OsmAnd/issues/15836#issuecomment-1347541167.

sonora:
Squashed and re-submited, hence closing this one.

vshcherb:
ok, request has UI so it will be reviewed separately once 4.3 issues are over

PR #15992, opened 2022-12-16
Fix 15949

0xRe1nk0:


sonora:
So why do we merge all fixes lately to the r4.3 branch only but not to master?

Chumva:
We merge r4.3  with fixes to master

sonora:
@Chumva Ok, thanks! But should maybe do that at least every evening  because many of our active testers use our nightly builds and we would bypass them if we don't update master often enough, I think?

Chumva:
Yes, we can do it, I merge it almost every day, but not always in the evening.

sonora:
Ok, cool!

Additional note: https://github.com/osmandapp/OsmAnd/pull/15943 is rather straightforward and solves a notorious issue (occurring even for OsmAnd default settings, i.e. for many users), may be worth reviewing and also (squash-) merging back into r4.3, maybe?

PR #16001, opened 2022-12-16
fix pointsGroups web

alisa911:


OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
```diff
@@ -507,6 +511,28 @@ private void updatePoint(double lat, double lon, String description, String name
 				setBackgroundType(backgroundType);
 			}
 		}
+		
+		private void updatePoint(double lat, double lon, String description, String name,
```
vshcherb, line 515: duplicate method

OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
```diff
@@ -1569,6 +1595,25 @@ public boolean deleteWptPt(WptPt point) {
 
 			return points.remove(point);
 		}
+		
+		public boolean deleteWptPt(String wptName) {
```
vshcherb, line 1599: class is overloaded with lines of code, refactoring is needed and this method is not a good example that should be here especially wptName is not unique key!

OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
```diff
@@ -1599,6 +1663,25 @@ public void updateWptPt(WptPt point, double lat, double lon, String description,
 			modifiedTime = System.currentTimeMillis();
 			pointsModifiedTime = modifiedTime;
 		}
+		
+		public void updateWptPt(WptPt point, double lat, double lon, String description, String name,
+		                        String category, String color, String iconName, String backgroundType, String address) {
```
vshcherb, line 1668: class is overloaded with lines of code, refactoring is needed and this method is not a good example that should be here especially wptName is not unique key!

OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
alisa911: this is overload method from android. i can't change safely such methods

OsmAnd-java/src/main/java/net/osmand/GPXUtilities.java
alisa911: wptName is unique key for group

PR #16006, opened 2022-12-17
Fix #15130

### LINKED ISSUE #15130: Search > Address screen is take 5 second to load
https://user-images.githubusercontent.com/9365308/185894603-1338c846-8f80-47bd-9006-c68713e96db2.mp4

sonora:


OsmAnd/src/net/osmand/plus/search/QuickSearchDialogFragment.java
```diff
@@ -1329,6 +1329,7 @@ private void reloadCitiesInternal() {
 		if (SearchUICore.isDebugMode()) {
 			LOG.info("UI >> Start loading nearest cities");
 		}
+		updateCitiesItems();
```
vshcherb, line 1332: this could lead to concurrentmodificationexception as the search is not finished and running when it starts updating... but probably not if it's sync

PR #16042, opened 2022-12-22
Factorization and improvement for bounding box computation

Codain:
Hi,

This pull request is a first modest attempt to propose some improvement to OsmAnd, an App I'm using since years now. Please be gentle at reviewing it :).

While crawling the source code on my spare time I noticed that some stuff could probably be implemented a little bit different for maintenance and/or performance reasons. This PR is a first attempt, proposing to group in a single place code used to compute bounding boxes. By doing it I believe that using an helper class will improve clarity and maintenance. Also, it allows some bit of performance improvement, first inclusion being transparent and iteration through the coordinates no longer calling another method. Performance benefit can be arguable, and I didn't measure it, but when thinking about the potential number of points in a route, I'm convinced it is a good direction.

Please note that I'm not fully satisfied about the solution as there is one open topic remaining: There is some code duplication that could be avoided in QuadRectHelper if Location, Node and TargetPoint (and possibly LatLon) would share a common Interface to provide getLatitude/Longitude. But I was not willing to propose such a change in the data structure out of the blue, so if there is an interest for it I propose to add it to this PR, otherwise this PR is already a good step IMHO.

Please share your thoughts.

If things are going well (including welcoming of this PR) I might propose some other improvements from time to time.

Disclaimer: Despite all the care I put in reviewing the PR I was not able to execute and validate the modifications. Consequently, I hope I did not introduce any error.

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/data/QuadRectHelper.java
+++ b/OsmAnd-java/src/main/java/net/osmand/data/QuadRectHelper.java
@@ -1,0 +1,91 @@
+package net.osmand.data;
+
+import androidx.annotation.NonNull;
+
+import net.osmand.Location;
+import net.osmand.data.QuadRect;
+import net.osmand.osm.edit.Node;
+import net.osmand.plus.helpers.TargetPointsHelper;
+
+import java.util.List;
+
+/* To do: Inclusion mechanism here will not be able to work well if a set of points is crossing the -180 / +180 longitude line.
+ * To do: An Interface with getLatitude and getLongitude applied to Location, Node and TargetPoint (and LatLon?) would help to avoid dupplicating methods.
+ */
+public class QuadRectHelper {
+	public static QuadRect createRectForBoundingBoxDetermination() {
+        /* With following initiale values, first position will automatically initialize boundaries of the QuadRect,
+         * thus avoiding the need to detect first insertion.
+         */
+		return new QuadRect(Double.MAX_VALUE, -Double.MAX_VALUE, -Double.MAX_VALUE, Double.MAX_VALUE);
+	}
+
+	public static void includeLatLon(QuadRect r, double latitude, double longitude) {
+		r.left = Math.min(r.left, longitude);
+        r.right = Math.max(r.right, longitude);
+        r.top = Math.max(r.top, latitude);
+        r.bottom = Math.min(r.bottom, latitude);
+	}
+
+    /* This function does the same thing than includeLatLon but on a set of Locations.
+     * Code is dupplicated for optimization purposes, the set of points possibly being huge.
+     */
+    public static void includeLocations(QuadRect r, List<Location> list) {
+		double latitude;
+        double longitude;
+        
+        for (Location l : list) {
+			latitude = l.getLatitude();
+            longitude = l.getLongitude();
+
+            r.left = Math.min(r.left, longitude);
+            r.right = Math.max(r.right, longitude);
+            r.top = Math.max(r.top, latitude);
+            r.bottom = Math.min(r.bottom, latitude);
+		}
+    }
+
+    /* This function does the same thing than includeLatLon but on a set of Nodes.
+     * Code is dupplicated for optimization purposes, the set of points possibly being huge.
+     */
+    public static void includeNodes(QuadRect r, List<Node> nodes) {
+		double latitude;
+        double longitude;
+        
+        for (Node n : nodes) {
+			latitude = n.getLatitude();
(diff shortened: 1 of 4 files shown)


OsmAnd-java/src/main/java/net/osmand/data/QuadRectHelper.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public class QuadRectHelper {
+	public static QuadRect createRectForBoundingBoxDetermination() {
+		/* With following initiale values, first position will automatically initialize boundaries of the QuadRect,
+		 * thus avoiding the need to detect first insertion.
+		 */
+		return new QuadRect(Double.MAX_VALUE, -Double.MAX_VALUE, -Double.MAX_VALUE, Double.MAX_VALUE);
+	}
+
+	public static void includeLatLon(@NonNull QuadRect r, double latitude, double longitude) {
```
vshcherb, line 23: This method could be part of QuadRect.

OsmAnd-java/src/main/java/net/osmand/data/QuadRectHelper.java
```diff
@@ -0,0 +1,91 @@
...
+	public static void includeLocations(@NonNull QuadRect r, @NonNull List<Location> list) {
+		double latitude;
+		double longitude;
+		
+		for (Location l : list) {
+			latitude = l.getLatitude();
+			longitude = l.getLongitude();
+
+			r.left = Math.min(r.left, longitude);
+			r.right = Math.max(r.right, longitude);
```
vshcherb, line 42: This is copypastet from include latlon

OsmAnd-java/src/main/java/net/osmand/data/QuadRectHelper.java
```diff
@@ -0,0 +1,91 @@
...
+	public static void includeNodes(@NonNull QuadRect r, @NonNull List<Node> nodes) {
+		double latitude;
+		double longitude;
+		
+		for (Node n : nodes) {
+			latitude = n.getLatitude();
+			longitude = n.getLongitude();
+
+			r.left = Math.min(r.left, longitude);
+			r.right = Math.max(r.right, longitude);
```
vshcherb, line 60: This is copypaste from includelatlon

OsmAnd-java/src/main/java/net/osmand/data/QuadRectHelper.java
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	public static void includeTargetPoints(@NonNull QuadRect r, @NonNull List<TargetPointsHelper.TargetPoint> points) {
+		double latitude;
+		double longitude;
+		
+		for (TargetPointsHelper.TargetPoint p : points) {
+			latitude = p.getLatitude();
+			longitude = p.getLongitude();
+
+			r.left = Math.min(r.left, longitude);
```
vshcherb, line 77: This is copypaste from includelatlon

vshcherb:
Usually a Helper class is always worse than standalone class. I would suggest to extend from QuadRect and make class QuadRectLatLon with methods:
```
boolean insetLocation(double lat, double lon);
boolean isEmpty();
```

All this for-loops will be oneliner and probably not needed to be prepared in advance.

Of course QuadRectLatLon should properly handle < -180 > 180 cause it's important of course going over -180 meridian is an issue cause we need to deal with 2 quadrects, in the end it's tricky. Than it's easier to support with some extent [-360, 360]

Codain:
Hello @vshcherb and thanks for your review, I will prepare a second set of pushes.

Regarding your comments on duplication I fully agree, which is why I proposed in my PR to create an Interface since some are already inheriting from other classes. This way only one method accepting any type of object implementing this interface would have to be maintained. There are also other places where such interface would help. Regarding the name of this interface unfortunately LatLon is taken so I was thinking of `HasLatLon`:

```
public interface HasLatLon {
    public void getLatitude();
    public void getLongitude();
}

```
Can you please give me a feedback on this one? Probably it should be handled in a separate and first PR, unless you prefer that I add it here. Please also share how you see it.

vshcherb:
I and team usually try to follow simple rules:
- No static methods
- No any kind of duplication
- No overengineering (if it's possible to do less code / less classes / less methods) - Better!
- And no need to write code in advance (like in your case non-used methods)
- Of course we apply maximum levels on class / file / function etc.

What I see, quadrect is very generic class which shouldn't has any connection to lat/lon, so it should be LatlonRect which could either inherit or encapsulate

Issue #16065, opened 2022-12-26
pre-calculated traveltime is always displayed in seconds 


Rowin63:
### Description
The travel time in the popup is always displayed in seconds (time xxxx sex)

### Your Environment
OsmAnd Version: 4.3.4 but was earlier too
Android/iOS version: 13
Device model: S20+

**Maps used (online or offline):**   
offline

![Screenshot_20221226_102907_OsmAnd+](https://user-images.githubusercontent.com/95540936/209533426-84a11503-6cfe-48f8-8f3a-9b788d6271c4.jpg)

dmpr0:
Hi!
The toast shows up only if **OsmAnd Development** plugin is On, and made for development purpose.

sonora:
Yes, that's not really the "pre-calculated travel time", it's a benchmark figure for debugging. German translation shown in the screenshot may be questionable. Also, with this being reported a second time already (#13501), maybe it's about time we make this an explicit setting in the Dev plugin and not activate it automatically when the user activates the plugin?

Rowin63:
Ah, thank you. Did not realize it is connectet to the plugin.

Zirochkabila:
@Rowin63 can we close this issue?

Rowin63:
Yes of course, thanks

sonora:
@vshcherb So how about https://github.com/osmandapp/OsmAnd/pull/16066?

Issue #13501, opened 2022-01-09
Toast notification showing driving time in seconds


jase88:
### Description
After route calculation was successfull, a toast message is shown with details about the route and calculation. It shows the driving time in seconds, which is a technical representation.
For users it needs to be represented in their (locale depending) duration format. For example: 1h 30mins 15secs (although seconds are probably not relevant at all) 

![Screenshot_20211221-180104~3.png](https://user-images.githubusercontent.com/804836/148675475-cc6f2531-47c0-4dd9-a864-f8ba51bed429.png)

### How to reproduce?
- define a start and end point for a route
- start navigation 
- view upcoming toast message

### Your Environment
OsmAnd Version: OsmAnd+ 4.1.11
Android/iOS version: Android 10
Device model: Nokia 8

**Maps used (offline):**   
Germany_schleswig-holstein_europe
01.01.2021

mattmccutchen:
Peanut gallery comment: That toast is a feature of the "OsmAnd development" plugin, so making it user-friendly probably isn't a priority.  The travel time is shown in user-friendly format in several other places in the UI.

vshcherb:
Well it's a dev plugin feature and it's more useful to capture like that , please disable the plugin and use once it's needed cause it's distracting ;-(

PR #16066, opened 2022-12-26
Fix debugging KPIs showing by default (#16065, #13501)

sonora:


OsmAnd/res/values/strings.xml
```diff
@@ -4293,8 +4293,8 @@ Download tile maps directly, or copy them as SQLite database files to OsmAnd\'s
     <string name="search_offline_clear_search">New Search</string>
     <string name="map_text_size_descr">Text size for names on the map:</string>
     <string name="map_text_size">Map font size</string>
-    <string name="trace_rendering">Rendering debug info</string>
-    <string name="trace_rendering_descr">Display the rendering performance.</string>
+    <string name="trace_rendering">Performance debug info</string>
```
vshcherb, line 4296: I don't think now rendering performance is used, it was very annoying toast on each render frame where navigation toast is just once per calculation

sonora: Well, both are "annoying" unless you specifically need them to debug something specific.  ;)

Hence I think they should be "opt-in", even when the Dev plugin is active, to avoid further user confusion and more issues opened about it.

So the question becomes: _Can the debug toasts be productively used for debugging if there is one combined switch to activate / deactivate both in sync (like this PR would do)?_ I have tested that things work ok, no toast is lost, and should there are parallel toasts for rendering and route calculation (which there often are), they show in succession. So it could really be a viable approach, feel free to test.

PR #16100, opened 2022-12-30
Remove temporary files at application startup

nazar-kutz:


OsmAnd/src/net/osmand/plus/settings/backend/backup/FileSettingsItemReader.java
```diff
@@ -44,24 +48,44 @@ public void readFromStream(@NonNull InputStream inputStream, String entryName) t
 		OutputStream output = new FileOutputStream(downloadFile);
 		byte[] buffer = new byte[SettingsHelper.BUFFER];
 		int count;
-		boolean success = false;
+		boolean fileMovedSuccessfully;
 		try {
+			notifyOnFileMoveStarted(downloadFile);
 			while ((count = inputStream.read(buffer)) != -1) {
 				output.write(buffer, 0, count);
 			}
```
alex-osm, line 56: Looks like we do not need copy files through streams in case if `inputStream` is `FileInputStream`.
I suggest to add `File sourceFile` to `readFromStream` and just call `FileUtils.replaceTargetFile`. Thus we will significantly speedup import and remove temp file passed to this method from above.

PR #16115, opened 2023-01-02
Fix backup temp files

alex-osm:


vshcherb:
Do we need it for 4.3 ? Weather issue is already fixed in 4.3

PR #16128, opened 2023-01-03
Check old language codes when checking is language correct

nazar-kutz:


vshcherb:
Alternative fix was made

PR #16217, opened 2023-01-12
Use Runnable lambdas.

Isira-Seneviratne:
Convert `Runnable` anonymous classes to [lambdas](https://www.oracle.com/technical-resources/articles/java/architect-lambdas-part1.html) where possible.


vshcherb:
That's a big refactoring that needs to be tested in Android.

PR #16241, opened 2023-01-14
Fix route chart points drawing in OpenGL

cepprice:


vshcherb:
Does it require updates in documentation for rendering layers?

cepprice:
@vshcherb There is no info about order of route layer turn arrows or chart points in orders table. It should be?

vshcherb:
Yes - it should be here https://docs.osmand.net/docs/technical/algorithms/map-rendering-layers

cepprice:
@vshcherb  https://github.com/osmandapp/web/pull/63

PR #16244, opened 2023-01-14
Android Auto - NavigationInfo panel theming and clear notification

RaceTheMaSe:
Intended improvement:
- Direction panel is themed not so distracting in dark mode
- Notification in AA is cleared after reaching destination or stopping navigation

Setup Phone:
Android 9
Android Auto 8.5.624594
Headunit Reloaded 6.3 RC3 in Self-Mode

Both HUR and AA have DarkMode set where possible and yet the direction panel in the AA screen is using the AA light secondary color, resulting in a distracting blue panel, like in day mode.

This implementation is probably not correct and against some guideline and will perhaps not show correctly on other devices, but it at least points at the things to rework.

Second change is that the notifcation on the bottom in the rail widget was not cleared when navigation is stopped. This change is done by introducing another state variable in NavigationSession and triggering the refresh of notifications. If there is a better way, please correct it. Its not ideal to have so many seperate variables tracking the state.


Zirochkabila:
OsmAnd~ 4.4.0#14208mqta, Released: 2023-02-28
Done

PR #16265, opened 2023-01-17
Measurement units for Vehicle parameters

0xRe1nk0:


vshcherb:
There is a good abstraction which should be extended - https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/src/net/osmand/plus/settings/bottomsheets/VehicleSizeAssets.java.
It should specify min / max that could override routing.xml, it should specify Units conversion i.e. inches for mi/ft and it should be more specific so it could override generic CAR width to TRUCK width - so it should be a class with that logic that provides UI chips & conversions.

PR #16292, opened 2023-01-20
Add zh-Hans and zh-Hant to Map language

SteveLz:
Add Simplified and Traditional Chinese to Map language


vshcherb:
https://taginfo.openstreetmap.org/keys/name:zh-Hant
https://taginfo.openstreetmap.org/keys/name:zh-Hans

vshcherb:
Even though strings are present but they are unlikely to be used... https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/res/values/strings.xml#L3512

lang_zh_hans will be searched but not found

PR #16296, opened 2023-01-20
Fix 15559 turn indication

Dima-1:


Chumva:
Should already be fixed [here](https://github.com/osmandapp/OsmAnd/pull/17354)

Issue #16058, opened 2022-12-25
Hiking trail(s) get(s) split


mdykierek:
Consider this point: https://www.openstreetmap.org/node/1741133570#map=19/50.09174/18.96211
On both left and right hand side there is a blue marked hiking trail Szlak Obrońców Polskiej Granicy.
However if you select on either side you get either 57 km section or 3 km sections while it is a single relation.
Please let me know why it happens so that I can fix it in OSM or fix it in OsmAnd.

### Your Environment
OsmAnd Version: OsmAnd+ 4.3.4
Android/iOS version: Android 11
Device model: Xiaomi Mi 9T Pro

**Maps used (online or offline):**   
December 2022 with Live updates.

Zirochkabila:
OsmAnd~ 4.4.0#14064mqta, released: 2022-12-30

`50.09168, 18.96214`
Most likely, these are two segments of one route.
By the [link](https://www.openstreetmap.org/relation/1363303#map=11/50.1839/18.8841), you can see that the total length reaches 60.9 km, and the segments visible in the video are 57.4 and 3.16 respectively, together these segments form a complete route


https://user-images.githubusercontent.com/104760013/210045530-5ed0547a-6832-48a5-83d8-a6055adc204f.mp4

mdykierek:
So how does one make it a single segment route?
The only difference between left and right is that on the left (West), the route has track type and on the right (East) doesn't. They also different in surface. As you noted this is a single relation I had recently fixed by adding a missing way.

mdykierek:
So how does one make it a single segment route?
The only difference between East and West side of this point is that West has track type while East doesn't. They also differ in surface. As you noted, this is a single relation I had recently fixed by adding a missing way somewhere else.

Dima-1:
The issue was due to this way consisting of one segment
![2023-01-25_11-49](https://user-images.githubusercontent.com/31807095/216564439-5341d9f1-308a-4310-a7a1-f1363375d62e.png)

Now the problem remains only with the shield on this segment
![2023-02-02_19-36](https://user-images.githubusercontent.com/31807095/216564459-2b046080-35fc-4b1a-b83a-bdb73e363cd3.png)

vshcherb:
Move back with fix - https://github.com/osmandapp/OsmAnd/commit/1e8187831cc53e7940af44ca1a8cd07509ee79dc

Zirochkabila:
[...]

PR #16321, opened 2023-01-25
Fix Hiking trail(s) get(s) split #16058

Dima-1:


vshcherb:
Queue is part of algorithm connection but loadRouteSegmentTile is generic method that should return generic list.

vshcherb (changes requested):

OsmAnd-java/src/main/java/net/osmand/router/network/NetworkRouteSelector.java
```diff
@@ -549,6 +550,28 @@ private void loadData(NetworkRouteSegment segment, RouteKey rkey, List<NetworkRo
 		}
 	}
 
+	private void addSegmentsBiggerThanOneTile(Map<RouteKey, List<NetworkRouteSegment>> tiles,
+	                                          long tileId, TLongArrayList queue) {
+		for (Map.Entry<RouteKey, List<NetworkRouteSegment>> entry : tiles.entrySet()) {
```
ivanPyrohivskyi, line 555: if queue is null just return

Dima-1: queue is not null by default

OsmAnd-java/src/main/java/net/osmand/router/network/NetworkRouteSelector.java
```diff
@@ -549,6 +550,28 @@ private void loadData(NetworkRouteSegment segment, RouteKey rkey, List<NetworkRo
...
+			for (NetworkRouteSegment segment : entry.getValue()) {
+				RouteDataObject rdo = segment.robj;
+				if (rdo == null) {
+					continue;
+				}
+				long startPointTileId = NetworkRouteContext.getTileId(rdo.getPoint31XTile(segment.start), rdo.getPoint31YTile(segment.start));
+				long endPointTileId = NetworkRouteContext.getTileId(rdo.getPoint31XTile(segment.end), rdo.getPoint31YTile(segment.end));
+				if (queue != null) {
+					if (tileId != startPointTileId) {
```
vshcherb, line 564: Missing ? 
if (!visitedTiles.add(tileID)) {
				continue;
			}

Dima-1: This check in the begin of next while cycle (line 531)

ivanPyrohivskyi: Agree with Dima

PR #16355, opened 2023-01-30
Added new strings to match app name with app store name

dmpr0:


vshcherb:
Probably shouldn't be in generic strings.xml file and shouldn't be copypasted in all languages.

dmpr0:
> Probably shouldn't be in generic strings.xml file and shouldn't be copypasted in all languages.

Probably all this long descriptions https://github.com/osmandapp/OsmAnd/blob/appstore_name/OsmAnd/res/values/strings.xml/#L3829, should be moved in generic file too?

vshcherb:
Main question if we want contributors to translate or we want to have manual edition, I think once it's translated it should be moved to separate file too. So it's separated Marketing / Functional / no_translate.xml

sonora:
How about we leave these key strings in _strings.xml_ where they are crowd-visible and for good translations, but for actual use we use "controlled values":
- the ones we actually use in production
- which we copy to no_translate
- and periodically review in a controlled fashion to see if there have been decisive changes (in e.g. specific translations).

If you want an update to to our English strings, I offer make suggestions or reviews, just let me know what you think should be included as message.

PR #16500, opened 2023-02-20
Fixed 1585

Corwin-Kh:


OsmAnd/build-common.gradle
```diff
@@ -3,12 +3,12 @@ task printc {
 }
 
 android {
-	compileSdkVersion 31
+	compileSdkVersion 33
 	buildToolsVersion "30.0.3"
 	// compileNdkVersion "android-ndk-r17b"
 
 	defaultConfig {
-		targetSdkVersion 31
```
vshcherb, line 11: ? I don't think we planned or we are ready to update to 33?

Corwin-Kh: per-app feature works only for 33+ api. in general, it would be good to smoke-test all application prior to updating, but there should be no problems

PR #16519, opened 2023-02-22
Fix_16453

Chumva:


alex-osm:
We already have implemented cancellation feature in the import/export classes backend. There is no reason to operate with AsyncTask(s) directly. Moreover - forced interruption of AsyncTask is bad practice and should not be used.

Issue #16315, opened 2023-01-24
Too many recalculations of route


Spielmops:
Version 4.2.7:
Previously, when I couldn't follow a route further and had to deviate, I'd got a short silence, then "You're deviating from route," then silence, then "Route recalculated, distance, etc." in the driving direction. The app often tried to guide me back, but without an announcement.

Now with version 4.3.5, the app announces this after 120 meters: "Route recalculated, distance..." etc. always one after the other, often so fast that the announcement is no longer complete. And again and again... Totally annoying. And the navigation always just wants to go back to the old route. And this continues until the app finds a new route which is shorter than the previous one.

My wish: an announcement "You have deviated from the route for ... meters", then the route will be recalculated as quickly as possible and in the direction of travel!

Spielmops

Zirochkabila:
@Spielmops 
Please clarify:
1. The device you use, OS version
2. Route, start and finish coordinates 
3. The presence of traffic restrictions
4. A short video and screenshots where we could see the error
5. Logcat file: Menu > Help > Send logcat log

Spielmops:
1. Cubot Max, LineageOS 14, Samsung S9+, LineageOS 18, Samsung Tab S6 Lite, LineageOS 19
2. independant of rout start and stop
3. no restrictions
4. I'm driving, so no video
5. Logcat of which device?

If I tell you something about a difference, why do you need any info about device, OS, route start or stop and so on? 

The difference is: it has been OK, it is with the new Osmand-Version not OK any more.

If you need more, then forget it and close this issue ...

Spielmops

vshcherb:
Recorded GPX track is needed at least part of it and a destination point, so we could simulate navigation part and  test it.

vshcherb:
> If you need more, then forget it and close this issue ...

The issue could be fixed only if it's reproducible and navigation issues are mostly location specific, so if you would like to get it fixed we need more details. If you want us to be notified of possible bug, then you can close the issue.

Spielmops:
https://hartmut.homelinux.org/Downloads/2023-01-21_11-41_Sat.gpx
gpx is not accepted here ...

The destination point is, where the track ends. The app wanted me to use the motorway A27 up to Walsroder Dreieck and then A7 back to Dorfmark.

Spielmops

That is not the only track ...

sonora:
The issue description sounds a bit like my observation here https://github.com/osmandapp/OsmAnd/issues/14466#issuecomment-1356832126.

[...]

PR #16548, opened 2023-02-27
Fix route recalculation

Dima-1:


sonora:
Possible that this also fixes https://github.com/osmandapp/OsmAnd/issues/16315.

PR #16564, opened 2023-02-28
Fix_16445

sonora:


Chumva (commented):
You can use this format for similar commits - "Fix #16455", it will make access to the issue easier by clicking on #number

sonora:
Will do, Vitaliy , thank you!

I had abandoned that lately because I remember that merging a commit (or PR, can't remember) named like that may automatically close the referenced Issue (without advancing it to our correct Review etc. stage). Not sure if that was desired. But I will gladly return to this easier github name convention.

PR #16830, opened 2023-03-28
Fix #16751 Accidental change of map rotation in "North is up (fixed)"…

### LINKED ISSUE #16751: Accidental change of map rotation
I believe that changing the dead zone for initiating the manual map rotation from 30 deg to just 5 deg here https://github.com/osmandapp/OsmAnd/pull/16746 increases the risk of accidentally  changing to manual map rotation when just pinch-zooming. Manual rotation is actually a rather rare use case for the majority of users...

sonora:
… mode

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
+++ b/OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
@@ -1621,6 +1621,7 @@ public class OsmandMapTileView implements IMapDownloaderCallback {
 
 		private static final float ANGLE_THRESHOLD = 5;
 		private static final float ANGLE_ZOOM_THRESHOLD = 15;
+		private static final float ANGLE_THRESHOLD_FIXED = 30;
 		private static final float ZOOM_THRESHOLD = 0.1f;
 		private static final float MAX_DELTA_ZOOM = 4;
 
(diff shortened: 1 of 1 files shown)


vshcherb:
This is default mode and I would say it should be the mode where angle threshold should definitely be 5 so user can freely rotate without a glitch

sonora:
Shall we perhaps amend the PR to apply for the 2D mode only?

The crux is exactly that
- "North is up" is the default mode every casual user gets (and many never intend to change it),
- and most users use OsmAnd simply like a map for orienteering purposes, expect North to be up,
- and the use case of twisting the map to view the landscape under a certain angle is comparatively rare, and possibly mostly needed in the 3D mode?

In addition,
- accidental map rotation can go unnoticed,
- or if noticed, some "normal" users are not aware how to reset it,
- or get tired of having to manually reset the orientation after many zoom operations.

I am really convinced by the feedback we see in issues, I get in person, and my own personal use last week that we have a cause for action here.

vshcherb:
This what need to be precisely tested first

Map gestures sensitivity in north up fixed:
- Rotating - 5 degrees
- Zooming - 0.15z
- Zooming & Rotating - 15 degrees, zoom > 0.3, 30 degrees zoom > 2

vshcherb:
Second we could speak about the sensitivity numbers but need to be precise

vshcherb:
Btw we re still fixing to have exact constants in the code, so please don't hurry with code

vshcherb:
Btw we re still fixing to have exact constants in the code, so please don't hurry with code, testing environment and numbers are much more important

PR #16835, opened 2023-03-29
Fixed min zoom

Corwin-Kh:


sonora:
Small hint: `Math.log(...)` is a rather expensive operation to perform multiple times at every simple gesture interaction.

If it is unavoidable to use code like `Math.log(relativeToStart)` instead of an approximation, I guess we should at least declare `Math.log(2)) * 1.5` as a constant rather than permanently recalculating it?

vshcherb:
@sonora sounds logical

sonora:
@vshcherb Not sure if this is still WiP?

If not, I will make a PR. Essentially `Math.log(2)*1.5` is about 1.04, so it's not very sensible to have such an expensive operation to multiply something whit 1.04. Chances are, we could drop entirely, and the zoom operation would not noticeably change...(?)

vshcherb:
ok, it's not in WiP but I think we need to increase threshold constants

Issue #16509, opened 2023-02-20
Track appearance not correctly indicated


sonora:
- [x] Recorded track in default appearance, i.e. appearance has never been customized: Tap on the track, then on its context menu tap the appearance button. You will receive this dialog, where neither the current color button is circled, nor is the current track width marked:

This is for the example of the OsmAnd Map style, but the issue is also present for all other Map styles.

- [x] Note: The new 'Default track appearance' dialog accessible via `Configure map > Tracks` does circle the correct color all right, but also has the issue that the track width in use is not correctly reflected.

![Screenshot_20230220-231353_OsmAnd~](https://user-images.githubusercontent.com/845735/220206523-bd657ef8-8167-4ca1-879d-1d2081a45ada.jpg)

Moved -> https://github.com/osmandapp/OsmAnd/issues/16584

Zirochkabila:
OsmAnd~ 4.4.0#14222mqta, Released: 2023-03-07
Record a new track > Options > Appearance > Current color and line width are not highlighted
<table style="width:100%">
  <tr>
<th><img src="https://user-images.githubusercontent.com/104760013/223477014-845ec736-f963-4b48-97c5-10fcca221647.jpeg"  width="300" height="auto"/> </th>
  </tr>
</table>

sonora:
@Zirochkabila Can I please ask you to verify: I think the color is now correctly indicated, but not he width? If e.g. I use the map style 'Touring view' and its track default  it shows me width "Default" and "1", but that' s certainly not correct?

sonora:
Here's how it looks for me, the color is now right, but "Custom" with "width="1 seems wrong?
![Screenshot_20230320-222427_OsmAnd~](https://user-images.githubusercontent.com/845735/226469704-732ae5ba-d413-4501-b62a-69d6b046e6ea.jpg)

Zirochkabila:
OsmAnd~ 4.4.0#14261mqta, Released: 2023-03-23
When switched to Touring view, the width of the route line changes (V1 and V2)

Configure map > Map style > Touring view

https://user-images.githubusercontent.com/104760013/227168141-8f2416b3-75c8-4c40-b889-fec38e73049b.mp4

sonora:
I could actually fix this via this commit: https://github.com/osmandapp/OsmAnd-resources/pull/949.

[...]

PR #16838, opened 2023-03-29
Track appearance fix

0xRe1nk0:


sonora:
Perhaps consider https://github.com/osmandapp/OsmAnd/issues/16509#issuecomment-1491100546 before merging.

vshcherb:
We shouldn't parse individual rules that's too complicated

OsmAnd/src/net/osmand/plus/track/TrackDrawInfo.java
```diff
@@ -143,6 +150,29 @@ public void updateParams(@NonNull OsmandApplication app, @NonNull GpxDataItem gp
 	@Nullable
 	private String getRenderDefaultTrackWidth(@Nullable RenderingRulesStorage renderer) {
 		if (renderer != null) {
+			float defaultWidth = 0.0F;
+			RenderingRule gpxRule = renderer.getRenderingAttributeRule("gpx");
+			Map<Float, String> trackWidthMap = new HashMap<>();
+			if (gpxRule != null && gpxRule.getIfElseChildren().size() > 0) {
+				List<RenderingRule> rules = gpxRule.getIfElseChildren().get(0).getIfChildren().get(0).getIfElseChildren();
```
vshcherb, line 157: ???

sonora:
@vshcherb Potential solution without much complexity could be we change our render files to indicate the resulting strokewidth _explicitly_, e.g.
`renderingProperty attr="currentTrackWidth" name="GPX width" description="Width of displayed GPX tracks" type="string" possibleValues="Fine (2),Thin (4),Medium (8),Bold (12)"`
which (as per our already existing code) will cause the button captons on our appearance dialogs to change accordingly, i.e. indicate the resulting strokewidth in parentheses. 

Could even parse these strings to derive the starting value for the "Custom" selection slider if we wanted.

Issue #16173, opened 2023-01-06
Wrong route with car profile


clementcontet:
# 🐞 routing report

### Routing engine
- [X] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile
Car (works fine with pedestrian or bicycle)

### Start and end points
45.77704, 4.82428 to 45.77699, 4.82360

### Actual and expected routes
Actual: 
<table style="width:100%">
  <tr>
<th><img src="https://user-images.githubusercontent.com/7261426/211017639-27ebc552-9ab4-4196-b729-4d9f68735ddd.jpg"  width="300" height="auto"/> </th>
  </tr>
</table>  

Expected:
https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=45.77704%2C4.82428%3B45.77699%2C4.82360

### Is this a regression?
I don't know

## 🌍 Your Environment

**OsmAnd Version:**
<pre><code>4.3.5</code></pre>

**Device and Android/iOS version:**
Android 

**Maps used (online or offline):**
- [X] Offline maps offered within the OsmAnd app for download (France Auvergne-Rhône-Alples from January 1st 2023)
- [ ] Online (tile / raster) maps

vshcherb:
Looks like there is a bug that road is impassable ? https://www.openstreetmap.org/way/22476166#map=19/45.77695/4.82450

clementcontet:
Indeed there is a barrier in the middle of the road (https://www.openstreetmap.org/node/9417509524) but it shouldn't make the whole road impassable

vshcherb:
https://www.openstreetmap.org/node/9417509524 - yes the issue is that it's 1st node and it misinterprets 1st node that 1st segment is fully impassable so it tries to  route to the east and not west which is correct.
![Screen Shot 2023-01-13 at 17 24 46](https://user-images.githubusercontent.com/1042025/212356208-464c3501-5d44-43d9-90cc-7a574b1ccd74.png)

xmd5a2:
@vshcherb yes, issue is in this bollard. If you remove it, then the route builds normally. Looks like a routing engine bug.

sonora:
Not sure if it's unavoidable,  but by my testing some commit in the last 5 days led to an increase in route calculation time by about 10-20%, could it be?

Zirochkabila:
@sonora What kind of commits are you talking about? 
[...]

PR #16847, opened 2023-03-30
Fix wrong route with car profile #16173

Dima-1:


vshcherb:
even if the idea correct it's definitely wrong request
1. why only for obstacle < 0?
2. why applicable only for non-reverse and not for reverse mode (not symmetric)?

And in general that it appears deep inside related to start / end logic looks conceptually wrong

PR #16857, opened 2023-03-31
add limit poi web

alisa911:


OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
```diff
@@ -550,7 +550,7 @@ protected void searchPoiIndex(int left31, int right31, int top31, int bottom31,
 						skipTiles.add(skipVal);
 					}
 					codedIS.popLimit(oldLim);
-					if (req.isCancelled()) {
+					if (req.isCancelled() || req.limitExceeded()) {
```
vshcherb, line 553: looks wrong

OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapIndexReader.java
```diff
@@ -1555,6 +1555,22 @@ public static SearchRequest<Amenity> buildSearchPoiRequest(int sleft, int sright
 
 		return request;
 	}
+	
+	public static SearchRequest<Amenity> buildSearchPoiWebRequest(int sleft, int sright, int stop, int sbottom, int zoom,
```
vshcherb, line 1559: PoiWeb ? looks weird in backend

PR #16896, opened 2023-04-05
Add train profile

0xRe1nk0:


williamdes:
Thank you @Chumva!

I am not sure why it's not working properly with France (it says it can not find a path when I specify two train stops) but I will try again when I will be in a train

PR #16988, opened 2023-04-18
Fix maxspeed:hgv doesn't work

Dima-1:


OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
```diff
@@ -769,7 +771,22 @@ public synchronized NextDirectionInfo getNextRouteDirectionInfo(NextDirectionInf
 	}
 
 	public synchronized float getCurrentMaxSpeed() {
```
vshcherb, line 773: nope this is not correct fix, cause maxspeed should be properly calculated inside route and not in Routing Helper.

So it should maxspeed should be propagated from RouteCalculateResult somehow and influenced by RouteResultPreparation.

PR #17010, opened 2023-04-21
Add OsmAndOwner to obf

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd-Issues/issues/1885


vshcherb:
- [ ] Changes here https://github.com/osmandapp/OsmAnd-resources/pull/962/files
- [ ] We need writer
- [ ] We need utility to add / update metainfo to the existing map

PR #17046, opened 2023-04-24
Hide "Shutter Sound" preference if it isn't allowed by system to disable it

nazar-kutz:


sonora:
@vshcherb From a support perspective it would actually be preferable to gray out unavailable options with a caption "Not available on this device", instead of hiding them, which typically wil lraise support requests that people cannot find certain options.

PR #17099, opened 2023-05-02
16725 add search to configure map tracks

Corwin-Kh:


sonora:
Just tested, this somehow still has issues. I think it has e.g.to do with the occurrence of the underscore character `"_"` in track file names.

I have e.g. many tracks recorded on a Monday, many of which have names like "yyyy-MM-dd_HH-mm_**Mon**_xyz.gpx", but none of these is found: If looking for "`Mon`" or "`_Mon`", the search returns no hits.

Issue #16725, opened 2023-03-17
Add Search to Configure Map > Tracks


dmpr0:
## ToDo
Add search to new **Tracks** screen in **Configure Map**.

## Tracks
- [x]  Add Search action to AppBar
- - [x] Icon `ic_action_search_dark`
- - [x] `android:contentDescription` = `Search`

<table style="width:100%">
  <tr>
    <th>AppBar</th>
  </tr>
  <tr>
    <td><img src="https://user-images.githubusercontent.com/9365308/231990460-3f498748-874c-43fd-ad1e-76cacaad845a.png"  width="360" height="auto"/> </td>
  </tr>
</table>   

#### AppBar
- [x]  Back button `android:contentDescription` = `Back`
- [x] Clear button
- - [x] Show when something is typed
- - [x] `android:contentDescription` = `Clear search field`
- - [x]  Touch target size = 48 dp
- [x] Start state Text hint **Search track by name**

<table style="width:100%">
  <tr>
    <th>Start state</th>
    <th>Few tracks selected</th>
    <th>All tracks selected</th>
    <th>Empty state</th>
  </tr>
  <tr>
    <td><img src="https://user-images.githubusercontent.com/9365308/232034615-d049c4f5-d430-4a44-a8c2-c45d03fb6002.png"  width="180" height="auto"/> </td>
    <td><img src="https://user-images.githubusercontent.com/9365308/231966460-f13bb581-73af-4223-abfe-1bec92eecdec.png"  width="180" height="auto"/> </td>
    <td><img src="https://user-images.githubusercontent.com/9365308/231966468-e4257d8b-f10c-409c-88a3-4bdd445390c6.png"  width="180" height="auto"/> </td>
    <td><img src="https://user-images.githubusercontent.com/9365308/231966470-d176b068-0b3e-4b6b-8bcb-57c9a074c2e1.png"  width="180" height="auto"/> </td>

  </tr>
</table>

dmpr0:
## Review
- [ ]  Open keyboard by default

- - - - 

- [ ]  Update **On map** track list, after hitting **Apply** on Search screen
<table style="width:100%">

  <tr>
    <td>
<video source src="https://user-images.githubusercontent.com/9365308/235642521-7cffff6d-2079-4cef-b293-0f05f519c2d1.mp4"  width="300" height="auto"/>  </td>
  </tr>
</table>

Zirochkabila:
[...]

PR #17146, opened 2023-05-09
Fixed search query; fixed sorting; #16725

Corwin-Kh:


OsmAnd/src/net/osmand/plus/configmap/tracks/SearchableTrackAdapter.kt
```diff
@@ -75,11 +76,17 @@ class SearchableTrackAdapter(
...
+            collator = CollatorStringMatcher(
+                filterTracksQuery,
+                CollatorStringMatcher.StringMatcherMode.CHECK_CONTAINS)
+        }
         for (itemObject in allItems) {
             if (itemObject is TrackItem) {
                 hasTrackItems = true
                 if (Algorithms.isEmpty(filterTracksQuery)
-                    || itemObject.name.contains(filterTracksQuery!!)) {
+                    || collator != null && collator.matches(itemObject.name.lowercase())) {
```
vshcherb, line 89: lowercase is not needed - probably it causes lags

Issue #17086, opened 2023-04-29
Introduce map rotation North fixed / Less sensitive calculate map rotation


billmaurer:
It is very easy to accidentally alter the map orientation by many different gestures on the screen. 

This could be done by adding a checkbox "Lock map to this orientation" to the Map Orientation dialog.

The dialog can be opened in Profile - General settings - Map orientation or by long pressing the Map Orientation button on the top left of the map display. 

My preference is to keep the map orientation to always point north while in hiking mode.

Zirochkabila:
Related to https://github.com/osmandapp/OsmAnd/issues/7282

billmaurer:
Not true. This has nothing to do with the compass or bearing mode or anything like that.

The gesture for zooming in and out is very similar to the gesture for rotating the map. If you let your fingers drift even slightly doing the zoom gesture the map will flip out of the orientation mode you had previously set and into rotation mode. This is highly undesirable.

Nine-Tailed:
https://github.com/osmandapp/OsmAnd/discussions/16836

billmaurer:
Yes, this is definitely the same as #16836.

So I'm not the first to be annoyed by this! :-)

arncor:
I (and a few friends too) use osmand a lot and we all want a "north up" only view. We almost never need manual rotation. 

In our opinion the low threshold is very very annoying.

Zirochkabila:
@billmaurer in item **4. A large button for locking/unlocking the maps turns** (7282) - this is exactly the feature you are asking about, isn't it?
@Nine-Tailed Regarding the zoom change thing - in the latest nightly version it works well, I think (video attached)
Regarding the map tilting - I agree, it's probably worth fixing, @alex-osm  - comment with video added https://github.com/osmandapp/OsmAnd/discussions/16836#discussioncomment-5764948

OsmAnd~ 4.5.0#35865m, Released: 2023-04-29

https://user-images.githubusercontent.com/104760013/235340759-8e7ed62a-4c72-40d1-9bc0-db6367ad9def.mp4

H--W:
I would also like to have this feature. Some underpowered devices are only usable in continuous north up orientation. Also many navigators expect a map to follow a paper map/chart paradigm. I would be willing to edit a text config file to get this feature. Perhaps this should be a profile setting? Rgds, -H- ver 4.4.7 Android

sonora:
[...]

PR #17197, opened 2023-05-16
Fix accidental map rotation in "North is up" mode (#16836, #17086)

sonora:


mnalis:
Thanks, I've [tried](https://github.com/mnalis/OsmAnd/actions/runs/5020936569) it quickly against recent `master`, and it seems to work as advertised:

https://github.com/osmandapp/OsmAnd/assets/156656/1c7d4865-23a4-491f-a72f-ef3e5844a4c3

sonora:
You can set the threshold to anything you like in `NORTH_LOCK_ANGLE_THRESHOLD = 30;`, 30 degrees was the value we had used for many years.

cyberman54:
Could we please merge this soon? The new UI behaviour is annoying, leads to permanent unwanted rotations of map in "North up fixed" mode.

Boothy99:
Looks like an excellent solution that should keep everyone happy, thanks @sonora 👌

vshcherb:
North up (fixed) mode should be removed as never existed in the app except the naming. I believe it should be Map Rotation lock which will apply to any rotated angle and could also prevent accidental map changes.

vshcherb:
PR changes default behavior that locks rotation in default mode, it's unacceptable. 
We need to either change rotation algorithm (as we now will explore again) or create a way to lock mode and explain to the users

cyberman54:
Why can't we just go back to the previous implementation? I never had a rotation UX problem with OSMand before. 

The current UX is very annoying. Do we really want to keep this for 6 months, until next regular release... ?

vshcherb:
Previous implementation also had accidental map rotation and caused friction for map rotation

mnalis:
People following this (including me :smile:) should try how 4.5.1beta behaves, as there have been changes, e.g. https://github.com/osmandapp/OsmAnd/issues/17086#issuecomment-1569285378

Nine-Tailed:
now you can’t use gestures without irritation, they just broke what worked for years without any complaints.

PR #17245, opened 2023-05-22
UI fixes for track folder screens

Chumva:


sonora:
@Chumva , @dmpr0 The chosen sorting seems not always applied correctly. When playing around with the two different Tracks screens, at some point you re-enter MyPlaces>Tracks and the folder sorting still shows A-Z, but the folders seem sorted randomly (or maybe by last access, not sure).

vshcherb:
@Chumva please include that issue in the hotfix list

sonora:
I've broken it out here: https://github.com/osmandapp/OsmAnd/issues/17278.

Issue #15559, opened 2022-10-25
Display C (Straight) instead of KL (if it's uniquely determined)


issteve:
# 🐞 routing report

### Routing engine

- [X] OsmAnd's in-app offline routing

### Routing Profile

CAR

### Start and end points

Schweizerhof, Abzweigung Jostal
https://www.openstreetmap.org/#map=17/47.98066/8.14132
Following street direction from west to east 

### Actual and expected routes

This is a real screen shot!

<table style="width:100%">
  <tr>
<th><img src="https://user-images.githubusercontent.com/63864296/197719447-1b2c4f10-8753-47d9-9784-eaea74c1618f.jpg"  width="300" height="auto"/> </th>
  </tr>
</table>  

The voice announcement is misleading "links einordnen" ("get left") like shown in the top left of the screenshot, but should be like it is at the exact same time(!) correctly shown in the top center. So there should be no announcement at all as it's just following the normal road in the normal direction straight forward. So the screen-navigation (top center) looks correct but the voice navigation says at the same time misleading things (top left).

### Is this a regression?

No, it's the same problem for months as it was(?) an error in the map data on openstreetmap.org. I (tried to?) fix it there around July/August(!). It does work with all the navigations on openstreetmap.org for months, but the misleading voice announcement keeps persisting even with the September and current October update of the map data. And even by showing it correct in small in the top center of the screen.

## 🌍 Your Environment

**OsmAnd Version:**

<pre><code>
OsmAnd+ 4.2.7
</code></pre>

**Device and Android/iOS version:**

Google Pixel 4a, Android 13

**Maps used (online or offline):**

- [x] Offline maps offered within the OsmAnd app for download.

Deutschland Baden-Württemberg 01.10.2022

**Anything else relevant?**

It happened on numerous other places before, that I changed an error or even couldn't find any error in the openstreetmap.org data. The navigation on openstreetmap.org even worked correctly and as expected, but the OsmAnd voice routing kept even after monthly updates saying the wrong misleading things. 
In the OsmAnd descriptions I found so far only the hint, it's using openstreetmap data for navigation, but obviously it doesn't (at least not the current downloaded one).

So how to get OsmAnd using the current downloaded offline data for voice navigation?

And if this on purpose, couldn't at least somewhere easier to find be a hint what data is used for voice navigation and when and how it gets updated?

vshcherb:
IF we reproduce we need to make small map example to test and fix it later

[...]

PR #17354, opened 2023-06-05
Avoid keep left/right on through moving

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/15559

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
@@ -234,6 +234,7 @@ public class RouteResultPreparation {
 		determineTurnsToMerge(ctx.leftSideNavigation, result);
 		ignorePrecedingStraightsOnSameIntersection(ctx.leftSideNavigation, result);
 		justifyUTurns(ctx.leftSideNavigation, result);
+		avoidKeepForThroughMoving(result);
 		addTurnInfoDescriptions(result);
 	}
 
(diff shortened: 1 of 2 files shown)


OsmAnd-java/src/main/java/net/osmand/router/TurnType.java
```diff
@@ -284,6 +284,35 @@ public static String lanesToString(int[] lns) {
...
+	public boolean isOnlyThroughActive() {
+		if (lanes == null) {
+			return false;
+		}
+		boolean hasThrough = false;
+		for (int h = 0; h < lanes.length; h++) {
+			boolean active = lanes[h] % 2 == 1;
+			int primary = TurnType.getPrimaryTurn(lanes[h]);
+			if (primary == 0) {
+				primary = 1;
```
vshcherb, line 296: 1 constant ?

OsmAnd-java/src/main/java/net/osmand/router/TurnType.java
```diff
@@ -284,6 +284,35 @@ public static String lanesToString(int[] lns) {
         return s.toString();
 	}
 
+	public boolean isOnlyThroughActive() {
```
vshcherb, line 287: method could be more generic and simpler ?
```
countTurnTypeDirections(TurnType checkActive, boolean onlyActive);
```

So check will look like
```
int cnt = countTurnTypeDirections(TurnType.C, true);
int cntAll = countTurnTypeDirections(TurnType.C, false);
if(cnt > 0 && cnt == cntAll) {
 .. 
}
```

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -1999,5 +2000,28 @@ private static double measuredDist(int x1, int y1, int x2, int y2) {
...
+			TurnType turnType = curr.getTurnType();
+			if (turnType == null) {
+				continue;
+			}
+			if (!turnType.keepLeft() && !turnType.keepRight()) {
+				continue;
+			}
+			RouteSegmentResult prev = result.get(i - 1);
+			String turnLanesPrev = getTurnLanesString(prev);
+			if (turnLanesPrev == null || !turnLanesPrev.contains("through")) {
```
vshcherb, line 2015: why do we check prev for through ?

ivanPyrohivskyi: We check if present lanes tag in previous segment for to be sure that "through" instruction is not code generated, but from real data.

OsmAnd-java/src/main/java/net/osmand/router/TurnType.java
ivanPyrohivskyi: Method countTurnTypeDirections will consist in most same code as isOnlyThroughActive.

PR #17427, opened 2023-06-15
fix getAttributes

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2053

PR #17477, opened 2023-06-22
Fix #1976 Async for add\delete favourite point

### LINKED ISSUE #1976: Bugs: The "Next turn" icon and "Lane guidance" icons have several bugs
I mention them together because together they provide safe driving.

As the attached screenshots show, they have several bugs:
1. The next turn indicator misses turn/fork on the road, and focuses far ahead.

2. OsmAnd provides no visible/audio hazard warnings at junctions, where incoming traffic can suddenly appear in front of the driver at high speeds.

3. At junctions, the driver should be warned to avoid the lanes closest to the merging side lane.
    OsmAnd does not do this.

4. The top bar often shows only a manoeuvre (such as U-turn icon), but does not show the road's name.

5. OsmAnd often fails to snap the route to the road. The route line often jumps across roads. 

6. Often a flyover/underpass has exit lanes just before it. In such cases, OsmAnd does not direct the driver to take the proper lanes.

7. Before exiting from a road, we have to change lanes. But OsmAnd does not guide the driver to execute this lane-changing operation well before the exit comes up. Changing lanes at the last minute may be dangerous!

[Bugs - Next turn indication, Lane guidance.pdf](https://github.com/osmandapp/Osmand/files/45203/Bugs.-.Next.turn.indication.Lane.guidance.pdf)

0xRe1nk0:


OsmAnd/res/values/strings.xml
```diff
@@ -11,6 +11,8 @@
 	Thx - Hardy
 -->
 
+    <string name="saving_points">Saving points…</string>
```
vshcherb, line 14: not used

PR #17594, opened 2023-07-11
Adding new parameter - custom gpx filename - to AIDL gpx navigation

pab2d:
Based on the current implementation all gpx filenames used bei gpx navigation routines have the "same" filename: route (x).gpx. Adding the ability to set the filename makes it a lot easier to delete specific gpx files.


Chumva:
We will use a similar approach, but the file name will be available in NavigateGpxParams and AGpxFile after it is saved because it can be changed during saving [duplicate files](https://github.com/osmandapp/OsmAnd/blob/1e536521c6fdff2c220c8d6c1185de1271799015/OsmAnd/src/net/osmand/plus/helpers/ExternalApiHelper.java#L669-L672)

PR #17639, opened 2023-07-17
Alpine hiking path

0xRe1nk0:


OsmAnd/res/values/strings.xml
```diff
@@ -10,7 +10,12 @@
 	- For wording and consistency, please note https://docs.osmand.net/docs/technical/contributions/translating-osmand
 	Thx - Hardy
 -->
-
+    <string name="rendering_value_sacScale_name">SAC</string>
+    <string name="rendering_value_sacScale_description">Swiss Alpine Club</string>
+    <string name="rendering_value_caiScale_name">CAI</string>
+    <string name="rendering_value_caiScale_description">Club Alpino Italiano (Italian Alpine Club)</string>
+    <string name="difficulty_classification_description">Difficulty classification types for the routes.</string>
+    <string name="difficulty_classification">Difficulty classification</string>
```
alex-osm, line 18: Why name is not "rendering_attr_alpineHiking_name" ?

sonora:
@vshcherb   cai_scale is an inactive proposal, or do I overlook anything? Is this https://wiki.openstreetmap.org/wiki/Proposal:Cai_scale what we are trying to support here?

What 's the standard rendering for it, where would that be documented? (In Touring view I am trying to stick to common standars where possible  so I am interested before supporting there also).

vshcherb:
Yes but it matches 20K tagging criteria

sonora:
Any idea how it should best be rendered? Looks like CAI once had a proposal but may have withdrawn it? On what basis did we decide how we render in default?

Please don't misunderstand: I love to support it, like I had supported the official SAC rendering for 10+ years, just want to get it as conventionally right as possible in Touring view.

vshcherb:
@xmd5a2 @sonora should be in the legend

PR #17725, opened 2023-07-26
Terrain menu category also for v1 engine

sonora:


vshcherb:
This is strange. We don't support terrain maps for v1, that's why old configure screen  should stay for v1. I think you've missed that.

Chumva:
This was added [here](https://github.com/osmandapp/OsmAnd/pull/17751)

dmpr0:
@Chumva add Todo if it was added.

PR #17728, opened 2023-07-26
Remove ambiguous uses of "Terrain" string

sonora:


OsmAnd/res/values/strings.xml
```diff
@@ -91,7 +92,7 @@
     <string name="verify_account">Verify your account</string>
     <string name="slope_legend_description">Slope uses colors to visualize terrain steepness. You can read more about Slope in %1$s.</string>
     <string name="terrain_category_description">Terrain contour lines, full hillshade map with dark shades and slope maps with colors to show peaks and lowlands. </string>
-    <string name="terrain_map">Terrain map</string>
+    <string name="terrain_map">3D Relief</string>
```
vshcherb, line 95: looks very confusing shouldn't be done UI should use relief_3d already trasnlated line

OsmAnd/src/net/osmand/plus/plugins/srtm/SRTMPlugin.java
```diff
@@ -507,7 +507,7 @@ public boolean onContextMenuClick(@Nullable OnDataChangeUiAdapter uiAdapter, @Nu
 		boolean terrainEnabled = TERRAIN.get();
 		TerrainMode terrainMode = TERRAIN_MODE.get();
 		adapter.addItem(new ContextMenuItem(TERRAIN_ID)
-				.setTitleId(R.string.shared_string_terrain, mapActivity)
+				.setTitleId(R.string.shared_string_hillshade_slope, mapActivity)
```
vshcherb, line 510: That would be like a duplicate line cause 2nd line is the same.

And I don't like it cause I also want in future hypsometric maps

OsmAnd/res/values/strings.xml
sonora: Yes, depends on what we finally decide to call that respective download file type. If "3D Relief" (instead of "Terrain map"), we would of course use the existing string "relief_3d". I can adjust the PR as you like. (This concerns only our naming of choice for  DownloadActivityType GEOTIFF_FILE.)

OsmAnd/src/net/osmand/plus/plugins/srtm/SRTMPlugin.java
sonora: No, below are the respective screenshots: Second line shows which one is selected or and selected status.

For "Hillshade / Slope" you can either select "Hillshade" or "Slope". So far we had called "Hillshade / Slope" "Terrain" in Configure map, but "Hillshade" and "Slope" on the Download UI (download types).

I think now it is appropriate we use Terrain as the overall header for all (currently max. 4) related map types and functionalities), and use the specific "Hillshade / Slope" for the sub-item, in line the respective download types, removing confusion which download is needed for which function? Keep in mind, the idea is to remove the word "Terrain" on any particular sub-item, use it as the overall section headline for our new overall functinality:


Screenshot_20230727-142534_OsmAnd~ Screenshot_20230727-142544_OsmAnd~

vshcherb: I think @dmpr0 is right. It should stay - https://docs.qgis.org/3.28/en/docs/training_manual/rasters/terrain_analysis.html (Terrain), here it says Raster Terrain. We don't plan to have any vector terrain there.

So it's better to change the card name to avoid name duplication or not change it.

sonora: @vshcherb, @dmpr0 But there's no contradiction in the link above at all? Also they use "Terrain" as the super-category for hillshade and other methods, exactly like U propose? Why would we not call our Hillshade and Slope maps consistently "Hillshade" and "Slope", and the issue is solved?

sonora: @dmpr0 @vshcherb   PS: In light of me now understanding our situation a bit better: Are we back to perhaps leaving things as they are now in v4.5.8, only simply renaming the section **headline**, perhaps to "Topography"?

sonora:
I am closing this, as it hss meanwhile been solved.

PR #17770, opened 2023-07-31
Fix 1449 Route without name

Dima-1:


vshcherb:
Won't be used in Android, so no need to keep it here - move back to https://github.com/osmandapp/OsmAnd-tools/pull/657/files

Issue #17376, opened 2023-06-07
Audible navigation instructions should be improved when following a curvy main road


sjvudp:
### Description
I was following a curvy main road when the audible instructions were quite confusing.
Here is a basic sketch to illustrate; the main road is the thicker one, and I was driving from bottom to top direction:

![Navigation](https://github.com/osmandapp/OsmAnd/assets/6734808/1f2cba90-fdba-4187-92af-22194a7bf941)

The distance between both turns of the main road is about 50 meters.

### Steps to reproduce
I had "normal" distances for announcements set, and when I was near point "1" the instructions were "take left turn".
I was driving in town at about 40-50 km/h.
That was very confusing, because the next left turn was *not* to be taken!
At point "2" there was no other announcement.
### Actual result
So I think the announcement for the actual turn was too early and confusing.

### Expected result
There should have been some other instructions at point "1", like "keep right", and only after the turn of the main road (near "2") there should have been a second announcement like "take left turn" (My TomTom does such and it's much easier to follow the correct course with the TomTom).

### Your Environment
OsmAnd Version: 4.4.7
Android/iOS version: 13.1
Device model: Oppo Find X5 Lite

### Update
As I was driving the same route again (with a newer version of the app, 4.5.9) I tried to make a screenshot while driving, so the situation is *after* the voice announcement. Also I rotated the screenshot to match my sketch (which wasn't quite exact, but anyway):
![OSMAnd-Map-rotated](https://github.com/osmandapp/OsmAnd/assets/6734808/a12c6e4e-9e0d-4e15-9e77-af83c08eb6e4)

I've added the two red annotations **1** (where the driver might think to turn left) and **2** (where the driver actually should turn left). Also the route displayed uses intermediate stops which has some issues of it's own (which I'm planning to report unless done already).  I was driving 46 km/h where 50 km/h were allowed.

DmitryAlexei:
Tested the work of voice prompts both on the ground (driving a car on narrow roads in the courtyards of multistory houses, on country and forest roads with many branches), as well as using the navigation simulation function. 

[...]

PR #17823, opened 2023-08-05
TURN_NOW timing - low-speed adjustment

sonora:


sonora:
Here's the corresponding trigger distance table: https://github.com/osmandapp/OsmAnd/issues/17376#issuecomment-1666603426.

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
```diff
@@ -189,13 +201,17 @@ private boolean isDistanceLess(float currentSpeed, double dist, double leadDist,
 	}
 
 	public float getSpeed(Location loc) {
-		boolean simulation = false;
-		if (locationProvider != null) {
-			simulation = locationProvider.getLocationSimulation().isRouteAnimating();
-		}
+		// Hardy 2023-08-05: I believe the limitation to the non-simulation case may be deprecated?
```
vshcherb, line 204: too much comments just makes code worse to read

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
```diff
@@ -189,13 +201,17 @@ private boolean isDistanceLess(float currentSpeed, double dist, double leadDist,
...
+		// Hardy 2023-08-05: Remove flooring the speed to facilitate low speed adjustment (Issue #17376)
+		//boolean simulation = false;
+		//if (locationProvider != null) {
```
vshcherb, line 207: simulation could be deleted

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
```diff
@@ -174,10 +174,22 @@ public boolean isTurnStateNotPassed(float currentSpeed, double dist, int turnTyp
...
 	private boolean isDistanceLess(float currentSpeed, double dist, double leadDist) {
-		return isDistanceLess(currentSpeed, dist, leadDist, (float) leadDist / DEFAULT_SPEED);
+		return isDistanceLess(currentSpeed, dist, leadDist, 0.0f);
 	}
 
 	private boolean isDistanceLess(float currentSpeed, double dist, double leadDist, float leadTime) {
+		if (leadTime != 0.0f) { 
+			// Issue #17376: low speed adjustment for TURN_NOW timing:
+			// Only required for car (but no big effect for bike/ped)
+			if ((currentSpeed > 0) && (currentSpeed < DEFAULT_SPEED)) {
```
vshcherb, line 184: code too complicated

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
sonora: Yes, true, these were just work notes for me ... :wink: - all removed now.

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
sonora: done

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
sonora: suggestions?

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
```diff
@@ -174,28 +174,35 @@ public boolean isTurnStateNotPassed(float currentSpeed, double dist, int turnTyp
...
-		return isDistanceLess(currentSpeed, dist, leadDist, (float) leadDist / DEFAULT_SPEED);
+		return isDistanceLess(currentSpeed, dist, leadDist, 0.0f);
 	}
 
 	private boolean isDistanceLess(float currentSpeed, double dist, double leadDist, float leadTime) {
+		if (leadTime != 0.0f) {    //identifies TURN_NOW case
+			// Issue #17376: low speed adjustment for TURN_NOW timing:
+			if (currentSpeed > 0 && currentSpeed < DEFAULT_SPEED) {
+				float scaleDown = (float) Math.sqrt(currentSpeed / DEFAULT_SPEED);
+				leadDist = Math.max(POSITIONING_TOLERANCE, scaleDown * leadDist);
```
vshcherb, line 185: Changing leadDist & leadTime bad practice cause these parameters are not working as intented

vshcherb:
Whole idea was different that Announcement is triggered ```DISTANCE < THRESHOLD || TIME < LEAD_TIME```
```
// car 50 km/h - 7 s, bicycle 10 km/h - 3 s, pedestrian 4 km/h - 2 s, 1 km/h - 1 s
		TURN_NOW_TIME = (float) Math.min(Math.sqrt(DEFAULT_SPEED * 3.6), 8);
```
Now concept needs to change for all ? or just for turn now

```DISTANCE < THRESHOLD && TIME < LEAD_TIME``` and it's very cumbersome in the code.

File: OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java [hunk 5]
```diff
@@ -174,28 +174,35 @@ public boolean isTurnStateNotPassed(float currentSpeed, double dist, int turnTyp
 	}
 
 	private boolean isDistanceLess(float currentSpeed, double dist, double leadDist) {
-		return isDistanceLess(currentSpeed, dist, leadDist, (float) leadDist / DEFAULT_SPEED);
+		return isDistanceLess(currentSpeed, dist, leadDist, 0.0f);
```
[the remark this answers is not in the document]
sonora: I need a way to identify the "TURN_NOW" case. And the easiest is to check if we called _isDistanceLess(...)_ with a leadTime, as for all but the TURN_NOW case we don't.

So using interim 0 here instead of a separate switch.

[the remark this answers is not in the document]
vshcherb: That means you need to create separate method for Turn now and not reuse this, this is a careful refactoring needed. But I still didn't understand the final concept...
Minimum distance is needed anyway, so why then it needs to be lower

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
sonora: Now they do, I timed it. The problem why they did not behave ok was we had floored _AnnounceTimeDistances.getSpeed()_ with DEFAULT_SPEED.

sonora:
> Whole idea was different that Announcement is triggered DISTANCE < THRESHOLD || TIME < LEAD_TIME

**(1)** It's still the very same concept - no change. (Note that only for the TURN_NOW case _leadDist_ and _leadTime_ are 2 **independent** triggers, for all other cases we derive _leadDist_ by multiplying a given _leadTime_ with DEFAULT_SPEED.)

(And do not mix up with us then having 2 separate checks, I have clarified our ([technical documentation](https://osmand.net/docs/technical/algorithms/voice-prompt-triggering/#trigger-behavior):)
```
dist <= leadDist + currentSpeed * voicePromptDelayTimeSec)
dist <= currentSpeed * (leadTime + voicePromptDelayTimeSec);
```

**(2)** The low speed scaleDown of _leadDist_ and _leadTime_ must only apply for the TURN_NOW case.
- For the TURN_IN begin/end limits we would run the risk of them moving too close to TURN_NOW, or moving too close together
- For the ALARM, APPROACH and TARGET cases we would run the risk of these radii shrinking too much, so the these prompts may never play, or the navigation reaching the target maybe less reliably identified.

All in all, this is the minimum impact change I can think of.

vshcherb:
No changes should be done here then only simplification.

Here leadTime / leadDistance needs to be changed. 
```
case STATE_TURN_NOW:
				return isDistanceLess(currentSpeed, dist, TURN_NOW_DISTANCE, TURN_NOW_TIME);
```

If lead distance is shrinking it's not a constant and shouldn't be defined as constant

sonora:
Here's the concept (only for TURN_NOW, blue like we had it, red is the new)

image

Yes, I can refactor to already finalize TURN_NOW_TIME and TURN_NOW_DISTANCE in method setArrivalDistances, valid idea. There we have no location&speed info yet, that's why made the red correction only in the final comparison.

vshcherb:
Yes, representing the graph is a good idea, though I don't understand how you can have variable Lead Distance, it's a bit weird it means that it could go down.

Probably for turn now it's not a big problem that you calculate 70m threshold and then you recalculate to 30m

vshcherb:
Also not clear why do we need to check leadTime now ? 

If you adjust it anyway to lead Dist - or may be I'm confused.
```
// Check only if speed > 0. (speed = 0 in cases where no time threshold is used.)
		return currentSpeed > 0 && (dist / currentSpeed - voicePromptDelayTimeSec) <= leadTime;
```

It looks like that only 1 check is needed though I'm not 100 % sure

vshcherb:
It looks like we have only 1 check
```
dist < (leadTime + voiceDelay) * Math.max(currentSpeed, DefaultSpeed) 
```

For turn now if speed < defaultSpeed you scaleDown and calculate only
```
dist < leadDist + voiceDelay * currentSpeed
```

vshcherb:
To summarize for 
```if (currentSpeed > 0 && currentSpeed < DEFAULT_SPEED) {```
you don't need to call isDistanceLess it only confuses and you can simply return dist < leadDist than

**Comment by sonora** [comment]
No, we indeed need two separate checks lead distance, and lead time, and either one can trigger  depending on _currentSpeed_. 

For the current parameterization the lead distance trigger will trigger for currentSpeed less than about 7m/s, the lead time trigger for above (while defSpeed is 12.5m/s).

This comes from our principle that we initialize with two _arbitrary_ (independent of each other) values like <45m  or <7sec). I do not want to change this, as it has always worked great except in the low speed case, so I only modify that.

Regarding if it solves the issue posted: I have tested for 2 days in rural areas and a small town  where it works as intended. May still tweak scaledown a bit after more feedback from other testers after merging so folks can test. But I am convinced this is systematically better than before.

**Comment by sonora** [comment]
PS: If you are interested, I could take the time to de-convolute our entire corresponding code, but it should become a separate PR then.

Because we currently do this:
(1) We use a lead time for orientation e.g. 22s
(2) We estimate, by multiplying with defSpeed, a corresponding lead distance, like TURN_IN_DISTANCE_(BEGIN/END)
(3) We call _isDistanceless(...)_ and there for some circumstances factor in currentSpeed (preferably for closer distances), for others not.

So essentially we go from time to distance to time. Plus  for TURN_NOW we set independent lead distance and lead time  which take precedence depending on currentSpeed.

I can tackle a PR where we call isDistanceLess with just "leadTime" and "speed", and where speed (depending on state) is either defSpeed or currentSpeed.

File: OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java [hunk 6]
```diff
@@ -174,28 +183,15 @@ public boolean isTurnStateNotPassed(float currentSpeed, double dist, int turnTyp
...
-		// (1) distance <= leadDistance?
-		if (dist <= leadDist + currentSpeed * voicePromptDelayTimeSec) {
-			return true;
-		}
-		// (2) time_with_current_speed <= leadTime? (high speed adjustment)
-		// currentSpeed = 0 in cases where no time threshold is used.
-		return dist <= currentSpeed * (leadTime + voicePromptDelayTimeSec);
+		// Check trigger. Lead distance is scaled up for high speeds. (For cases without such scaling we pass currentSpeed=0.)
+		//return dist <= Math.max(1, currentSpeed / DEFAULT_SPEED) * leadDist + currentSpeed * voicePromptDelayTimeSec;   with division avoided:
+		return dist * DEFAULT_SPEED <= Math.max(DEFAULT_SPEED, currentSpeed) * leadDist + currentSpeed * voicePromptDelayTimeSec * DEFAULT_SPEED;
```
vshcherb, line 188: Something to take a break may be... 
```
+ currentSpeed * voicePromptDelayTimeSec * DEFAULT_SPEED;
```
This is clearly a typo, and it looks like 2 parts of equastion could be divided by DEFAULT_SPEED to get it simpler

```
dist <= Math.max(distToTrigger, currentSpeed / DEFAULT_SPEED * distToTrigger)  + currentSpeed * voicePromptDelayTimeSec;
```
Naming
dist -> leadDist, leftDist
leadDist -> distanceTrigger

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
```diff
@@ -138,7 +138,16 @@ public int getImminentTurnStatus(float dist, Location loc) {
...
 		switch (turnType) {
 			case STATE_TURN_NOW:
-				return isDistanceLess(currentSpeed, dist, TURN_NOW_DISTANCE, TURN_NOW_TIME);
+				if (dist <= currentSpeed * TURN_NOW_TIME) {
+					return true;
+				}
+				int leadDist = TURN_NOW_DISTANCE;
+				// Issue #17376: low speed adjustment for TURN_NOW timing:
+				if (currentSpeed > 0 && currentSpeed < DEFAULT_SPEED) {
+					float scaleDown = (float) (currentSpeed / DEFAULT_SPEED);
```
vshcherb, line 147: hm.. here you'vot something obvious
newLeadDist = currentSpeed / DEFAULT_SPEED * leadDist
Math.max(newLeadDist, currentSpeed / DEFAULT_SPEED * newLeadDist)
// but you don't need it cause your currentSpeed < DEFAULT_SPEED - so no need to call it

return dist <= Math.max(POSITIONING_TOLERANCE, scaleDown * leadDist);

vshcherb: I see it even 1 liner:
```
if (currentSpeed > 0 && currentSpeed < DEFAULT_SPEED) {
   return dist <= Math.max(POSITIONING_TOLERANCE, currentSpeed / DEFAULT_SPEED * TURN_NOW_DISTANCE);
}
```

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
sonora: I don't see a typo here? The 'simple' equation is the one in the comment in line 187. But it involves a costly float division, I avoid that by multiolying both sides by DEFAULT_SPEED ?

OsmAnd/src/net/osmand/plus/routing/data/AnnounceTimeDistances.java
sonora: Yes, 1-liner is fine - done. ;)

vshcherb: Where?

PR #17831, opened 2023-08-07
Sorting options for same folders consolidated and use 1 preference

0xRe1nk0:


Chumva:
Moved to 4.5 in a separate pull request

PR #17884, opened 2023-08-10
Allow to completely disable map rotation using gesture.

milan11:
The map rotation using gesture was possible even when "North is up" is set as Map orientation. So, create a separate independent setting to disallow the gesture. Actually, the setting can be useful even in other Map orientation modes.

By default, the gesture is enabled, so the app behaves as before.

Disabling the rotation gesture allows users to avoid accidental rotation when zooming was intended. The problem manifests itself e.g. in these scenarios:
- (1) I zoom the map. The map rotates while zooming. I need to click the compass icon to revert to North is up. Actually I needed to build the habit of "checking if north is up" after each zooming, which is annoying.
- (2) I zoom the map. The map rotates while zooming. I do not notice this. Later, I notice "something is wrong with the map". Click the compass icon.
- (3) During some activity (geocaching, bike, sport) I accidentally rotate the map without the intention to do anything. Usually I do not notice this and it continues as in (2).

This supports the issue 15660 and some related issues.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -1541,6 +1541,8 @@ You need to activate the sensor so OsmAnd can find it.</string>
     <string name="please_provide_point_name_error">Please provide a name for the point</string>
     <string name="use_volume_buttons_as_zoom">Volume buttons as zoom</string>
     <string name="use_volume_buttons_as_zoom_descr">Control the map-zoom level using the volume buttons on the device.</string>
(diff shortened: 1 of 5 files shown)


sonora:
And alternatively: This would be a solution to provide an initial 30 degree twist threshold in "North is up" mode, allowing users who really need to directly rotate the map from that state to start doing this **by a big twis**t, rather than having to go to any menu or widget: https://github.com/osmandapp/OsmAnd/pull/16830/files.

SergeyP2:
Version 4.6.9. This annoying BUG is still HERE. Are you doing anything???

PR #17938, opened 2023-08-16
Help fixes

Chumva:


sonora:
It looks like when you are on a sub-menu in the Help screen, e.g. on Help>Map legend, the "back" arrow (top left on the scren) has no effect.
 (The device back ley works, and also otworks on the top screens it seems to work, though.)

Chumva:
This toolbar arrow should be already fixed

sonora:
Yes, alteady fixed, had not synced the second commit yet! sorry!

PR #17947, opened 2023-08-17
Put Version info under About item

sonora:


sonora:
Hi Dmytro, my thinking is the following:
1. It is customary in software to always find the exact installed version under "About".
2. Our "What's new" articles more coarsely refer to the _major release_ (and which one that is is stated in the article headline), so the exact version code is less needed under "What's new" .

Can you agree?

vshcherb:
Not 100% sure...

vshcherb:
This button is about company not about version.

vshcherb:
Then we will have 2 versions. About OsmAnd 4.6 and What's new in 4.6.

They are both logical places but they duplicate. You are right About OsmAnd doesn't speak about product but speaks about Company or Project which will be version independent

Issue #13298, opened 2021-12-04
Announcement Time - Off Route


duiveldoder:
### Description
When going to any navigation profile
Then "configure profile"
then "navigation settings"
then "voice prompts"
then scroll down to where it says "Announcement time"
expand the information panel below the slider

it all makes sense - early, normal, late you get early notifications "xxx" meters before passing
or normal notications "xx" meters before passing
or late "x" meters before passing

EXCEPT the "off router warnings"
under early I get these for my profile "walking" at 80-85 meters
under normal I get these 55-60m
under late I get these 25-30m

but the thing is - this is OFF ROUTE meaning you get these AFTER and not BEFORE you pass anything..

it means AFTER you get away from the track x, xx, xxx meters you get the warning.

So the values here should actually be increasing and not decreasing as you move the slider from left to right.

### How to reproduce?
just try and see

### Your Environment
OsmAnd Version: latest
Android/iOS version: 10
Device model: Mate 20 pro

**Maps used (online or offline):**   
If you have an issue related to offline maps, tell us the exact name of the map file where the issue occurs and its edition date.

Always latest up to date.

vshcherb:
Late means that you've closer to the "Action time", so if you have very precise GPS and slow speed you need to use "Late" but if you experience lots of errors of position detection then it's better to use "Early"

duiveldoder:
I think you haven't read what I wrote..

OFF ROUTE - LATE - means I get them within a shorter deviation from the real route..

That's opposite of what to be expected when selecting LATE

PR #17950, opened 2023-08-17
Fix_17948 Off route scaling

sonora:


sonora:
For the Off-Route case, the "event" announced is already behind you, so the lead distances scaling needs to be inverted.

cridemichel:
> For the Off-Route case, the "event" announced is already behind you, so (lingustically) the lead distance scaling needs to be inverted.
> 
> But it's debatable, see Victor's argumentation here [#13298 (comment)](https://github.com/osmandapp/OsmAnd/issues/13298#issuecomment-998125427).

Note that with this fix the "in the last meters" string related to the slider which is used to set "arrivalDistanceFactor"
is now misleading...

sonora:
There"s pros and cons here. On the pro side, I believe that in this fashion now it is more understandable to the user.

Regarding "In the last meters", that wording had always been misleading for the "You have been off the route..." case. It  could be fixed by simply rewording to "Very late".

cridemichel:
> There"s pros and cons here. On the pro side, I believe that in this fashion now it is more understandable to the user.
> 
> Regarding "In the last meters", that wording had always been misleading for the "You have been off the route..." case. It could be fixed by simply rewording to "Very late".

I agree with rewording as you suggested. About splitting a single slider into several ones I do not think that makes less understandable and from a coding point of view it can be easily implemented. OsmAnd is intended to be as an highly customizable navigation system and I do not see to point to struggle with keeping this setting as simple as possible...

PR #17981, opened 2023-08-20
Added check to prevent rotation if in North Up mode.

### LINKED ISSUE #17979: Compass mode "North is up (fixed)" doesn't work
## Description

I was excited to see in the 4.4.7 version: "Map orientation added new fixed mode" as this is a feature I have been hoping for for a really long time. Unfortunately, after testing this it isn't working for me and I signed up a github account to submit a bug report. I am new at this so I hope I am doing this rite. 

The new "North is up (fixed)" mode works exactly the same as the old "Manually rotated" mode for me. If I rotate the map with two fingers the map manually rotates, in both modes. I'm not sure if it is just my device. But at the moment there is no difference between the two modes that I can tell. 

I never want the map to rotate and I expect that most other people don't either. But if manual rotation is the default setting that would be OK, as long as "fixed" mode does truly fix the map so that rotation is not possible. This is highly important for safety as the map rotating unintentionally can easily get you lost and / or cause you to be distracted while operating a vehicle. 

### How to reproduce

1. In Settings > Compass mode, select "North is up (fixed). 
2. Go to the main map. 
3. With two fingers, make a rotation gesture.

### Actual result

Map manually rotates, just the same as if you are in "Manually rotated" mode.

### Expected result

Map does not rotate and north remains fixed to be the top of the screen when in "fixed" mode.

**Environment**
OsmAnd Version: 4.7.7
Android/iOS version: Android 12
Device model: Pixel 4

osmandlover:
Fixes, resolves and closes #17979 , #16836 , #17884 , #17086 

Disallow rotation when in North-Up compass mode. Being able to rotate when in north-up mode is a contradiction and makes OsmAnd unintuitive to use. This change still allows users who want manual rotation to select that mode and manually rotate, or, if you have the compass icon active, to toggle through the different modes without having to fiddle in settings by tapping the compass. 

I'd be ok with making the default compass mode "manually rotated" so that OsmAnd behaves how some users have become accustomed to using it, but having a separate "north-up" mode is pointless if rotation is still allowed in this mode.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
+++ b/OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
@@ -2089,7 +2089,7 @@ public class OsmandMapTileView implements IMapDownloaderCallback {
 			} else {
 				startZooming = true;
 			}
-			if (mapGestureAllowed(MapGestureType.TWO_POINTERS_ROTATION) && isAngleOverThreshold(Math.abs(relAngle), Math.abs(deltaZoom))) {
+			if (settings.getCompassMode() != CompassMode.NORTH_IS_UP && mapGestureAllowed(MapGestureType.TWO_POINTERS_ROTATION) && isAngleOverThreshold(Math.abs(relAngle), Math.abs(deltaZoom))) {
 				startRotating = true;
 			} else {
 				relAngle = 0;


Dollarius:
Thank you, thank you, thank you!!
56 characters of code that will save hours of trouble of many users!
I hope it gets approved soon :)

olesenm:
This is very close to a feature request that I had, not sure if this would also cover it, or is similar enough to be extended.

RFE: a "sticky north" Mode or behaviour.
Reasoning: when displaying with fixed north, minor inadvertent movements should *not* kick it into rotated north mode. Keeping north as north ensures that the map direction (ie, fixed) remains consistent and avoids slowness since a rotation then triggers re-rendering of the text labels.

osmandlover:
Yes, this PR covers the behavior you describe. It disables map rotation (major _and_ minor) when in fixed-north-up mode, which is the behavior that is implied by the name "fixed-north-up".

Fixed-north-up mode currently isn't useful, because making a rotation gesture is a "shortcut" to changing the mode, which is non intuitive because surely the point of fixing north up is to keep it "fixed", which is how I prefer to use maps at all times. 

The alternative solution presented is to increase the threshold for how much you have to rotate the map before the shortcut to manually-rotated mode is invoked. However this is a poor solution because:

1. awkward look-and-feel, as rotating appears to do nothing until it does a lot with a big jump.
2. large thresholds in particular make this feature undiscoverable. 
3. there is an easy alternative shortcut - tapping on the compass changes the mode so you can easily change to manual if you want.
4. it's very easy to invoke accidentally, evidenced by the numerous duplicate bug reports on this topic. I find that even a large threshold does not help when operating the map while in motion / with wet hands / pocket dialing. Pinch-zoom is much easier and more common gesture than rotation; although there are zoom buttons these are difficult to operate when in motion. Difficult = dangerous. 
5. the consequences of accidentally invoking the manual rotation shortcut are disproportionately large: slowness with the re-render, disorientating the user & getting them lost, and safety issues if they are either confused by it or try to correct it while operating a vehicle. 
6. is this a shortcut that anybody even wants or needs? In fact, is "manual rotation mode" even a _feature_ that anybody wants or needs? It's useless on a mobile device, where you can physically rotate the device or your body if you want to orient the map. Even so, I don't know of any use case to rotate a map arbitrarily without reference to either north or bearing.

I can say that my experience of OsmAnd has been dramatically improved since running this fork, and I hope that it can be upstreamed soon, or a more robust implementation.

olesenm:
@osmandlover - it might be useful to rebase this on top of the current master (and force update) to reduce some noise. Not sure what you have to do to get the attention of a maintainer to look at it (without appearing shout), but please persevere.

Dollarius:
I believe the problem is that this solution (that many people are requesting) is not to the liking of the developers.
As far as I understand, developers want to give unexperienced users the chance to rotate the map without digging into the settings (i.e., keeping the usability simple), and to restore the North up by just pressing the compass (which would instead permanently fix it with this commit).

I could propose to add an additional option in the General settings to "force North up".
@osmandlover would that be possible/doable for you?

It would be nice to solve this issue: I find my map inadvertently rotated every single day (phone in the pocket, map zoom with two hands, etc.)

milan11:
@Dollarius 
I have already created a PR probably similar to what you are requesting - a separate setting which will prevent the rotation.

https://github.com/osmandapp/OsmAnd/pull/17884

This setting would not disturb anyone (because the default behavior is as before), but an argument against it may be that "additional configuration possibilities may unnecessarily complicate the app and its testing".

osmandlover:
I have rebased @olesenm , thanks for the tip. I'm not sure what I can do to promote it other than keep the branch up to date. There's only 19 open PRs so I'm sure it has been seen. I feel I've made my case as best I can and it's up to the maintainers now whether/how they want to solve this issue. 

> I believe the problem is that this solution (that many people are requesting) is not to the liking of the developers.
As far as I understand, developers want to give unexperienced users the chance to rotate the map without digging in the settings (i.e., keeping the usability simple), and to restore the North up by just pressing the compass (which would instead permanently fix it with this commit).

I get that the solution is not to the liking of the developers. I think it's about creating a consistent model where the map behaves in an intuitive material kind of way, with the fingers fixed to the map, the theory being that if you can zoom in and out with a pinch gesture, you should also be able to rotate in the same way. 

To be clear, this commit does not permanently fix north. Tapping the compass still restores north from manual rotation mode. Switching to manual rotation mode is there by default on the home screen and you just need to tap it again to reenable rotation.  By default this is available on the screen for inexperienced users (personally I turn the compass off because I like north to always be up).

> This setting would not disturb anyone (because the default behavior is as before), but an argument against it may be that "additional configuration possibilities may unnecessarily complicate the app and its testing".

Your alternative solution would work for me, and I'm be very happy if they merged your PR, but IMO the biggest weakness of OsmAnd is the complexity of the settings (especially for inexperienced users).

There's already an appropriate setting: compass mode. North is up, Movement direction, Compass direction, and Manually rotated. It's very easy to change between the four modes: tapping on the compass toggles between them.

Currently "north is up mode" isn't actually a mode, it's just a way to set "manual rotation" temporarily up. Because as soon as you rotate, the mode is switched to "manual rotation". What this commit does is remove this method/shortcut to switch to "manual rotation" mode simply by rotating. It's unnecessary and unintuitive. The "north is up" mode is intuitive and what most people expect from this mode is for north to be up. Tapping the compass and changing to "manual rotation" is also really intuitive. 

To me this is a win/win/win. I hope the maintainers can see that. But until then just build and install the fork!

mikehgentry:
One possibility I don't think I've seen mentioned is to make the list of available rotation modes configurable, ideally per profile. There are some profiles I use for planning where I'm never going to want to use compass direction / movement direction, and some profiles I use for riding where they're all I'm likely to want to use, so that makes sense to me on a basic level.

Then 'North always up' could be an available mode, and you wouldn't have to put it in the list for any given profile by default.

MikeTango:
Hi, 

i am currently experiencing the same issue with two fingers frisbeeing the map. It's infuriating, especially because i use osmand on a tablet for routing while driving. Why even label it "north locked"?  Is there any way to get this into a regular update?

SergeyP2:
Version 4.6.9. This annoying BUG is still HERE. Are you doing anything???

osmandlover:
This is still unfortunately an issue in the main branch of OsmAnd. I have rebased again, rebuilt and retested - this branch is working very well and correctly prevents the undesirable shortcut from "north fixed up" mode to the "manually rotated" mode. 

@MikeTango I believe the reason that it's labelled "North Locked" is because the mode does indeed have north up, however the app maintainers believe that a two-finger rotation is usually an _intentional_ attempt at rotation, and therefore, an appropriate shortcut to switch the mode to the "manually rotated" mode. 

I note also that Google Maps behaves similarly to this, without being explicit about the "mode" like OsmAnd is, so I believe that OsmAnd is trying to mimic the Google behaviour. Which is a crying shame, because in most ways OsmAnd is vastly superior and shouldn't use Google as a guide. 

It's a bit frustrating that some comments in support of this (namely @SergeyP2 's comment) are being deleted. However to their credit, the maintainers haven't closed all of these repeated issues and this PR remains open. So I think there must _at least_ be some internal disagreement between them around this point. 

I'll reiterate that I haven't seen a single user who says they want the "manual rotation" mode and logically I can see no reason for such a mode. Even if having a manual rotation mode is desirable, having a true north-fixed-up mode is evidently also (more?) desirable. Lastly, I haven't seen a single person arguing that they find the shortcut to switch from fixed north to manual rotation useful or desirable, and removing this shortcut seems to provide a win/win/win for all concerned, keeps the default behaviour as is, and avoids adding any extra settings.  

It's a shame that we have to maintain and build a branch to fix this basic issue, but at least there is a fix. The current behaviour does feel dogmatic rather than pragmatic, in the absence of any argument to the contrary except the two points I myself made (manual rotation is analogous to Google + makes the map behave like material). 

I really hope that the maintainers can see sense and, if not merge this PR, provide a way for the legions (majority?) of OsmAnd users who need north to remain (truly) fixed "up" to use the app in the way that is the most intuitive, safe, consistent and performant.

mikehgentry:
> I'll reiterate that I haven't seen a single user who says they want the "manual rotation" mode and logically I can see no reason for such a mode.

It is useful for getting a good look at 3D terrain particularly. Find a steep East/West valley in 3D mode and the close side is compressed and hard to read, same goes for the far side of hills / mountains.

Aside from that I use it to get a good look at prospective routes when touring. Small screen devices don't give you much choice.

MikeTango:
> @MikeTango I believe the reason that it's labelled "North Locked" is because the mode does indeed have north up, however the app maintainers believe that a two-finger rotation is usually an _intentional_ attempt at rotation, and therefore, an appropriate shortcut to switch the mode to the "manually rotated" mode.

North locked should mean north locked and lock north. It's completely distracting while driving.

> I note also that Google Maps behaves similarly to this, without being explicit about the "mode" like OsmAnd is, so I believe that OsmAnd is trying to mimic the Google behaviour. Which is a crying shame, because in most ways OsmAnd is vastly superior and shouldn't use Google as a guide.
> 
> It's a bit frustrating that some comments in support of this (namely @SergeyP2 's comment) are being deleted. 

That comment was rude and didn't add anything to the discussion.

> I really hope that the maintainers can see sense and, if not merge this PR, provide a way for the legions (majority?) of OsmAnd users who need north to remain (truly) fixed "up" to use the app in the way that is the most intuitive, safe, consistent and performant.

Well, we're here and making this feature request known. I also don't see anything that would speak against making north locked truly locked, while you can still click the compass icon to switch to a free-floating map movement. 

Customizing is good, but the main question when building an interface that is used in moving vehicles is ergonomics of a moving vehicle. No one in their right mind would put that kind of gesture congestion into the controls of an airplane or excavator. Touchscreens are inherently bad for that application, but it's what we have.

mikehgentry:
I wouldn't use it in motion, no (hence my suggestion above - have a list of available rotation modes per profile).

Dollarius:
> But how much do use it once on the move, especially when operating any kind of vehicle?
@MikeTango I use it for cycling/hiking and almost every single time I put the phone in my pocket the North gets rotated with the sweat.
Another common scenario is upon zooming on the map while riding, where gloves and vibrations end up rotating the map.
The most frustrating thing is that I often don't notice about the accidental rotation and cannot figure out why I'm going "down" on the map while I'm supposed to be going (i.e.) North, or East.

SergeyP2:
It's a miracle how fix of such an obvious bug requires months of discussions. 1-2 lines of code I believe. "North Locked" is North locked, what can be unclear here?? Looks like a devs team has lost connection with common sense and reality.

osmandlover:
> It is useful for getting a good look at 3D terrain particularly. Find a steep East/West valley in 3D mode and the close side is compressed and hard to read, same goes for the far side of hills / mountains.

Thanks for the rationale, I don't use 3D mode so it's useful to hear that manual rotation is useful in 3D mode.

Note that this fix does not change anything about 3D mode. Manual rotation is still easily accessible. If you are in "north locked" mode, tapping on the compass toggles to manual rotation mode. This fix simply prevents changing to "manual rotation" mode when a rotation gesture is made in "north locked" mode (which to me, and others, seems like the entire point of a "north locked" mode).

> Aside from that I use it to get a good look at prospective routes when touring. Small screen devices don't give you much choice.

In 2D mode, one can "manually rotate" small screen devices by... manually rotating the physical device. This works in all modes, including "north fixed up"  mode! However, to reiterate, it doesn't make sense to gimp the "north fixed up" use case with this shortcut to the "manual rotation" use case. Even if they were equally useful (they're not), whyever would we not support both use cases?

> have a list of available rotation modes per profile).

This would work well with the way I personally use OsmAnd, which is to set it up differently for different profiles (different styles, shortcuts etc for walking than driving). This would be quite a lot more work than just disabling rotation in "north fixed up" mode, but if the maintainers would agree to this then I'd be happy to have a crack at it. 

> That comment was rude and didn't add anything to the discussion.

Well, it was only a little rude, and to be honest I understand @SergeyP2 's frustration with the situation. I do think it adds something: it adds another voice to show that this is a really, really important issue that is being ignored. 

> why I'm going "down" on the map while I'm supposed to be going (i.e.) North, or East.

In a nutshell this is the the problem. Maps are complex and it's a feat of human ability to be able to mentally match what is happening on the ground with the reality around them. My brain doesn't work in a way that can process direction based on the compass point vs what's happening. I struggle to point towards north at the best of times! If others can mentally rotate and process an arbitrarily rotated map in their minds, that's great. But not all of us have that ability! It's an accessibility issue, really.

If the affordance/intuition arguments, the safety arguments, and the usability arguments don't sway you, perhaps the accessibility argument will? The simple solution in this PR is for "north fixed up" to do what it says on the tin, and it doesn't take anything away from anyone who likes the other modes.  But if the simple solution doesn't appeal - some other options:
1.  include a setting to allow the user to turn the rotate-map-to-engage-manual-rotation-mode shortcut on / off
2. a separate button to "find north" vs "switch compass mode"  (as these are actually kind of two different things), and a setting to be able to add/remove these two buttons from the screen
3. a fifth mode, "north up for now" which has the behaviour of the current "north up" mode, but then a "north fixed up" mode as well that does what the name suggests
4. a way to allow/disallow the different modes per profile (and then directionally-challenged folks like me can remove the "manually rotated" mode (and hence the rotate-map-to-engage-manual-rotation-mode shortcut would not do anything for those profiles).

If any of those solutions are strongly likely to be merged, please let me know and I will try to implement it.

mikehgentry:
> In 2D mode, one can "manually rotate" small screen devices by... manually rotating the physical device. This works in all modes, including "north fixed up" mode! However, to reiterate, it doesn't make sense to gimp the "north fixed up" use case with this shortcut to the "manual rotation" use case.

I'm sitting in a pub, planning a route going South East. My phone has a small screen, so panning around without aligning the route to the screen is a lot more work. What do you want me to do, rotate my chair so I'm pointing South East before scrolling through the route? 

> Even if they were equally useful (they're not)

Your use case is different to mine, so that's not really a helpful comment. I have very little use for 'North up' most of the time.

Dollarius:
I totally second the introduction of a fifth mode: "North always up" while keeping the current "North now up" that seems useful for some.
The modes could be ordered so that:
1. The default mode for new users is "North now up", like now
2. Rotating the map beyond threshold, the mode switches to "Manually rotated", like now
3. One click on the compass restores "North now up", like now
4. For geeks-in-need, multiple clicks on the compass lead to the NEW mode "North always up". This could also be a setting only accessible from the menu, but from my understanding it would be a usability issue according to developers.

osmandlover:
> What do you want me to do, rotate my chair so I'm pointing South East before scrolling through the route?

No, I want you to tap the compass button to toggle onto "manual rotation" mode, then rotate to your heart's content.

> Your use case is different to mine, so that's not really a helpful comment. I have very little use for 'North up' most of the time.

The difference is that a shortcut to your use case is ruining our use case. I don't want to take your use case away - I just want to take away the shortcut. Because at the moment, the shortcut makes "north fixed up" meaningless.

You're painting me as being unreasonable here, and I don't appreciate it.

osmandlover:
> I totally second the introduction of a fifth mode: "North always up" while keeping the current "North now up" that seems useful for some.

Let's see how this goes down with the community, if positive, I'll have a crack at implementing this.

alberto4github:
Is one incredible ofuscation not have this fixed north😭. If you start the zoom perfectly diagonal then it doesn't rotate during the zoom even if you twist your fingers, but if you make the mistake of having human fingers and the diagonal is not perfect at the beginning and you exceed that low threshold of degrees they have put, the map is already rotated somewhat and you have to hit the compass to be north. 

I still don't understand why they don't put that fixed north option. It doesn't cost anything, it doesn't take away any functionality from anyone and it would satisfy many people who have asked for it.

dmpr0:
I still don't understand why they don't put that fixed north option. It doesn't cost anything, **it doesn't take away any functionality** from anyone and it would satisfy many people who have asked for it.

Map rotation?

osmandlover:
I can not understand why either. But at this point it is obviously intentional design decision not overlooking or problem with developer capacity - there is clearly strong demand for fixed north option, simple PR has been open for over 6 months, and my offer to fix it in different way to be suggested still stands. 

I think problem lies with the compass doing two jobs. 

One, it resets north to up. 

Two, it changes the map rotation mode. 

This is quite standard map pattern, i notice it is present in Google maps too. I don't use google maps but it would annoy me there if I did too and I would want fixed north up. In Google Maps, you can manually rotate map at any time. then compass appears. Tap compass, and it goes back to north. Tap again and it goes into tilt mode. So tapping compass toggles between tilt mode (which osmand calls 3D) and flat mode, UNLESS map is rotate, then it recentres to north.

So problem is in Google too, but it is worse in osmand however, i think because there are more modes, so once you get lost it is much harder to get back to what you want. 

The other reason is because Osmand "remembers" your last saved rotattion, so if you toggle back to the rotated mode, then you will tilt to your previous rotation which almost certainly didnt mean anything to you and was probably accidental in first place. So it would be big improvement for Osmand to just to forget previous rotation amount.  although i hestitate to suggest that because it still doesn't get us what we really want, which is a true fixed north up.

This whole problem is especially annoying because i dont want any other modes EVER, and although Osmand is so customisable that you can even turn off the compass icon, you _cannot turn off the rotation_, sorry no you can customise osmand down to last widget but no this very basic request is not possible, you must always want rotate available, who wouldn't Hm!?

This is the number 1 UX problem with Osmand IMO and it will require a custom build of this branch to fix unless one day osmand maintainers see that they should be making BETTER map than google, not simply copying their design.

alberto4github:
In 2022, @vshcherb says: "Later we could think for a small fraction of user who will be unhappy.".

"small fraction" .. I have taught Osmand to many people, both in person and in (free) courses at the mountain club. I am 60 years old and have a life dedicated to the mountains and maps (OSM edition and even scanning and calibration when mobile phones did not even exist).

A lot of people complain about this. People who are new, who have human fingers and who zoom in on maps when they go in the mountains. Maybe if someone is in front of a screen all day they don't have the problem, but if you use it, they do.

I have also given many orienteering courses. In orienteering it is BASIC to have the map to the north. At the beginning it is easier to turn it, but it is a problem with time, because we don't visualize the mountain, reliefs, etc... It is basic for many people to have the map FIXED to north. I, like the previous colleague and many other people, never use any other orientation.

Sometimes, not being anchored in an opinion is smarter than insisting on maintaining an error, and even more so when the solution is as simple as putting a "Fixed North" option (real). As the previous colleague said, this is by far the biggest problem that Osmand currently has for people who use it in the mountains. 



Translated with DeepL.com (free version)

vshcherb:
The issue is that we don't want to introduce another setting by default and we don't want to block map rotation by default.

If we introduce 4-5th option then it's even more confusing cause we will have long cycle rotation. In this issue we need to be consistent with other maps apps cause most of the users should have same Map gestures experience.

vshcherb:
Closed: cause request won't be merged.

The only direction I see, it would be to have it configurable which options will be available. Or an option to prevent map rotation in North up...

SergeyP2:
> Closed: cause request won't be merged.
> 
> The only direction I see, it would be to have it configurable which options will be available. Or an option to prevent map rotation in North up...

You are a stupid idiot. I will never recomend your app to anybody.

osmandlover:
> If we introduce 4-5th option
vs
> we need to be consistent with other maps apps

Problem is OsmAnd is NOT consistent with other map apps precisely BECAUSE it has so many options. If you want to make it consistently as bad as Google Maps, but not worse, then remove the "remembering previous rotation" would help a lot.

> we don't want to introduce another setting by default

Can you explain what "another setting by default" means? do you just mean you don't want another setting? I don't think any of us are saying north up has to be the default setting (though personally i think that would be sensible).

> we don't want to block map rotation by default.

Because why? Do you have reason to think that most users want to rotate maps by default? I think you are dogmatically sticking to this idea without actually considering whether users want it. But still, at least a way to turn off the shortcut into "manual rotation mode" would be useful...


> The only direction I see, it would be to have it configurable which options will be available. Or an option to prevent map rotation in North up...

Ok, so if I make setting "make fixed north up mean fixed north up" (off by default) would PR be considered? I can do that if so?

osmandlover:
I think you overestimate impossibility of (1) even with large threshold. This rotation doesnt just happen _sometimes_ when zooming. it happens to me all. the. time... with wet fingers or in or on a moving vehicle especially.

I find threshold solution very ugly and poor ux, and if your suggestion was implemented i would still run custom build with a true "fixed north" mode.

P.s. i want fixed north at ALL times, in urban environments, car mode, everything. not just orienteering. I never ever ever ever ever ever ever want to ever ever ever ever see any rotation. Off please. Fixed north up ALWAYS please. 

I find refusing to add setting to  "really make fixed north really mean fixed north" to be really weird hill to die on to not want more options. This is Osmand!! App which has separate settings for 5 different profiles and separate settings for configure map vs configure screen!!! There are so many settings in Osmand i genuinely get lost. They never cared about adding or organising settings before, but THIS is the setting they refuse to add??

Honestly, it just feels spiteful at this point.

MikeTango:
Please reopen. This kind of running after features that another company implemented because of this and that reason isn't good, especially given the track record of google to just discontinue products and randomly implement various feature. Your userbase is complaining that this feature is annoying, you can go the not google way and listen to your users.

sonora:
> I think you overestimate impossibility of (1) even with large threshold.

I'm just reiterating here, it's all been exchanged already in e.g. https://github.com/osmandapp/OsmAnd/discussions/16836. But I wanted to react to the 'new' information that this is to avoid "a 5th mode", 'cause I think that's easy. :wink:

And regarding an initial twist threshold (really just an initial one, after reaching it, the map starts rotating and can in the same gesture be twisted backwards if less rotation is desired): We have had this in OsmAnd for perhaps 8 years or so prior to 2022, can't remember much complaining. Not in the coldest bivouacs had I ever experienced and unwanted map rotation prior to https://github.com/osmandapp/OsmAnd/pull/16746.  But could even up the threshold to 45deg, if desired ...  :wink:

alberto4github:
For me, the old 30º threshold would be a "Solomonic" solution if they don't want to put that option of "fixed North with no possibility of rotation".

That's what we had about 2 years ago and it worked well. Nowadays the threshold is about 15º. With 30º there would be much less "human error" and, whoever wants to rotate it, has that possibility without having to go and change the configuration.

@sonora's  solution, please 🙏

vshcherb:
There is no sense in exchanging arguments anyway it tend to very same. Nobody who likes current behavior is going to come and start arguing and all mainstream map apps allow to rotate map easily while zooming and all have same issue. Nobody doesn't do things that doesn't make things especially if it's present in mainstream apps.

I would shutdown this conversation cause it doesn't bring any point. Anyway we're thinking implement some sort of fix Q2 2024.

Issue #17723, opened 2023-07-26
OsmAnd giving frequent and consecutive warnings for closely spaced speed bumps


jol70:
## Description
L exemple ce sont les passage piéton, la remarque est vraie aussi pour les rond points ou les changements de directions. Bien sûr on peut supprimer l information, toutefois si on l utilisé c est que c'est utile. Les gps type waze ou maps on limite la répétions à environ 3 fois et c est suffisant. 

### Expected result
Que l information sonore soit limitée en nombre de rappel et alors que le véhicule est à une distance fonction de sa vitesse et de son poids. Premiere alerte très eloignée  si la vitesse et le poids du véhicule sont important ou fixe à 250m hors ville et 50m en ville  ou réglable Deuxième alerte 20m Troisieme alerte 5m. 
Merci 
**Environment**
OsmAnd Version:
Android/iOS version:
Device model:

vshcherb:
We need to translate issue to english so everybody could follow

letypequividelespoubelles:
I can do the translation, but even the French is not very understandable. 
@jol70 combien de notifications tu as maintenant ? Et merci d'indiquer la version d'osmand, etc
Translated and resumed:
Tok much sound notification for pedestrian crossing, direction change, etc. Three notifications should be enough, first one, far from the poi, at a distance deoemding on the vehicle and city ) outside city, a second one 20m before, and a third one 5m before.

yuriiurshuliak:
Reproducing involves OsmAnd giving frequent and consecutive warnings for closely spaced speed bumps, resulting in an excessive number of audible crosswalk alerts. In contrast, other navigation apps handle this situation differently by providing a more concise instruction like "attention, several (few) bumps ahead."

OsmAnd~ 4.6.0#497m, released: 2023-07-28
OsmAnd 4.4.10

<table style="width:100%">
  </tr>
  <tr>
    <th>iOS</th>
    <th>Android</th>
  </tr>
  <tr>
    <td> <video source src="https://github.com/osmandapp/OsmAnd/assets/127092082/1a27129e-2ab4-402a-94dd-3516baca7930"  width="300" height="auto"/>   </td>
    <td><video source src="https://github.com/osmandapp/OsmAnd/assets/127092082/ef916964-cca3-403d-b7bf-d4d748795b3e"  width="300" height="auto"/>   </td>
  </tr>
</table>

ivanPyrohivskyi:
iOS needed

[...]

PR #18090, opened 2023-09-04
Avoid series of same types alarms

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/17723

### DIFF
--- a/OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
+++ b/OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
@@ -4,7 +4,6 @@ import static net.osmand.plus.routing.data.AnnounceTimeDistances.STATE_LONG_ALAR
 import static net.osmand.plus.routing.data.AnnounceTimeDistances.STATE_LONG_PNT_APPROACH;
 import static net.osmand.plus.routing.data.AnnounceTimeDistances.STATE_SHORT_ALARM_ANNOUNCE;
 import static net.osmand.plus.routing.data.AnnounceTimeDistances.STATE_SHORT_PNT_APPROACH;
-import static net.osmand.plus.settings.backend.ApplicationMode.*;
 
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
```diff
@@ -464,6 +464,8 @@ public void announceVisibleLocations() {
 										if (atd.isTurnStateActive(atdSpeed, dist, STATE_SHORT_ALARM_ANNOUNCE)) {
 											locationPointsStates.put(point, ANNOUNCED_DONE);
 											proceed = false;
+											//update lastAlarm location for avoid series of same types alarms
```
vshcherb, line 467: comment is not needed, location should be updated once it's in the list ? AlarmInfo lastAlarm = lastAnnouncedAlarms.get(t);

OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
```diff
@@ -357,8 +357,9 @@ public AlarmInfo calculateMostImportantAlarm(RouteDataObject ro, Location loc, M
 					if (info != null) {
 						if (info.getType() != AlarmInfoType.SPEED_CAMERA || showCameras) {
 							long ms = System.currentTimeMillis();
-							if (ms - announcedAlarmTime > 50 * 1000) {
-								announcedAlarmTime = ms;
+							Long timeLastAlarm = lastAnnouncedAlarmsTime.get(info.getType());
```
ivanPyrohivskyi, line 360: From OpenGL thread

ivanPyrohivskyi: Essentially this block do the same as in announceVisibleLocations but from another thread. These blocks supress each other.

PR #18141, opened 2023-09-12
Fix Infer Wikimedia Commons category from Wikidata to show photos 12303

Dima-1:


OsmAnd-java/src/main/java/net/osmand/wiki/WikiCoreHelper.java
```diff
@@ -116,6 +133,9 @@ public static class Claims {
 		@SerializedName("P18")
 		@Expose
 		private final List<P18> p18 = null;
+		@SerializedName("P373")
+		@Expose
+		private final List<P18> p373 = null;
```
vshcherb, line 138: Should be renamed to generic class ```P18 -> ?? ```

vshcherb:
Also Obj-C will be needed, request is generally approved

OsmAnd-java/src/main/java/net/osmand/wiki/WikiCoreHelper.java
ivanPyrohivskyi: WikibaseProperty

Chumva:
Looks like it is [fixed in iOS](https://github.com/osmandapp/OsmAnd-iOS/pull/3278), probably we have to sync code

PR #18145, opened 2023-09-12
convert issue templates into yml

TheSnoozer:
Hello, many thanks for OsmAnd!

I finally found time to make my first contribution here, so let's go "big" ;)

Github allows to encode an issue template as yml-file (see: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema). I believe that format suits better to OsmAnd, since the issue templates here are somewhat encoding instructions in comments. Instead the yml format allows a much greater control of what elements should be required (so one can enforce a "valid" issue) and one can somewhat properly encode them into a nice and more human friendly way.

You can try the suggestion directly here:
https://github.com/TheSnoozer/OsmAnd/issues/new/choose

Be warned that I slightly rephrased some of the wording and made it more consistent.


TheSnoozer:
Ohh and if you like with the yml format you can auto assign labels once an issue is created.

Something along the lines of:
```
labels: [ "bug", "to-triage" ]
```
would give the ticket the "bug" label and a "to-triage". I'm not sure how github handels the case when a label would be missing. Maybe it's simply not attached to the (newly) created ticket then.

Issue #16489, opened 2023-02-17
When creating OSM notes, they should include the application name - OsmAnd


angoca:
# 🚀 feature request

### Description

Most of the applications that allow the creation of OSM notes are including the application name in some way. Some use a hashtag (#Mapsme, #organicmaps) or some with a text (via StreetComplete XX.Y)

This will be useful for filtering the notes from OsmAnd and help us understand the context of the creation. Also, it allows separating the notes from erratic websites or applications from the correct ones from OsmAnd.

There is a thread in OSM Communities discussing this: https://community.openstreetmap.org/t/osm-notes-should-include-the-application-including-notes-from-osm-website/9064.

### Describe the solution you'd like

I would like the notes created from OsmAnd to include a hashtag like #OsmAnd or #OsmAnd4.3.8)

### Describe alternatives you've considered

Just appending a text does not affect the content of the note - https://resultmaps.neis-one.org/osm-notes-country-custom?c=Germany&query=open

franco999:
I am an OSM user and use the notes add-on from OsmAnd a lot. In my opinion, I would really like this minor addition as it would better sort and clarify the note context, its usage and its search filtering.

ceever:
Please also add some version information and datetime.

lint3:
Any way to disable this?

vshcherb:
Nope it was done generically, any issue with it?

lint3:
No issue, I just personally would prefer not to have it added. But it's not a big deal.
What do you mean that it was done generically?

vshcherb:
I mean it's done for all notes, the only solution would be to revert PR and it won't be added

mnalis:
Please keep the `#OsmAnd` hashtag in notes, it is useful

PR #18254, opened 2023-09-30
Add hashtag with appname in notes #16489

deevroman:
https://github.com/osmandapp/OsmAnd/issues/16489

Example note: https://master.apis.dev.openstreetmap.org/note/46911

### DIFF
--- a/OsmAnd/src/net/osmand/plus/plugins/osmedit/helpers/OsmBugsRemoteUtil.java
+++ b/OsmAnd/src/net/osmand/plus/plugins/osmedit/helpers/OsmBugsRemoteUtil.java
@@ -81,7 +81,7 @@ public class OsmBugsRemoteUtil implements OsmBugsUtil {
 				b.append(getNotesApi()).append("?");
 				b.append("lat=").append(point.getLatitude());
 				b.append("&lon=").append(point.getLongitude());
-				b.append("&text=").append(URLEncoder.encode(text, "UTF-8"));
+				b.append("&text=").append(URLEncoder.encode(text + "\n\n#OsmAnd", "UTF-8"));
 				msg = "Creating bug";
 			} else {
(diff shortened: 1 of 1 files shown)


vshcherb:
Is it a common practices for the apps? Or will it be just annoying trash as you can see in a conversation. 

Osm Notes is a conversation tool and not an editor, so it shouldn't matter which messenger do you use

deevroman:
> Is it a common practices for the apps?

More like yes. The most popular editors actively [using](https://github.com/Zverik/every_door/pull/642#issuecomment-1735680249) notes (StreetComplete, MapsMe/OrganicMaps) add such information. And even the version + date of the maps used in the application. 

> Osm Notes is a conversation tool

In most cases, no. By my count, only 13% of notes have more than two comments (including the note itself). More than 5 is 1.6%, more than 10 is 0.01%. But note that such meta information is only added when the note is created. It certainly doesn't interfere with the discussion.

----

In Osmand's case, there are two other good reasons to add meta-information:

1. OsmAnd allows you to add anonymous notes. Often such notes make little sense and the author cannot be contacted. Having the name of the app, will add credibility.
2.  OsmAnd's maps can lag months behind the database. As the mappers who process the notes say, it is very useful to understand the state of the OSM database as seen by the user.

mmd-osm:
In fact hashtags in notes are somewhat controversial, maybe check for yourself: https://github.com/openstreetmap/openstreetmap-website/issues/3932

deevroman:
In this case, the only thing in dispute is how the metadata about the application should look (with or without a hashtag). 

The main argument in https://github.com/openstreetmap/openstreetmap-website/issues/3932 is whether the notes data model should be changed. Something tells me that there will be no changes in the near future... And until there are no changes, we can only use the text of the note.

Issue #2159, opened 2016-01-16
More natural handling of the POI/Favorite along the route


raindropsfromsky:
When we are navigating along a route, OsmAnd can (optionally) display the POIs/Favorites along that route. Let's call it the "POI/Favorite card" for the sake of this discussion.

This card only allows the user to dismiss the next suggested place. 

But what if we want to accept the suggestion and stop over at that point? There is no such provision.

**_Desired:**_ 
Let the user tap on the left side of the card (area that displays (a) the icon of the place, (b) the distance+offset and (c) the name of the place.)

Immediately, OsmAnd should add that point to the route, and recalculate the route.

---

The  **_overflow**_ button in the card is meant to be used while the user is driving. But this is not useful at all: If he taps this button, he is presented with a bewildering list of destinations, POIs and Favorites. Clearly, he cannot deal with the complexity while driving. So why is this quick control provided??

If he pulls aside his car, he can simply go back to the **_Waypoints**_ dialog in just two steps: First he taps on the blue arrowhead, and then the white triangular flag button. Now he can see all the POIs/Favorite settings.

vshcherb:
I agree nice suggestion.

vshcherb:
Let's add to 3 points context menu navigate to (add as first waypoint) and show the list

raindropsfromsky:
Yes, the context menu requires affirmative action, so it ensures that an accidental tap on the card does not automatically add it to the route. (Once added, the user cannot remove it easily, especially while he is driving).

sonora:
Please note that this is logically very close to suggestion #1862 (only there not with respect to a route, but with respect to your position alone)). In m mind such a "map discoverability" feature s one of OsmAnd's big business opportunities, developing our app to some sort of electronic tour guide, facilitating e.g live discovery of POIs along a route (e.g. on a tourist walking tour through a city, where your decide if what you want to see next is rather a cathedral, or a pub for lunch).  :-)

vshcherb:
Yes I remember #1862 and immediately made mental connection. I also have couple examples where closest poi would be great.

sonora:
Quick implementation would be having a button "Show nearby POI", which would put you on the existing Search/POI/categories screen. (Very similar to the already existing "Search near hear" menu option, which puts you on the Search screen also, but does not force the "POI" tab there).

While this new "Show nearby POI" option may actually be worth a button of its own, a menu item under "3-dots" may suffice?

raindropsfromsky:
[...]

PR #18340, opened 2023-10-13
#2159: Fixes after review and crashes

Corwin-Kh:


OsmAnd/src/net/osmand/plus/track/helpers/GPXDatabase.java
```diff
@@ -235,7 +237,7 @@ public class GPXDatabase {
 	private static final String GPX_MIN_CREATE_DATE = "SELECT " +
 			"MIN(" + GPX_COL_FILE_CREATION_TIME + ") " +
 			" FROM " + GPX_TABLE_NAME + " WHERE " + GPX_COL_FILE_CREATION_TIME +
-			" != 0";
+			" > " + CHECKED_UNKNOWN_TRACK_CREATION_TIME;
```
vshcherb, line 240: Doesn't look good needs to be fixed

OsmAnd/src/net/osmand/plus/track/helpers/GPXDatabase.java
```diff
@@ -640,6 +642,9 @@ public boolean updateLastUploadedTime(@NonNull GpxDataItem item, long fileLastUp
 	}
 
 	public boolean updateCreateTime(@NonNull GpxDataItem item, long fileCreatedTime) {
+		if (fileCreatedTime == 0) {
```
vshcherb, line 645: All created time needs to be updated to latest modified time if it's unknown

PR #18392, opened 2023-10-20
Fix 16434 Highway shield steals numbers from the route/ref/name

Dima-1:


OsmAnd/src/net/osmand/plus/routing/RoutingHelperUtils.java
```diff
@@ -27,23 +30,40 @@ public class RoutingHelperUtils {
...
+		return formatStreetName(name, ref, destination, towards, null);
+	}
+
+	@NonNull
+	public static String formatStreetName(String name, String originalRef, String destination, String towards, RoadShield shield) {
+		StringBuilder formattedStreetName = new StringBuilder();
+		if (originalRef != null && originalRef.length() > 0) {
+			String[] refs = originalRef.split(";");
+			for (String ref : refs) {
+				if (shield != null && !isRefEqualsShield(shield.getText(), ref)) {
```
ivanPyrohivskyi, line 42: Why shield != null do not move to the top if statement ?

PR #18413, opened 2023-10-23
Auto zoom draft

cepprice:


Chumva:
It looks like it was implemented [here](https://github.com/osmandapp/OsmAnd-Issues/issues/2056)

PR #18437, opened 2023-10-26
Revert "Fix 16434 Highway shield steals numbers from the route/ref/name"

vshcherb:
Reverts osmandapp/OsmAnd#18392


vshcherb:
We need to revalidate structure

PR #18443, opened 2023-10-27
Fix bicycle private

Dima-1:


OsmAnd-java/src/main/java/net/osmand/binary/RouteDataObject.java
```diff
@@ -827,15 +829,15 @@ public String getHighway() {
...
 		for (int i = 0; i < sz; i++) {
-			RouteTypeRule r = region.quickGetEncodingRule(types[i]);
-			if ("motorcar".equals(r.getTag())
-					|| "motor_vehicle".equals(r.getTag())
-					|| "vehicle".equals(r.getTag())
-					|| "access".equals(r.getTag())) {
-				if (r.getValue().equals("private")) {
+			RouteTypeRule rule = region.quickGetEncodingRule(types[i]);
+			String tag = rule.getTag();
+			if ("vehicle".equals(tag) || "access".equals(tag)
```
vshcherb, line 837: Faster to switch if's
```if (rule.getValue().equals("private")) {
 String tag = rule.getTag();
			if ("vehicle".equals(tag) || "access".equals(tag)
```

vshcherb: Code is hard to read...
Probably it's time to split by many if's and group by profiles

Dima-1: split by profiles

OsmAnd-java/src/main/java/net/osmand/binary/RouteDataObject.java
```diff
@@ -827,17 +829,20 @@ public String getHighway() {
...
-			if ("motorcar".equals(r.getTag())
-					|| "motor_vehicle".equals(r.getTag())
-					|| "vehicle".equals(r.getTag())
-					|| "access".equals(r.getTag())) {
-				if (r.getValue().equals("private")) {
-					return true;
-				}
+	public boolean hasPrivateAccess(GeneralRouterProfile profile) {
+		for (int type : types) {
+			RouteTypeRule rule = region.quickGetEncodingRule(type);
```
vshcherb, line 834:
```
if(rule.getValue().equals("private")) {
   if ("vehicle".equals(tag) || "access".equals(tag)) {
     return true;
   }
   if (profile == GeneralRouterProfile.CAR) {
      ...
   } 
}
```

PR #18450, opened 2023-10-27
Fix highway shields 2

Dima-1:


OsmAnd/src/net/osmand/plus/routing/CurrentStreetName.java
```diff
@@ -77,12 +83,78 @@ public static CurrentStreetName getCurrentName(@NonNull RoutingHelper routingHel
...
 			streetName.showMarker = true;
 		}
 		return streetName;
 	}
+
+	public static class RoadShield {
+		private final RouteDataObject rdo;
+		private final String tag;
+		private final String value;
+		private StringBuilder additional;
```
vshcherb, line 99: String is better?

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/StreetNameWidget.java
```diff
@@ -212,30 +214,23 @@ public boolean updateWaypoint() {
...
-		}
-		for (int i = 0; i < object.nameIds.length; i++) {
-			String key = object.region.routeEncodingRules.get(object.nameIds[i]).getTag();
-			String val = object.names.get(object.nameIds[i]);
-			if (key.startsWith("route_road") && key.endsWith("_ref")) {
-				boolean visible = setRoadShield(object, key, val, additional);
-				if (visible) {
-					return true;
-				}
+	private boolean setRoadShield(@NonNull List<RoadShield> shields) {
```
vshcherb, line 217: do we need limit to have only 2-3 max shields

OsmAnd/src/net/osmand/plus/routing/CurrentStreetName.java
Dima-1: Further down the code we add more tags to the field "additional" in the loop. StringBuilder more efficient.

PR #18477, opened 2023-11-01
Fix Keybindings after 1st UI Review (2)

nazar-kutz:


OsmAnd/src/net/osmand/plus/keyevent/InputDeviceHelper.java
```diff
@@ -228,9 +242,38 @@ private void notifyListeners(@NonNull EventType event) {
...
+	 *                   from where the public method was called.
+	 */
+	private void updateAppModeIfNeeded(@NonNull ApplicationMode appMode) {
+		if (this.appMode != appMode) {
+			this.appMode = appMode;
+			updateCachedInputDevices();
+		}
+	}
+
+	private void updateCachedInputDevices() {
```
Chumva, line 263: Probably we need separate cache for settings UI and backend

PR #18479, opened 2023-11-01
Speed up haversine getDistance() by ~20%

RZR-UA:


sonora:
@RZR-UA Well spotted, thanks!!

Issue #18284, opened 2023-10-04
routing for bike doesn't use height information to estimage duration of trip


starbrights:
### Description

The router returns a much to less time for bicycle rides and doesn't seem to use information of height.

### Steps to reproduce

Using 4.5.10 (latest version from FDroid).
Select a trip of around 6.5km and 400 high meters and I get a result of 21min. That's far to less.
Or 11km and 900hm in 20min. This is far to less.

Used the offline map of Piemont/Italy, 3D map of (hole) Italy and lines of hight are loaded.

### Actual result

Time is like drive in flat environment

### Expected result

brouter returns a much longer time

### Your Environment (required)

OsmAnd Version: 4.5.10
Android/iOS version: LOS 20 / A13
Device model: Samsung S10

sonora:
Perhaps the issue title is mislesding  because elevation data (if present on your device) is actually used when "Use elevation data" is active, right? Please activate/deactivate (there's a shortcut button on the route config screen) and observe the difference in travel time estimate.

scaidermern:
For me the chosen route changes but the travel time doesn't (after adding via points so that the chosen route is the same in both cases). So it looks like height data doesn't influence the displayed the travel duration.

sonora:
Yes, that's due to the "Hilly", "Less hilly" etc. route optimization setting for bicycles. Perhaps to start debugging this, select a short route up a road where there's no alternatives. For me the travel time does change, like e.g. between 11 and 14 minutes (1.9 km  60 m uphill).

starbrights:
For me the route indeed changes  when activate that, but not travel time. 
I would expect that travel time always uses hight information if available, even if route is not "optimized" for that, right?

BTW: Do you know what this height optimzation does? It can reduce the max number of hight, but sometimes it might be better to limit the steepness.

scaidermern:
> Perhaps to start debugging this, select a short route up a road where there's no alternatives. For me the travel time does change, like e.g. between 11 and 14 minutes (1.9 km 60 m uphill).

Nope, the travel time is the same, despite having to go 133 meters up.

![photo_2023-10-05_09-53-43](https://github.com/osmandapp/OsmAnd/assets/1040784/c76047e5-9331-420a-bf68-363f9fe7ab83)

https://osmand.net/map?start=47.593422%2C10.346116&end=47.608791%2C10.374468&mode=bicycle#13/47.576386/10.360292

From: 47.59342, 10.34612
[...]

PR #18549, opened 2023-11-13
Use Naismith's/Scarf rules for bicycle routing complete_time #18284

RZR-UA:


sonora (commented):
@RZR-UA I have added an extra commit changing the comments a bit and streamlining the code. Should be ok now - thanks!

Issue #18093, opened 2023-09-04
Navigation issue: reason circular road is treated as roundabout


Nine-Tailed:
## Description
Osmand is building too big of a loop. There are no restraining tags on the road. Navigation was tested for both car and pedestrian use
<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/12549744/63eca1db-ac8d-46f7-a8e4-8b5e04cc3307"  width="300" height="auto"/> </td>
  </tr>
</table>

Route built by OSRM:
<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/12549744/d3f4b359-3748-4c28-b258-46b9bff343bd"  width="300" height="auto"/> </td>
  </tr>
</table>

Place:
https://osmand.net/map/?start=51.556292,39.489428&end=51.555255,39.489878&type=osmand&profile=car#18/51.55571/39.49006

**Environment**
OsmAnd Version: 4.6.0#718m
Android version: 13
Device model: poco f3

mic140:
![IMG_20230904_195407](https://github.com/osmandapp/OsmAnd/assets/69987154/d4c7c2b4-40ac-4f7a-98bf-e73da76a3417)

RZR-UA:
https://github.com/osmandapp/OsmAnd-resources/pull/1022

RZR-UA:
> Related to #17715? Although this one here doesn't involve oneway tags.

#17715 is not connected to the current Issue.

#18093 is a routing calculation bug (already fixed in testing-branch).

#17715 seems to be related to the road/turn type detection/display.

RZR-UA:
Road/Turn type bug is confirmed:

https://osmand.net/map/?start=60.518789,5.239019&finish=60.518953,5.240178&type=osmand&profile=car#19/60.51898/5.23973

RZR-UA:
**Status:**

Already fixed in HH-routing branch and should be finally fixed in the next releases.

Routing Test # 63 to check this bug was created.

ivanPyrohivskyi:
Wait new implementation of routing (HH routing), therefore task is not close!
After this need remove "ignore"="true" in the test "63. Reason circular road is treated as roundabout. https://github.com/osmandapp/OsmAnd/issues/18093"

RZR-UA:
Finally fixed.

PR #18565, opened 2023-11-15
Tidy up routing tests (use distinct directory) Issue#18093

RZR-UA:


RZR-UA:

PR #18688, opened 2023-12-07
Reduce excessive yawing during navigation

Andrewschenko:
Use current location to calculate map rotation during navigation


Chumva:
It should already be implemented [here](https://github.com/osmandapp/OsmAnd/pull/18693)

PR #18694, opened 2023-12-08
Fix nearest cities search for tracks

Chumva:


OsmAnd/src/net/osmand/plus/track/helpers/GpxReaderTask.java
```diff
@@ -158,7 +160,11 @@ private void sortAmenities(@NonNull List<Amenity> amenities, @NonNull LatLon lat
 		Collections.sort(amenities, (o1, o2) -> {
 			double distance1 = MapUtils.getDistance(latLon, o1.getLocation());
 			double distance2 = MapUtils.getDistance(latLon, o2.getLocation());
-			return Double.compare(distance1, distance2);
+
+			double d1 = distance1 / (Algorithms.stringsEqual("town", o1.getSubType()) ? TOWN_SEARCH_RADIUS : CITY_SEARCH_RADIUS);
```
vshcherb, line 164: Code could support all CityType and radius from it. why have it's own constants?

Chumva: If we use radius from cityType we wint find nearest cities in many cases, for example [this track](https://github.com/osmandapp/OsmAnd-Issues/issues/2282#issuecomment-1838469455) did not get city because of small search radius

PR #18698, opened 2023-12-08
fix German translation for board:title.xml

Elefant-aus-Wuppertal:
The German translation for board:title was not completely correct until now, so I changed it.


sonora:
Thanks! You may in the future also use Weblate to do this more easily.

PR #18700, opened 2023-12-09
Add support for more capacity:* subkeys and service:bicycle:stand

Henry00572:
This is supposed to add some tags to the POI information (shown when you click on a POI). I added English and German translations. The added tags are:

Tag | Uses
---|---
`capacity:charging` | 6.2k
`capacity:cargo_bike` | 1k
`service:bicycle:stand` | 1.6k


vshcherb:
@xmd5a2 I'm not sure it's a correct place to merge these changes, please double check

xmd5a2:
This is only iOS translations. In order to add these tags you should modify https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/poi_types.xml and https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/res/values/phrases.xml

PR #18718, opened 2023-12-11
Fix search results UI

Dima-1:


OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
```diff
@@ -852,7 +854,36 @@ public boolean publish(SearchResult object) {
...
+		public List<String> getTranslatedAdditionalInfo(Amenity amenity) {
+			List<String> result = new ArrayList<>();
+			Map<String, String> additionalInfo = amenity.getInternalAdditionalInfoMap();
+			if (!Algorithms.isEmpty(additionalInfo)) {
+				MapPoiTypes poiTypes = MapPoiTypes.getDefault();
+				for (Map.Entry<String, String> entry : additionalInfo.entrySet()) {
+					String key = entry.getKey();
+					String value = entry.getValue();
+					AbstractPoiType pt = poiTypes.getAnyPoiAdditionalTypeByKey(key);
+					if (pt == null && !Algorithms.isEmpty(value) && value.length() < 50) {
```
vshcherb, line 867: Is it a copy paste another method?

OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
```diff
@@ -852,7 +854,36 @@ public boolean publish(SearchResult object) {
 			}
 			return false;
 		}
-		
+
+		public List<String> getTranslatedAdditionalInfo(Amenity amenity) {
+			List<String> result = new ArrayList<>();
```
vshcherb, line 859: Old method used to exclude ```boolean excludeZipped``` now it doesn't filter anything. Implementation looks very strange it can't be reviewed without deep context understanding

OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
ivanPyrohivskyi: It is partially copied from [here](https://github.com/osmandapp/OsmAnd/blob/08ea213514837408f0d68394d5081bc76a8e0979/OsmAnd/src/net/osmand/plus/helpers/AmenityExtensionsHelper.java#L93)  and no idea how to refactoring.
Better to left as is.

vshcherb: No idea too but it's definitely not correct to just copy, this code is used for editing or for displaying context menu and it doesn't look logical to reuse in search. 
So code is not clear and I couldn't accept it.

OsmAnd/src/net/osmand/plus/search/listitems/QuickSearchListItem.java
```diff
@@ -119,9 +121,41 @@ public static String getName(OsmandApplication app, SearchResult searchResult) {
 
 	public String getTypeName() {
 		String typeName = getTypeName(app, searchResult);
+		if(searchResult.object instanceof Amenity) {
+			searchResult.alternateName = getTranslatedAdditionalInfo();
```
vshcherb, line 125: Assignment in get method - bad practice

OsmAnd/src/net/osmand/plus/search/listitems/QuickSearchListItem.java
```diff
@@ -119,9 +121,41 @@ public static String getName(OsmandApplication app, SearchResult searchResult) {
...
+		}
 		return (searchResult.alternateName != null ? searchResult.alternateName + " • " : "") + typeName;
 	}
 
+	public String getTranslatedAdditionalInfo() {
+		Map<String, String> additionalInfo = ((Amenity) searchResult.object).getInternalAdditionalInfoMap();
```
vshcherb, line 131: > getInternalAdditionalInfoMap();
Shouldn't be called outsdie Amenity, so this code should be restructured

vshcherb (changes requested):
Some code changes needed

ivanPyrohivskyi (approved):

vshcherb:
- New method ```	public String getTranslatedAdditionalInfo(String alternateName) {``` is not used anywhere before code?
- Make private ``` getInternalAdditionalInfoMap``` and refactor code

ivanPyrohivskyi (approved):

vshcherb (approved):

PR #18733, opened 2023-12-13
draft [BRP] Fix NativeLib call (intermediatesEmpty VS precise X/Y)

RZR-UA:


RZR-UA:
closed. intermediatesEmpty should be removed and the intermediate points shall be passed into C++ code directly as is (to work same way as in iOS)

Issue #18512, opened 2023-11-06
HHRoutePlanner / Tests / C++


vshcherb:
HHRoutePlanner
- Create tests (random test points on Germany) - pseudorandom 1 month same (?) 1) check all routing combinations > 10 (same routing time / ...) and geometry - 0.1% 2) Check how many ^visitedSegments (~speed) -> to monitor that speed doesn't get worse [car/bicycle]
- Use tests for C++/ Java

RZR-UA:
RandomRouteTest tech plan:

Prepare:

- [ ] read *.obf from maps dir (use MapCreator's way)

- [ ] think how to select and validate points from obf(s)

- [ ] choose random profile, start, finish, and 0-2 intermediate points

- [ ] create route-id (reproducible), web-route-url, incl. profile, points, params

Classic (with refreshed BinaryRouterPlanner):

- [ ] run BinaryRouterPlanner searchRoute() in Java A* mode

- [ ] run BinaryRouterPlanner searchRoute() in Java Dijkstra mode

- [ ] run BinaryRouterPlanner searchRoute() with C++ native lib (TODO cpp-port first)

- [ ] run C++ RoutePlannerFrontEnd::searchRoute() to simulate iOS OsmAnd routing flow (TODO?!)

HH:

- [ ] run HHRoutePlanner (MapRouterLayer/hhRoute as example) in Java A* mode (TODO interpoints)

- [ ] run HHRoutePlanner (MapRouterLayer/hhRoute as example) in Java Dijkstra mode (TODO interpoints)

- [ ] run HHRoutePlanner (MapRouterLayer/hhRoute as example) with C++ native lib (TODO cpp-port first)

- [ ] run HHRoutePlanner / iOS (later)

Finally:

- [ ] assume BinaryRouterPlanner (default) as trusted

- [ ] check % of difference: cost, distance (geometry)

- [ ] write GPX-files (multicolor, showing differences)

- [ ] compare pure run-time of each function (call twice, count 2nd run only)

vshcherb:
@RZR-UA - Web Tasks (4) / C++
@ivanPyrohivskyi - C++
@vshcherb - HH Routing bugs (Java)

------------------------------------

RZR-UA:
BRP C++ is almost fixed:

Was:

![was](https://github.com/osmandapp/OsmAnd/assets/90444451/d564e226-6463-4156-be9d-5dd32f67342c)

Now:

[...]

PR #18753, opened 2023-12-18
[BRP] Use QuadPointDouble to fix bug x31 (int<->float) --- proposal only

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18512

RZR-UA:
**TODO: RouteSegmentSearchResult : QuadPoint -> QuadPointDouble (and check Android build)**

RZR-UA:
Decided to use int pair for xy31

PR #18764, opened 2023-12-18
Fix unnecessary map redrawing after device unlock

Andrewschenko:
Don't redraw map after unlocking the device


AndyB69:
Hello, can I ask if this issue #18764 will also resolve unneccessary redraw after the app comes to the foreground (for any reason, i.e. from screen off, or other app in foreground first)?

Issue #17898, opened 2023-08-11
OsmAnd live: POI data displayed


scaidermern:
The map still shows the old POI name (Istrien Gourmet) while the POI info already shows the new name (Verde 22):
https://www.openstreetmap.org/node/9123988124
<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/1040784/666354a7-7d87-4b96-8839-c66aadca98f1"  width="300" height="auto"/> </td>
  </tr>
</table>

Map and POI info show a brewery although the brewery tags have been removed:
https://www.openstreetmap.org/way/316383846
<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/1040784/e033adc5-8a0c-466e-b659-3c9b7588f423"  width="300" height="auto"/> </td>
  </tr>
</table>

Map Germany_bayern_europe.obf is from August. No overlapping maps activated. Deleting and redownloading live data doesn't help.

OsmAnd 4.5.10.

(This is actually the same report as #17893. However, in my case I can still reproduce this bug.)

Nine-Tailed:
See https://github.com/osmandapp/OsmAnd/issues/17887

Citro12:
Problem is solved by deleting the base map of the respective federal state and downloading it again

scaidermern:
> Problem is solved for me by deleting the base map of the respective federal state and downloading it again

Didn't work for me. I already tried re-downloading Bavaria and live data, two times. The old data is still displayed.

DmitryAlexei:
https://github.com/osmandapp/OsmAnd/issues/17887

DmitryAlexei:
@UserMicha try to clear the live updates cache. Sometimes opening and closing OsmAnd multiple times can help.

UserMicha:
Unfortunately, that doesn't help. I also deleted the cache of the app in the Android settings and ended it hard. Map is rebuilt, but the same appearance. As before, only switching to Engine 1 helps. It is Android 10 on a Samsung XCover.

vshcherb:
To be fixed with this issue - https://github.com/osmandapp/OsmAnd/issues/14714, https://github.com/osmandapp/OsmAnd/issues/17834

vshcherb:
Tasks need to be retested (probably related to https://github.com/osmandapp/OsmAnd/issues/18153)

ivanPyrohivskyi:
[...]

PR #18777, opened 2023-12-19
Fix live-updates for amenities

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/17898

Best place for test in Berlin is here (added opening hours):
https://www.openstreetmap.org/node/886088999

### DIFF
--- a/OsmAnd/src/net/osmand/plus/resources/ResourceManager.java
+++ b/OsmAnd/src/net/osmand/plus/resources/ResourceManager.java
@@ -83,6 +83,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/resources/ResourceManager.java
```diff
@@ -1150,10 +1151,10 @@ public List<AmenityIndexRepository> getAmenityRepositories() {
 		return getAmenityRepositories(true);
 	}
 
-	public List<AmenityIndexRepository> getAmenityRepositories(boolean includeTravel) {
-		List<String> fileNames = new ArrayList<>(amenityRepositories.keySet());
+	public LinkedList<AmenityIndexRepository> getAmenityRepositories(boolean includeTravel) {
+		LinkedList<String> fileNames = new LinkedList<>(amenityRepositories.keySet());
 		Collections.sort(fileNames, Algorithms.getStringVersionComparator());
-		List<AmenityIndexRepository> res = new ArrayList<>();
+		LinkedList<AmenityIndexRepository> res = new LinkedList<>();
```
vshcherb, line 1157: ? Linked list worse by performance than ArrayList, don't see any reason below

OsmAnd/src/net/osmand/plus/resources/ResourceManager.java
```diff
@@ -1194,16 +1196,31 @@ public List<Amenity> searchAmenities(SearchPoiTypeFilter filter, double topLatit
...
 		} finally {
 			searchAmenitiesInProgress = false;
 		}
-		return amenities;
+
+		// filter amenities for live-updates map
+		HashSet<Long> amenitiesHash = new HashSet<>();
+		List<Amenity> res = new ArrayList<>();
+		for (Map.Entry<Integer, List<Amenity>> entry : sortedResult.entrySet()) {
+			for (Amenity a : entry.getValue()) {
```
vshcherb, line 1212: >> sortedResult.entrySet()
Is nto sorted at all what the reason?

How is it different than old approach

ivanPyrohivskyi: List<Amenity> r = index.searchAmenities(...) - synchronize method
amenities.addAll(r); - at the adding moment r is empty, so adding only link of r, but not r values(), values are adding randomly (depend from speed of *.obf reading).
Complete list amenities available only _try { }_ block
So I strictly set order by reading indexes.

vshcherb (changes requested):
Code changes needed

PR #18785, opened 2023-12-20
not change wpt time if hidden was changed

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2328

OsmAnd-java/src/main/java/net/osmand/gpx/GPXFile.java
```diff
@@ -203,13 +203,17 @@ private GPXUtilities.WptPt getWptPt(String wptName, int wptIndex) {
 		return currentWpt;
 	}
 
-	public void updateWptPt(GPXUtilities.WptPt existingPoint, GPXUtilities.WptPt newWpt) {
+	public void updateWptPt(GPXUtilities.WptPt existingPoint, GPXUtilities.WptPt newWpt, boolean hiddenWasChanged) {
```
vshcherb, line 206: more generic - variable updateTimestamp and default true

Issue #18788, opened 2023-12-20
Make PENALTY_FOR_REVERSE_DIRECTION profile dependent and configurable in routing.xml


vshcherb:
- [x] DEFAULT_PENALTY_FOR_REVERSE_DIRECTION - 500 seconds which is too much for cycling / pedestrian (which should be probably ~ 10-20 or 0)
- [x] **Partially** (merged to routingHHBase) Fix for C++ as well
- [x]  Check / think about intermediate points (probably PENALTY_FOR_REVERSE_DIRECTION / 2 still makes sense)
- [x]  Test about Plan Route where it shouldn't be used

vshcherb:
С++ needed

yuriiurshuliak:
The route is built correct, proceeding directly towards the red endpoint without any looping involved.

https://github.com/osmandapp/OsmAnd/assets/127092082/74dbcb43-069c-43ab-be27-36d2b653aab8

PR #18888, opened 2024-01-10
Add penaltyForReverseDirection defined in routing.xml

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18788

PR #18896, opened 2024-01-11
Fix #18337 Missing time data for created or imported tracks

### LINKED ISSUE #18337: Missing time data for created or imported tracks
### Description

Good morning, I am writing to you to understand if there is some bug with Osmand or if there is some function to activate.

I found the application splendid and full of all the features necessary to undertake an outdoor activity.

I tried importing some of my gpx tracks and they correctly show elevation gain, slope and route length. However, what seems to be missing in all of them is travel time. The clock icon tells me 00:00 for all the tracks, and it seems strange to me that such an important data is missing.

I tried importing at least 9 tracks and the result is always the same. Trying to create the route directly from the application, setting the "walking" profile, identical problem.

### Steps to reproduce

The bug occurs following these operations:

1) Open the OsmAnd application.

2) Press the three lines at the bottom left and enter the menu.

3) Press Plan a route.

4) Create a route using the walking profile, then press save.

5) Return to the menu and press "my places"

6) Once pressed, you will see a summary of the route and the time will be 00:00

The same procedure is possible by importing the track. To do this instead of creating it, just press plan a route and then import track

### Actual result

The route time will be 00:00

### Expected result

See the estimated time to travel the tracks

```
OsmAnd Version: 4.6.2
Android/iOS version: Android 13
Device model: Samsung S20
Crash-Logs: ?
```

nazar-kutz:


OsmAnd-java/src/main/java/net/osmand/gpx/GPXTrackAnalysis.java
```diff
@@ -344,6 +348,26 @@ private void checkUnspecifiedValues(long fileTimeStamp) {
...
+			for (RouteSegment routeSegment : routeSegments) {
+				result += (long) (1000 * parseFloatSilently(routeSegment.segmentTime, 0.0f));
+			}
+			return result;
+		}
+		return 0;
+	}
+
+	public int getDurationInSeconds() {
+		return (int) (getDurationInMs() / 1000 + 0.5f);
```
alex-osm, line 364: Looks like `getDurationInMs() / 1000` should be float or double before addition

PR #18903, opened 2024-01-12
Relax (0.5x) reverse-penalty for HH with inter-points

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18788

OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
```diff
@@ -830,6 +830,9 @@ public RouteCalcResult searchRoute(final RoutingContext ctx, LatLon start, LatLo
 			HHRoutePlanner<NetworkDBPoint> routePlanner = HHRoutePlanner.create(ctx, null);
 			HHNetworkRouteRes r = null;
 			Double dir = ctx.config.initialDirection ;
+			if (targets.size() > 1) {
+				ctx.config.penaltyForReverseDirection /= 2; // relax reverse-penalty in case of inter-points
```
vshcherb, line 834: 1) not correct for 1st point 2) not applicable for other type of routings ?

RZR-UA: 1. It was correct and now is clarified additionally.

3. Yes, this penalty is basically applied for inter-points routes for HH-engine only. If you decide to have the same for BRP-engine, I'll implement it.

vshcherb: It's not correct, for the first point. It should be applied only to intermediates and revert later

RZR-UA: Fixed: 1) inter-points only, from second target; 2) reverted once after calculateHHRoute()

PR #18929, opened 2024-01-19
Fix useOnlyHHRouting (hhoff should be prioritized)

RZR-UA:


OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
```diff
@@ -268,9 +268,6 @@ public void setDefaultHHRoutingConfig() {
 	
 	public RoutePlannerFrontEnd setUseOnlyHHRouting(boolean useOnlyHHRouting) {
 		this.useOnlyHHRouting = useOnlyHHRouting;
-		if (useOnlyHHRouting && hhRoutingConfig == null) {
```
vshcherb, line 271: Why this code is deleted?

PR #18934, opened 2024-01-19
[HH] Add disableHHRoutingConfig/isHHRoutingConfigured - for HH-SERVER

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-tools/pull/744

PR #18951, opened 2024-01-25
New app strings reworked

comradekingu:


sonora:
@comradekingu 
Thanks, I like it! Will merge.

Perhaps you can check on one thing: We mostly use, I think in line with common practice, "auto zoom" (without hyphen), so that change seems questionable to me.

I think there is one other occurrence in the file of "auto-zoom", which would also deserve fixing, and all others exhibit the correct "auto zoom"?

PR #18952, opened 2024-01-25
Implement"Disable battery optimization" feature

nazar-kutz:


OsmAnd/src/net/osmand/plus/settings/bottomsheets/CustomizableQuestionV1BottomSheet.java
```diff
@@ -0,0 +1,144 @@
...
+		getDialogButton(0).onButtonClick(super::onThirdBottomButtonClick);
+	}
+
+	@Override
+	public int getFirstDividerHeight() {
+		return getResources().getDimensionPixelSize(R.dimen.dialog_content_margin);
+	}
+
+	@NonNull
+	private DisplayDialogButtonItem getDialogButton(int index) {
```
Chumva, line 122: Probably we need to support different number of buttons (1-3)

OsmAnd/src/net/osmand/plus/settings/bottomsheets/CustomizableQuestionV1BottomSheet.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	@Override
+	protected int getThirdBottomButtonTextId() {
+		return getDialogButton(0).getTitleId();
+	}
+
+	@Override
+	protected DialogButtonType getThirdBottomButtonType() {
+		return getDialogButton(0).getButtonType();
```
Chumva, line 108: Confusing order, we need third button, but use 0 index

PR #18953, opened 2024-01-26
Spelling: Auto-zooming/ed

comradekingu:
https://english.stackexchange.com/questions/7657/use-of-hypens-with-auto-autopopulate-auto-populate-or-auto-populate

https://grammarhow.com/auto-populate-or-autopopulate/

https://books.google.com/ngrams/graph?content=auto+zoom%2Cauto-zoom%2Cautozoom&year_start=1800&year_end=2019&corpus=en-2019&smoothing=3

Using a dash is consistent with other use of dashes in OsmAnd, it isn't wrong, only one source say to avoid "auto-" followed by a consonant, other than that cited as correct and conservative.
It is a bit of a wash on the whole.

"Auto zoomed past me", meaning either a car flying by, or auto-focus not doing its thing is what to avoid.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -2970,7 +2970,7 @@ You need to activate the sensor so OsmAnd can find it.</string>
     <string name="nearest_cities">Nearest cities</string>
     <string name="select_city">Select city</string>
     <string name="select_postcode">Postcode search</string>
-    <string name="quick_action_auto_zoom">Auto-zoomed map on/off</string>
+    <string name="quick_action_auto_zoom">Auto-zoom map</string>
     <string name="quick_action_auto_zoom_desc">Button to turn speed-controlled auto-zooming on or off.</string>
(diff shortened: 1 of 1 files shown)


sonora:
In my mind itt's like this, would you agree?
- auto zoom as a noun is just like automatic zoom, no hyphen.
- for auto-zoom as a verb or adjective you can hyphenate auto- as a prefix. In analogy to e.g. auto-tune in Merriam-Webster

sonora:
Based on that, this [Suggestion prefix-hyphenation](https://github.com/osmandapp/OsmAnd/pull/18953/commits/8c0cb10525c3bd735d95e3a80040e76fa7a1247d) would be the result, if you concur.

I wanted to avoid introducing yet more versions like "auto-zoomed", also because this would have to be aligned in the documentation again also.

This would also create consistency here:
- `<osmand_accessibility_description>`: "..._for example to auto-announce_...",
but we keep the noun cases like already have here:
- `<auto_announce_on>`: "_Start auto announcing_"
- `<auto_announce_off>`: "_Stop auto announcing_"

Issue #19038, opened 2024-02-06
Crash in Android Auto on ACTION_NAVIGATE intent


ntruchsess:
### Description

When in Android-Auto starting the navigation from a third-party POI-app OSMAnd+ receives the intent and crashes. Using the same functionality with GMaps starts up navigation as expected.

### Steps to reproduce

1) in AA make sure OsmAnd is the most recently used navigation-app.
2) in AA open a third-party POI-app that allows to trigger naviation. E.g. EVMaps (https://github.com/ev-map/EVMap) or EnbW mobility+ (both available in Google Play store).
3) Open a POI in that app and click that navigation-button the app provides.

### Actual result

OsmAnd opens in AA but instead of starting the navigation to the coordinates of the POI being send via ACTION_NAVIGATE intent it crashes.

### Expected result

start OsmAnd and show the coordinates or address that were included in the ACTION_NAVIGATE intent in the search-result panel to start a navigation to that point.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 4.7.0
Android/iOS version: 12
Device model: Samsung Galaxy S10 (SG973F)
Crash-Logs: ?
```

yuriiurshuliak:
Please attach [Crash logs](https://osmand.net/docs/user/troubleshooting/crash-logs) files after application crash: Menu → Help → Send crash log (Send logcat log)

ntruchsess:
crash log:
```
2024-02-07 17:40:09.220 19056-19056 AndroidRuntime          net.osmand.plus                      E  FATAL EXCEPTION: main
                                                                                                    Process: net.osmand.plus, PID: 19056
                                                                                                    java.lang.RuntimeException: java.lang.IllegalArgumentException: Template is in a loading state but a list is set, or vice versa
                                                                                                    	at androidx.car.app.utils.RemoteUtils.lambda$dispatchCallFromHost$0(RemoteUtils.java:155)
                                                                                                    	at androidx.car.app.utils.RemoteUtils$$ExternalSyntheticLambda1.run(Unknown Source:6)
                                                                                                    	at androidx.car.app.utils.ThreadUtils.runOnMain(ThreadUtils.java:39)
                                                                                                    	at androidx.car.app.utils.RemoteUtils.dispatchCallFromHost(RemoteUtils.java:147)
[...]

PR #19039, opened 2024-02-06
Fix Android Auto navigation intent handling (solves #19038)

ntruchsess:
this PR solves #19038 . It implements propper parsing of ACTION_NAVIGATE geo-url data, propagating either coordinates or address-query to the existing search.
In case no result is found it fixes a crash that would happen due to invalid state (itemlist being null while result is marked as finished loading) during call to onGetTemplate().


ntruchsess:
setting to draft as I'm going to change some details.

PR #19058, opened 2024-02-08
Send `lang` to API /userdata/{user-register,send-code} to use with localized email templates

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2380

Issue #18158, opened 2023-09-14
Lane guidance has difficulty suggesting to go straight ahead


TheSnoozer:
## Description

Lane guidance has difficulty suggesting to go straight ahead

### How to reproduce

Start: 48.58836°N, 7.75573°E
End: 48.59019°N, 7.74927°E

https://osmand.net/map/?start=48.588429,7.755468&end=48.590190,7.749270&type=osrm&profile=car#18/48.58895/7.75273

### Actual result
No lane in lane guidance is marked yellow

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/6849390/8c7ef8bc-960a-433e-85ef-22970ab1f514"  width="300" height="auto"/> </td>
  </tr>
</table>

### Expected result
straight ahead should be yellow in lane guidance

**Environment**
```
Device : RE513CL1
Brand : realme
Model : RMX3241
Product : RMX3241EEA
Build : RMX3241_11_F.07
Version : 13
App Version : OsmAnd+
Apk Version : 4.5.10 4510
```
Maps from 01 Sept 2023

yuriiurshuliak:
When inputting the provided coordinates into the simulation and activating the navigation, the "Lanes" widget fails to show the current road's lane and does not illuminate the lane for straight traffic during active navigation.

OsmAnd~ 4.6.0#770m, released: 2023-09-16

https://github.com/osmandapp/OsmAnd/assets/127092082/bfe884b5-4357-4739-942f-845a6a1aae38

vshcherb:
- [MUTE] Go ahead (TL|C|C,TR) onto Avenue des Vosges and go 0.2 km. 
- https://osmand.net/map/?start=48.588429,7.755468&finish=48.590589,7.751155&profile=car#16/48.5890/7.7515 - also this is not correct it's not a TR
- REGRESSION since 2023

yuriiurshuliak:
The bug has been fixed. The "Lanes" widget shows the current lane of the road.

OsmAnd~ 4.7.0#1969m, released: 2024-04-03

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/127092082/5b1a2a94-cb86-40dc-a85d-8f3fc3bb8e71"  width="300" height="auto"/> </th>
[...]

PR #19068, opened 2024-02-09
Fix turn lanes 18158

ivanPyrohivskyi:
To issue [18158](https://github.com/osmandapp/OsmAnd/issues/18158)

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
@@ -1507,7 +1507,7 @@ public class RouteResultPreparation {
 		int[] lanes = t.getLanes();
 		String turnLanes = TurnType.convertLanesToOsmString(lanes, true, false);
 		if (turnLanes != null) {
-			int[] uniq = getUniqTurnTypes(turnLanes);
+			int[] uniq = getUniqDirections(turnLanes);
 			// if we have 2 and more active directions
 			if (uniq.length >= 2) {
 				return turnLanes;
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -1808,21 +1816,13 @@ protected TurnType createSimpleKeepLeftRightTurn(boolean leftSide, RouteSegmentR
...
-			}
-		});
-
 		int size = rs.attachedAngles.size();
 		boolean allStraight = rs.allAreStraight();
 		int[] lanes = new int[1];
 		int extraLanes = 0;
 		double prevAngle = Double.NaN;
 		// iterate from left to right turns
-		for (int i = size - 1; i >= 0; i--) {
```
vshcherb, line 1825: What the difference? Situation should be symmetric cause we have the right driving countries and left driving driving countries, fix couldn't rely on changing iteration order

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -2264,7 +2288,10 @@ private int[] findActiveIndex(RouteSegmentResult prevSegm, RouteSegmentResult cu
 		if (rs == null) {
 			return pair;
 		}
-		int[] directions = getUniqTurnTypes(turnLanes);
+		int[] directions = getUniqDirections(turnLanes);
+		if (rs.roadsOnLeft + rs.roadsOnRight >= directions.length) {
```
vshcherb, line 2292: This is too broad condition which not correct for many correctly mapped intersections, turn:lanes could have "TL|C|TR" even for 2 way intersections cause TL could be further away, so replacing it with syntetic is not a good thing

ivanPyrohivskyi: It is not this case. Your case when we have less roads than a turn lanes on intersection. Such situation is working fine in our code.
Function findActiveIndex need for range of active lanes.

Issue #18969, opened 2024-01-29
Progress bar for new HH Routing


vshcherb:
Progress bar should be approximately smooth for most of the routes. 

Reference map: full Germany.
Profiles: **Bicycle**/ **Car**
Parameters: no parameters tweaks for now
Length of routes: 20-1000 km (3-4 types of routes)

The simple reference code is left RouteCalculationProgress
```
public float getLinearProgress() {
```

- [x] Implement interpoints progress
- [x] C++ (main part, calculations, bugs)
- [x] refactor Java enum to pure int-enum
- [x] RouteCalculationProgressWrapper

- [x] Test in Android
- [x] Test in iOS

Fix crashes after isCancelled() - moved to https://github.com/osmandapp/OsmAnd-Issues/issues/2412

yuriiurshuliak:
When HH Routing is enabled, a progress bar is displayed when routes are calculated.

OsmAnd~ 4.7.0#1698m, released: 2024-02-28

<table style="width:100%">
  </tr>
  <tr>
    <th>Android</th>
    <th>iOS</th>
  </tr>
  <tr>
    <td> <video source src="https://github.com/osmandapp/OsmAnd/assets/127092082/a81395ed-1799-434e-986c-3cc6e1fd6805"  width="300" height="auto"/>   </td>
    <td><video source src="https://github.com/osmandapp/OsmAnd/assets/127092082/83e84f60-dadc-47e4-98cd-e5fa817417a0"  width="300" height="auto"/>   </td>
  </tr>
</table>

PR #19082, opened 2024-02-12
Refactor hhIterationStep from enum to int (required for JNI to exchange progress with Native lib)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18969

PR #19088, opened 2024-02-12
add info to RenderingGenerationResult

alisa911:
This pull request is related to other requests and should be merged together with them.
https://github.com/osmandapp/OsmAnd-core-legacy/pull/172
https://github.com/osmandapp/OsmAnd-tools/pull/756
https://github.com/osmandapp/web/pull/330


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2355

OsmAnd-java/src/main/java/net/osmand/NativeLibrary.java
```diff
@@ -44,11 +46,40 @@ public NativeLibrary() {
 	}
 
 	public static class RenderingGenerationResult {
```
vshcherb, line 48: Do we need this class in Android

PR #19117, opened 2024-02-15
[HH] Avoid double call the prepareResult() (fix 1st/last segments + distance)

RZR-UA:


RZR-UA:

PR #19124, opened 2024-02-16
Fix Android Studio "prepareKotlinBuildScriptModel" error

RZR-UA:
**Fix Android Studio problem when `android` and `tools` projects are opened at the same time.**

Error: Task 'prepareKotlinBuildScriptModel' not found in project ':OsmAnd'.

Before:

After:

Links:

https://github.com/gradle/gradle/issues/14889#issuecomment-988134777

https://stackoverflow.com/questions/64290545/task-preparekotlinbuildscriptmodel-not-found-in-project-app/64418715


Chumva:
Unnecessary changes in master, these issues should be fixed locally

Issue #19125, opened 2024-02-16
regression: Android Auto, crash parsing geo-intent


ntruchsess:
### Description

PR #19115 attempted to fix issue #19038 but introduced a new crash parsing the intent.
crash happens in [String query = uri.getQueryParameter("q");](https://github.com/osmandapp/OsmAnd/blob/4c75209b4a4eb9749d79d16c458c76d28951e448/OsmAnd/src/net/osmand/plus/auto/NavigationSession.java#L250). The class android.net.Uri doesn't support parsing query-parameters of geo-urls. The parsing must be done in own code.

### Steps to reproduce

same steps as issue #19038

### Actual result

application crashes.

### Expected result

SearchResultScreen would open with coordinates respectivly searchphrase or address from intent. Unique search-result would trigger calculation of route and directly open NavigationScreen.

### Your Environment (required)

```
OsmAnd Version: 4.7.0 (build master commit 81267a773d913abfa7e856ea0344b865bfe46de7)
Android/iOS version: 12
Device model: Samsung Galaxy S10 (SG973F)
Crash-Logs: 
2024-02-16 18:38:53.213  3491-3491  AndroidRuntime          net.osmand.plus                      E  FATAL EXCEPTION: main
                                                                                                    Process: net.osmand.plus, PID: 3491
                                                                                                    java.lang.RuntimeException: java.lang.UnsupportedOperationException: This isn't a hierarchical URI.
                                                                                                    	at androidx.car.app.utils.RemoteUtils.lambda$dispatchCallFromHost$0(RemoteUtils.java:155)
                                                                                                    	at androidx.car.app.utils.RemoteUtils$$ExternalSyntheticLambda1.run(Unknown Source:6)
                                                                                                    	at androidx.car.app.utils.ThreadUtils.runOnMain(ThreadUtils.java:39)
                                                                                                    	at androidx.car.app.utils.RemoteUtils.dispatchCallFromHost(RemoteUtils.java:147)
                                                                                                    	at androidx.car.app.utils.RemoteUtils.lambda$dispatchCallFromHost$1(RemoteUtils.java:186)
                                                                                                    	at androidx.car.app.utils.RemoteUtils$$ExternalSyntheticLambda2.run(Unknown Source:8)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:938)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:99)
[...]

PR #19126, opened 2024-02-16
#19125 Android Auto, fix parsing coordinates of geo-intents, directly open RoutePreview if coordinates are valid

ntruchsess:
this PR implements issue #19130
What has been implemented:
1) class android.net.Uri cannot directly parse the scheme-specific part of geo-urls (This is the format that is being used in intents of type ACTION_NAVIGATE). This PR implements parsing the coordinates and query-part by using a regular-expression.
2) in case the geo-url already contains coordinates different than 0,0 those coordinates are used to directly open the RoutePreview-screen.
3) in case the coordinates are 0,0 the parameter q is retrieved from the query-string. As the query-string is url-encoded this encoding must be removed before passing on the query-parameter 'q' to the SearchResult-screen. As android.net.Uri does only decode the %-encoded parts but leaves '+' as it is the replacement of '+' by ' ' also has to be done in own code. (The search-screen doesn't recognise '+' as delimiter and will not find addresses encoded this way)

### DIFF
--- a/OsmAnd/src/net/osmand/plus/auto/NavigationSession.java
+++ b/OsmAnd/src/net/osmand/plus/auto/NavigationSession.java
@@ -46,6 +46,8 @@ import net.osmand.plus.views.OsmandMapTileView;
 import net.osmand.util.Algorithms;
 
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Session class for the Navigation sample app.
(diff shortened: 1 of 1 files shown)


Chumva:
This issue should be fixed [here](https://github.com/osmandapp/OsmAnd/pull/19312). I had to use parsing via GeoPointParserUtil because local parsing via pattern can lead to issues because it is not tested completely, also some pattern methods require newer android API. Can you check if it is working properly?

ntruchsess:
I just compiled and tested it, the fix works great for both Navigation-
intents with valid coordinates and those containing a search-query.
Reusing the existing GeoPointParserUtil is certainly a much better
solution than mine.

- Norbert

PR #19129, opened 2024-02-16
Added title (textPrefix) to Architect field in Amenity menu UI.

osmandlover:
This is a fix for the issue reported here: https://github.com/osmandapp/OsmAnd/discussions/19099

This adds the "Architect" label to POIs that have this label. It has an unintended side effect of changing the icon to an Architect-specific one, I can't understand why - but I think this is a good change. 

A cleaner fix would be to remove the code that blanks out the textPrefix (title) for these items, which results in all the text fields getting a title, not just the Description and those without icons. But, I guess there are reasons it is done that way - if so I would be interested to hear them. If there are other suggestions from maintainers about how to solve this problem or how to improve this UI I would be happy to implement them.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
+++ b/OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
@@ -186,6 +186,7 @@ public class AmenityUIHelper extends MenuBuilder {
 			boolean isWiki = false;
 			boolean isText = false;
 			boolean isDescription = false;
+			boolean isArchitect = false;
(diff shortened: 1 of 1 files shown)


vshcherb:
I don't think this is a correct fix, probably we should create a proper issue and have description tag for each like even in your example it looks bad for dates.

laramaceo:
Screenshot 2024-02-27 9 06 24 AM

For me the person icon is fine for both the artist and the architect, adding their respective titles.  The title for me is the most important thing, the icon is a support.
This is my opinion, but it is up to you to decide.

vshcherb:
I don't like the code looks even before you, I think we should always display hint as it might be confusing in the future for different kind of tags. So team need to clean it up

PR #19163, opened 2024-02-21
[HH, BRP] Refactor makeStartEndPointsPrecise + prepareResult

RZR-UA:
TODO:

- [ ] check OsmAnd Anrdoid (Java/Native) + gpxApprox

This request fixes the sequence of calls:

1) `makeStartEndPointsPrecise()` - to make precision - **firstly**

2) `prepareResult()` - to calculate segments (time, distance) and load obf-regions - **finally**

`makePrecise` (boolean) used because `searchRouteInternalPrepare()` method might be called by `findGpxRouteSegment()` where `makeStartEndPointsPrecise()` is not required

iOS/native refactor: https://github.com/osmandapp/OsmAnd-core-legacy/pull/183


RZR-UA:

ivanPyrohivskyi:
Generally I would be simplify code. Instead of adding many confused flags, need to return result from different branches (Java/C++/Java HH/C++ HH) without call makeStartEndPointsPrecise + prepareResult and do it in the end of calling.
For example in the method searchRoute we have three results in RouteCalcResult (Java/C++/C++ HH) and one HHNetworkRouteRes (Java HH). In the end of searchRoute we can call sequentially makeStartEndPointsPrecise + prepareResult (only for Java HH - separate).
For transfer to C++. For iOS have separate method RoutePlannerFrontEnd::searchRoute where we can also call makeStartEndPointsPrecise + prepareResult after call HH routing.

RZR-UA:
Will be refactored as a new request soon. Kept as draft.

RZR-UA:
closed

PR #19182, opened 2024-02-25
Simplify routing code flow, easily remove searchRouteWithInterSmartRecalc (later)

vshcherb:


OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
```diff
@@ -851,103 +867,76 @@ public RouteCalcResult searchRoute(final RoutingContext ctx, LatLon start, LatLo
...
 			}
-		}
-		
-		double maxDistance = MapUtils.getDistance(start, end);
-		if (!intermediatesEmpty) {
-			LatLon b = start;
-			for (LatLon l : intermediates) {
-				maxDistance = Math.max(MapUtils.getDistance(b, l), maxDistance);
-				b = l;
+			r = runJavaHHRoute(ctx, start, targets);
```
ivanPyrohivskyi, line 880: Perhaps else { here?
Otherwise after C++ routing we call Java routing

OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
```diff
@@ -1078,53 +1067,75 @@ private void makeSegmentPointPrecise(RouteSegmentResult routeSegmentResult, LatL
...
-			ctx.startX = start.preciseX;
-			ctx.startY = start.preciseY;
-			ctx.startRoadId = start.road.id;
-			ctx.startSegmentInd  = start.segStart;
-			ctx.targetX = end.preciseX;
-			ctx.targetY = end.preciseY;
-			ctx.targetRoadId = end.road.id;
-			ctx.targetSegmentInd  = end.segStart;
-			return runNativeRouting(ctx, recalculationEnd, null);
+			RouteSegmentResult[] res = runNativeRouting(ctx, null);
```
ivanPyrohivskyi, line 1128: Lost ctx.targetRoadId etc. for native routing

OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
```diff
@@ -1246,58 +1248,47 @@ private boolean hasSegment(List<RouteSegmentResult> result, RouteSegment current
...
-								firstPartRecalculatedRoute.add(prev.get(k));
-							} else {
-								restPartRecalculatedRoute.add(prev.get(k));
-							}
+		}
+		List<RouteSegmentResult> firstPartRecalculatedRoute = null;
+		List<RouteSegmentResult> restPartRecalculatedRoute = null;
+		List<RouteSegmentResult> prev = ctx.previouslyCalculatedRoute;
+		RouteSegmentPoint first = pnts.get(1);
+		for (int i = 0; i < prev.size(); i++) {
```
ivanPyrohivskyi, line 1265: NPE ?

OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
vshcherb: Yes if routing result or  not correct we call further (but not necessary java) C++ as well?

OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
vshcherb: > Lost ctx.targetRoadId etc. for native routing
Target road id in Java couldn't be set because addSegment should happen in C++, to speed up routing. It was working before as well.

OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
RZR-UA: In this case, runJavaHHRoute() is always called after runNativeRouting(). Please fix.

OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
RZR-UA: Actually, not set in C++. The start/end RoadId & start/end SegmentInd were set from Java side because java_wrap calls BRP directly w/o C++ RoutePlannerFrontend functions (where is the setters located). So we need this vars passed directly from Java.

RZR-UA: fixed

ivanPyrohivskyi (approved):

PR #19184, opened 2024-02-26
Android Auto navigation improvements

linuxrocks123:
This is the result of a weekend project I did to make Android Auto work better for me.  It does the following:

- Uses sunrise/sunset to determine whether to use dark mode or not, not the data from the car.  The reason for this change is that my personal Android Auto instance erroneously tells mapping applications that it is night even during the day.  The car is fine: the headlights correctly come on at night and not during the day.  Android Auto must be buggy.  I see from searching the Internet that others have experienced this bug as well, and the symmetrical bug where Android Auto is erroneously rendering day mode at night.  So, I'd say better to just use sunrise/sunset than rely on Android Auto to give correct information.

- Enables "turn after next" information reporting and gives "turn after next" information unconditionally, not just when the next turn is imminent.  I find this very useful so I can pick the correct lane when making a turn.

I know that Google Play has idiotic PITA requirements for Android Auto navigation apps, so it may be that the features in this PR can't be implemented in an APK distributed on Google Play.  If that is the case, I'd suggest conditionally enabling the features in this PR such that they are at least available in the F-Droid APK.

I'd also suggest maybe adding a menu option for sunrise/sunset versus "use what the car says about day/night" and for "display turn after next".  Having a toggle might be helpful to some people.

Any such modifications, however, I'll have to leave as an exercise for the maintainers :)  Now that my copy of OsmAnd does what I want, I don't have the free time to make further revisions.  However, I do hope this code is useful to others, and I will be posting my APK on my personal website so that others can get it if they want.  (I will post the link in a comment later.)

Thanks for making a great app.  You guys rock.

### DIFF
--- a/OsmAnd/build.gradle
+++ b/OsmAnd/build.gradle
@@ -29,10 +29,10 @@ android {
 		}
 
 		publishing {
-			storeFile file("/var/lib/jenkins/osmand_key")
-			storePassword System.getenv("OSMAND_APK_PASSWORD")
-			keyAlias "osmand"
-			keyPassword System.getenv("OSMAND_APK_PASSWORD")
+			storeFile file("../keystores/debug.keystore")
+			storePassword "android"
+			keyAlias "androiddebugkey"
+			keyPassword "android"
 		}
 	}
 
--- a/OsmAnd/src/net/osmand/plus/NavigationService.java
+++ b/OsmAnd/src/net/osmand/plus/NavigationService.java
@@ -318,7 +318,7 @@ public class NavigationService extends Service {
 							false/*routingHelper.isRouteWasFinished()*/,
 							destinations, trip.getSteps(), destinationTravelEstimate,
 							lastStepTravelEstimate != null ? lastStepTravelEstimate.getRemainingDistance() : null,
-							false, true, null);
+							true, true, null);
 				}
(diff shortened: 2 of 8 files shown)


vshcherb:
> Uses sunrise/sunset to determine whether to use dark mode or not, not the data from the car.

It sounds but it's not good according to guardrails. We can't accept it cause we had initially implemented but then it won't pass Google checks to publish on Google Play.

I don't think you will be able to use Android Auto on F-Droid build without magic tricks... At least we didn't succeed for most of our F-Droid users ...

vshcherb:
So it's probably good to stay though I can't approve it for now.

linuxrocks123:
As promised, here's a link to my website with the APK: https://moongate.ydns.eu/gps.html

@vshcherb you might want to at least look at merging the "next turn" part of the PR for now, possibly controlled by an option if some people don't want the extra information.  Android Auto's guidelines shouldn't care about that since their API explicitly supports passing that information around.

Also, I am an F-Droid APK user who was successfully able to use "magic tricks" to get Android Auto working.  Those magic tricks are detailed on my website (can't guarantee Android 11 and up work, but it should if rooted). So, it wouldn't be a useless feature to add for that build :)

Issue #19083, opened 2024-02-12
Crash when opening the Track Appearance screen


zipav:
### Description

Our user complained that while opening Track Appearance screen to change track colour the app crashes.

### Steps to reproduce

-  Show track on the map
- Open Track Context menu
- Appearance 
- Check the result

### Actual result

Crash
[exception (4).log](https://github.com/osmandapp/OsmAnd/files/14237460/exception.4.log)

### Expected result

Normal work with track in the application

### Your Environment (required)

```
OsmAnd Version: OsmAnd 4.6.11, released: 2023-12-28
Android/iOS version: 
Device model: Samsung Galaxy S22 Ultra
Crash-Logs: ?

```

PR #19194, opened 2024-02-26
#19083: Fixed indexd out of bounds

Corwin-Kh:


OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapRouteReaderAdapter.java
```diff
@@ -362,7 +362,11 @@ public int getRefTypeRule() {
 		}
 
 		public RouteTypeRule quickGetEncodingRule(int id) {
-			return routeEncodingRules.get(id);
+			if (routeEncodingRules.size() > id) {
```
alex-osm, line 365: Wrong fix. Should be fixed at higher level.

PR #19221, opened 2024-03-01
Fix #19049 Tracks sort button is half hidden

### LINKED ISSUE #19049: Tracks sort button is half hidden
This appeared together with the new Tracks UI and has never been fixed, see below. I only see this on some devices, probably dependent on screen resolution and/or font size selected.

PS: As can be seen from the spinner and the still "empty" list, this screenshot was taken during the 12-15 second wait I experience upon every opening of this screen, or every scroll operation to the top of the list on this screen which also initiates a 12-15 sec list refresh, see #17805.

Screenshot_20240208-063342_OsmAnd~

0xRe1nk0:


sonora:
Thanks! Have you seen the second screenshot I posted? That may require a similar fix?

Issue #17805, opened 2023-08-03
opening new My Tracks very slow with Android 12


pebogufi:
## Description
With new My tracks my phone S10 is very slow, but not my Tab with Android 10.
I use  two devices, which are nearly complete identicaly in OsmAnd folders and files.
+Samsung S 10, Android 12, installed on internal, Android/media
+Samsung Tab S5e, Android 10, installed on Sd card, Android/obb
The S10 is normally **much** faster, example: install of osmand 
S10 needs 5-8 sec, 
Tab needs more than 20 sec.

BUT opening my tracks needs (total files in tracks subfolders ~500)
S10 5-7 sec
Tab ~1 sec.
I suppose the problem is the file access restrictions in Android 12 which uses extended file acces when creating statistics (tracks, distance, uphill, downhill, duration, size) which are at bottom line in tracks page.

### proposed change
Please add an option/checkbox to decide whether this automatic calculation of statstics should be done. I think this will not only help a lot of users with new Android versions > 10, but make them really happy !

**Problem 2**
Opening My tracks / "Visible on map"  with S10, Android 12 needs so much time, that i often get a message from system, whether to continue or abort the app.
With Tab it is also very slow, 5-10 sec, but no system warning. Visible files with both devices max 5, total files in tracks subfolders ~500.

**Environment**
OsmAnd Version: OsmAnd~ 4.6.0#537m, veröffentlicht: 2023-08-02
Android version: 12
Device model: Samsung S 10

pebogufi:
With S10:
After removing all but 50 files in tracks subfolders and no tracks visible the function My tracks / "Visible on map" needs 3-5 sec to show only one line "currently recording track" while recording is off.

Corwin-Kh:
This issue should be checked on latest build as it could be already fixed. Aside from this issue, there was another one that is related to SmartFolders and affected opening My places -> tracks. It is fixed in [#2159](https://github.com/osmandapp/OsmAnd-issues/issues/2159)

pebogufi:
> tracks. It is fixed in [#2159](https://github.com/osmandapp/OsmAnd-issues/issues/2159)

Link reports error 404, not found

After using again my folders with 767 gpx files
opening track folder needs 10 sec, that is no improvement to the initial report.

Corwin-Kh:
[...]

PR #19228, opened 2024-03-02
Speed up tracks screen when no Smart Folders are used (partial fix for #17805)

sonora:


Chumva:
Won`t there be issues if we add new smart folders, but allAvailableTrackItems in SmartFolderHelper is empty?

sonora:
I have tested bot creating a the first smartfolder and removing the last smart folder with that patch, and both worked flawlessly.

PR #19239, opened 2024-03-04
Speed up tracks screen

sonora:


OsmAnd/src/net/osmand/plus/myplaces/tracks/filters/SmartFolderHelper.kt
```diff
@@ -197,10 +193,11 @@ class SmartFolderHelper(val app: OsmandApplication) {
...
 	fun addTrackItemToSmartFolder(item: TrackItem) {
-		LOG.debug("addTrackItemToSmartFolder " + item.name)
-		val newSet = HashSet(allAvailableTrackItems)
-		newSet.add(item)
-		allAvailableTrackItems = newSet
+//		LOG.debug("addTrackItemToSmartFolder " + item.name)
+//		val newSet = HashSet(allAvailableTrackItems)
+//		newSet.add(item)
+//		allAvailableTrackItems = newSet
+		allAvailableTrackItems.add(item)
```
Chumva, line 200: Will this lead to crashes with multithreading?

sonora: That I am not sure. We can try without that first, I think it may not cause the majority of the improvement. I believe the biggest issue was one of the log commands executed very often.

PR #19280, opened 2024-03-08
Fix NPE when useOnlyHHRouting && HH-failed

RZR-UA:


RZR-UA:

vshcherb:
https://github.com/osmandapp/OsmAnd/commit/638884a2747504c5055921118715c8e7e13dbadf

Issue #19016, opened 2024-02-04
HH C++ Routing missing most of the turns info


mdykierek:
### Description

I had 2 situations where OsmAnd+ shows only one, very distant direction while there were plenty of intersections that need guidance. One of them I describe here, the other one was https://www.openstreetmap.org/way/47460763 that showed only Head with straight arrow, while I was supposed to take A4 towards east.

### Steps to reproduce

Going from Mikulov in Czech Republic to Ruda Śląska in Poland.
The highway intersection in Brno has a closed link to D1 (https://www.openstreetmap.org/way/142312586).
The closure was after offline maps update in January 2024 so I had to take a detour.

### Actual result

After taking the detour, OsmAnd+ re-calculated the route and the only direction was to take slight right in a couple hundreds of kilometers - leave A4 in Ruda Śląska.

### Expected result

There should be plenty of guidance for many highway intersections e.g. Vyskov, Prostejov, Olomouc etc.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: OsmAnd+ 4.6.13
Android/iOS version: 13
Device model: Moto Edge 40 Pro
Crash-Logs: ?
```

mdykierek:
Today I had the same misbehavior on a shorter route from Dworcowa 25 in Gliwice Poland to Wyzwolenia 11 in Ruda Śląska, Poland. I had to follow one way thus required route re-calculation. Then it skipped the A4 exit towards 925 and showed the left turn on the bridge. It means it didn't show the exit but the turn left that would come next. I stopped the navigation and started again. The exit appeared just fine.

sonora:
It sounds like this may actually duplicate #18824.

vshcherb:
I think it's related to omitting unnecessary turn announcements on highways. Thought it's not clear.
1. https://test.osmand.net/map/?start=45.803704,12.818105&finish=45.804804,12.900674&profile=car#18/45.79641/12.82641 (https://www.openstreetmap.org/way/47460763#map=15/45.7878/12.8351) Actually here as I see it shuold be displayed Go ahead (+C|TR). 
2.https://test.osmand.net/map/?start=49.157145,16.600288&finish=49.159137,16.604880&profile=car#16/49.1584/16.6009 - also works fine

vshcherb:
This route:
[...]

PR #19302, opened 2024-03-13
Add attached to native routing

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/19016


OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -2147,12 +2147,12 @@ public RouteSegment next() {
 				public void remove() {
 				}
 			};	
-		} else if (ctx.nativeLib == null) {
+		} else /*if (ctx.nativeLib == null)*/ {
```
vshcherb, line 2150: native lib should attached all segments otherwise it doesn't make any sense

Issue #18919, opened 2024-01-16
Countries, capitals, large objects from Worldwide should be searchable without increase


gtrus21:
### Describe the idea (required)

Add a specific list of objects to search for.  This should include countries, capitals of countries, oceans and similar objects.  When searching, you must first check this list, then add objects from the regular search to it

### Tell us about the expected behaviour (required)

This is necessary to quickly search for the most global and important objects, regardless of the distance to them

### Tell us about alternatives you've considered (required)

-

vshcherb:
Please be more specific with examples

yuriiurshuliak:
I would like to highlight that the efficiency of the search feature can be further enhanced by ensuring that you have the relevant offline vector map files. Initially, the search is based on data located on the map in the visible area of the device screen. If you don't find what they are looking for initially, OsmAnd suggests increasing the search radius, which can be helpful in locating more distant objects.

To ensure the most accurate results, we recommend using the Online Search feature in the app. Online Search allows you to find locations, addresses, and points of interest in real-time.

gtrus21:
> Please be more specific with examples

For example, I couldn’t find Guatemala on the map, and I couldn’t understand what was wrong.  It seems that it is not displayed because the distance to it is more than 10,000 km

Sonwon1:
Being able to define the search radius would be a very good feature.

yuriiurshuliak:
The bug was not reproduced. I tried to reproduce this bug by conducting searches from Eastern Europe and Australia. To address this, the search radius needs to be expanded once for the 10,185-kilometer distance from Eastern Europe and twice for the 15,311-kilometer distance from Australia. It's important to note that online search wasn't employed in this process.

OsmAnd~ 4.7.0#1376m, released: 2024-01-17

<table style="width:100%">
  </tr>
  <tr>
    <th>Eastern Europe</th>
    <th>Australia</th>
  </tr>
  <tr>
    <td> <video source src="https://github.com/osmandapp/OsmAnd/assets/127092082/3fd27b4e-9e66-44ba-b252-5552ef98ab6a"  width="300" height="auto"/>   </td>
    <td><video source src="https://github.com/osmandapp/OsmAnd/assets/127092082/d7aadaaf-5fd7-4974-be1b-06dc225c4129"  width="300" height="auto"/>   </td>
  </tr>
</table>

gtrus21:
[...]

PR #19305, opened 2024-03-13
Cover Earth for basemap search (radius = 20k km)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18919

PR #19328, opened 2024-03-16
Backport C++ car 2-phase fix legacy/pull/185

RZR-UA:
Original C++ request: https://github.com/osmandapp/OsmAnd-core-legacy/pull/185

C++ code needs that fixes to fix broken match of 2-phase start/end and to fix incorrect 2-phase detection at all.

Java methods have correct 2-phase startPoint/endPoint (for today, but who knows what might happen later).

Reasons to approve:

1. This request will sync Java and C++ code

3. Addionally, BRP h() method will be faster because old implementation was not optimal for 2-phase.

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/BinaryRoutePlanner.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/BinaryRoutePlanner.java
@@ -349,6 +349,12 @@ public class BinaryRoutePlanner {
 
 	private void initQueuesWithStartEnd(final RoutingContext ctx, RouteSegmentPoint start, RouteSegmentPoint end,
 			PriorityQueue<RouteSegmentCost> graphDirectSegments, PriorityQueue<RouteSegmentCost> graphReverseSegments) {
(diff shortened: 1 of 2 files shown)


ivanPyrohivskyi:
In RoutePlannerFronted need to check searchRouteInternalPrepare and apply ctx.precalculatedRouteDirection = routeDirection.adopt(ctx); before runNativeRouting
This code fully repeat method adopt.

RZR-UA:
The problem IS actual only for:

```
if (intermediatesEmpty && ctx.nativeLib != null) {
...
  ctx.precalculatedRouteDirection = routeDirection.adopt(ctx);
}
```

The reason of problem is that we do not have **precise start** here -> adopt() sets non-precise startPoint -> timeEstimate() can not match precalculated route

So basically we do not need to have `updatePreciseStartEnd` in Java but:

- [ ] think how to fix this issue in C++ better than using updatePreciseStartEnd method
- [ ] accept fix for h() in BRP-java because it can improve Java heuristics performance

PR #19331, opened 2024-03-16
Correct distance & time after making points precise

vshcherb:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2470

PR #19357, opened 2024-03-19
Fix empty lanes () after native turns

RZR-UA:


vshcherb:
Strange commits: 1 line change and 3 commits? https://github.com/osmandapp/OsmAnd/pull/19357/commits/5edaafb4798fcf186e84b959f9545131de808a63#diff-8f942a710e9ffc319c084fcb297626284d468f5f116a7b2ce8a6606424ee337b - also this is not merged yet.
Something fishy with PR

RZR-UA:
That was a commit to different branch, and that brance was merged before commit applied. Will re-created it with 1-line commit only, sorry.

PR #19372, opened 2024-03-22
Get back part of refactor (failed basemap search)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2470

PR #19384, opened 2024-03-24
Change / Default appearance for selected tracks

nazar-kutz:


nazar-kutz:
At this point "Change appearance" implemented

Issue #19422, opened 2024-03-30
Bug parsing finish/end URL parameter IntentHelper


OscarDorronzoro:
### Description

Android app doesn't show intended location and route when opening a link with OsmAnd containg the finish parameter.

### Steps to reproduce

Web app original link:
https://osmand.net/map/?start=-34.603722,-58.381594&finish=-34.607709,-58.370587&profile=car#16/-34.6063/-58.3798
<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/48965382/4fc329e7-b5e9-4996-b223-24165d1de9fe"  width="500" height="auto"/> </th>
  </tr>
</table>

The IntentHelper class looks for an "end" parameter when parsing URL
https://github.com/osmandapp/OsmAnd/blob/4b8107f44fda6d425e83675b2566817783309a43/OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java#L93

Workaround:
I tried to rename the "finish" URL parameter to "end", like this:
https://osmand.net/map/?start=-34.603722,-58.381594&end=-34.607709,-58.370587&profile=car#16/-34.6063/-58.3798
and it works, now android app opens the route correctly, plus web app is still working, it actually rewrites the URL with the "finish" parameter name instead of "end".

### Actual result

Original link led the app to open normally without any information from the link.

### Expected result

It should open on indicated location and show navigation menu, like the following (workaround applied):
<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/48965382/d62d8349-7c6b-4f2c-a77c-d5726238b70d"  width="300" height="auto"/> </th>
  </tr>
</table>

Could you guys add the "finish" parameter to IntentHelper class, or rather modify web app to accept "end" parameter?
Thanks.

### Your Environment (required)

```
OsmAnd Version: 4.6.13
Android/iOS version: Android 11
```

vshcherb:
Obviously we need to be backward compatible, so we will update web version I think

RZR-UA:
There are few other discrepancies:

1. intermediate points Android: `ipoints`, which web never used (actually: "inter" and "via")
2. mode-string (profile,params) Android `mode`, which web never used (actually: "profile", "params")

@vshcherb please decide how to do a better fix:

- [ ] fix Android for all actual parameters (lose backward compatibility)
- [ ] fix Web/random-tester for "end" parameter only
[...]

PR #19450, opened 2024-04-02
Sync routing URL params with OsmAnd Web

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19422

PR #19465, opened 2024-04-03
Implement useGeometryBasedApproximation APPROXIMATION_TYPE option (Android)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2411

PR #19580, opened 2024-04-16
Reduce MAX_COUNT_REITERATION from 500 to 50

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2471

PR #19605, opened 2024-04-18
Change angle on every zoom

0xRe1nk0:


Andrewschenko:
It's a simple linear function - why is "for" there?

Issue #19591, opened 2024-04-17
Wrong estimated time for trip with intermediate step


Guylby:
### Describe the routing engine used (required)

- [X] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Describe your routing Profile (required)

Standard Osmand offline car profile

### Describe your start and end points (required)

Start point: 43.117566, 5.934752
Intermediate point: 44.37926, 4.732222
End point: 45.572210, 5.918950

https://test.osmand.net/map/?start=43.117566,5.934752&end=45.572210,5.918950&via=44.37926,4.732222&profile=car#14/43.9636/5.9025

### Describe the actual route (required)

Without the intermediate stop, routing time is accurate and states 4:30h travel time.

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/151167525/892ee31a-21b2-42e5-84e9-76c30f0136f0"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/osmandapp/OsmAnd/assets/151167525/a1c84b1d-8d85-44e2-be90-034ffc05d745"  width="300" height="auto"/> </th>
  </tr>
</table>

With an intermediate stop on the route (same distance, same everything), travel times jumps to 9h32:
This is new in 4.7.9, didn't happen on 4.7.6

However, the test.osmand/map doesn't show this bug:
https://test.osmand.net/map/?start=43.117566,5.934752&end=45.572210,5.918950&via=44.37926,4.732222&profile=car#14/43.9636/5.9025

### Describe the expected route (required)

Routing time are correct even with intermediate steps

### Describe what Maps you used (online or offline) (required)

- [X] Offline maps offered within the OsmAnd app for download, all updated to current (April). 
- [ ] Online (tile / raster) maps 

### Your Environment (required)

```
OsmAnd Version: 4.7.9
Android/iOS version: 13
Device model: Sony Xperia 10 III
```

RZR-UA:
**The bug is confirmed using default HH-native routing.**

Other kind of routing (HH-Java, A* 2-phase) are ok.

I'm figuring out how to fix it.

![hh-native](https://github.com/osmandapp/OsmAnd/assets/90444451/20eb7534-99b6-4bf2-b345-a2c8ea261039)

Guylby:
I've found some interesting behaviour if that can help pinpoint the bug:
[...]

PR #19612, opened 2024-04-18
Fix avgSpeed in calculateIntermediateIndexes Issue#19591

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19591

PR #19614, opened 2024-04-19
Required Maps – Route preparation

nazar-kutz:


OsmAnd/res/values/strings.xml
```diff
@@ -10,6 +10,12 @@
 	- For wording and consistency, please note https://docs.osmand.net/docs/technical/contributions/translating-osmand
 	Thx - Hardy
 -->
+    <string name="maps_that_also_be_used">Maps that also be used: %1$s</string>
+    <string name="getting_list_of_required_maps">Getting a list of required maps...</string>
```
Chumva, line 14: There is a separate symbol for …

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
```diff
@@ -107,7 +107,7 @@ public RoutingHelper(OsmandApplication context) {
 		app.getAppCustomization().addListener(customizationListener);
 	}
 
-	RouteProvider getProvider() {
+	public RouteProvider getProvider() {
```
Chumva, line 110: Is this necessary?

Issue #19483, opened 2024-04-05
Slow Route Calculation (bicycle profile)


anten1222:
### Describe the routing engine used (required)

- [ ] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Describe your routing Profile (required)

In 4.7 Nightly, the calculation of the route over longer distances was quick some time ago. Now it is very slow again. 10 minutes or more

### Describe your start and end points (required)

In 4.7 Nightly, the calculation of the route over longer distances was quick some time ago. Now it is very slow again. 10 minutes or more

### Describe the actual route (required)

In 4.7 Nightly, the calculation of the route over longer distances was quick some time ago. Now it is very slow again. 10 minutes or more

### Describe the expected route (required)

In 4.7 Nightly, the calculation of the route over longer distances was quick some time ago. Now it is very slow again. 10 minutes or more

### Describe what Maps you used (online or offline) (required)

- [ ] Offline maps offered within the OsmAnd app for download. 
- [ ] Online (tile / raster) maps 

### Your Environment (required)

```
OsmAnd Version:4.7
Android/iOS version:
Device model:
```

yuriiurshuliak:
Please provide a detailed steps to reproduce this issue.  Include a video demonstrating the issue occurrence. Additionally, specify the starting and ending coordinates of the route that is showing slow computation.

anten1222:
With Osmand Nightly

For example

Profile Bike

From 53.82286° N, 9.28587° E

To 50.38578° N, 6.65635° E

Some time ago 45 seconds

Now 10 minutes or more

zipav:
OsmAnd Nightly 4.7.0#2046m, released: 2024-04-12

Not reproduced on OsmAnd Nightly. The route is built in about 60 seconds. 

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/123936521/de0d5a7d-7801-43dc-9513-501a7573b5b3"  width="300" height="auto"/> </th>
  </tr>
</table>

anten1222:
Version 4.7.9 Route by car, calculation under 10 seconds. By bike over 10 minutes or crash

vshcherb:
@anten1222 did you update all the maps? did you use specific bicycle settings? Please try to switch off all the settings for navigation and try. 
If you use any specific avoids please let us know

RZR-UA:
[...]

PR #19622, opened 2024-04-19
[HH-native] Fix empty parameterValues in selectBestRoutingFiles()

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19483

PR #19675, opened 2024-04-26
[4.7] Improve GpxSegmentsApproximation, add JNI support

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2411

PR #19735, opened 2024-05-02
[4.7] Implement fast and simple applyExternalTimestamps(), cleanup old setPoints(), fix GPX parsing of time with fractional seconds, refactor approximation code

RZR-UA:


RZR-UA:

PR #19744, opened 2024-05-02
Increase compatibility to JavaVersion.VERSION_1_8

RZR-UA:


vshcherb:
4.7 not going to be merged

Issue #19456, opened 2024-04-03
Support multiparameters - Avoid motorways + "prefer unpaved roads" 


sinkoneri:
### Describe the routing engine used (required)

- [x] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Describe your routing Profile (required)

I have a motorcycle profile where I enable "prefer unpaved roads"  bcos if I can I would love to ride offroad, but if no offroad route available I would like to go on paved backroads as a fallback option and awoid motorways in any circumstances.
But seems if no offroad route available, the navigation planner fallback logic doesn't take into account the avoid road settings.

### Describe your start and end points (required)

Test1
From: Budapest, Hungary
To: Velence, Hungary

Test2
From: Berhida, Hungary
To: Velence, Hungary

### Describe the actual route (required)

Test1 OsmAnd route (should not contain M7 which is a Motorway)
https://youtu.be/nO2mWsiIGa4?si=4c78oV-eYp3Gwl6M

Test2 OsmAnd route  (should not contain M7 which is a Motorway)
https://youtu.be/jokYmFxpHjo?si=V8yk6GLhCI2vI1sU 

### Describe the expected route (required)

Test1:
At 0:18 avoid motorways enabled, than route calculated fine
At 0:38 see the excpected route in the linked video 0:38 via route 7 which is paralel with M7 motorway. Expected route.
At 1:07 I enble "prefer unpaved roads" then the route gets recalculated but the resulting route is wrongly goes via M7 motorway bcos avoid motorways is still enabled.

Test2:
Avoid motorways enabled and prefer unpaved roads also enabled.
Navigation wrongly goes via M7 motorway
0:25 prefer unpaved roads disabled route gets recalculated and the route is fine via E66.

### Describe what Maps you used (online or offline) (required)

- [x] Offline maps offered within the OsmAnd app for download. 
- [ ] Online (tile / raster) maps 

### Your Environment (required)

```
OsmAnd Version: 4.6.13
Android: 14
Device model: Samsung S21 Phone

OsmAnd Version: 4.6.13
Android: 13                          
Samsung S6 lite tablet
```

yuriiurshuliak:
[...]

PR #19756, opened 2024-05-03
Add multi parameters

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/19456

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/RoutingConfiguration.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/RoutingConfiguration.java
@@ -469,7 +469,17 @@ public class RoutingConfiguration {
 	private static void addSubclause(RoutingRule rr, RouteAttributeContext ctx) {
 		boolean not = "ifnot".equals(rr.tagName);
 		if(!Algorithms.isEmpty(rr.param)) {
-			ctx.getLastRule().registerAndParamCondition(rr.param, not);
+			if (rr.param.contains(",")) {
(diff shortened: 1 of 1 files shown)


RZR-UA (approved):

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19456

OsmAnd-java/src/main/java/net/osmand/router/RoutingConfiguration.java
```diff
@@ -469,7 +469,17 @@ private static boolean checkTag(String pname) {
...
+				String[] params = rr.param.split(",");
+				for (String p : params) {
+					p = p.trim();
+					if (!Algorithms.isEmpty(p)) {
+						ctx.getLastRule().registerAndParamCondition(p, not);
+					}
+				}
+			} else {
+				ctx.getLastRule().registerAndParamCondition(rr.param, not);
+			}
```
RZR-UA, line 482: It could be simplified with just-split-iterator, without contains() and if() block.

Additionally, it's need to verify that `if=motorcycle, avoid_motorway, prefer_unpaved` would work as:

if (motorcycle **AND** avoid_motorway **AND** prefer_unpaved)

RZR-UA: if (motorcycle AND avoid_motorway AND prefer_unpaved) -- verifed

It's OK to accept this request.

PR #19759, opened 2024-05-03
Increase compatibility to JavaVersion.VERSION_1_8

RZR-UA:
It's incorrect to build Android with disabled sourceCompatibility/targetCompatibility.

Sooner or later, somebody may include incompatible code into OsmAnd-java classes.

For example, Java-tools is allowed to use Java 11 and its higher limit is inherited to OsmAnd-java in Tools project.

Therefore, the Compatibility should be limited to Android SDK max allowed = Java 8.

### DIFF
--- a/OsmAnd-java/build.gradle
+++ b/OsmAnd-java/build.gradle
@@ -6,11 +6,11 @@ configurations {
 	android
 }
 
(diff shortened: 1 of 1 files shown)


RZR-UA:
declined

PR #19763, opened 2024-05-03
3x faster TrackFolderLoaderTask by removing recursion

sonora:


sonora:
@vshcherb 2 things are I still need to investigate here:

1. Accessing the "Visible tracks" screen (also via Configure map) is not equally accelerated by this, perhaps it uses yet another async task, or there's too much obsolete/redundant sorting involved there
2. On the MyPlaces>Tracks screen the bottom statistics seem incomplete for folders with subfolders. Not sure how that comes from this commit, perhaps some unrelated bug.

sonora:
With https://github.com/osmandapp/OsmAnd/commit/1cca915034b162872b4f45e0060df600111b5b17 and merging https://github.com/osmandapp/OsmAnd/pull/19772 both are fixed.

PR #19772, opened 2024-05-04
Bulk load GPX cache

sonora:


sonora:
```
05-04 23:41:04.025 17853 18146 I net.osmand: GpxDbHelper Time to loadGpxItems 1956 ms items count 3445 {Initializing app}
05-04 23:41:04.046 17853 18146 I net.osmand: GpxDbHelper Time to loadGpxDirItems 21 ms items count 3445 {Initializing app}
05-04 23:41:04.046 17853 18146 I System.out: Initialized GPX_DB_INITIALIZED in 1977 ms
```
:wink:

vshcherb:
let @Chumva review the code have no idea what gives such improvement

vshcherb:
What's the advantage of that code? Before vs after

sonora:
@vshcherb , @Chumva

- The bulk of the improvement is to **one-time create a map of available files by using a stream I/O** instead of before using the countless single `file.exists()` operations.

- As for a minor improvement, I think writing to a `HashMap` and then bulk-updating the `ConcurrentHashMap` is slightly faster than directly accessing the `ConcurrentHashMap` in all single cases.

Been testing this for some hours now, have very good results.

sonora:
I've timed it and the second effect is negligible, so we go with less code, as I have adjusted it now.

Issue #19683, opened 2024-04-27
Load GPX Tracks Optimally (Refactoring)


vshcherb:
- [x]  Load new GPX tracks from FS in background only when user enters My places (data will be refreshed only when reenters or enters folders)
- [x] On entering My Places / Configure Map data should be displayed immediately from Memory (cause track items are loaded during start) - NO DB, NO FS.
- [x]  Pull to refresh could or could not wait full read from DB or from FS - depends on implementation - please specify how it works

_Originally posted by @vshcherb in https://github.com/osmandapp/OsmAnd/issues/19373#issuecomment-2053602225_

sonora:
Here's a UI observation: When you have enough Favorites  and Track  folders, there is a narrow scrollbar visible on the right screen edge .

While on MyPlaces>Favorites, this scrollbar is briefly visible after each scrolling sof the list,  then disappears when the list is left untouched.

If on the MyPlaces>Tracks tab, that scrollbar pulses visible<->invisible about once per second, indefinitely even if the screen is left untouched,  as if there is some never ending refresh process perpetually busy on that screen...

vshcherb:
- Missing technical plan / estimate

alex-osm:
https://github.com/osmandapp/OsmAnd/issues/20844

PR #19792, opened 2024-05-06
Try to fix root folder stats update

Chumva:


sonora:
@vshcherb Not sure this really needs fixing in the first place, because all it achieves is immediately start showing folders, something probably much better achieved anyway via https://github.com/osmandapp/OsmAnd/issues/19683, i.e. de-coupling database and file I/O from showing the Trackss screens.

Issue #19768, opened 2024-05-04
Rendering GPX Tracks is slow due to constant checks


vshcherb:
## Description
Rendering GPX Tracks is slow due to constant checks

## To do: 
- [x] Implement proper fix that check for gpx attributes happens only once something has changed in appearance, probably Global version of Appearance should be maintained or Version per track (int variable )
- [x] Auto test

alex-osm:
Hot fix - https://github.com/osmandapp/OsmAnd/commit/9556fd1bb630d3d98308f0ac33ed6919fb5486f2

DmitryAlexei:
OsmAnd~ 4.8.0#2539m, released: 2024-06-20
on my setup works well

PR #19793, opened 2024-05-06
#19768: Implemented test to check updating track analysis

Corwin-Kh:


OsmAnd/src/net/osmand/plus/track/helpers/GPXDatabase.java
```diff
@@ -40,7 +40,7 @@ public class GPXDatabase {
 
 	public static final Log LOG = PlatformUtil.getLog(GPXDatabase.class);
 
-	protected static final int DB_VERSION = 23;
+	protected static final int DB_VERSION = 24;
```
Chumva, line 43: Do we really need to update it for test?

OsmAnd/src/net/osmand/plus/track/helpers/GpxDbHelper.java
```diff
@@ -37,6 +37,11 @@ public class GpxDbHelper implements GpxDbReaderCallback {
 	private final Map<File, GpxDataItemCallback> readingItemsCallbacks = new ConcurrentHashMap<>();
 
 	private GpxReaderTask readerTask;
+	private DBTestCallback dbTestCallback;
```
vshcherb, line 40: static counter will be enough for test, how many reads are done from DB

vshcherb:
Definitely not 4.7 branch

OsmAnd/src/net/osmand/plus/track/helpers/GpxDbHelper.java
Corwin-Kh: using callback we ensure we perform this action only in test, checking only specific track and skip analysis update on initial track loading.

PR #19808, opened 2024-05-07
changed default wall coloring type

Corwin-Kh:


vshcherb:
Too many changes for 4.7. 1 line / 1 constant to change default would be accepted

PR #19814, opened 2024-05-08
Fix_19333

sonora:


OsmAnd/src/net/osmand/plus/myplaces/tracks/dialogs/BaseTrackFolderFragment.java
```diff
@@ -344,6 +348,7 @@ protected String getSortEntryName() {
 	@Override
 	public void setTracksSortMode(@NonNull TracksSortMode sortMode, boolean sortSubFolders) {
 		if (sortSubFolders) {
+			removeSurplusTabsSortModes();
```
Chumva, line 351: Do we need to call this method every time we change sort mode & perform settings.saveTabsSortModes() two times?

sonora: It's true that I have not found the ideal place(s) yet when to call this. It's comparatively expensive, hence perhaps should not be called after every folder move, delete or rename

 In a way I 'parked' it here under the 'sort subfolders' action where it can be triggered manually via the corresponding menu option.  Will come back to this as I find some time, but it's not a priority 1 issue, I guess.

PR #19822, opened 2024-05-10
Fix_19804 (at least interim)

sonora:


Chumva:
Do we still need this after [fix](https://github.com/osmandapp/OsmAnd/commit/fa12fce532b8033ee1b03bfdf872e1a6aa48a147) ?

sonora:
@Chumva No, it's outdated now, let me close it.

However: commenting out the progress update would still speed things up, as it it one way (currently the only one I find easy to implement) to suppress that the folder statistics being calculated twice (once interim, then final in the end), so this does actually speed up things by like 30% or so...

PR #19828, opened 2024-05-10
Implement alignGpxTimeMilliseconds() for gpx-time

RZR-UA:
API level 24 does not support Java 8 DateTimeFormatter()
Old SimpleDateFormat() requires fixed-length milliseconds.
Shorter or longer ms-length must be aligned to a fixed length.
Without align, parseTime() fails and results 1-second precision.


RZR-UA:

RZR-UA:
https://github.com/osmandapp/OsmAnd/pull/19790

RZR-UA:
moved to 4.7

PR #19887, opened 2024-05-16
Avoid shortcutting of loops (GpxRouteApproximation)

RZR-UA:


RZR-UA:

Issue #16397, opened 2023-02-05
Brands – Search results prioritize streets over POI (not sorted by distance overall) - Aldi


Norbert80:
Search results are not properly sorted by distance. If I scroll down the list some near results are displayed below results that are several hundred kilometres away.

<table style="width:100%">
  <tr>
<th><img src="https://user-images.githubusercontent.com/6921696/216839329-b5661578-4850-429b-8646-18c43ead1e77.png"  width="300" height="auto"/> </td>
  </tr>
</table>

OsmAnd~ 4.3.8

I am not sure if the following report is duplicate:
https://github.com/osmandapp/OsmAnd/issues/8677

vshcherb:
> we need completely restructure the search cause search doesn’t have mechanism to understand what’s the meaning of the words.  So I think you can adjust how you use it and type Supermarket Aldi and save as a filter - so then it’s prioritized
 
You could see that results are sorted 1) Cities / Streets 2) POI - in each group results sorted by distance

scaidermern:
Same issue for me (OsmAnd+ 4.3.7) with similar POIs like Aldi, Netto (supermarkets) and Aral (fuel station):
![foo](https://user-images.githubusercontent.com/1040784/216905389-9832e6a4-c465-4e09-b1f1-dfe224d33c16.jpg)

Looks like it could have something to do with partival vs exact matches?

vshcherb:
Exactly cause it's a deliberate decision to prioritize exact matches over partial matches, unless there is a language or vocubalary understanding behind that issue won't be fixed. 

**Please use**: Supermarket (category) Aldi to achieve necessary result

scaidermern:
> **Please use**: Supermarket (category) Aldi to achieve necessary result

Unfortunately this leads to the exact same behavior or is even worse.

Searching for "Convenience store and supermarket Aldi" (_without_ space at the end) returns exactly the same results.
Searching for "Convenience store and supermarket Aldi " (_with_ space at the end) only returns "ALDI Nord" but not "Aldi".

vshcherb:
Similar issue with Lidl

![telegram-cloud-photo-size-4-5983060546691381751-y](https://user-images.githubusercontent.com/1042025/216977746-e013c266-dec1-4719-a1de-999b0972f1d6.jpg)


If it enables search by category you should see button increase radius 
1. Works correctly
[...]

PR #19898, opened 2024-05-17
Support osmand_brand_filter in search

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/16397

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapIndexReader.java
+++ b/OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapIndexReader.java
@@ -1617,9 +1617,14 @@ public class BinaryMapIndexReader {
 		request.resultMatcher = resultMatcher;
 		return request;
 	}
-	
+
+	public static SearchRequest<Amenity> buildSearchPoiRequest(int sleft, int sright, int stop, int sbottom, int zoom,
+	                                                           SearchPoiTypeFilter poiTypeFilter, ResultMatcher<Amenity> matcher) {
+		return 	buildSearchPoiRequest(sleft, sright, stop, sbottom, zoom, poiTypeFilter, null, matcher);
+	}
+
 	public static SearchRequest<Amenity> buildSearchPoiRequest(int sleft, int sright, int stop, int sbottom, int zoom, 
-			SearchPoiTypeFilter poiTypeFilter, ResultMatcher<Amenity> matcher){
+			SearchPoiTypeFilter poiTypeFilter, SearchPoiSubcategoryFilter subcategoryFilter, ResultMatcher<Amenity> matcher){
 		SearchRequest<Amenity> request = new SearchRequest<Amenity>();
 		request.left = sleft;
 		request.right = sright;
@@ -1627,6 +1632,7 @@ public class BinaryMapIndexReader {
 		request.bottom = sbottom;
 		request.zoom = zoom;
 		request.poiTypeFilter = poiTypeFilter;
+		request.subcategoryFilter = subcategoryFilter;
 		request.resultMatcher = matcher;
 
 		return request;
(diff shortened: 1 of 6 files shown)


OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapIndexReader.java
```diff
@@ -1784,6 +1794,7 @@ public static class SearchRequest<T> {
 		SearchFilter searchFilter = null;
 
 		SearchPoiTypeFilter poiTypeFilter = null;
+		SearchPoiSubcategoryFilter subcategoryFilter;
```
vshcherb, line 1797: it's not subcategory. Category is key for Amenity, subcategory would be "cinema", ... however it's called poitype.

So it's definitely bad name for poi Additional tags

OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -735,6 +737,44 @@ protected static class PoiTypeResult {
 		public Set<String> foundWords = new LinkedHashSet<String>();
 	}
 
+	public static  class SearchAmenityBrandAPI extends SearchBaseAPI {
+		@Override
```
vshcherb, line 741: 1. This is perfectly part of SearchAmenityTypesAPI  - addPoiTypeResult. 
2. I'm not sure search code should depend on brands, it should be generic poisubtypes (https://github.com/osmandapp/OsmAnd/issues/16397) in task it doesn't specify only brand is problem. Could be operator or any searchable field.

OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -735,6 +739,66 @@ protected static class PoiTypeResult {
 		public Set<String> foundWords = new LinkedHashSet<String>();
 	}
 
+	public static  class SearchPoiTopIndexAdditionalAPI extends SearchBaseAPI {
+		@Override
+		public boolean search(SearchPhrase phrase, SearchResultMatcher resultMatcher) throws IOException {
+			List<BinaryMapIndexReader> offlineIndexes = phrase.getOfflineIndexes();
```
vshcherb, line 745: bbox?

vshcherb: We need cache and no search for brand because we assume < 1000 per map

OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -735,6 +739,66 @@ protected static class PoiTypeResult {
...
+							@Override
+							public String getName() {
+								return topIndexName;
+							}
+
+							@Override
+							public String getIconResource() {
+								//Example: brand_mcdonalds, operator_bank_of_america
+								String val = finalTopIndexValue.toLowerCase().replaceAll("'", "");
+								val = val.replaceAll("\\W+", "_");
```
vshcherb, line 785: ? Forbidden - ':' , other symbols not a problem.

OsmAnd-java/src/main/java/net/osmand/data/Amenity.java
```diff
@@ -65,6 +65,7 @@ public class Amenity extends MapObject {
 	// duplicate for fast access
 	private String openingHours;
 	private Map<String, String> additionalInfo;
+	private List<String> topIndexAdditional;
```
vshcherb, line 68: Discuss

OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
```diff
@@ -778,7 +786,16 @@ private Amenity readPoiPoint(int left31, int right31, int top31, int bottom31,
 			case OsmandOdb.OsmAndPoiBoxDataAtom.SUBCATEGORIES_FIELD_NUMBER:
 				int subtypev = codedIS.readUInt32();
 				retValue.setLength(0);
+				hasSubcategoriesField = true;
 				PoiSubType st = region.getSubtypeFromId(subtypev, retValue);
+				if (req.poiTopIndexAdditionalFilter != null) {
```
vshcherb, line 791: poiTopIndexAdditionalFilter -> poiAdditionalFilter (make it work for non top index ? )

Issue #19841, opened 2024-05-11
OsmAnd requesting unneeded Oauth2 permission "Redact map data"?


mnalis:
### Description

When logging in via OsmAnd recently, it requests OAuth2 permissions. One of those permissions is moderators-only `Redact map data`  which should not be needed (none of the other OSM editing apps ever needed that AFAIK):

![osmand-mod-perms-crop](https://github.com/osmandapp/OsmAnd/assets/156656/68263535-1bca-4524-b8e1-83acafdd3a67)

(This might be result of moving from obsoleted OAuth1 to Oauth2)

### Steps to reproduce

- enable _"OpenStreetMap editing"_ plugin
- on its settings log in the first time using `Sign in with OpenStreetMap`
- enter username and password on openstreetmap.org webpage that opens
- see the list of permissions required 

### Actual result

While most permissions are normal and used by other OSM editing apps too, there is one named `Redact map data` with blue star. Hovering over that blue star in https://www.openstreetmap.org/oauth2/authorized_applications says `This permission is for actions available only for moderators` 

### Expected result

I would not expect `Redact map data` to be requested. 
Other permission that is also requested `Modify the map` is enough for actually editing the map (adding/deleting/changing POIs etc).

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: OsmAnd~ 4.7.10 (latest F-droid)
Android/iOS version: Android 14
Device model: Samsung Galaxy S23+
Crash-Logs: no
```

DmitryAlexei:
OsmAnd~ 4.8.0#2286m, released: 2024-05-13
- unlike other apps OsmAnd asks for permission to *Read map data* (see screenshot)

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/116079541/e70fbc87-7d8f-4a91-9b6d-431c46c9e32d"  width="300" height="auto"/> </td>
  </tr>
</table>

vshcherb:
Updated OAuth2 application

RZR-UA:
https://wiki.openstreetmap.org/wiki/OAuth **write_redactions** is not listed in the `Supported scopes:` section

https://wiki.openstreetmap.org/wiki/API_v0.6 requires either write_redactions or write_api OAuth scope; _**write_redactions** is being phased out_

RZR-UA:
[...]

PR #19966, opened 2024-05-28
Remove write_redactions from oauth2_scope

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19841

mnalis:
@vshcherb : It would be good to make a release with this change **very soon**, because:

- it seems OSM admins have very recently disabled that scope, so attempting to login with newer OsmAnd which supports Oauth2 but request this `write_redactions` (like e.g. mine `4.7.17` from F-droid) no longer works (it worked few weeks back), and now fails with `An error has occurred. The requested scope is invalid, unknown, or malformed.`
- there are [brownouts active for old Oauth1.0a](https://community.openstreetmap.org/t/oauth-1-0a-and-http-basic-auth-shutdown/108490/17) (and [full disabling of Oauth1.0 on June 1st 2024](https://wiki.openstreetmap.org/wiki/2024_authentication_update)) which will force all users (who have not done it already) to logout and login again.

Combined result is that in **about 3 days time basically no user would be able to upload changes with OsmAnd**, as they won't be able to login again after logout.

PR #20028, opened 2024-06-07
[RT-4.7] Refactor Online Routing to avoid double-requests

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2560

OsmAnd/src/net/osmand/plus/routing/RouteCalculationParams.java
```diff
@@ -28,6 +29,7 @@ public class RouteCalculationParams {
 	public boolean inPublicTransportMode;
 	public boolean extraIntermediates;
 	public boolean initialCalculation;
+	public GPXFile gpxFile;
```
alex-osm, line 32: Could `gpxRoute` be used for getting gpxFile?

RZR-UA: It seems possible to convert GPXRouteParams.routePoints to GPXFile.points but there are a lot of other members and usages in GPXFile. Another less possible option is to remove GPXFile from online-routing-flow and try to replace it by GPXRouteParams everywhere, but it might be time-consuming task.

PR #20055, opened 2024-06-10
Try to fix #19804

### LINKED ISSUE #19804: Regression: Track folder statistics preliminary, never updated
This is a regression, reverting this commit (i.e. preventing interim progress publishing) seems to solve the issue: https://github.com/osmandapp/OsmAnd/commit/f0e141b9df0965a38b3ea9f64a26a37278f2f096 (at the cost of the Adapter only filled once complete).

To reproduce: Simply have many tracks, open the MyPlaces>Tracks screen and check the stats at the bottom of the screen. I e.g. have 3000 tracks, the stats may only show 100 or so (likelya n interim count). This gets never updated. Except when I do a pull refresh, then I may get yet another count.

(The issue is not only for the root folder, but fo all reasonably large folders.)

Chumva:


sonora (commented):
@Chumva Quick test result: Looks like not a good commit. The time it takes to pull up the Tracks screen for me is back up from 2 sec to a bout 12-15 sec. in a new build, very likely due to this. 

Potential issues:
- `items.add(new TrackFolderAnalysis(rootFolder));` instead of `items.add(rootFolder.getFolderAnalysis());` and 
- `items.add(new TrackFolderAnalysis(selectedFolder));` instead of `items.add(selectedFolder.getFolderAnalysis());`

add new instantiations every time called, rather than re-using the old ones?  (Becasue `items.add(ARG)` will simply replace if the `ARG` is already present, but `items.add(new ARG)` will blow up the list by creating more entries...)

sonora:
Just have tested it out: With this commit https://github.com/osmandapp/OsmAnd/pull/20062 things are fast again.

(Main performance killer was the change in `AvailableTracksFragment.java`)

Issue #20019, opened 2024-06-06
HH-Routing not automatically enabled after update


scaidermern:
### Description

Multiple users (including myself) report that after updating to OsmAnd 4.7.x, HH routing wasn't configured for them. They had to enable the development plugin to change the routing type to HH (https://osmand.net/docs/user/navigation/guidance/navigation-settings/#development-settings).

My case was upgrading via the Playstore from 4.6.x to the 4.7.x beta.
Two other user upgraded via F-Droid from 4.6.x to 4.7.x.
The upgrade path from other affected users is unknown.

### Steps to reproduce

Upgrade from 4.6.x to 4.7.x

### Actual result

HH not always automatically enabled.

### Expected result

HH always automatically enabled.

### Your Environment (required)

.

yuriiurshuliak:
The bug has been reproduced. To reproduce this issue, join the Beta test, perform the update, then open and check the routing algorithm. Navigate through Menu - Settings - Driving - Navigation settings - Route Parameters - Development - Routing algorithm. The HH routing option will not be selected by default.

OsmAnd~ 4.8.0#2454m, released: 2024-06-07

https://github.com/osmandapp/OsmAnd/assets/127092082/f2cae285-493e-4364-a497-adf26785ddd1

vshcherb:
we can reset setting (rename key) to enable everywhere - renaming should also happen in iOS

RZR-UA:
Android request created. Additionally, it seems that OsmAnd iOS is not affected by this issue (option use_old_routing was created with correct default "NO" initially).

DmitryAlexei:
OsmAnd~ 4.8.0#2598m, released: 2024-06-26
HH x C++ was enabled after fresh installation

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/116079541/210cdb69-561f-4424-b043-319680916424"  width="300" height="auto"/> </td>
  </tr>
</table>

dmpr0:
# Review

<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/ef64f39c-7f9e-4ebc-b1b4-297e578cdb9a"  width="300" height="auto"/> </td>
[...]

PR #20056, opened 2024-06-10
Reset ROUTING_TYPE / APPROXIMATION_TYPE settings keys

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20019

PR #20060, opened 2024-06-10
Update strings.xml

flexmak:
I fixed a grammar mistake in the German localization file. I'm a native speaker.

"Öffnen" means to open. "Geöffnet" means that sth is open.


sonora:
Thanks! You can directly correct  stuff like this in weblate, no need to even go via PR.

PR #20099, opened 2024-06-13
Fix two rare HH bugs (lost heuristicCoefficient and broken start/finish reiteration in BRP after HH-failed)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19456

Issue #19982, opened 2024-05-30
Not calculated a route from Qal’ai Khumb to Rushon 


EugeneZmeuk:
### Description

No calculated route from Qal’ai Khumb to Rushon (177 km) - Tajikistan (Car Profile)

### Steps to reproduce

* Download **Tajikistan** map
* Switch to **Car** profile
* Set **Start point** – Qal’ai Khumb `38.45728, 70.78783`
* * End point – Rushon `37.94753, 71.55744`
* * Route https://test.osmand.net/map/navigate/?start=38.457285,70.787830&end=37.947527,71.557437&profile=car#9/38.3783/71.4221

### Actual result

<table style="width:100%">
  <tr>
    <th>Missed maps</th>
    <th>Empty Suggested maps dialog</th>
  </tr>
  <tr>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/e6414b84-cb80-4721-a47e-99a56879385e"  width="150" height="auto"/> </td>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/12a9fd9a-c943-424a-87c3-1eaaa72cbdb0"  width="150" height="auto"/> </td>
  </tr>
</table>   

* No route. 
* But when I downloaded the Afghanistan map, this route was calculated at once.

I can calculate the route from Qal’ai Khumb to https://osmand.net/map?pin=38.30532,71.33596#17/38.30532/71.33596
I think the road points from this place were on the Afganistan map during map generation. 

### Expected result

* The route from these two points is calculated. 

https://test.osmand.net/map/navigate/?start=38.457285,70.787830&end=37.947527,71.557437&profile=car#10/38.2269/71.0472

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 4.8 (Nightly)
Android/iOS version: 13
Device model:
Crash-Logs: ?
```

Max1234-Ita:
The problem was reported by an user who's currently in Tagikistan; I reproduced it on my Android Device: attaching some screenshot.


Adding some screenshot:

<table>
<tr>
 <td>
![image](https://github.com/osmandapp/OsmAnd/assets/15315381/03e454db-f989-45a4-b77b-9ec80050bee4)

 <td>
[...]

PR #20130, opened 2024-06-19
Missing Maps improvements and bugfixes

RZR-UA:


RZR-UA:

Chumva (approved):

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculationResult.java
```diff
@@ -14,6 +15,7 @@ public class MissingMapsCalculationResult {
 
 	private final RoutingContext missingMapsRoutingContext;
 	private final List<LatLon> missingMapsPoints;
+	private boolean onlineResultPoints = false;
```
ivanPyrohivskyi, line 18: naming ?

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculationResult.java
```diff
@@ -24,14 +26,23 @@ public class MissingMapsCalculationResult {
...
+	                                    boolean calledByCalcOnline) {
 		this.missingMapsRoutingContext = missingMapsRoutingContext;
 		this.missingMapsPoints = missingMapsPoints;
+		this.onlineResultPoints = calledByCalcOnline;
 	}
 	
-	public void addMissingMaps(String region) {
-		missingMaps.add(region);
-		mapsToDownload.add(region);
+	public void addMissingMapsByRegionsWithDistinctCountries(List<String> regions) {
```
vshcherb, line 36: Please separate request, it's not part of original issue

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculator.java
```diff
@@ -58,7 +59,8 @@ public MissingMapsCalculator(OsmandRegions osmandRegions) {
 	}
 
 
-	public boolean checkIfThereAreMissingMaps(RoutingContext ctx, LatLon start, List<LatLon> targets, boolean checkHHEditions)
+	public boolean checkIfThereAreMissingMaps(RoutingContext ctx, LatLon start, List<LatLon> targets,
```
vshcherb, line 62: flag calledByCalcOnline is not requested

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculator.java
```diff
@@ -188,10 +189,15 @@ private void addPoint(RoutingContext ctx, Map<String, RegisteredMap> knownMaps,
 		boolean onlyJointMap = true;
 		List<String> regions = new ArrayList<String>();
 		for (BinaryMapDataObject o : resList) {
-			regions.add(or.getDownloadName(o));
-			if (!or.isDownloadOfType(o, OsmandRegions.MAP_JOIN_TYPE)
-					&& !or.isDownloadOfType(o, OsmandRegions.ROADS_JOIN_TYPE)) {
-				onlyJointMap = false;
+			boolean hasMapType = or.isDownloadOfType(o, OsmandRegions.MAP_TYPE);
```
vshcherb, line 192: not clear what it fixes

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculationResult.java
RZR-UA: It's definitely a part of original issue because Pamir Highway (issue's route belongs to this highway) is located along the border of countries. All the points along the border belong to both regions (in this case, Tajikistan and Afghanistan) and our "too simple" algorithm offers only one country with longer name (Afghanistan) evern when the points are located at Tajikistan's side.

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculator.java
RZR-UA: The problem is the method checkIfThereAreMissingMaps() is re-used and original targets[] is overwritten by online result when the "Calculate online" button is pressed. calledByCalcOnline (and its followers passed between classes) are required to detect the fact that the Missing Maps fragment is called after calc online resulted - and finally to hide Calc Online button against double-request.

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculator.java
RZR-UA: This is also directly connected to original issue. The problem. We have SRTM-only (sub)regions such as Afghanistan, Egypt, Libya, Chile, etc. That SRTM-only (no map, no routing!!!) regions are added here to regions List and a few lines later it is sorted to get 1st only String from regions.

In case of Afghanistan + Tajikistan, the sorted result is:

afghanistan_baghlan_asia - the longest name, sorted to top = get(0)
afghanistan_asia (ignored)
tajikistan_asia (ignored)

As a result, afghanistan_baghlan_asia is the only mapsToDownload region, which is OBVIOUSLY not a map/roads, not allowed to download as a map/roads and finally we have the EMPTY list of Missing Maps and disabled Download button (as displayed in this issue indeed).

That's it.

PR #20136, opened 2024-06-20
Missing Maps: avoid adding SRTM-only regions

RZR-UA:
The problem. We have SRTM-only (sub)regions such as Afghanistan, Egypt, Libya, Chile, etc.

That SRTM-only regions (no map, no routing, only contours) are added regions List and a few lines later the list is sorted to get 1st only String from regions finally.

In case of Afghanistan + Tajikistan, the sorted result is:

afghanistan_baghlan_asia - the longest name, sorted to the top = get(0)
afghanistan_asia (ignored)
tajikistan_asia (ignored)

As a result, afghanistan_baghlan_asia is the only mapsToDownload region, which is obviously NOT A MAP/roads, not allowed to download as a map/roads, and finally we have the EMPTY list of Missing Maps and the disabled Download button.

The bugfix just excludes NOT map/roads regions from List.


RZR-UA:

PR #20138, opened 2024-06-20
Missing Maps: allow several countries for 1 point

RZR-UA:
There are some interesting locations, such as Pamir Highway, that are located along the border between countries.

All the points along the border belong to both regions (in example case, Tajikistan and Afghanistan) and our current algorithm offers only one country with longer name (Afghanistan) even when the points are located at Tajikistan's side.

For example:

Tajikistan Qal’ai Khumb and Rushon are located at Tajikistan's side.

Both points belong to 2 regions:

afghanistan_asia
tajikistan_asia

Current algorithm consider only 1st region with the longest name - afghanistan_asia.

Finally, on Tajikistan's side Pamir Highway, users will be always offered to download Afghanistan instead of Tajikistan.

The request solves this problem by adding all regions by distinct countries (skipping already added countries).

The fine point might be a way of parsing country: /^(XXXXXXXX)_.*/ but in the worst case the method will result with only 1st region, which is absolutely the same result as original method.

I've tested many border points with new method in Asia (Pamir) and Europe (Austria, Switzerland, Germany) and have not found any problems.

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculationResult.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculationResult.java
@@ -6,6 +6,7 @@ import net.osmand.map.WorldRegion;
 import net.osmand.util.Algorithms;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


vshcherb:
This issue is minor for now and it shouldn't be any real case when people download start/end region manually and only need intermediate regions

RZR-UA:

PR #20139, opened 2024-06-20
Missing Maps: Decrease DISTANCE_SPLIT to 15km

RZR-UA:
This request fixes some cases in Europe such as Austria (Salzburg) -> Germany -> Austria (Innsbruck), Switzerland -> France -> Switzerland where the fastest route should run by a highway not via mountain roads.

I've tested the performance on 2000+ km distance and Missing Maps calc speed was not degraded.


RZR-UA:

PR #20140, opened 2024-06-20
Missing Maps: improve visibility of "Calc online" and "Use existing maps sections", avoid double calc-online call

RZR-UA:


OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculationResult.java
```diff
@@ -14,6 +14,7 @@ public class MissingMapsCalculationResult {
 
 	private final RoutingContext missingMapsRoutingContext;
 	private final List<LatLon> missingMapsPoints;
+	private boolean onlineResultPoints = false;
```
vshcherb, line 17: This had bad name & bad intention, I would like to understand the gist of this field and why it's necessary here

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculator.java
```diff
@@ -158,13 +159,14 @@ public boolean checkIfThereAreMissingMaps(RoutingContext ctx, LatLon start, List
 				}
 			}
 		}
-		
-//		System.out.println("Used maps: " + usedMaps);
+
+		if (result.hasMissingMaps() || calledByCalcOnline) {
```
vshcherb, line 163: This is confusing as well

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculationResult.java
RZR-UA: refactored

OsmAnd-java/src/main/java/net/osmand/router/MissingMapsCalculator.java
RZR-UA: refactored

RZR-UA:

OsmAnd/src/net/osmand/plus/routepreparationmenu/RequiredMapsController.java
```diff
@@ -222,4 +218,14 @@ public static void showDialog(@NonNull FragmentActivity activity) {
...
 	}
+
+	public boolean shouldShowOnlineCalculation() {
+		return !onlineCalculationRequested && !isLoadingInProgress() && !isAlreadyCalcOnline();
+	}
+
+	private boolean isAlreadyCalcOnline() {
+		RoutingHelper helper = app.getRoutingHelper();
+		MissingMapsCalculationResult result = helper.getRoute().getMissingMapsCalculationResult();
+		return result != null && result.getMissingMapsPoints().size() > 2 + helper.getIntermediatePoints().size();
```
alex-osm, line 229: Not a direct sign of calculation type - online or not. 
Perhaps a `String tag` field could be added to `MissingMapsCalculator` and `MissingMapsCalculationResult` to differentiate between types.

RZR-UA: This is the most obvious method to detect the fact that Calc Online is successfully done already.

We have a number of `missingMapsPoints` and a number of original routing points (`start+end+intermediates = 2 + intermediates`).

Before Calc Online, missingMapsPoints are the same as start+end+intermediates. After, it have a lot of calculated points (hunderds-thousands). Just compare!

The method is based on data that is already exist and, in my opinion, in this particular case we should not expand classes with new useless members just to copy-paste already-existing information.

RZR-UA: refactores to strange tag

alex-osm (approved):

vshcherb:
- Separate requests
- Don't hide button + no changes in missing maps

Issue #19932, opened 2024-05-23
Location-dependent address tab cold start delay (was: Every first run of 'Nearest cities' is slow)


sonora:
Often, users open OsmAnd to specifically look up an address.

When going to Search>Address, the first time you visit there in any app session the list of 'Nearest cities' takes a long time to show up, I always experience in excess of 10 sec.

This is not necessarily attributable to any app initialization going on, as it also happens if the app had already been launched a long time ago.

But any subsequent opening of that screen is near instant (~1 sec), **also if you had panned the map completely elsewhere so the list needed a fresh compilation**.

vshcherb:
I don't see this particular slow obviously cache of cities is built. I have ~2 sec first start.

I don't think we're going to address that issue cause is not that slow on 4 years phone. You can try but I have other issues to invest to ;-)

sonora:
@vshcherb  Perhaps it scales badly with the number of maps you have or something. Yes, I may analyze as I find time.

What I find suspicios is that all subsequent searches are fast, so there seems true potential to improve here, in particular since address search is a core functionality of OsmAnd, so I find it hard to ignore this entirely. 😉

It would help if other users affected could quickly post here, so I can judge if fixing is really worth any effort.

sonora:
It feels as if we compile ALL streets belonging to ALL cities upon the first visit of this screen. Although we would need this only ever if the user taps on a city, and then only for the tapped city...

vshcherb:
If we compile that's probably not efficient and could be changed. 

I suspect it has something to do with that generic code but it's too complicated to be sure without debugging
```
	updateCitiesItems();
		startNearestCitySearch();
		runCoreSearch("", false, false, new SearchResultListener() {
```

Then it probable goes internally to 
```
resArray.clear();
				resArray = townCitiesQR.queryInBox(bbox, resArray);
```

But this one I don't see reading streets ... The only potential issue that's reading all cities from files, so probably it could be improved somewhere here by checking bbox
```
BinaryMapIndexReader r = offlineIndexes.next();
				if (!townCities.containsKey(r)) {
```

sonora:
Two observations which may help: debug:

1. The gravity of the issue is very sensitive to where the map is panned to when you try it. In remote or extremely rural areas it may take only a second or two for the first time the "Nearest cities" list takes to appear. While in other areas, not limited to but including big cities, it may take more like 10 seconds. Try the center of Rome for an example.

[...]

PR #20173, opened 2024-06-24
Decrease ADDRESS_BBOX_RADIUS from 2000km to 500km

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19932

PR #20192, opened 2024-06-27
Improve minCost assertion due to higher inaccuracy of squareRootDist31() on long distances

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2597

PR #20197, opened 2024-06-28
Outdate 4.6 RoutingType migration (fix bug "HH is not default after upgrade 4.6.9 to 4.8.0")

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20019

PR #20240, opened 2024-07-03
Make pebble integration optional 

### LINKED ISSUE #4686: Voice notifications on Pebble even if voice output turned off
Hi,

I am using OSMAnd version 2.8.2 with offline maps on an HTC One M8 with Android 6.0. A very common use-case for me is to navigate along a custom GPX track with the offline-maps, in pedestrian navigation mode. This works well concerning the navigation instructions on the phone. 

If I couple my phone with my Pebble 2 HR through BT, I am getting a lot of duplicate notifications among which each time one that starts with "**Voice:**" and then the navigation text. Another notification starts with "OSMAnd:" and then the routing instruction. 

Now I would like to suppress the "Voice:" notifications, so I have turned "Sound: off" in OSMAnd navigation settings for pedestrian, speech engine to "not used", switched the phone loudness down, set the phone to silent mode. What occurs strange to me is that the voice notifications sill make it through to Pebble. Is there a way to turn the "Voice:" notifications off? 

Thanks,
Stefan

Enovale:
Add a setting in the voice prompts settings tab to disable the Pebble Voice notification added in 3f9a611b125adf6c49ff451e4675504ac205fb95

This hopes to resolve #4686 where the "integration" Voice notification shows less information and is generally less helpful than the normal notification, but that notification cannot be disabled normally.

I wonder why this integration exists at all, but in good faith I just want it to be optional. If someone knows a good use case for it please let me know as I'm curious.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -905,6 +905,7 @@ You need to activate the sensor so OsmAnd can find it.</string>
     • Default-speed adjustable forecast for elevation-dependent walking/hiking/running trip times\n\n
     • Fixed audio output and volume slider behavior\n\n
     </string>
+    <string name="pebble_voice">Send Pebble Voice notifications</string>
     <string name="speak_gps_signal_status">Announce GPS signal loss and recovery</string>
     <string name="speak_route_recalculation">Announce route recalculation</string>
     <string name="map_center_widget">Coordinates: Map center</string>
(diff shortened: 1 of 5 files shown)


vshcherb:
This might be too old and we need to disable pebble completely better

sonora:
@vshcherb Yes, time to deprecate our Pebble alerts, I would say.

Enovale:
Integrating an actual navigation app for pebble could be interesting but that's probably out of the scope of osmand? An example of this would be Nav Me, in fact the existing Nav Me app could probably be interfaced with and told nav data except using OsmAnd data instead

Enovale:
Since this was created, [pebble is back](https://repebble.com/watch) and there are a few [navigation app](https://github.com/bquelhas/pebble-steer)s built to work with OsmAnd already. Not sure where to go from here, but thought i'd get an email out!

PR #20267, opened 2024-07-09
Display correct arrow direction during navigation

Andrewschenko:
Use track direction to display correct marker bearing during navigation


Jerome-Herbinet:
Thank you very much for fixing this problem, as it was causing me to be unable to use OSMand.

PR #20287, opened 2024-07-11
Android part of iOS Issue 3724 (MM calc online fix)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-iOS/issues/3724

PR #20289, opened 2024-07-11
Implement AUTO_DETECT_MICROAMPERES (energy stats)

RZR-UA:
https://forum.developer.samsung.com/t/charging-current-battery-property-current-now/13517


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2528

Issue #20302, opened 2024-07-12
Wrong turn indication in lanes widget (connectivity)


DmitryAlexei:
### Description

Wrong turn indication in route details and lanes widget on web, Android and iOS when routing through this street https://www.openstreetmap.org/way/447752286 
Left lane is proposed when driving straight. 

### Steps to reproduce

- create a route on the OsmAnd web, Android and iOS apps using https://osmand.net/map/navigate/?start=48.501362,9.215598&end=48.504628,9.223682&profile=car#17/48.50261/9.22172
- open route details and see turn-by-turn information

### Actual result

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/8fc4df03-8cfa-46a8-b34b-43958df42425"  width="600" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/9a8bb8ed-53a4-41a7-b4bd-a90fb2348fe8 "  width="300" height="auto"/> </td>
  </tr>
</table>

<table style="width:100%">
  </tr>
  <tr>
    <th>android</th>
    <th>ios</th>
  </tr>
  <tr>
 <th><img src="https://github.com/user-attachments/assets/05b00a1e-3519-4eac-8098-4d0aff851bf1"  width="300" height="auto"/> </th>
  <th><img src="https://github.com/user-attachments/assets/9f7238d0-4c96-4940-ac50-a42a22b57725"  width="300" height="auto"/> </th>
  </tr>
</table>

### Expected result

- Correct indication of turns in route details and lanes widget

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version:
Android/iOS version:
Device model:
Crash-Logs: ?
```

vshcherb:
https://github.com/osmandapp/OsmAnd/issues/18853

aceman444:
How does this depend on connectivity relation support? There does not seem to be such relations on this particular road segment. Also, there are turn:lanes tags, which should be helpful to decide which lane to take. connectivity relations are to be added only to problematic cases.

ivanPyrohivskyi:
Test: [Route1](https://osmand.net/map/navigate/?start=48.501362,9.215598&end=48.504628,9.223682&profile=car)

PR #20316, opened 2024-07-15
Improve active lanes check in muteAndRemoveTurns(), rename method

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20302

RZR-UA:
wrong approach

Issue #20280, opened 2024-07-10
Abnormal work of Attach to the roads 


Velomeloman:
### Description

Attach to the roads doesn't work correctly 
<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/109347561/e8a7f0a0-cf1f-4274-902a-4dc933a19a59"  width="300" height="auto"/> </td>
  </tr>
</table>

Related [message](https://t.me/ruosmand/154181).

### Steps to reproduce

Try attach to the roads [Обручева тест.gpx.txt](https://github.com/user-attachments/files/16162122/gpx.txt)

### Actual result

Incorrect work

### Expected result

Correct Attach to the roads like in stable 4.7.17.

### Your Environment (required)

```
OsmAnd Version: OsmAnd~ 4.8.0#2705m, released: 2024-07-09
```

Nine-Tailed:
I had this
![IMG_20240710_193441_859.jpg](https://github.com/osmandapp/OsmAnd/assets/12549744/d52c915c-c234-4db9-8db4-1d10c2360e1d)

yuriiurshuliak:
The bug has been reproduced. To reproduce it, import the provided GPX file and attempt to attach the track to the road. You'll observe that at a threshold distance of 50 meters, the blue dotted line fails to properly connect with the track. However, reducing the threshold distance to around 5 meters resolves the issue.

OsmAnd~ 4.8.0#2722m, released: 2024-07-12

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/2abfbdb6-a34d-4843-a1ac-424cc467fb41"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/154946ac-e46a-46ce-b703-f06f04479caf"  width="300" height="auto"/> </th>
  </tr>
</table>

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/9c0ecdc0-38c5-42e9-9e05-4c5d6cc1c5c8"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/9076801f-68ae-4dc2-a41e-acfffdc34162"  width="300" height="auto"/> </th>
  </tr>
</table>

RZR-UA:
[...]

PR #20318, opened 2024-07-15
Hotfix ugly misplaced resultMatcher call

RZR-UA:


RZR-UA:

PR #20324, opened 2024-07-16
Fix crash

ivanPyrohivskyi:


OsmAnd/src/net/osmand/plus/search/SearchUtils.java
```diff
@@ -16,6 +16,7 @@
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
+import net.osmand.binary.BinaryMapIndexReader;
```
vshcherb, line 19: Not needed

OsmAnd/src/net/osmand/plus/search/SearchUtils.java
```diff
@@ -66,6 +67,8 @@ public static PoiUIFilter getShowOnMapFilter(@NonNull OsmandApplication app, @No
 				if (!Algorithms.isEmpty(searchPhrase.getFirstUnknownSearchWord())) {
 					filter.setFilterByName(searchPhrase.getFirstUnknownSearchWord());
 				}
+			} else if (searchPhrase.getLastSelectedWord().getResult().object instanceof BinaryMapIndexReader.SearchPoiAdditionalFilter) {
+				filter = new PoiUIFilter(null, app, "");
 			} else {
```
vshcherb, line 72: add instanceof here

PR #20328, opened 2024-07-16
Implement BRP/HH ignoreObstaclesOnStartEndPoints

RZR-UA:


RZR-UA:
Fixed in easier way: https://github.com/osmandapp/OsmAnd-core-legacy/commit/24a290f1cbd3e2fa3f0605336782c8f587b75337

PR #20348, opened 2024-07-18
Increase minCost assertion const from 0.1 to 2.0

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2597

PR #20353, opened 2024-07-19
Fix dex

Corwin-Kh:


OsmAnd/src/net/osmand/plus/OsmandApplication.java
```diff
@@ -894,8 +894,14 @@ public AssetManager getAssets() {
 
 	@Override
 	public Resources getResources() {
-		Resources localizedResources = localeHelper.getLocalizedResources();
-		return localizedResources != null ? localizedResources : super.getResources();
+		OsmandMap map = getOsmandMap();
```
vshcherb, line 897: I wrote this part, it doesn't look very nice as it access activity from application, though it looks it much simpler approach however it's bad to use resources from application but it requires refactoring for the whole application

OsmAnd/src/net/osmand/plus/utils/AndroidUtils.java
```diff
@@ -1438,6 +1439,14 @@ public static boolean isBluetoothEnabled(@NonNull Context context) {
 
 	public static Display getDisplay(@NonNull Context context) {
 		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+			OsmandApplication app = (OsmandApplication) context.getApplicationContext();
+			OsmandMap osmandMap = app.getOsmandMap();
```
vshcherb, line 1443: here we need to use context.getDisplay()

OsmAnd/src/net/osmand/plus/views/layers/ContextMenuLayer.java
```diff
@@ -175,6 +175,12 @@ private void updateContextMarker() {
 		contextMarkerImage = getScaledBitmap(R.drawable.map_pin_context_menu, scale);
 	}
 
+	@Override
```
vshcherb, line 178: Other layers missing like - 1) Destination intermediate points,  2) Avoid roads . Something else ? Parking?

PR #20390, opened 2024-07-23
[4.8] GpxSegmentsApproximation improvements (gpx angle, bidirectional segments, distance to projection, crush long segments)

RZR-UA:


RZR-UA:

RZR-UA:
will be improved later

PR #20434, opened 2024-07-28
[draft] Prepare class for GPX-approximation tests

RZR-UA:


RZR-UA:

RZR-UA:
will be replaced soon

PR #20472, opened 2024-08-01
Compass fix

0xRe1nk0:


OsmAnd/src/net/osmand/plus/base/MapViewTrackingUtilities.java
```diff
@@ -159,7 +161,11 @@ public void setLastResetRotationToNorth(long lastResetRotationToNorth) {
 		this.lastResetRotationToNorth = lastResetRotationToNorth;
 	}
 
-	public boolean allowRotationAfterReset(){
+	public boolean allowRotationAfterReset() {
+		CompassMode currentMode = settings.getCompassMode();
+		if (currentMode == MANUALLY_ROTATED || currentMode == NORTH_IS_UP) {
```
vshcherb, line 166: north is up allows rotation?

Issue #19785, opened 2024-05-06
Improvements for 3D track view


dmpr0:
## Description
Previous task https://github.com/osmandapp/OsmAnd-Issues/issues/2436

## ToDo

<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th>Visualized by</th>
        <th>Wall color</th>
      </tr>
      <tr>
        <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/49e3198c-d386-4c10-bbb1-917a170ecc8a"  width="300" height="auto"/> </td>
        <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/eab928a4-b525-4c1b-852b-31ebc090c6ec"  width="300" height="auto"/> </td>
      </tr>
    </table>   
</details>   

- [x] Place **Wall color** as third parameter below **Track line**

### Wall color
- [x] Solid / Downward gradient / Upward gradient
- - [x] Track line – user selected color
- - [x] Wall color – user selected color
- [x] Altitude / Slope / Speed
- - [x] Track line – user selected color
- - [x] Wall color – selected gradient (Altitude / Slope / Speed)

### Visualized by

<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th>Altitude</th>
        <th>Fixed height</th>
      </tr>
      <tr>
        <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/c7235824-8a04-4340-9196-43326937cad6"  width="300" height="auto"/> </td>
        <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/8998c25f-c753-4046-9826-6b430d047b57"  width="300" height="auto"/> </td>
        <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/353530ad-afdd-4142-abeb-0bb75a0f1a13"  width="300" height="auto"/> </td>
      </tr>
    </table>   
</details>   

- [x]  If **Visualized by** = **Fixed height**, change **Vertical exaggeration** to **Wall height**

**Add new options:**

- [x] Speed
- [x] Heart rate
- [x] Bicycle Cadence
- [x] Bicycle Power
- [x] Temperature
- [x] And all other with numers
- [x] Add Height from Relief (altitude from track) - if 3D relief is ON for for all types of visualization except "By height"

DmitryAlexei:
[...]

PR #20473, opened 2024-08-01
Fix 3d tracks height with 3d maps for #19785

Chumva:


OsmAnd/src/net/osmand/plus/views/layers/GPXLayer.java
```diff
@@ -172,6 +172,7 @@ public class GPXLayer extends OsmandMapLayer implements IContextMenuProvider, IM
 	private net.osmand.core.jni.MapMarker highlightedPointMarker;
 	private LatLon highlightedPointLocationCached;
 	private long trackMarkersChangedTime;
+	private boolean needRedraw;
```
alex-osm, line 175: Rename needRedraw > invalidated and move it to OsmandMapLayer

Issue #20426, opened 2024-07-27
Route calculation is so slow in +4.8.4


Rowin63:
### Description

Since the last update to the playstores +4.8.4 the route calculation is very slow, even if the finish is only one km away. There is always a huge break at around 20% progress. Processing a route to a finish 2km away needs 15-20 sec. Longer routes like 20km seem to never finish calculating. 

### Steps to reproduce

As described. 

### Actual result

Calculating is very slow. 

### Expected result

Calculate quick as usual. 

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version:
Android/iOS version:
Device model:
Crash-Logs: ?
```

Rowin63:
Deleted the Austrian map and re-downloaded it. Also tried the standard map as well as the road-only-map. No difference.

Rowin63:
Happens in the HHxC++, but not in the A*classic.

MarioSwitch:
Not reproduced 🤔

OsmAnd 4.8.4, Google Pixel 7 (Android 14), HH x C++ routing

Can get a ~1000km route in 15 seconds, as quick as older versions...

Rowin63:
This is the point where the calculation stops, it´s always the same. 
(Very) short routes contiue after a while, longer ones don´t (or I dont wait long enough, but it takes minutes).

Happens only on this device (S20+). 
Happens since the last update (+4.8.4). 
There was a crash at the first app start after the update (I sent the crashlog to your support mail address).

Does not happen on my Tab Act3 (also running A13 and Osmand +4.8.4).

![Screenshot_20240727_130138_OsmAnd+](https://github.com/user-attachments/assets/cb8203cb-94f8-4b88-a869-ace60425bce4)

Rowin63:
Did a de- and re-install (with "keep the app's data), did not help.
Did a de- and re-install (without keeping the data), this helped.
Restored my data from a backup I made before the de-install, and still everything is fine. So the issue shouldn't be hidden in my data, it's the same before and now. Perhaps it was caused by the crash at the update? I don't know. You can see the crashlog I sent. 

By now the problem seems to be solved.

vshcherb:
As I noticed it could be if some maps are not updated (even that are not used in calculation), however if you update all maps it should be fast

Rowin63:
I had looked at that, and all maps have been up to date (also world map, ...)
[...]

PR #20484, opened 2024-08-02
Fix HH routing with map with different timestamp

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/20426


OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -741,7 +743,15 @@ private HHRoutingContext<T> selectBestRoutingFiles(LatLon start, LatLon end, HHR
 			}
 		}
 		for (HHRouteRegionsGroup<T> g : groups) {
-			g.containsStartEnd = g.contains(start) && g.contains(end);
+			boolean cStart = g.contains(start);
```
vshcherb, line 746: If group of HH route regions is selected only points from these group should be used. So it means there is a bug in route calculation but that check doesn't make sense

PR #20517, opened 2024-08-06
Fix hh routing

ivanPyrohivskyi:
Add filter for avoid unnecessary route searching (findRouteSegment for start/end, searchRouteInternal for initStart) in RoutingContext.

To issue https://github.com/osmandapp/OsmAnd/issues/20426


OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -174,6 +174,10 @@ public HHNetworkRouteRes runRouting(LatLon start, LatLon end, HHRoutingConfig co
 		if (hctx == null) {
 			return new HHNetworkRouteRes("Files for hh routing were not initialized. Route couldn't be calculated.");
 		}
+		hctx.rctx.mapIndexReaderFilter = new HashSet<>();
```
vshcherb, line 177: 1. move to initHCtx
2. This method requires to call RoutingContext.unloadAllData
3. That's why it needs to be checked whether there is something to filter or not. So it's better to create mapIndexExclude map so it's easy to check if there is something to exclude to clear cache.
4. **It's incorrect** doesn't include live update maps for same region

ivanPyrohivskyi:
1. Move to selectBestRoutingFiles
2. RoutingContext.unloadAllData is call always after run hh routing [RoutePlannerFrontend runHHRoute](https://github.com/osmandapp/OsmAnd/blob/7cb46f71fac4534e82d6328db4d9383fdcee6ef4/OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java#L526)
3. Check count of found groups in selectBestRoutingFiles . groups.size() == 1 mean that all maps are updated.
4. live-updates maps is not include HH section, so we pass such files
5. Added RoutePlannerFrontEnd.CALCULATE_MISSING_MAPS , that means user is not use option "Use downloaded maps" in missing maps dialog.

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -797,6 +797,12 @@ public int compare(HHRouteRegionsGroup<T> o1, HHRouteRegionsGroup<T> o2) {
 		if (allMatched) {
 			return currentCtx;
 		}
+		if (RoutePlannerFrontEnd.CALCULATE_MISSING_MAPS && groups.size() > 1) {
```
vshcherb, line 800: We need filter always if groups > 1.  Why here is the check for CALCULATE_MISSING_MAPS

PR #20523, opened 2024-08-07
Do migration for navigation and static location profile icons

nazar-kutz:
Implement migration from old to new Navigation Icons ids to escape possible bugs when we override static icons' ids on runtime when algorithm thinks those ids are old.


vshcherb:
Supercede https://github.com/osmandapp/OsmAnd/pull/20527/files

PR #20735, opened 2024-09-02
Prevent writing Group appearance to every single point

sonora:


vshcherb:
Simply smaller file size

PR #20775, opened 2024-09-07
Fix approximate test free

vshcherb:


RZR-UA:
superseded with https://github.com/osmandapp/OsmAnd/pull/20842

PR #20890, opened 2024-09-23
Retrieve amenity for rendering polygon

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd-Issues/issues/2393


Chumva (approved):
getSyntheticAmenity looks strange, but other places are fine

PR #20933, opened 2024-09-27
fling fix

0xRe1nk0:


OsmAnd/res/layout/fragment_edit_poi_normal.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.core.widget.NestedScrollView
```
vshcherb, line 2: Here was no problem as I remember

OsmAnd/res/layout/fragment_edit_poi_advanced.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.core.widget.NestedScrollView
+<ScrollView
```
vshcherb, line 2: We can't use ScrollView / NestedScrollView coudln't be used here cause it's already inside scrollview. Also we need to use recyclerview for tag/ values on advanced tabs

OsmAnd/src/net/osmand/plus/plugins/osmedit/dialogs/NewBasicEditPoiFragment.java
Chumva, line 1: Rename to BasicEditPoiFragment

OsmAnd/src/net/osmand/plus/plugins/osmedit/dialogs/NewBasicEditPoiFragment.java
```diff
@@ -0,0 +1,268 @@
...
+			openingHours = new OpeningHoursParser.OpeningHours();
+		}
+		openingHoursAdapter.replaceOpeningHours(openingHours);
+		updateViews();
+		openingHoursAdapter.updateHoursData();
+
+		basicTagsInitialized = true;
+	}
+
+	public class OpeningHoursAdapter {
```
Chumva, line 202: Copy pasted class from the previous screen?

OsmAnd/src/net/osmand/plus/plugins/osmedit/dialogs/NewBasicEditPoiFragment.java
```diff
@@ -0,0 +1,268 @@
...
+import net.osmand.plus.utils.AndroidUtils;
+import net.osmand.plus.utils.ColorUtilities;
+import net.osmand.util.OpeningHoursParser;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+public class NewBasicEditPoiFragment extends BaseOsmAndFragment implements EditPoiDialogFragment.OnFragmentActivatedListener {
```
Chumva, line 38: It looks like there is issue with scroll height 

<table style="width:100%">
  <tr>
<th> </th>
  </tr>
</table>

Issue #2993, opened 2016-08-23
show city next to search result


schiessle:
If you search for something which exists multiple times in your region, e.g. a shopping center like Ikea, you get multiple results and it is hard to judge which one ist he right one. On the search result, the only additional information you have is the distance to your current location. But that's not always enough, especially if you are in a foreign location. Therefore it would be really helpful to add some additional information to the search result, like for example the city.

vshcherb:
It's kind of duplicate nice to have.  Though it doesn't require full results geocoding and could be implemented just for city name.

sonora:
This is requested very frequently, guess it should be in our 3.0 milestone.

vshcherb:
@sonora Very unlikely it will happen soon. The frequency of requests shows popularity but doesn't show the complexity. Ideally all POIs should have Street, City. City is not enough but in order to achieve it addresses should be prepared in the data which makes that story difficult to implement

schiessle:
@vshcherb why not doing it step-by-step if showing only the city would be easier? Sure, showing the complete address would be the perfect solution but having just the city would make search in many cases already much easier.

vshcherb:
If we are talking about popular POI in large cities than their name is mostly irrelevant. If we are talking about correct information it requires address lookup for each poi.

jfburke3:
Would displaying the value for the "branch" tag be easy, if the value is not null? E.g., a displayed result might appear as "Wells Fargo (Wildwood)" for a Wells Fargo bank location with `branch=Wildwood` set in OSM.

I don't know about the rest of the world, but in the US and Canada, it is fairly common for some types of businesses to identify various locations by branch.

EvanCarroll:
@jfburke3  that be preferable imho and no more difficult. We track that here https://github.com/osmandapp/OsmAnd/issues/6947 For reference the OP is from Stuttgart, Germany. In Stuttgart there are five Starbucks. Knowing that they're all in Stuttgart is not very useful, however the branch names seem much more valuable:

* Koenigstrasse 44
* Stuttgart, Koenigsbau Passagen
* Stuttgart HB Kiosk
* Milaneo Kiosk
* Milaneo

What would you rather have @schiessle those names above or just "Stuttgart". If you want the branch names, would it be cool to close this as a duplicate of this https://github.com/osmandapp/OsmAnd/issues/6947 ?

schiessle:
[...]

PR #21039, opened 2024-10-14
Change priority in search for city / alternate name

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/2993#issuecomment-2397118471


vshcherb:
Nobody asked to hide alternate name, we will have another issue. Search by brand is not alternate name essentially... This needs to be discussed

PR #21077, opened 2024-10-18
Enable battery saving mode in Android Auto

Andrewschenko:
Enable battery saving mode while using Android Auto


OsmAnd/src/net/osmand/plus/auto/SurfaceRenderer.java
```diff
@@ -371,7 +371,7 @@ public void setupOffscreenRenderer() {
 							offscreenMapRendererView.addListener(this);
 							mapView.getAnimatedDraggingThread().toggleAnimations();
 						}
-						mapView.applyBatterySavingModeSetting(offscreenMapRendererView);
```
vshcherb, line 374: disable happens in another view? or not needed?

Andrewschenko: it reverts back to app setting when android auto goes off and switches back to device screen

PR #21079, opened 2024-10-18
Add region identificator to amenity

ivanPyrohivskyi:


OsmAnd-java/src/main/java/net/osmand/data/Amenity.java
```diff
@@ -11,23 +9,13 @@
...
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Set;
-import java.util.TreeSet;
 
-import gnu.trove.list.array.TIntArrayList;
+import static net.osmand.gpx.GPXUtilities.AMENITY_PREFIX;
```
vshcherb, line 17: Constantly changing... however no imports changed actually, java.util.* not by default old guidelines

Issue #20620, opened 2024-08-15
Wikipedia links lead to main page 


mitkin89:
### Description

When saved in favorites pressing to read Wikipedia article online leads not to the article but to the main wiki page 

### Steps to reproduce

Open a favorited wiki article from on the map and press to view it online

https://github.com/user-attachments/assets/82a5a9cf-639d-4876-b3ce-f1126ae99708

### Actual result

Leads to main wiki page 

### Expected result

Leads to the actual article 

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 4.8.6
Android/iOS version: Emui 14
Device model: Mate x3 
Crash-Logs: ?
```

yuriiurshuliak:
The bug has been reproduced. To reproduce: enable the Wikipedia plugin, download the Wikipedia packages, display the Wikipedia POIs, add one of the Wikipedia POIs to your favorites, select this favorite to read the article, click "Read full article," and then click "Read full article online."

As a result, instead of being redirected to the specific article, you'll be taken to the main page of Wikipedia in the browser.

OsmAnd~ 4.9.0#2997m, released: 2024-08-16


https://github.com/user-attachments/assets/dc53f012-a742-4ac9-a5ec-dedc28173340

vshcherb:
@yuriiurshuliak is it a regression (4.5)? is it for all languages / articles that are not inside obf files?

yuriiurshuliak:
Since the issue also occurs in version 4.5, it is a regression. The problem relates to downloaded OBF files and has been verified in Polish, German, and English languages.

When you add Wikipedia POIs to your favorites and then try to read the article online in a browser, you are redirected to the main page. However, if you don't add the Wikipedia POI to favorites and attempt to read the article online, you are correctly redirected to the specific article.

OsmAnd 4.5.10, released: 2023-08-09

<table style="width:100%">
  </tr>
  <tr>
    <th>Polish</th>
    <th>German</th>
    <th>English</th>
  </tr>
  <tr>
    <td> <video source src="https://github.com/user-attachments/assets/deea8e2d-e25d-4145-9e23-b8010d0e518a"  width="300" height="auto"/>   </td>
    <td><video source src="https://github.com/user-attachments/assets/9b0f79f2-fcbc-41b9-9d7b-ad556ed0e59c"  width="300" height="auto"/>   </td>
[...]

PR #21171, opened 2024-10-28
Fix name for Wiki-Articles-in-Favorites, fix multi-lang tag prefix, fix lost infoRow due to already defined hasWiki

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20620

PR #21176, opened 2024-10-29
Optimize another hasWiki try for infoRow at return of createLocalizedAmenityInfoRow

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20620

PR #21183, opened 2024-10-29
Remove delay for speed alarm and simplify announce-speed-alarm logic by doing it

Matombo:
This PR removed the 5 secounds wait before a speed alarm sound is played because going 130km/h on the Autobahn you make 180m before the sound starts playing -> more then enough for a speeding camera to get you.

It still waits at least 120s before repeating the speeding limit warning.

I would like this 120s timer to reset once you go under the speed limit so that then it can play instant again if you go over again, even if 120s haven't passed yet, but I don't know how to do that as I'm not familiar with the codebase.

I have no testing setup set-up and the README doesn't include a quick guide on how to get one running, so this code is untested and should more be seen as a suggestion. Let me know if there is a developer guide for OsmAnd that I'm simply not aware of.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/routing/VoiceRouter.java
+++ b/OsmAnd/src/net/osmand/plus/routing/VoiceRouter.java
@@ -52,8 +52,6 @@ public class VoiceRouter {
 	private static final int STATUS_TOLD = 5;
 
 	private static final int SPEEDING_ANNOUNCEMENTS_INTERVAL_MS = 120_000;
-	private static final int SPEEDING_ANNOUNCEMENT_CANCEL_MS = 30_000;
-	private static final int SPEEDING_ANNOUNCEMENT_WAITING_MS = 5_000;
 
 	public static final String TO_REF = "toRef";
 	public static final String TO_STREET_NAME = "toStreetName";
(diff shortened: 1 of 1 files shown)


Matombo:
> I would like this 120s timer to reset once you go under the speed limit so that then it can play instant again if you go over again, even if 120s haven't passed yet, but I don't know how to do that as I'm not familiar with the codebase.

Just looking at the original function It only makes sense when it gets triggered continiously while you are over the speed limit. Honest question: Am I right?

But it doesn't get triggered when you are under the speed limit, and you don't know if you went under the speed limit since the last time it got triggered.

One Idea that I have to detect if we are in the same continous triggering loop or the next one just withing the function, is to set a timestamp everytime the function gets triggered and have a threshhold that, for example, the function didn't get triggered for 1000ms, we assume we are in a new continues trigger loop and reset lastAnnouncedSpeedLimit to 0 at the beginning of the function, so that the sound is triggered instant again.

Matombo:
> Let me know if there is a developer guide for OsmAnd that I'm simply not aware of.

Found it https://www.osmand.net/docs/technical/build-osmand/

vshcherb:
> wait before a speed alarm sound is played because going 130km/h on the Autobahn you make 180m before the sound starts playing -> more then enough for a speeding camera to get you.

That's what the code suppose to do, there is no need to play it immediately. I think it works by default exactly as it's written today. The version you are describing was very annoying to many users, so it was improved by this part of the code

Matombo:
What about making it configurable?

Currently there is the "Tolerance for speedlimit" setting which sets the amount of speed to ecxeed before the warning. My quick idea is to aditionaly have a "Time tolarance for speed limit" setting SPEEDING_ANNOUNCEMENT_WAITING_MS (maybe renaming the other one to "Speed tolarance for speed limit" and give a good short description for both explaining their relation).

PR #21191, opened 2024-10-30
Fix obd reconnect

Corwin-Kh:


OsmAnd/src/net/osmand/plus/plugins/odb/VehicleMetricsPlugin.kt
```diff
@@ -418,52 +416,40 @@ class VehicleMetricsPlugin(app: OsmandApplication) : OsmandPlugin(app),
...
+		if (currentConnectingState != OBDConnectionState.DISCONNECTED) {
+			disconnect()
+		}
+		if (BLEUtils.isBLEEnabled(activity)) {
+			if (AndroidUtils.hasBLEPermission(activity)) {
+				onConnecting(deviceInfo)
+				OBDDispatcher.useInfoLogging =
+					PluginsHelper.getPlugin(OsmandDevelopmentPlugin::class.java)?.isEnabled == true
+
+				if (settings.SIMULATE_OBD_DATA.get() && deviceInfo.address.isEmpty()) {
```
vshcherb, line 428: Should be part of of connectToDevice - same code

OsmAnd/src/net/osmand/plus/plugins/odb/VehicleMetricsPlugin.kt
```diff
@@ -418,52 +416,40 @@ class VehicleMetricsPlugin(app: OsmandApplication) : OsmandPlugin(app),
...
-						if (settings.SIMULATE_OBD_DATA.get() && deviceInfo.address.isEmpty()) {
-							onDeviceConnected(deviceInfo)
-							val simulator = ODBSimulationSource()
-							val input = simulator.reader
-							val output = simulator.writer
-							OBDDispatcher.setReadWriteStreams(input, output)
-							return true
-						}
+		if (currentConnectingState != OBDConnectionState.DISCONNECTED) {
+			disconnect()
```
vshcherb, line 420: Not part of this request but connect disconnect should be moved to the thread of OBDDispatcher, so they can work asyncrhonously and all operations with socket will happen in 1 thread - in a thread of OBDDispatcher

OsmAnd/src/net/osmand/plus/plugins/odb/VehicleMetricsPlugin.kt
Corwin-Kh: moved same code to function

PR #21194, opened 2024-10-31
Fixed route preview zoom

Corwin-Kh:


OsmAnd/src/net/osmand/plus/views/Zoom.java
```diff
@@ -46,6 +46,23 @@ public void zoomOut() {
 		changeZoom(-1);
 	}
 
+	public void smallZoomIn() {
+		zoomFloatPart += 0.01f;
```
vshcherb, line 50: what this constants about, looks really wrong. We need to adjust exactly proper zoom up to 0.1 not to 0.01 delta

OsmAnd/src/net/osmand/plus/views/Zoom.java
```diff
@@ -46,6 +46,23 @@ public void zoomOut() {
...
+		if (zoomFloatPart > 0.99) {
```
vshcherb, line 51: 0.99 -> 0.5

PR #21196, opened 2024-10-31
try fix multithreading issues

alisa911:


OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
```diff
@@ -103,15 +103,19 @@ public void setTranslator(RegionTranslation translator) {
 		this.translator = translator;
 	}
 
+	private final Object fileLock = new Object();
+
 	public BinaryMapIndexReader prepareFile() throws IOException {
 		File regions = new File("regions.ocbf");
 		// internal version could be updated
-//		if (!regions.exists()) {
```
vshcherb, line 109: This is definitely not part of this class, multithreading should be solved where it's called

Issue #20846, opened 2024-09-15
Can't update POI data


luigifab:
### Description

I'm displaying swimming pools POI on the map. I would like to add opening hours  but I can't edit it, I can only add a new POI.

### Steps to reproduce

Go to 55.52213 25.08419 and display swimming pools POI. Then try to edit it.

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/5086f23b-0b1e-424b-8299-1be3fe673e1b"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/d2c6d83e-0946-4fbd-a0ea-b2aaf0a2877f"  width="300" height="auto"/> </th>
  </tr>
</table>

### Actual result

Can't edit POI details.

### Expected result

Allow to edit the POI.

### Your Environment (required)

```
OsmAnd Version: 4.8.6
Android/iOS version: 11
Device model: Wiko
```

yuriiurshuliak:
I was unable to reproduce the issue. To try reproducing it, go to the coordinates 55.52213, 25.08419, click on the POI of the pool, select "Actions," then choose "Modify POI" and "Add Opening hours."

Please ensure that the OpenStreetMap editing plugin is enabled.

OsmAnd~ 4.9.0#3193m, released: 2024-09-16

https://github.com/user-attachments/assets/f15ae662-dafd-4a22-81d3-586ca26816ed

luigifab:
yes plugin is enabled, and I tried again and I still can't edit this POI, for example next to the swimming pool there are bus stops, and I can edit them

vshcherb:
Bug reproduced; https://www.openstreetmap.org/way/322243903.
Context menu displays correct url to way. 
To fix: all nodes and ways with correct url should be editable in OSM.

nazar-kutz:
@vshcherb The problem seems to be caused by the "no_edit=true" attribute on the poi type of this point. Maybe we should fix it in xml.

Line in XML:
<poi_type name="swimming_pool" tag="leisure" value="swimming_pool" no_edit="true" poi_additional_category="nudism,fee,access_general"/>
https://github.com/osmandapp/OsmAnd-resources/blob/master/poi/poi_types.xml#L4196

Line in code that filtering this point:
isEditable = !amenity.getType().isWiki() && poiType != null && !poiType.isNotEditableOsm();
[...]

PR #21198, opened 2024-10-31
Fix "Can't update POI data #20846"

nazar-kutz:


OsmAnd/src/net/osmand/plus/plugins/osmedit/OsmEditingPlugin.java
```diff
@@ -343,8 +343,12 @@ public void registerMapContextMenuActions(@NonNull MapActivity mapActivity,
 			} else {
 				amenity = ((TransportStop) selectedObj).getAmenity();
 			}
-			PoiType poiType = amenity.getType().getPoiTypeByKeyName(amenity.getSubType());
-			isEditable = !amenity.getType().isWiki() && poiType != null && !poiType.isNotEditableOsm();
+			if (settings.isInternetConnectionAvailable() && ObfConstants.isOsmUrlAvailable(amenity)) {
+				isEditable = true;
+			} else {
+				PoiType poiType = amenity.getType().getPoiTypeByKeyName(amenity.getSubType());
+				isEditable = !amenity.getType().isWiki() && poiType != null && !poiType.isNotEditableOsm();
```
Chumva, line 350: Do we need to ignore amenity.getType().isWiki(), or only the second part with poiType != null && !poiType.isNotEditableOsm()?

PR #21211, opened 2024-11-02
Fixed fuel tank default value

Corwin-Kh:


OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java
```diff
@@ -1134,7 +1134,7 @@ public VolumeUnit getProfileDefaultValue(ApplicationMode mode) {
 
 	// fuel tank capacity stored in litres
 	public final OsmandPreference<Float> FUEL_TANK_CAPACITY = new FloatPreference(this,
-			"fuel_tank_capacity", 0).makeProfile();
+			"fuel_tank_capacity", 52f).makeProfile();
```
vshcherb, line 1137: hard coded? link?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/obd/OBDDataComputer.kt
```diff
@@ -15,11 +15,12 @@ object OBDDataComputer {
 
 	private val log = LoggerFactory.getLogger("OBDDataComputer")
 
+	private const val DEFAULT_FUEL_TANK_CAPACITY = 52f
```
vshcherb, line 18: how setting is  used here?

PR #21269, opened 2024-11-07
add hasLongDistance

alisa911:


alisa911:

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxParameter.kt
```diff
@@ -13,6 +13,7 @@ enum class GpxParameter(
 	FILE_NAME("fileName", "TEXT", String::class, null, false),
 	FILE_DIR("fileDir", "TEXT", String::class, null, false),
 	TOTAL_DISTANCE("totalDistance", "double", Double::class, 0.0, true),
+	IS_LONG_DISTANCE("isLongDistance", "int", Boolean::class, false, true),
```
RZR-UA, line 16: `maxDistanceBetweenAdjacentPoints` would be better to increase flexibility at caller's side

RZR-UA (changes requested):

PR #21273, opened 2024-11-07
add temperature units to obd widgets

0xRe1nk0:


OsmAnd/src/net/osmand/plus/plugins/odb/OBDWidgetSettingFragment.kt
```diff
@@ -81,18 +88,29 @@ class AverageModeSettingFragment : BaseSimpleWidgetSettingsFragment() {
...
+				selectedAverageMode,
+			) { showMarkerModeDialog() })
+
+			if (selectedAverageMode) {
+				buttonsCard?.addView(createButtonWithDescription(
+					getString(R.string.shared_string_interval),
+					availableIntervals[selectedIntervalMillis],
+					false,
+				) { showSeekbarSettingsDialog() })
+			}
```
Chumva, line 113: It looks complicated, can we simplify it? Because adding new parameters will be harder...
 
What if we use list  - inner class ButtonItem(title: String, desc: String, listener: View.OnClickListener), showShortDivider will be for last item

OsmAnd/src/net/osmand/plus/plugins/odb/OBDTextWidget.kt
```diff
@@ -24,18 +26,20 @@ open class OBDTextWidget(
 	private val fieldType: OBDTypeWidget,
 	customId: String?,
 	widgetsPanel: WidgetsPanel?) :
-	SimpleWidget(mapActivity, widgetType, customId, widgetsPanel) {
+	SimpleWidget(mapActivity, widgetType, customId, widgetsPanel), OBDWidgetOptions {
 	private val plugin = PluginsHelper.getPlugin(VehicleMetricsPlugin::class.java)
```
Chumva, line 30: There is a new requirePlugin method, preferable for places that works only with it

OsmAnd/src/net/osmand/plus/plugins/odb/OBDTextWidget.kt
```diff
@@ -214,4 +235,8 @@ open class OBDTextWidget(
 			else -> false
 		}
 	}
+
+	override fun getTemperatureUnit(): TemperatureUnit {
+		return temperatureUnitPref?.get() ?: app.weatherHelper.weatherSettings.weatherTempUnit.get()
```
Chumva, line 240: Is it ok to get a temperature unit from another plugin?

PR #21276, opened 2024-11-08
fix exit\remove symbol in street name

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/NextTurnBaseWidget.java
```diff
@@ -242,6 +239,13 @@ public void setStreetName(@Nullable CurrentStreetName streetName) {
 		}
 	}
 
+	public static String removeSymbol(String input) {
+		if (input.matches("^»\\s.*")) {
```
Chumva, line 243: Do we need regex here? won`t startsWith be enough?

PR #21279, opened 2024-11-08
add time formatters

alisa911:


alisa911:

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
```diff
@@ -888,19 +916,41 @@ object GpxUtilities {
...
 	private fun parseTime(text: String, format: DateTimeFormat<DateTimeComponents>): Long {
-		var time: Long = 0
-		try {
-			time = flexibleGpxTimeParser(text, format)
-		} catch (e: Exception) {
+		val processedText = preprocessGpxTime(text)
+		val formats = listOf(
+			format,
+			getTimeNoTimeZoneFormatter(),
+			getTimeFormatterTZ(),
```
RZR-UA, line 933: getTimeFormatterTZ() could be excluded because it's already duplicated with:

	fun parseTime(text: String): Long {
		return parseTime(text, getTimeFormatterTZ())
	}

Issue #21252, opened 2024-11-05
Approximate size of map source cache


Nine-Tailed:
### Description

Recently osmand shows the size of the cache of online map sources if it exceeds 50 MB as > 50 MB. I would like to see the exact size of the cache as before.

<table>
    <tr>
        <th>5.0</th>
        <th>4.8.6</th>
    </tr>
    <tr>
        <td>
<img src="https://github.com/user-attachments/assets/78c58ae3-2e75-47d7-b92a-fe30ce185529">
</td>
        <td>
<img src="https://github.com/user-attachments/assets/ccd4e9a4-5b87-471b-b6d8-ef764df755b5"

</td>
    </tr>
</table>

### Steps to reproduce

.

### Actual result

The size of large files is shown as >50 Mb

### Expected result

Show the exact size of the online maps cache.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version:OsmAnd~ 5.0.0#3647m, Дата выпуска: 2024-11-04
Android/iOS version:
Device model:
Crash-Logs: ?
```

dmpr0:
## ToDo
<table style="width:100%">
  <tr>
    <th>Context menu</th>
    <th>File size, Calculating…</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/e6fdd3c8-3ab2-4277-a67f-256ea20fe174"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/4827653f-1c2e-4e26-bc47-dfcac972f093"  width="300" height="auto"/> </td>
  </tr>
</table>

Nine-Tailed:
For map sources in sqlite.db format it might be worth making an exception, since it is only one file.

vshcherb:
Of course you didn't mention you have sqlitedb in bug description.

dmpr0:
# Review

- [ ] **Calculate size** action visible only after some action? App Restart?

vshcherb:
Action should be visible always

nazar-kutz:
We made "Calcualte size" action permanently visible here: https://github.com/osmandapp/OsmAnd/pull/21662

Zirochkabila:
# Testing 2
`OsmAnd~ 5.0.0#4062m, Released: 2025-01-06`

1. Calculate size action - **Fixed**
[...]

Issue #20974, opened 2024-10-03
Opening Maps & Resources / Local takes very long


scaidermern:
### Description

Opening the "Local" tab at Maps & Resources takes very long (about 40 seconds). This happens both right after starting the app and also when coming back again later. It doesn't happen after loading finished and switching between the tabs Downloads and Local.

I tried to move all my maps, geotiff, heightmap and tracks to a different location so that OsmAnd doesn't have to index them. The behavior stays the same.

I already cleared OsmAnd's cache. The behavior stays the same.

Map panning/zooming speed is normal (within OsmAnd's abilities, at least :)).

Not sure if this is a new behavior with OsmAnd 4.9.0 since I had to replace my old phone and there is no easy way to switch back to the old, stable version.

I can provide a video but I don't think it will be of much use.

### Steps to reproduce

Open Maps & Resources / Local.

### Actual result

A loading icon appears for about 40 seconds.

### Expected result

The content is shown almost immediately.

### Your Environment (required)

```
OsmAnd Version: 4.9.0
Android version: 13
Device model: Fairphone 4
```

scaidermern:
I can't reproduce this behavior with the current nightly. Was there any fix for this problem?

I'm not sure if reproducing it with the nightly really tells me if this problem got fixed since it has a different storage path etc.

DmitryAlexei:
This information worth adding to documentation (Troubleshooting section or [Maps and resources > local](https://osmand.net/docs/user/personal/maps-resources#local)  )

scaidermern:
> was a sqlite big ?

Which sqlite?

From the video I recorded "Maps sources" was 0.6 MB. However, I don't really believe that. I had tiles from various different online map sources (OsmAnd online, Microsoft Earth, ...) in various geographic regions. Must have been more than just 0.6 MB, also considering that removing the tiles directory took quite some time.

Rowin63:
The "tiles" folder is growing with every use (XX-thousand files) and can not be cleared easily. It´s ojnly possible by using a file commander, and then one never knows what to delete or not (it´s not possible to delete just the tiles/areas/countries not used/wanted anymore).

vshcherb:
Tiles folder should not calculating at all if it's takes > 2-3 seconds and in background, so we need than kind of a link "Calculate size" as we did in Storage analyzer

scaidermern:
[...]

PR #21316, opened 2024-11-12
Don't limit display size for tiles stored as SQLite DB

nazar-kutz:
This request has to fix problem described here: https://github.com/osmandapp/OsmAnd/issues/21252#issuecomment-2462754169

The problem comes from the new conditions we set to calculate the size of the tiles (Issue: https://github.com/osmandapp/OsmAnd/issues/20974), however we shouldn't set these limits for tiles those are stored as one SQLite file, because there is only one file and it calculates anyway.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/download/local/CollectLocalIndexesAlgorithm.java
+++ b/OsmAnd/src/net/osmand/plus/download/local/CollectLocalIndexesAlgorithm.java
@@ -104,8 +104,7 @@ public class CollectLocalIndexesAlgorithm {
 	}
 
 	private void addSeparatelyCalculationItemIfNeeded(@NonNull LocalItem item) {
-		LocalItemType type = item.getType();
-		Long limit = rules.getCalculationSizeLimit(type);
+		Long limit = rules.getCalculationSizeLimit(item);
 		if (limit != null && !separateSizeCalculationItems.contains(item)) {
 			separateSizeCalculationItems.add(item);
 			item.setSizeCalculationLimit(limit);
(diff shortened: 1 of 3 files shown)


OsmAnd/src/net/osmand/plus/download/local/CollectLocalIndexesAlgorithm.java
```diff
@@ -104,8 +104,7 @@ private void addLiveItem(@NonNull LocalCategory category, @NonNull File file, @N
 	}
 
 	private void addSeparatelyCalculationItemIfNeeded(@NonNull LocalItem item) {
-		LocalItemType type = item.getType();
-		Long limit = rules.getCalculationSizeLimit(type);
+		Long limit = rules.getCalculationSizeLimit(item);
 		if (limit != null && !separateSizeCalculationItems.contains(item)) {
```
Chumva, line 108: Will this work if we return limit or -1?

OsmAnd/src/net/osmand/plus/download/local/LocalItemsLoaderTask.java
```diff
@@ -52,7 +53,14 @@ public Map<CategoryType, LocalCategory> loadAllFilesByCategories() {
 				.addDirectoryIfNotPresent(internalDir, false)
 				.addDirectoryIfNotPresent(externalDir, true)
 				.addForcedAddUnknownDirectory(noBackupDir)
-				.addTypeToCalculateSizeSeparately(TILES_DATA, TILES_SIZE_CALCULATION_LIMIT)
+				.setSizeLimitCondition(localItem -> {
+					if (localItem.getType() == TILES_DATA) {
+						File file = localItem.getFile();
+						boolean isSQLite = file.getName().endsWith(IndexConstants.SQLITE_EXT);
+						return !isSQLite ? TILES_SIZE_CALCULATION_LIMIT : -1;
```
Chumva, line 60: can be single line

PR #21325, opened 2024-11-13
Improve OBD data screen; removed used devices from found bounded devi…

Corwin-Kh:
…ces list

### DIFF
--- a/OsmAnd/src/net/osmand/plus/plugins/odb/dialogs/OBDDevicesSearchFragment.kt
+++ b/OsmAnd/src/net/osmand/plus/plugins/odb/dialogs/OBDDevicesSearchFragment.kt
@@ -107,7 +107,7 @@ class OBDDevicesSearchFragment : OBDDevicesBaseFragment(),
 
 	override fun onStart() {
 		super.onStart()
-		vehicleMetricsPlugin?.let { plugin ->
+		vehicleMetricsPlugin.let { plugin ->
 			val usedDevices = plugin.getUsedOBDDevicesList()
 			pairedDevicesAdapter.items = plugin.getPairedOBDDevicesList(requireActivity())
 				.filterNot { pairedDevice ->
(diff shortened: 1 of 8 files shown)


OsmAnd/src/net/osmand/plus/plugins/odb/dialogs/OBDDevicesSearchFragment.kt
```diff
@@ -107,9 +107,14 @@ class OBDDevicesSearchFragment : OBDDevicesBaseFragment(),
 
 	override fun onStart() {
 		super.onStart()
-		pairedDevicesAdapter.items =
-			vehicleMetricsPlugin?.getPairedOBDDevicesList(requireActivity()) ?: emptyList()
-		updateCurrentStateView()
+		vehicleMetricsPlugin?.let { plugin ->
```
Chumva, line 110: We could get vehicleMetricsPlugin via requirePlugin & remove "?"

OsmAnd/src/net/osmand/plus/plugins/odb/dialogs/OBDMainFragment.kt
```diff
@@ -27,10 +27,29 @@ import net.osmand.util.Algorithms
...
+	}
+
+	data class OBDDataItem(val dataType: OBDDataType, val widget: OBDComputerWidget)
+
 	private val handler = Handler(Looper.getMainLooper())
 	private val items = mutableListOf<Any>()
 
-	private lateinit var adapter: OBDMainFragmentAdapter
+	private var adapter: OBDMainFragmentAdapter? = null
 	private var progress: View? = null
```
Chumva, line 53: Why we make it nullable?

Corwin-Kh: updateNightMode() could be called before adapter is initialized

OsmAnd/src/net/osmand/plus/plugins/odb/dialogs/OBDMainFragment.kt
```diff
@@ -301,11 +321,17 @@ class OBDMainFragment : OBDDevicesBaseFragment(), VehicleMetricsPlugin.Connectio
...
 
 	override fun onForgetSensorConfirmed(deviceId: String) {
 		vehicleMetricsPlugin?.removeDeviceToUsedOBDDevicesList(deviceId)
 		view?.let { setupUI(it) }
 	}
+
+	override fun updateNightMode() {
+		super.updateNightMode()
+		adapter?.nightMode = nightMode
+		adapter?.notifyDataSetChanged()
```
Chumva, line 335: It looks somewhat strange, correct nightMode should be set in onCreateView, also theme change should force layout recreation after this [method](https://github.com/osmandapp/Osmand/blob/fecbd55b6d6cd4b30a232b8ab0e1b2ba2bceadf8/OsmAnd/src/net/osmand/plus/helpers/MapFragmentsHelper.java#L117)

PR #21330, opened 2024-11-13
change mode on click\fix formatting

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/dialogs/AddWidgetFragment.java
```diff
@@ -163,6 +163,7 @@ private void setupContent() {
 		List<WidgetType> widgets = widgetsDataHolder.getWidgetsList();
 		AidlMapWidgetWrapper aidlWidgetData = widgetsDataHolder.getAidlWidgetData();
 		if (widgets != null) {
+			widgets.sort((w1, w2) -> app.getString(w1.titleId).compareTo(app.getString(w2.titleId)));
```
Chumva, line 166: Better to use collator.compare() (OsmAndCollator.primaryCollator())

OsmAnd/src/net/osmand/plus/plugins/odb/OBDRemainingFuelWidget.kt
```diff
@@ -103,5 +115,10 @@ class OBDRemainingFuelWidget(
 				return app.getString(R.string.percent_unit)
 			}
 		}
+
+		fun next(): RemainingFuelMode {
```
Chumva, line 119: We have several enums with the same code, assuming that we slowly increase kotlin usage we could add universal methods for enums, something like this - package net.osmand.plus.utils

inline fun <reified T : Enum<T>> T.next(): T {
	val values = enumValues<T>()
	return values[(ordinal + 1) % values.size]
}

inline fun <reified T : Enum<T>> T.previous(): T {
	val values = enumValues<T>()
	return values[(ordinal - 1 + values.size) % values.size]
}

OsmAnd/src/net/osmand/plus/utils/OsmAndFormatter.java
```diff
@@ -160,15 +160,15 @@ public static String getFormattedTimeShort(long seconds, boolean useCurrentTime)
...
-	public static String getFormattedTimeShort(long seconds, boolean useCurrentTime, boolean shouldShowDay) {
+	public static String getFormattedTimeShort24(long seconds, boolean useCurrentTime, boolean shouldShowDay) {
 		Calendar calendar = Calendar.getInstance();
 		if (useCurrentTime) {
 			calendar.setTimeInMillis(System.currentTimeMillis() + seconds * 1000);
 		} else {
 			calendar.setTimeInMillis(seconds * 1000);
 		}
 		Date date = calendar.getTime();
-		String formattedTime = shortTimeFormatter.format(date, twelveHoursFormat);
```
Chumva, line 171: Won't this break something in UI?

PR #21366, opened 2024-11-18
Fix not all obd widgets added via reorder widgets fragment

0xRe1nk0:


OsmAnd/src/net/osmand/plus/plugins/odb/VehicleMetricsPlugin.kt
```diff
@@ -139,6 +139,21 @@ class VehicleMetricsPlugin(app: OsmandApplication) : OsmandPlugin(app), OBDReadS
 		val engineCoolantTempWidget: MapWidget =
 			createMapWidgetForParams(mapActivity, WidgetType.OBD_ENGINE_COOLANT_TEMP)
 		widgetsInfos.add(creator.createWidgetInfo(engineCoolantTempWidget))
+		val engineRuntimeWidget: MapWidget =
```
Chumva, line 142: Can we make widgets dynamic here based on WidgetType? add something like WidgetType.getObdTypes()

PR #21380, opened 2024-11-19
delayed switch app profile

0xRe1nk0:


OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java
```diff
@@ -446,22 +448,57 @@ public boolean switchAppModeToPrevious() {
...
-			ctx.showShortToastMessage(message);
-			return true;
+		return enabledModes.get(indexOfNext);
+	}
+
+	private final Handler delayedSwitchProfileHandler = new Handler();
+	private ApplicationMode delayedSwitchProfile;
+	private Toast delayedSwitchProfileToast;
+
+	public void delayedSwitchAppMode(boolean next) {
```
Chumva, line 477: Probably better to store this behaviour in quick action, because we need it only there

PR #21391, opened 2024-11-20
review fix

0xRe1nk0:


OsmAnd/src/net/osmand/plus/plugins/monitoring/VehicleMetricsRecordingAdapter.java
```diff
@@ -199,6 +203,13 @@ public void bindView(@NonNull VehicleMetricsRecordingCategory category) {
 			int dp36 = getDimen(app, R.dimen.showAllButtonHeight);
 			itemView.setMinimumHeight(dp36);
 			textView.setText(category.titleId);
+
+			LinearLayout.LayoutParams currentImageParams = (LinearLayout.LayoutParams) imageView.getLayoutParams();
+			int size = app.getResources().getDimensionPixelSize(R.dimen.standard_icon_size);
+			LinearLayout.LayoutParams params = new LinearLayout.LayoutParams(size, size);
```
Chumva, line 209: We could reuse same params, just update it

OsmAnd/src/net/osmand/plus/plugins/monitoring/MonitoringSettingsFragment.java
```diff
@@ -272,12 +272,18 @@ private void setupShowTripRecNotificationPref() {
 
 	public void setupObdRecordingPref() {
 		Preference preference = findPreference(RECORD_OBD_DATA);
+
+		@ColorRes int iconColor = isNightMode() ? R.color.icon_color_default_light : R.color.icon_color_default_dark;
```
Chumva, line 276: its ok, but annotations in local variables make code harder to read (if there are many of them)

PR #21399, opened 2024-11-20
Fix AA navigation restored started on the phone

vshcherb:


OsmAnd/src/net/osmand/plus/auto/NavigationSession.java
```diff
@@ -415,17 +419,23 @@ public boolean requestLocationNavigation() {
...
+			Screen top = screenManager.getTop();
+			if (top instanceof NavigationScreen) {
+				return;
+			}
+		}
 		if (navigationScreen == null) {
 			navigationScreen = new NavigationScreen(getCarContext(), settingsAction, this);
 			navigationCarSurface.setCallback(navigationScreen);
 		}
+		startCarNavigation();
```
vshcherb, line 434: Is it correct?

Issue #21362, opened 2024-11-17
Fix shields & destination in widgets (Android & iOS)


vshcherb:
- [ ] Duplicate shields (from destination:ref and ref) are displayed
- [ ] Destination:ref causes empty commas in Destination. 

Solution: destination:ref should be combined with ref and used a common united set.

Should be tested with:
- Current Street Widget 
- AA / CarPlay Widget 
- Next turn Widget 
- Route Details
- Voice instruction

----------
`4.8.4`
`OsmAnd~ 5.0.0#3746m, Released: 2024-11-18`

[OsmAnd](https://osmand.net/map/search/#17/51.28857/22.44057)
[OSM](https://www.openstreetmap.org/search?query=51.2871697317902%2C%2022.444053620061002#map=17/51.287104/22.445229)
`51.2871697317902, 22.444053620061002`

Duplicate names in Street name widget - Shields duplicate with names. 51.29146 22.43126

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/3c8cac14-bdbb-4059-b367-49f2a99f03b1"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/827003fc-cfc8-4b9b-b8cd-4bf3406a5ca2"  width="310" height="auto"/> </th>
  </tr>
</table>

Zirochkabila:
iOS **4.8.4** - same
[OsmAnd](https://osmand.net/map/search/#17/51.28857/22.44057)
`51.2871697317902, 22.444053620061002`


<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/827003fc-cfc8-4b9b-b8cd-4bf3406a5ca2"  width="300" height="auto"/> </th>
  </tr>
</table>

ivanPyrohivskyi:
Problem with UI in iOS.
Show only one shield (12) instead three (17, S17, 12)
In [code](https://github.com/osmandapp/OsmAnd-iOS/blob/a1f620175653d4dabe843ecb8b3e27eac8fa1c97/Sources/Controllers/Map/Widgets/OATopTextView.mm#L669) shields are set three times, but show only once.
<img width="300" src="https://github.com/user-attachments/assets/a97196d3-c794-448d-bfcf-825fc791a48d" />

vshcherb:
![image](https://github.com/user-attachments/assets/b7807317-e0ee-43b4-99cb-42ece0fdd247)

[...]

PR #21413, opened 2024-11-21
Avoid show ref is shields was created

ivanPyrohivskyi:
Avoid to create street name based on _ref_ if shields was already created. In most cases shields already based on the _ref_ in different combinations.

To issue https://github.com/osmandapp/OsmAnd/issues/21362

### DIFF
--- a/OsmAnd/src/net/osmand/plus/routing/CurrentStreetName.java
+++ b/OsmAnd/src/net/osmand/plus/routing/CurrentStreetName.java
@@ -62,14 +62,14 @@ public class CurrentStreetName {
 			RouteSegmentResult rs = routingHelper.getCurrentSegmentResult();
 			if (rs != null) {
 				text = getRouteSegmentStreetName(routingHelper, rs, false);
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/routing/CurrentStreetName.java
```diff
@@ -62,14 +62,12 @@ public void setupCurrentName(@NonNull RoutingHelper routingHelper, @NonNull Next
...
-				if (Algorithms.isEmpty(text)) {
-					text = getRouteSegmentStreetName(routingHelper, rs, true);
-					isSet = !Algorithms.isEmpty(text);
-				} else {
-					isSet = true;
-				}
 				showMarker = true;
 				shields = RoadShield.create(rs.getObject());
+				if (Algorithms.isEmpty(text) && shields.isEmpty()) {
+					text = getRouteSegmentStreetName(routingHelper, rs, true);
```
vshcherb, line 68: looks like streets completely ignored

PR #21421, opened 2024-11-22
show aa navigation screen when starting navigation from phone; remove…

Corwin-Kh:
…d debug logs


OsmAnd/src/net/osmand/plus/auto/NavigationSession.java
```diff
@@ -468,7 +467,12 @@ public void updateLocation(Location location) {
 
 	@Override
 	public void newRouteIsCalculated(boolean newRoute, ValueHolder<Boolean> showToast) {
-		showRoutePreview();
+		if (routingHelper.isFollowingMode() && routingHelper.isRouteCalculated()) {
+			startNavigationScreen();
+			updateCarNavigation(getApp().getLocationProvider().getLastKnownLocation());
```
vshcherb, line 472: minor - could be inside startNavigationScreen.

PR #21425, opened 2024-11-23
 New Module for smartwatches display #18502 

RomaricKc1:
From #18502, I decided to make the app usable for `wear os` devices. 

I've been stalling quite a bit, the project is quite big and I don't have a lot of exp in android dev but here it is. 

A **heads-up**: There's a tons of changes add/edit. Mostly adds. See, I think it's possible to have a separate `res` files when targeting `wear os` devices, but I went the easiest way. A new module called `OsmAndWear`. 

This is not a bait. I known there's ` 6,316 files changed` but
It's basically a clone of the `OsmAnd` from  [54f673d](https://github.com/osmandapp/OsmAnd/commit/54f673dc3d7da02038fd198b3802580a7dde6f45), with the res folder edited. 
One can use some `diff` tool to check that.

Of course, this doesn't fix all the layout. I am sure there are some ui stuff that aren't usable on the watches yet. I believe this pr is the right way to go, in order to make the dev more easier once it's in the pipeline. Having all the src files and stuff, to only focus on res.
Again, I am not sure that the new module is necessary, maybe it can be done without copying almost line by line the other `src` etc. from the main `OsmAnd`.

Also, an important thing: I had `disabled swipe to dismiss` in my previous test, to make it easier to navigate on the map. But using the same directives in the `theme` `values>styles.xml` in the current state of the app doesn't work. I have "duplicated res error".

That being said, can't wait to hear from you whether this is a welcome change or not.
Thanks


vshcherb:
Many thanks for pushing smartwatch display forward.

Though  we can't merge this pull request as is. We need to investigate whether we need to create common library or incorporate changes in OsmAnd. We plan to do it in Q4 2024 - Q1 2025.

RomaricKc1:
Thank you for the update! 

Looking forward for it.

seregamuravyov:
Any updates on WearOS?

sfields-forever:
> We need to investigate whether we need to create common library or incorporate changes in OsmAnd. We plan to do it in Q4 2024 - Q1 2025.

What is the result of the investigation? It is such a shame that OsmAnd does not work on WearOS.

iCaitlyn:
> Many thanks for pushing smartwatch display forward.
> 
> Though we can't merge this pull request as is. We need to investigate whether we need to create common library or incorporate changes in OsmAnd. We plan to do it in Q4 2024 - Q1 2025.

It's now Q2 2026, any changes so far?

wombatpandaa:
I just got a Pixel Watch 3 and would love to use osmand on it instead of Google Maps!

PR #21434, opened 2024-11-25
Fixed crash in track reload

Corwin-Kh:


OsmAnd/src/net/osmand/plus/shared/SharedUtil.kt
```diff
@@ -249,7 +249,7 @@ object SharedUtil {
 		jSegment.name = segment.name
 		jSegment.generalSegment = segment.generalSegment
 		val jPoints: MutableList<GPXUtilities.WptPt> = ArrayList()
-		for (point in segment.points) {
+		segment.points.forEach { point ->
```
Chumva, line 252: Looks strange for concurrentmodificationexception fix, is it really working?

PR #21459, opened 2024-11-27
Volume unit behavior as units of length

0xRe1nk0:


OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java
```diff
@@ -1047,6 +1047,7 @@ public boolean setValue(Object prefs, DrivingRegion val) {
 			boolean overrideMetricSystem = !DRIVING_REGION_AUTOMATIC.getValue(prefs, DRIVING_REGION_AUTOMATIC.getDefaultValue());
 			if (overrideMetricSystem && val != null) {
 				METRIC_SYSTEM.setValue(prefs, val.defMetrics);
+				UNIT_OF_VOLUME.set(val.volumeUnit);
```
vshcherb, line 1050: Incorrect, driving region should override only default value. If the value is set by user it should never be overridden

Chumva: Do we still need to change the preference for METRIC_SYSTEM here, or is it fine to remove it?

Issue #21419, opened 2024-11-22
Freshly installed OsmAnd is confusing new users


readonly24:
### Describe the idea

When you install OsmAnd for the first time, you end up with an empty screen and just nothing happens. For first time users this is very confusing.

### Expected behaviour

After OsmAnd was installed

1. Ask for GPS permission. 
2. Set the zoom level, that one actually can see something of the basic world map.
3. Ask for downloading the more detailed world map, show a toast message that the map is being downloaded in the backgroud. After download has finished load the map and display it.
4. If GPS is fixed, ask for download the detailed map of the current location, load the map and display it.

### Alternatives you've considered

None

### Context

Demo video

https://github.com/user-attachments/assets/dcbd370b-8458-42b6-83d2-8f839537085f

yuriiurshuliak:
We appreciate your feedback and will review these ideas to improve the onboarding experience for new users.

vshcherb:
We need to understand what location by default is displayed, probably it should be worldwide overview and not a blank screen? Like (0,0) - zoom 5

dmpr0:
> Ask for GPS permission.
OsmAnd request GPS when it needed. 

> Ask for downloading the more detailed world map
Most users doesn't need these maps, basic is included with the app

> Ask for downloading the more detailed world map
Use declined, and _end up with an empty screen and just nothing happens._

DmitryAlexei:
Probably we can show online map at a first start?

Nine-Tailed:
> Probably we can show online map at a first start?

Then users will remain on the online map without figuring out how to switch to the offline map ;)

readonly24:
> OsmAnd requests GPS access only when it is needed.
Yes, but I think this should be the first question, because I think almost every new user wants to use GPS.

> Most users don't need these maps; the basic version is included with the app.
That's ok, but set the zoom level, that one actually can see something of the basic world map.

There are not only technically skilled users, so I think they should have a more user friendly start.

readonly24:
If I start from myself, when I install a navigation app, I expect to see a map and my location or so right after the first run, but for sure not just an blank screen.

vshcherb:
Is it reproduced? @DmitryAlexei - it's still in reproducing stage to figure out if there are other conditions involved

vshcherb:
Intitial zoom should be 7.

Zirochkabila:
`OsmAnd~ 5.0.0#3830m, Released: 2024-11-28`

[...]

PR #21464, opened 2024-11-27
fix set zoom depending on map download state #21419

0xRe1nk0:


OsmAnd/src/net/osmand/plus/firstusage/FirstUsageWizardFragment.java
```diff
@@ -627,8 +628,19 @@ public void showAppropriateWizard(FragmentActivity activity, boolean updateWizar
 		}
 	}
 
+	private void setProperZoom(){
+		int zoom;
+		if ((wizardType == WizardType.MAP_DOWNLOAD || wizardType == WizardType.MAP_DOWNLOADED) && mapIndexItem != null) {
+			zoom = DOWNLOAD_MAP_ZOOM_LEVEL;
+		} else {
+			zoom = NO_MAP_ZOOM_LEVEL;
+		}
```
Chumva, line 637: Probably we should also check for app.getResourceManager().isAnyMapInstalled()?

PR #21484, opened 2024-11-29
Show first activity in gpx export

0xRe1nk0:


OsmAnd/src/net/osmand/plus/plugins/osmedit/dialogs/SendGpxBottomSheetFragment.java
```diff
@@ -92,9 +95,25 @@ private void setupTagsRow(@NonNull View view) {
 		}
 	}
 
+	@Nullable
+	private String getFirstActivity() {
+		String activity = null;
+		if (!Algorithms.isEmpty(files)) {
```
Chumva, line 101: Probably we need to find first available activity

PR #21490, opened 2024-11-29
Removed <50

Corwin-Kh:


OsmAnd/src/net/osmand/plus/plugins/odb/VehicleMetricsPlugin.kt
```diff
@@ -738,7 +738,7 @@ class VehicleMetricsPlugin(app: OsmandApplication) : OsmandPlugin(app), OBDReadS
 			var isNan = (data is Float) && data.isNaN()
 			isNan = isNan || (data is Double) && data.isNaN()
 			if (isNan) {
-				return if (computerWidget.type == OBDDataComputer.OBDTypeWidget.FUEL_LEFT_KM) ">50" else "<50"
+				return "-"
```
vshcherb, line 741: 1. Where is first computation? we also should display - on first tick
2. where is check for > 100

Issue #21424, opened 2024-11-23
Short way routing - weird unnecessary turns - not enough penalty for turns


abclop99:
### Routing engine

- [X] OsmAnd's in-app offline routing
- [X] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile

Routing profile: Car
Restriction: None?

Enabled:
- Fuel-efficient way
- Consider temporary limitations
- Minimal distance to recalculate route: 120m
- In case of reverse direction

### Start and end points

An example route on [openstreetmap.org](https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=45.290465%2C-75.694099%3B45.292685%2C-75.693967#map=18/45.291566/-75.694031).

Example route on [osmand.net](https://osmand.net/map/navigate/?start=45.290935,-75.693934&end=45.292460,-75.694111&profile=car&params=car,short_way#19/45.29178/-75.69407)

Any route that goes through that intersection in that direction has the same problem.

### Actual route

The actual route turns onto Leitrim Road then turns immediately back onto River Road.

![Screenshot of actual route in the app](https://github.com/user-attachments/assets/68ccb37f-c8aa-4780-b4ab-eb411691112e)

It also shows up on [osmand.net](https://osmand.net/map/navigate/?start=45.290935,-75.693934&end=45.292460,-75.694111&profile=car&params=car,short_way#19/45.29178/-75.69407) with "Short way" enabled

![image](https://github.com/user-attachments/assets/aea9f8aa-a9a4-49b2-bd65-8d695bce02b3)

### Expected route

The route should go straight through the intersection along River Road as on [openstreetmap.org](https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=45.290465%2C-75.694099%3B45.292685%2C-75.693967#map=18/45.291566/-75.694031).

Also works properly on [osmand.net](https://osmand.net/map/navigate/?start=45.290935,-75.693934&end=45.292460,-75.694111&profile=car#18/45.29151/-75.69318) without "Short way" enabled.

![image](https://github.com/user-attachments/assets/1433fc17-9f1f-4d91-8c7d-046b53a97847)

### Maps you used (online or offline)

- [x] Offline maps offered within the OsmAnd app for download. 
  - 2024-11-23 10:30
- [x] Online (tile / raster) maps 
[...]

PR #21491, opened 2024-11-29
Get back maxRouteLength for RouteTestingTest (required for tests 49,63)

RZR-UA:
This reverts commit 6ce98f8e074a5247843676d5738ff028803d569d.

### DIFF
--- a/OsmAnd-java/src/test/java/net/osmand/router/RouteTestingTest.java
+++ b/OsmAnd-java/src/test/java/net/osmand/router/RouteTestingTest.java
@@ -167,7 +167,10 @@ public class RouteTestingTest {
 				System.out.println("This is test on hanging routing");
 				break;
 			}
+
+			checkRouteLength(params, routeSegments);
 			checkRoutingTime(ctx, params);
+
 			for (Entry<String, String> es : expectedResults.entrySet()) {
 				long id = RouterUtilTest.getRoadId(es.getKey());
 				switch (es.getValue()) {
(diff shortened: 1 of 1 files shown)


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21424

vshcherb:
Usually it's unnecessary if we can specify which segments are not allowed cause distance is highly dependent on first/last segment which might be added sometimes and logic check behind, so distance fluctuates over time where geometry coudln't

RZR-UA:
When we do min/max check, exact distance value is not important and fluctuation over time is OK.

Just check the old issue:

https://www.openstreetmap.org/way/508244839

In both routes (with/without the bug) the way 508244839 is present and exact segment is present too.

We could check an absence of other segments, but currently we do not have this feature in tests.

A presence check exist:

```
"expectedResults": {
  "4692930": "3"
}
```

An absence check as it could be implemented:

```
"expectedResults": {
  "4692930": "-1",
  "4692930": "-2",
  "4692930": "-3"
}
```

At the moment, since removing of `maxRouteLength` broke tests 49 and 63, the `maxRouteLength` should have been restored for compatibility reasons.

vshcherb:
I think adding non-present segments more important than length

RZR-UA:
https://github.com/osmandapp/OsmAnd/pull/21503/files
https://github.com/osmandapp/OsmAnd-resources/pull/1164/files

PR #21503, opened 2024-12-02
Implement present/absent points check in the routing test

RZR-UA:


RZR-UA:

ivanPyrohivskyi (approved):

PR #21510, opened 2024-12-03
Introduce turn penalties for short_way=yes routing

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21424

RZR-UA:
- [ ] C++ transfer

PR #21539, opened 2024-12-06
[Kotlin jvmMain] Add forgotten override fun write(str: String?)

RZR-UA:


RZR-UA:
approved by @alex-osm

Issue #18924, opened 2024-01-16
US Addresses - not found (Tiger data)


davew99:
### Describe the idea (required)

OsmAnd should accept a standard address format when I want to search for an address.  Such as Street number, Street Name, City, State, zip code.  Along with this, OsmAnd should give me the option as whether to do an online or offline search.

### Tell us about the expected behaviour (required)

On my android phone if I try to use a standard address (Street number, street name, city, state) link, and I try to use OsmAnd to search for and display the address the search generally fails.  On rare occasions, OsmAnd does find the address when using a standard address link, but very rarely.  Usually, it may find the street, but not the address and only suggests a cross street.  The rest of the time OsmAnd can't even find the street.  The online search option is not much better than the offline search.  Plus, the online search I must copy and paste the address into OsmAnd, I cannot use a link.  Manually entering an address into OsmAnd yields no better results than when using a link.

The behavior that I would expect is that when I use a standard address link (or manually enter a standard address) that OsmAnd would be able use the standard address format to find the address and display it and then allow a navigation to that address.  It would be great if OsmAnd could do this offline, but online would be fine as well.  But, if offline is not possible, the OsmAnd should either ask if would like to go online and or OsmAnd should have a menu option as whether offline or online is the default address search.

### Tell us about alternatives you've considered (required)

The only alternative I have is to use Google maps.  I would much rather use OsmAnd, but it is totally unreliable at finding and address.

### Context (optional)

No.

goldfndr:
Related: #6824

yuriiurshuliak:
To ensure the most accurate results, we recommend using the Online Search feature in the app. To use Online Search, please refer to this link: https://osmand.net/docs/user/search/search-poi#online-search
 
Additionally, we would like to highlight that the OsmAnd app functions primarily offline, which means the address search process may have its own specificities. To perform a search using Search, follow this sequence: Select city > Select street > Postcode. Here is the user guide on address search: https://download.osmand.net/docs/user/search/search-address/

If this does not help, keep in mind that we use an address database, and in your situation, this usually means that the address is not in the database.

Also please clarify what you mean by standard address link?
And provide us with specific coordinates or addresses of objects that you cannot find.

davew99:
Thread? #6824 is expressing the same issue that I am complaining about.  I did not see any acceptance in that thread that this is an issue?  Let alone any resolve?

Getting to the online search is overly convoluted and does not address my primary issue.

[...]

PR #21567, opened 2024-12-13
Process tiger

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/18924

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/osm/edit/Entity.java
+++ b/OsmAnd-java/src/main/java/net/osmand/osm/edit/Entity.java
@@ -14,7 +14,7 @@ import java.util.Set;
 
 
 public abstract class Entity implements Serializable {
-	public enum EntityType {
+    public enum EntityType {
 		NODE,
 		WAY,
 		RELATION,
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/osm/edit/Entity.java
```diff
@@ -110,6 +110,7 @@ public boolean equals(Object obj) {
 	private int version;
 	private double latitude;
 	private double longitude;
+	private boolean isTiger;
```
vshcherb, line 113: Generic class for isTiger flag is not suitable

OsmAnd-java/src/main/java/net/osmand/osm/edit/Entity.java
```diff
@@ -200,6 +201,9 @@ public String putTagNoLC(String key, String value) {
 		if (tags == null) {
 			tags = new LinkedHashMap<String, String>();
 		}
+		if (key.equals("tiger:osmand") && value.equals("yes")) {
```
vshcherb, line 204: Constant for tag is needed

Issue #3533, opened 2017-02-27
Bring back search by intersecting street


TimMagee:
Because of OpenStreetMap's often limited address data, it was very nice to be able to search by intersecting streets in previous versions of OsmAnd. However, that functionality was entirely removed in OsmAnd 2.5. Is there any plan to bring back that feature?

vshcherb:
It is available if you select a street and start typing street name you will see it.

PR #21584, opened 2024-12-18
#3533: Fixed getting track analysis; fixed display track info accordi…

Corwin-Kh:
…ng to join segments setting

### DIFF
--- a/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
+++ b/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
@@ -24,7 +24,6 @@ class GpxTrackAnalysis {
 	}
 
 	var name: String? = null
-	var totalDistanceWithoutGaps = 0f
 	var timeSpanWithoutGaps: Long = 0
 	var expectedRouteDuration: Long = 0
 	var timeMovingWithoutGaps: Long = 0
@@ -169,6 +168,10 @@ class GpxTrackAnalysis {
 		get() = (getGpxParameter(GpxParameter.TOTAL_DISTANCE) as Double).toFloat()
 		set(value) = setGpxParameter(GpxParameter.TOTAL_DISTANCE, value.toDouble())
 
+	var totalDistanceWithoutGaps: Float
+		get() = (getGpxParameter(GpxParameter.TOTAL_DISTANCE_WITHOUT_GAPS) as Double).toFloat()
+		set(value) = setGpxParameter(GpxParameter.TOTAL_DISTANCE_WITHOUT_GAPS, value.toDouble())
+
 	fun isTimeSpecified(): Boolean {
 		val startTime = startTime
 		val endTime = endTime
@@ -295,6 +298,7 @@ class GpxTrackAnalysis {
 		var totalSensorCadenceSum = 0.0
 
 		var _totalDistance = 0.0f
+		var _totalDistanceWithoutGaps = 0.0f
 
 		points = 0
 
@@ -402,7 +406,7 @@ class GpxTrackAnalysis {
 						distanceMovingOfSingleSegment = 0f
 					}
 					if (point.lastPoint) {
-						totalDistanceWithoutGaps += distanceOfSingleSegment
+						_totalDistanceWithoutGaps += distanceOfSingleSegment
 						timeMovingWithoutGaps += timeMovingOfSingleSegment
 						totalDistanceMovingWithoutGaps += distanceMovingOfSingleSegment
 					}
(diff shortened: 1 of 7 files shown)


OsmAnd/src/net/osmand/plus/track/GpxBlockStatisticsBuilder.java
```diff
@@ -173,7 +174,6 @@ public void initItems(@Nullable GpxTrackAnalysis initAnalysis) {
 				withoutGaps = !selectedGpxFile.isJoinSegments()
 						&& (Algorithms.isEmpty(currentGpx.getTracks()) || currentGpx.getTracks().get(0).isGeneralTrack());
 			} else {
-				GpxDisplayItem displayItem = getDisplayItem();
 				if (displayItem != null) {
 					analysis = displayItem.analysis;
 					withoutGaps = !selectedGpxFile.isJoinSegments() && displayItem.isGeneralTrack();
```
Chumva, line 179: Looks like copy-pasted logic

OsmAnd/src/net/osmand/plus/configmap/tracks/viewholders/TrackViewHolder.java
```diff
@@ -252,20 +253,27 @@ private void appendDistanceDescription(@NonNull SpannableStringBuilder builder,
 		appendFolderName(builder, trackItem, shouldShowFolder);
 	}
 
+	private float getDistance(@NonNull TrackItem trackItem, @NonNull GpxTrackAnalysis analysis) {
+		GpxDataItem dataItem = trackItem.getDataItem();
+		boolean joinSegments = dataItem != null && (boolean) dataItem.getParameter(GpxParameter.JOIN_SEGMENTS);
+		return joinSegments ? analysis.getTotalDistance() : analysis.getTotalDistanceWithoutGaps();
+	}
```
Chumva, line 260: We may need the same info in other places, we could add a universal method to GpxTrackAnalysis or TrackItem instead of UI

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
```diff
@@ -24,7 +24,6 @@ class GpxTrackAnalysis {
 	}
 
 	var name: String? = null
-	var totalDistanceWithoutGaps = 0f
 	var timeSpanWithoutGaps: Long = 0
 	var expectedRouteDuration: Long = 0
 	var timeMovingWithoutGaps: Long = 0
```
Chumva, line 29: Do we have the same issue with other variables here?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
```diff
@@ -446,8 +450,11 @@ class GpxTrackAnalysis {
 			}
 			processElevationDiff(s)
 		}
-
+		totalDistanceWithoutGaps = _totalDistanceWithoutGaps
 		totalDistance = _totalDistance
+		if(splitSegments.size == 1 && !splitSegments[0].segment.isGeneralSegment()) {
+			totalDistanceWithoutGaps = totalDistance
+		}
```
Chumva, line 457: It looks strange. Is it needed because there are different values in totalDistance and totalDistanceWithoutGaps when we do not have multiple segments?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
```diff
@@ -446,8 +450,11 @@ class GpxTrackAnalysis {
 			}
 			processElevationDiff(s)
 		}
-
+		totalDistanceWithoutGaps = _totalDistanceWithoutGaps
 		totalDistance = _totalDistance
+		if(splitSegments.size == 1 && !splitSegments[0].segment.isGeneralSegment()) {
```
alex-osm, line 455: Formatting

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
Corwin-Kh: yes. in case when we have only one segment we don't have "general segment" and totalDistanceWithoutGaps is not calculated, but it can be referenced in code

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
Corwin-Kh: have not faced such problems with others.

PR #21585, opened 2024-12-18
TravelGpx: restore localized names for tracks and points

RZR-UA:


RZR-UA:

OsmAnd/src/net/osmand/plus/wikivoyage/data/TravelGpx.java
```diff
@@ -77,9 +77,13 @@ public GpxTrackAnalysis getAnalysis() {
 	@Override
 	public WptPt createWptPt(@NonNull Amenity amenity, @Nullable String lang) {
 		WptPt wptPt = new WptPt();
+		wptPt.setName(amenity.getName());
 		wptPt.setLat(amenity.getLocation().getLatitude());
 		wptPt.setLon(amenity.getLocation().getLongitude());
-		wptPt.setName(amenity.getName());
+
+		Map<String, String> wptPtExtensions = wptPt.getExtensionsToWrite();
+		amenity.getNamesMap(true).forEach((key, value) -> wptPtExtensions.put("name:" + key, value));
```
Chumva, line 85: Do we need to check also here for "ref" && "sym"?

RZR-UA: The tags name:ref and name:sym are synthetic and used only for Ways (not for WptPt).

PR #21591, opened 2024-12-19
OsmAnd-shared: support "text" inside "link" in Metadata, WptPt and Author (class Link)

RZR-UA:


RZR-UA:

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
```diff
@@ -1164,6 +1179,11 @@ object GpxUtilities {
 									}
 
 									"link" -> parse.link = parser.getAttributeValue("", "href")
+									"text" -> {
+										if (insideTagDepth["link"]!! > 0) {
```
alex-osm, line 1183: Introduce class `Link` instead of `insideTagDepth`

RZR-UA: done

PR #21595, opened 2024-12-20
Fix OsmRouteType for route_piste (OSM ski routes)

RZR-UA:


RZR-UA:

Issue #19787, opened 2024-05-06
Display GPX image / desc  


vshcherb:
### Default GPX with image / desc
- [x] Desc should be written and read from https://www.topografix.com/GPX/1/1/ - metadataType - desc (now Wikivoyage saves to wrong extensions:osmand:desc)
- [x] gpx track: If link contains image it should be displayed in Description
- [x] gpx waypoints: If link contains image it should be displayed in Description
- [x] AmenityMenuBuilder: process description, url/link-image in a similar way to waypoints
- [x] No additional should be displayed (what's additional? duplicated description removed)

### Wikivoyage (note: no changes - it all seems OK initially)

- [x] Wikivoyage gpx should contain special id in extension to wikivoyage title, lang. By this GPX "Read full" should open Wikivoyage Article or suggest to download
- [x] If this special id not present, GPX should be treated as default gpx with image desc
- [x] Wikivoyage GPX should contain short description (non-html) - consult Backend team

### Dev Notes (there was a discord session with @ivanPyrohivskyi )

- [x] fix http->https and 404 URLs in XML schema in GPX header
- [x] fix unsupported `<text>` in `<link>` for waypoints
- [x] fix HTML description display for Wikivoyage articles

Zirochkabila:
# Testing 1

`OsmAnd~ 5.0.0#4062m, Released: 2025-01-06` - Fixed

Data for testing 
[test-html-and-link.zip](https://github.com/user-attachments/files/18321028/test-html-and-link.zip)

PR #21609, opened 2024-12-23
Fix 404 and non-https links in GPX XML header

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19787

PR #21616, opened 2024-12-25
Fix map shifts on change orientation in download tiles fragment

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
```diff
@@ -1828,6 +1828,40 @@ public void fitLocationToMap(double clat, double clon, int zoom,
 		}
 	}
 
+	public void fitLocationToMap(double clat, double clon, int zoom,
```
Chumva, line 1831: Copy-pasted method?

OsmAnd/src/net/osmand/plus/plugins/rastermaps/DownloadTilesFragment.java
```diff
@@ -134,6 +135,13 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
 			selectedMinZoom = savedInstanceState.getInt(KEY_SELECTED_MIN_ZOOM);
 			selectedMaxZoom = savedInstanceState.getInt(KEY_SELECTED_MAX_ZOOM);
 			downloadType = AndroidUtils.getSerializable(savedInstanceState, KEY_DOWNLOAD_TYPE, DownloadType.class);
+			if (savedInstanceState.containsKey(KEY_LAT_LON)) {
+				LatLon savedLatLon = AndroidUtils.getSerializable(savedInstanceState, KEY_LAT_LON, LatLon.class);
+				if (savedLatLon != null) {
+					lat = savedLatLon.getLatitude();
```
Chumva, line 141: We could store LatLon in class, instead of two Double

PR #21630, opened 2024-12-26
Fix "Extra option lost for custom rendering style"

nazar-kutz:


OsmAnd-java/src/main/java/net/osmand/render/RenderingRulesStorage.java
```diff
@@ -410,7 +410,7 @@ public void startElement(Map<String, String> attrsMap, String name) throws XmlPu
 				if (attrsMap.get("possibleValues") != null) {
 					prop.setPossibleValues(attrsMap.get("possibleValues").split(","));
 				}
-				PROPS.registerRule(prop);
+				PROPS.registerRule(prop, !addon);
```
alex-osm, line 413: Not clear - why we should not overwrite rule in case of addon?

Issue #21577, opened 2024-12-16
Brand search results are duplicated after increasing the search radius


DmitryAlexei:
### Description

When user searches for a brand and increases search radius, multiple Brand search labels appear on top of the search results

### Steps to reproduce

- install the latest nightly build
- go to 50.44305 30.52311 and 
- download the latest Kyiv map
- search for "Ciльпо" and don't leave search screen
- scroll down the search results and tap "increase search radius" button
- see how many brand search labels are available

### Actual result

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/d2070fc6-c827-46c6-9fa8-38fc4a86e1e1"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/75248a7b-d155-4d91-b927-83a53e05ebf0"  width="300" height="auto"/> </td>
  </tr>
</table>

### Expected result

- [ ] Only one brand search is available after the search radius was increased

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version:
Android/iOS version:
Device model:
Crash-Logs: ?
```

RZR-UA:
The same duplication problem happens when radius increases (try search "Hiking" routes - new category in TravelGpx task). Video:

https://github.com/user-attachments/assets/8d7eb045-40f1-4252-bc01-99a3da8161c6

Zirochkabila:
# Testing 1
`OsmAnd~ 5.0.0#4259m, Released: 2025-01-29`
`4.9.5.4958`
Fixed both examples

dmpr0:
# Review


<details>
    <summary>iOS</summary>
    <table style="width:100%">
      <tr>
        <th>сільпо</th>
        <th>silpo</th>
      </tr>
      <tr>
        <td><img src="https://github.com/user-attachments/assets/951b48e5-95aa-4a50-85b6-ceaf71ab0649"  width="300" height="auto"/> </td>
        <td><img src="https://github.com/user-attachments/assets/ebbc08e2-3400-4ab6-b121-df346434a56c"  width="300" height="auto"/> </td>
      </tr>
    </table>   
</details>   

<details>
    <summary>Android</summary>
    <table style="width:100%">
      <tr>
        <th>silpo</th>
        <th></th>
      </tr>
      <tr>
[...]

PR #21701, opened 2025-01-10
Fix duplications in search

ivanPyrohivskyi:
To issue [Brand search results are duplicated after increasing the search radius](https://github.com/osmandapp/OsmAnd/issues/21577)


OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
```diff
@@ -304,7 +304,7 @@ public boolean sameSearchResult(SearchResult r1, SearchResult r2) {
 					return MapUtils.getDistance(r1.location, r2.location) < similarityRadius;
 				}
 			} else if (r1.object != null && r2.object != null) {
-				return r1.object == r2.object;
+				return r1.object.equals(r2.object);
```
vshcherb, line 307: it's comparision for same object not for similar object, so this is incorrect and we don't support equals on all types

Issue #21692, opened 2025-01-09
Colors in <trk> not displayed since 4.9.10


yuriiurshuliak:
### Discussed in https://github.com/osmandapp/OsmAnd/discussions/21690

**Description:**
In OsmAnd 4.9.10, colors assigned to `<trk>` elements in GPX files are not displayed correctly. This issue was not present in version 4.8.6. While colors for `<wpt>` elements remain functional, track colors default to a single color instead of their assigned values.

**Steps to Reproduce:**
1. Create a GPX file with multiple tracks (`<trk>`) and assign different colors using the `<osmand:color>` tag within the `<extensions>` section.
2. Open the GPX file in OsmAnd 4.8.6 and confirm that different colors are correctly displayed for each track.
3. Open the same GPX file in OsmAnd 4.9.10.
4. Observe that only one color is applied to all tracks, regardless of their individual color settings.
5. Verify that colors in `<wpt>` elements are still displayed correctly.

**Actual Behavior:**
All tracks appear in a single color, despite being assigned different colors in the GPX file. Waypoint (`<wpt>`) colors continue to function correctly.

**Expected Behavior:**
Each track (`<trk>`) should retain its assigned color as defined in the GPX file.

**Environment:**
- **OsmAnd Version:** 4.9.10 (F-Droid)
- **Last Known Working Version:** 4.8.6
- **Device:** [Specify device model]
- **OS:** [Specify OS version]

```
...
<wpt lat..lon>
  <name>...
  <desc>...
   <extensions>
     <show_arrows>true</show_arrows>
     <osmand:color>#FFff0000</osmand:color>
     <osmand:icon>special_trekking</osmand:icon>
     <osmand:background>square</osmand:background>
  </extensions>
</wpt>
<wpt..>
...
...
<trk>
  <name>...
  <desc>...
  <extensions>
    <show_arrows>true</show_arrows>
    <osmand:color>#59ff0000</osmand:color>
  </extensions>
 <trkseg>
  ....
 </trkseg>
...
</trk>

<trk>
...
...
```

yuriiurshuliak:
The bug can be **partially reproduced**. Here are the steps to verify it:  

1. **Download and install** OsmAnd **4.8.6** (older version).  
2. **Import the provided GPX file**  [test_tracks.zip](https://github.com/user-attachments/files/18374912/test_tracks.zip)
3. **Verify that different colors are displayed** for the three tracks when importing.  
[...]

PR #21711, opened 2025-01-12
Fix broken multisegment colors (incorrect copyExtensions target in GpxFile clone methods), add color for trkseg

RZR-UA:


RZR-UA:

PR #21713, opened 2025-01-13
[4.9] fix broken colors (incorrect target of copyExtensions / GpxFile clone)

RZR-UA:


RZR-UA:

Issue #11709, opened 2021-05-19
Enhanced Arabic Search Functionality: Addressing Diacritics, Character Variants, and Other Search-Related Improvements


abdullahO2:
## Subject: Arabic Search Issue: Diacritics and Variant Characters Preventing Accurate Results

Hello OsmAnd Developers,

I am writing to report a significant issue impacting Arabic search functionality within OsmAnd on all Android/iOS versions.  The current search implementation appears to be overly sensitive to Arabic diacritics (tashkeel), hamza variations, and other orthographic nuances, preventing users from finding desired locations when searching without these markings.

**Problem Description:**

Arabic script utilizes diacritics (fatḥa َ, kasra ِ, ḍamma ُ, shadda ّ, sukun ْ), hamza variations (أ، إ، ئ، ؤ), tanween ( ٌ  ً ٍ ), alif madda (آ), and kashida (ـ) to indicate pronunciation and sometimes disambiguation. While important for proper Arabic grammar, these markings are often omitted in everyday searches.  OsmAnd's current search fails to match entries containing these markings when a search term omits them.

**Reproducing the Issue:**

1. **Example 1:**  Navigate to the following OpenStreetMap node: [https://www.openstreetmap.org/node/8663596893](https://www.openstreetmap.org/node/8663596893)

    * Searching for "الخُشيَّش" (with diacritics) correctly returns the location.
    * Searching for "الخشيش" (without diacritics) fails to return the location.

2. **Example 2:** Navigate to the following OpenStreetMap node: [https://www.openstreetmap.org/node/8613191675](https://www.openstreetmap.org/node/8613191675)

    * Searching for "العَاساء" (with diacritics) correctly returns the location.
    * Searching for "العاساء" (without diacritics) fails to return the location. (Note: The example seems to be the same with and without diacritics - please provide an example where removing the diacritics causes the search to fail).

**Proposed Solution:**

* **I recommend normalizing both the indexed location names and user search queries using the following function:**

```java
import java.text.Normalizer;
import java.util.regex.Pattern;

public class ArabicNormalizer {

    private static final Pattern DIACRITICS_PATTERN = Pattern.compile("\\p{Mn}");

    public static String normalizeArabic(String text) {
        if (text == null) {
            return null;  // Handle null input
        }

        String normalized = Normalizer.normalize(text, Normalizer.Form.NFD);
[...]

PR #21724, opened 2025-01-14
Add arabic normalizer for search

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/11709

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/CollatorStringMatcher.java
+++ b/OsmAnd-java/src/main/java/net/osmand/CollatorStringMatcher.java
@@ -58,16 +58,13 @@ public class CollatorStringMatcher implements StringMatcher {
 		return cmatches(collator, name, part, mode);
 	}
 
-	public static boolean cmatches(Collator collator, String fullName, String part, StringMatcherMode mode){
-		String withoutDiacritic = ArabicNormalizer.normalize(fullName);
-		boolean matchDiacritic = false;
-		if (!fullName.equals(withoutDiacritic)) {
-			matchDiacritic = cmatchInternal(collator, withoutDiacritic, part, mode);
+	public static boolean cmatches(Collator collator, String fullName, String part, StringMatcherMode mode) {
+		if (ArabicNormalizer.isSpecialArabic(fullName)) {
+			fullName = ArabicNormalizer.normalize(fullName);
 		}
-		return matchDiacritic || cmatchInternal(collator, fullName, part, mode);
-	}
-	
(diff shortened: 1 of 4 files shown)


OsmAnd-java/src/main/java/net/osmand/CollatorStringMatcher.java
```diff
@@ -55,9 +57,17 @@ public Collator getCollator() {
 	public boolean matches(String name) {
 		return cmatches(collator, name, part, mode);
 	}
-	
-	
+
 	public static boolean cmatches(Collator collator, String fullName, String part, StringMatcherMode mode){
+		String withoutDiacritic = ArabicNormalizer.normalize(fullName);
```
vshcherb, line 62: Very ineffective too many calls for comparision plus overhead of 10 replaceAll methods

sonora: Yes, exactly. Some hints:

- Simplify the Kashida replacement by using 'replace' instead of 'replaceAll', more efficient (when no regex is needed).
- Use a single single map-based lookup to reduce the redundancy of multiple 'replace' calls
- Use a Stream to initialize the character replacements, makes the code easier to maintain.

sonora: How's this:
```
public class ArabicNormalizer {

    private static final Pattern DIACRITICS_PATTERN = Pattern.compile("\\p{Mn}");

    private static final Map<String, String> CHAR_REPLACEMENTS = Stream.of(new String[][] {
        {"إ", "ا"}, {"أ", "ا"}, {"ئ", "ي"}, {"ؤ", "و"},
        {"آ", "ا"}, {"ى", "ي"}, {"ة", "ه"}
    }).collect(Collectors.toMap(data -> data[0], data -> data[1]));

    public static String normalize(String text) {
        if (text == null) {
            return null;  // Handle null input
        }

        String normalized = Normalizer.normalize(text, Normalizer.Form.NFD);
        normalized = DIACRITICS_PATTERN.matcher(normalized).replaceAll(""); // Remove diacritics efficiently

        // Hamza and other normalizations
        for (Map.Entry<String, String> entry : CHAR_REPLACEMENTS.entrySet()) {
            normalized = normalized.replace(entry.getKey(), entry.getValue());
        }

        // Kashida
        normalized = normalized.trim().replace("\u0640", ""); // Kashida

        return normalized;
    }
}

```

PR #21761, opened 2025-01-17
Prioritize assigned GPX width over trk/trkseg width

RZR-UA:


RZR-UA:

sonora:
Shouldn't it be the other way round: Higher hierarchy levels (e.g. on the top `gpx` level) are there as the fallback for where no more specifics are present, but are overruled by more specific properties like on the `trk`, `trkseg` or `wpt` levels...

RZR-UA:
@sonora, I agree with you; that's the right way. But at the moment, we do not have features for editing lower-level colors/widths, so we have to consider the color/width set by the user (on the top level). Waypoints are processed individually. Finally, by this request (and previous 1 (https://github.com/osmandapp/OsmAnd/pull/21727) 2 (https://github.com/osmandapp/OsmAnd/pull/21711) 3 (https://github.com/osmandapp/OsmAnd/pull/21722)), we do not break the existing feature of editing a track's color/width, and at the same time, it allows the user to set `trk` and `trkseg` color/width manually, just to prettify track appearance on the map in a peculiar way.

PR #21775, opened 2025-01-20
Avoid using SPEED as the secondType if there is no speed data in the track

RZR-UA:


RZR-UA:

Issue #21248, opened 2024-11-05
Climbing routes: Support for `climbing=area` and `climbing=crag`


alensiljak:
### Describe the idea

From what I've seen, only `sport=climbing` seems supported when looking for POIs. Considering there is a growing number of climbing areas and crags marked with a relation of type `climbing`, as per [Climbing page](https://wiki.openstreetmap.org/wiki/Climbing) in the OSM wiki, it would be useful to (be able to) show these on a map.
Currently, these are shown as POIs since they are also tagged with `sport=climbing`, but a broader support on a pedestrian map would be great. 

### Expected behaviour

Similar to https://openclimbing.org/, crags and areas could be marked with specific icons in a map profile.
`climbing=crag`
`climbing=area`

The tags are ordered hierachically:
Crags (normally) contain routes and route_bottom. Areas contain (multiple) crags. Areas can be nested (to represent sub-areas of a bigger area).

### Alternatives you've considered

n/a

yuriiurshuliak:
https://taginfo.openstreetmap.org/tags/climbing=area#overview
https://taginfo.openstreetmap.org/tags/climbing=crag#overview

vshcherb:
https://openclimbing.org/#18.82/46.3215/7.9813
https://overpass-turbo.eu/?key=climbing&value=crag&template=key-value

![Screenshot 2024-11-07 at 17 40 06](https://github.com/user-attachments/assets/d3eb7abc-db50-4582-ae2f-54e5f956a173)

dmpr0:
## ToDo

**Core**
- [ ] Add support for routes

**Android** 
- [ ] Use icon `ic_action_hill_climbing`

RZR-UA:
- [ ] fix icon

![Image](https://github.com/user-attachments/assets/b07231b3-26b1-443d-bb27-a44ebda263f0)

DmitryAlexei:
- [ ] add osm_climbing and other strings to translations

alensiljak:
Thanks, guys! You are likely the first to support the climbing taxonomy. 
This should make OsmAnd attractive to the climbing community and also make access to climbing areas easier.

Zirochkabila:
# Testing 1
`OsmAnd~ 5.0.0#4361m, Released: 2025-02-12`
`4.9.5.4972`

1. The icon is not displayed in iOS

<details><summary>iOS</summary>
<table style="width:100%">
  <tr>
[...]

PR #21781, opened 2025-01-22
Use ic_action_hill_climbing for Climbing routes

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21248

PR #21784, opened 2025-01-22
Fix tools build (migrate RouteExporter to Kotlin lib, etc)

RZR-UA:


RZR-UA:
approved by @alex-osm

PR #21792, opened 2025-01-24
Show default value if current style value is missing

nazar-kutz:


OsmAnd/src/net/osmand/plus/plugins/srtm/SRTMPlugin.java
```diff
@@ -571,7 +571,8 @@ public void toggleTerrain(boolean isChecked, Runnable callback) {
 	}
 
 	public String getPrefDescription(Context ctx, RenderingRuleProperty p, CommonPreference<String> pref) {
-		if (!Algorithms.isEmpty(pref.get())) {
+		String value = pref.get();
+		if (!Algorithms.isEmpty(value) && p.containsValue(value)) {
```
Chumva, line 575: Is it acceptable to change only the UI description? If a non-existent value from preferences is passed to rendering, will the map behave the same as it does with the default value?

PR #21804, opened 2025-01-27
fix next turn widget disappear

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/NextTurnWidget.java
```diff
@@ -76,7 +76,7 @@ public void updateNavigationInfo(@Nullable DrawSettings drawSettings) {
 				setDeviatePath((int) routingHelper.getRouteDeviation());
 			} else {
 				NextDirectionInfo info = routingHelper.getNextRouteDirectionInfo(nextDirectionInfo, true);
-				if (info != null && info.distanceTo > 0 && info.directionInfo != null) {
+				if (info != null && info.distanceTo >= 0 && info.directionInfo != null) {
```
Chumva, line 79: Do we need the same changes in TripHelper.setupTrip()?

PR #21807, opened 2025-01-27
Show default value if current style value is missing p.2

nazar-kutz:


Chumva:
Moved to [universal fix](https://github.com/osmandapp/OsmAnd/pull/21854)

PR #21847, opened 2025-02-03
Enable map grid in development plugin

Andrewschenko:
Enable showing grid on the map in development plugin settings

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -431,6 +431,11 @@ You will be able to pair this scanner again at any time.</string>
     <string name="show_debug_info">Show debug information</string>
     <string name="show_debug_info_description">Display graphical info about placement of each map text</string>
     <string name="map_text">Map texts</string>
+    <string name="show_grids">Show map grid</string>
+    <string name="show_grids_description">Allow displaying grid over the map</string>
+    <string name="show_utm_grids">Use UTM instead of WGS84</string>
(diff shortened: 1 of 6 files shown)


sonora:
Looks like this works only in v2 engine, why?

vshcherb:
Yes, no plan to support v1. May be once feature is complete we could backport it, for now there is no priority as v1 doesn't work on IOS

sonora:
@vshcherb OK!

But off-topic, I am wondering if we have any stats on how many of our Android users use v1 vs. v2. And: I get to know a significant number of users out there where their device defaulted to v2, but they experience lagginess and it resolved after I made them try v1.

So the story here is that quite possibly OsmAnd's reputation of being a bit laggy is in pats caused by users (unknowingly) not using the engine which would perform better on their device.

vshcherb:
By crash stats I see only v2 and negligible 1-2% from v1 from rendering, we also saw situations lots of crashes from arm-v7 devices where v2 doesn't work at all, so we did a hard-switch to v1 on these devices but these devices only 1% of all devices that don't support arm-v8, arm-v64

vshcherb:
According to my tests all devices 5- years older have v2 running faster than v1

sonora:
Yes, I followed how we do things. But there could be gaps, like e.g. my current workhorse device the outdoor Samsung Xcover 4s is 2019, 64bit, and runs v2 without errors. But v1 is noticeably faster.

PR #21849, opened 2025-02-04
Fix exit size and text format

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/NextTurnBaseWidget.java
```diff
@@ -516,6 +517,23 @@ public void recreateView() {
 		view.setVisibility(oldContainer.getVisibility());
 	}
 
+	private void formatSubText() {
+		String subText = distanceSubView.getText().toString();
+
+		if (distanceSubView != null && !Algorithms.isEmpty(subText)) {
```
Chumva, line 523: Access distanceSubView & after that check it for null?

PR #21897, opened 2025-02-13
Don't save elevation correction without approving

nazar-kutz:


Chumva:
The main issue should already be fixed

PR #21907, opened 2025-02-15
Implement obd ble tmp

Corwin-Kh:


Chumva:
This should already be implemented

PR #21961, opened 2025-02-23
Add new "Select Location" screen

nazar-kutz:


OsmAnd/src/net/osmand/plus/plugins/audionotes/TakeAudioNoteAction.java
```diff
@@ -32,19 +31,24 @@ public TakeAudioNoteAction(QuickAction quickAction) {
...
 	@Override
-	public void drawUI(@NonNull ViewGroup parent, @NonNull MapActivity mapActivity) {
-		View view = LayoutInflater.from(parent.getContext())
-				.inflate(R.layout.quick_action_with_text, parent, false);
-		((TextView) view.findViewById(R.id.text)).setText(
-				R.string.quick_action_take_audio_note_descr);
-		parent.addView(view);
+	@Nullable
+	protected Object getLocationIcon(@NonNull MapActivity mapActivity) {
+		AudioVideoNotesPlugin plugin = PluginsHelper.getPlugin(AudioVideoNotesPlugin.class);
```
Chumva, line 44: Probably here plugin should already be available, we could get it via require

OsmAnd/src/net/osmand/plus/quickaction/actions/RouteAction.java
```diff
@@ -34,17 +32,19 @@ public RouteAction(QuickAction quickAction) {
...
 	@Override
-	public void drawUI(@NonNull ViewGroup parent, @NonNull MapActivity mapActivity) {
-		View view = LayoutInflater.from(parent.getContext())
-				.inflate(R.layout.quick_action_with_text, parent, false);
-		((TextView) view.findViewById(R.id.text)).setText(
-				R.string.quick_action_add_route_descr);
-		parent.addView(view);
+	@Nullable
+	protected Object getLocationIcon(@NonNull MapActivity mapActivity) {
+		return null;
```
Chumva, line 42: Should be implemented in second pr

PR #22013, opened 2025-03-05
RouteInfoCard improvements, translation for tag names (Issue: Ski slopes - fixing)

RZR-UA:


OsmAnd/res/values/strings.xml
```diff
@@ -10,6 +10,13 @@
 	- For wording and consistency, please note https://docs.osmand.net/docs/technical/contributions/translating-osmand
 	Thx - Hardy
 -->
+    <string name="tag_lit">Lighting</string>
+    <string name="tag_piste_name">Name</string>
+    <string name="tag_piste_type">Type</string>
+    <string name="tag_piste_oneway">Oneway</string>
+    <string name="tag_piste_ref">Reference</string>
+    <string name="tag_piste_difficulty">Difficulty</string>
+    <string name="tag_piste_grooming">Piste grooming</string>
```
Chumva, line 19: Is it correct? do we need to get tag translations from phrases? for example [Piste grooming](https://github.com/osmandapp/Osmand/blob/1e9ff389468859653c320fcc9a5fdb16921e0209/OsmAnd/res/values/phrases.xml#L52)

RZR-UA: Yes, it is correct.

`phrases.xml` contains translation for OsmAnd POI types, not for the arbitrary OSM tags.

For instance, according to the Issue, tag `piste:difficulty` should translate as "Difficulty" whereas `poi_piste_difficulty` is "Piste difficulty."

Another example is that `tag_lit` ("Lighting" according to the Issue) is not the POI type as is.

RZR-UA: 0f40412452ab5717dd2c02b09ba27b5802d8c616

PR #22020, opened 2025-03-05
Fix EPERM (Operation not permitted) in getGpxFileName

RZR-UA:


RZR-UA:

PR #22021, opened 2025-03-06
Revert strings with tag_ prefixes from strings.xml

RZR-UA:


RZR-UA:

Issue #21873, opened 2025-02-08
Regression: Wikimedia ignores images in categories, if there is no wikidata tag.


elgaard:
### Description

It works for `wikimedia_commons=File:...`
But `wikimedia_commons=Category:...`
no longer works, unless there is a wikidata tag (and I assume that requires that wikidata references the correct Wikimedia Category but I have not tested that part).

### Steps to reproduce

* Find **Esbjerg Strand Lystbådehavn** `caravan_site` in OsmAnd https://www.openstreetmap.org/way/1190666459
* Press "details".  Then "Online photos"

### Actual result

It says "No photos available"

### Expected result

Gallery of photos from Wikimedia category.

Similar to Reersø Havn Autocamperplads, https://www.openstreetmap.org/way/1311769980

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 4.9.7
Android version:9
Device model: SM-G950F
Crash-Logs: ?
```

DmitryAlexei:
Related
- https://github.com/osmandapp/OsmAnd/issues/21579
- https://github.com/osmandapp/OsmAnd/issues/21797

yuriiurshuliak:
https://taginfo.openstreetmap.org/tags/amenity=charging_station

yuriiurshuliak:
The bug can be reproduced. To verify:  

1. On the nightly or latest release version, navigate to **coordinates 55.483813, 8.412003**.  
2. Load the map and select the POI **Esbjerg Strand Lystbådehavn caravan_site**.  
3. Press **Details**, then **Online photos**.  

**Actual Result:** You will see the "No photos available" message.  

To compare, try installing OsmAnd version **4.8.6** and repeat the steps.  
**Expected Result:** A gallery of photos from the Wikimedia category is correctly displayed.  

OsmAnd~ 5.0.0#4355m, released: 2025-02-11
OsmAnd 4.8.6, released: 2024-08-08

<table style="width:100%">
  </tr>
  <tr>
    <th>OsmAnd~ 5.0.0#4355m</th>
    <th>OsmAnd 4.8.6</th>
  </tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/762cb342-692e-410e-845e-4ac53e3b9157"  width="300" height="auto"/> </th>
[...]

PR #22032, opened 2025-03-07
Wikimedia

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/21873

PR #22034, opened 2025-03-07
Create places db in cache dir

Corwin-Kh:


Chumva:
It should already be implemented

Issue #16808, opened 2023-03-26
Search OSM routes by type (lwn, lcn, etc) and by name


Talkless:
It seems in order to follow cycling routes, one has to:

1. Open route in https://cycling.waymarkedtrails.org/#route?id=9184289
2. Save GPX track on PC, as this site does not seem to be mobile-friendly
3. Import GPX track into phone.
4. Import GPX track into OSM, follow...

This is completely not user friendly.

Maybe it would be possible to search for lwn/lcn networks, import it as "GPX" and start to follow it?

Also it seems OsmAnd search does not find "Kurtuvėnų žiedas 33" route (network), only segments of it like "Bubiai MTB XC Route" https://www.openstreetmap.org/way/660452030#map=16/55.8442/23.1332

Zirochkabila:
@Talkless Everything can be done in the shortest way:

- Bicycle profile
- Configure map
- Cycle routes 
- Enabled
- Select route
- Navigation
- Set start


https://user-images.githubusercontent.com/104760013/227780492-0146b1e0-657c-4fff-a68b-41fdf1343528.mp4

Talkless:
Thanks @Zirochkabila !

OK. first of all, it seems you need to use SHORT tap on the route, but it works _only_ one some sort of label, like name text "Bubiai MTB XC" or "[E11]" sign. So this is first complication. I was trying long tap, and it produced different popup.

Secondly, to make it work with round-trip routes, you have to manually select as  "From:" point .

Now question is: can you enter this route anywhere, or have to start from the "official" starting point?

Talkless:
@Zirochkabila I cant find a way to select this cycling route: https://cycling.waymarkedtrails.org/#route?id=9210538&map=12.0/55.7894/23.0989

There's no labels. Is it mapped incorrectly..?

Zirochkabila:
@Talkless The route can be selected by clicking on the route name - please try as shown in the video


https://github.com/osmandapp/OsmAnd/assets/104760013/39d3895f-474c-4b75-b108-e528558e9e40

Talkless:
It's just street name, not name of the whole route.

Talkless:
So if route goes on paths, or uncategorized ways with noname=yes, you could not select it?

vshcherb:
I think we need to split into 2 issues:
1. Search should be able to list all routes as we have Search and we have Routes category there 
[...]

PR #22044, opened 2025-03-10
Implement OSMC icons (shields) in Track Card, Amenity Menu, clean up internal tags in Track Card

RZR-UA:


RZR-UA:

PR #22061, opened 2025-03-12
Fix lost osmc_textcolor for new OSM routes

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/16808#issuecomment-2718577709

PR #22075, opened 2025-03-14
Remove useless GPX reloading for isTempFile tracks (avoid double-call to getAnalysis, etc)

RZR-UA:
Tested on NetworkRouteSelector, TravelGpx and ClickableWay.

Click from map = OK
Click from search = OK
Save to permanent GPX = OK


RZR-UA:

Issue #21788, opened 2025-01-23
Website and contact:website


OpenSourceForEver:
### Description

When a POI has "website" "and "contact website" tags Osmand only displays "contact:website" but not "website"  tag.
Examples:
https://osmand.net/map?pin=40.62594,-4.00726#17/40.62594/-4.00726

https://osmand.net/map?pin=40.40597,-3.68877#19/40.40597/-3.68877

https://osmand.net/map?pin=40.47229,-3.68293#19/40.47229/-3.68293

https://osmand.net/map?pin=40.69270,-4.03543#19/40.69270/-4.03543

### Steps to reproduce

Clicking a POI with "website" and "contact:website" tags

### Actual result

Osmand does not show "website" tag

### Expected result

It would be usefull to show both tags, "website" and "contact:website"

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version:
Android version:
Device model:
Crash-Logs: ?
```

vshcherb:
For long time `contact:website` was converted to `website` by default, in that case it leads to this issue. 
Options:
1. Probably we should leave by default `website` if `contact:website`  is present. 
2. Completely separate `contact:website` and don't convert it to `website`
```
[out:json][timeout:25];
// gather results
nwr["contact:website"]["website"]({{bbox}});
// print results
out geom;
```

vshcherb:
From @dmpr0 we incline to separate 2 tags

xmd5a2:
POI to test: https://www.openstreetmap.org/node/10086218202
Map to test: Ukraine_kyiv
Map will be updated in a couple of days after scheduled maps will be done.

Zirochkabila:
# Testing 1
`OsmAnd~ 5.0.0#4322m, Released: 2025-02-06`
`4.9.5.4966`

Implemented in the generated map, waiting for the map update
@dmpr0 Maybe we should add the description “Website” as it is done for “Contact:website” if we display both tags?

[Fora.obf.zip](https://github.com/user-attachments/files/18692661/Fora.obf.zip)


<details><summary>Fora</summary>
<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/02989998-153e-42d4-87ef-086bb17ca600"  width="600" height="auto"/> </th>
[...]

PR #22114, opened 2025-03-19
Remove setWebsiteUrl completely

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21788

Issue #20226, opened 2024-07-02
Hide "For long distances: Please add intermediate destinations" empty state and toast for HH routing


dmpr0:
### How to reproduce
* Open **Menu** – **Navigation**
* **Set destination** on the place where no offline maps loaded
* Missed maps empty state should appears, tap **Details**
* Tap on **Use downloaded maps**

### Actual result
<table style="width:100%">
  <tr>
    <th>Toast</th>
    <th>Empty state</th>
  </tr>
  <tr>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/7d987d8f-baab-4491-b303-6bb6a8aa4549"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/ef3cac6c-71d4-4130-8556-2f548d37d9f4)"  width="300" height="auto"/> </td>
  </tr>
</table>   

### Expected result

- [x] If Routing type is set to HH x CC++ or HH X Java
- - [x] Don't show toast "For long distances..."
- - [x] Don't show empty state "For long distances..."

**Environment**
OsmAnd Version:
Android: 15
Device model: Pixel 6A

RZR-UA:
To Do:

- [ ] iOS check/fix

Test:

- [ ] Routing >400km using car/bicycle profiles should not display "Long-distance" toaster
- [ ] The same using pedestrian profile should still display the toaster

Zirochkabila:
# Testing 1
`OsmAnd~ 5.0.0#4768m, Released: 2025-03-21`
`4.9.5.4998`

**The only difference between the two systems is that the route progress bar does not move in iOS**

It was tested
1. Android “HH x CC++” and “HH X Java” routing - toast are not displayed, the route progress bar reaches 99% and nothing else happens
2. iOS “A*” and “HH” - toast is not displayed, the route progress bar does not move at all

dmpr0:
## UI Review 

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/7afd193e-4b92-44e2-84c4-307732979162"  width="300" height="auto"/> </td>
  </tr>
</table>   


- [ ] Remove toast from iOS version

PR #22122, opened 2025-03-20
Disable "long distances" toaster for HH+car/bicycle

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20226

PR #22137, opened 2025-03-21
Fix crash when try to change zoom levels when current zoom isn't supported

nazar-kutz:


OsmAnd-java/src/main/java/net/osmand/util/Algorithms.java
```diff
@@ -1130,6 +1130,12 @@ public static int compare(String str1, String str2, boolean nullIsLess) {
         return str1.compareTo(str2);
     }
 
+	public static <T extends Comparable<T>> T clamp(T value, T min, T max) {
```
Chumva, line 1133: [MathUtils.clamp()?](https://developer.android.com/reference/androidx/core/math/MathUtils)

PR #22171, opened 2025-03-27
Better fix for click on unsupported road shields

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-iOS/issues/4437

PR #22185, opened 2025-03-29
Synchronize text size for left and right widget sides

nazar-kutz:


OsmAnd/src/net/osmand/plus/utils/AndroidUtils.java
```diff
@@ -629,9 +630,42 @@ public static void setTruncatedText(TextView textView, String text) {
 		textView.setText(text);
 	}
 
-	public static int getTextWidth(float textSize, String text) {
+	public static int getMaxPossibleTextSize(@NonNull String text, @Nullable Typeface typeface,
```
Chumva, line 633: Do we need this as a universal method, or could it be moved to widgets?

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/routeinfo/RouteInfoWidget.java
```diff
@@ -61,12 +61,13 @@ public class RouteInfoWidget extends MapWidget implements ISupportVerticalPanel,
...
+	private DisplayValue cachedDefaultView;
 	private DisplayPriority cachedDisplayPriority;
 	private Integer cachedMetricSystem;
 	private boolean forceUpdate = false;
 	private boolean hasEnoughWidth;
 	private boolean hasSecondaryData;
+	private int widgetWidth;
 
 	// views
 	private MultiTextViewEx tvPrimaryLine1;
```
Chumva, line 73: better avoid unnecessary numbers in variables names

PR #22198, opened 2025-03-31
Fixed crash on reset filters

Corwin-Kh:


OsmAnd/src/net/osmand/plus/myplaces/tracks/SearchMyPlacesTracksFragment.java
```diff
@@ -134,7 +134,8 @@ protected void updateButtonsState() {
 					}
 				}
 			}
-			saveButton.setEnabled(filtersChanged);
+			final boolean enableSaveButton = filtersChanged;
+			app.runInUIThread(() -> saveButton.setEnabled(enableSaveButton));
```
Chumva, line 138: Will other UI changes in updateButtonsState()  lead to similar chash?

PR #22201, opened 2025-04-01
adding widgets ui review

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/panel/ConfigureWidgetsFragment.java
```diff
@@ -191,7 +225,9 @@ private void setupToolbar() {
 				toggleEditMode(true);
 				selectedFragment.updateEditMode();
 			} else if (selectedFragment != null) {
-				selectedFragment.resetToOriginal();
+				v.postDelayed(() -> {
+					selectedFragment.resetToOriginal();
+				}, 200);
```
Chumva, line 230: Why do we need this delay?

PR #22233, opened 2025-04-03
Enabled logging for picasso

Corwin-Kh:


OsmAnd/src/net/osmand/plus/plugins/development/OsmandDevelopmentPlugin.java
```diff
@@ -259,6 +261,13 @@ public void disable(@NonNull OsmandApplication app) {
 		super.disable(app);
 	}
 
+	@Override
+	public void setEnabled(boolean enabled) {
+		super.setEnabled(enabled);
+		PicassoUtils.getPicasso(app);
+		Picasso.get().setLoggingEnabled(enabled);
```
Chumva, line 268: Probably we need similar logs for new [NetworkImageLoader](https://github.com/osmandapp/Osmand/blob/master/OsmAnd-shared/src/androidMain/kotlin/net/osmand/shared/util/NetworkImageLoader.kt#L15-L14)

PR #22246, opened 2025-04-04
Coordinates Grid Iteration 3

nazar-kutz:


OsmAnd/src/net/osmand/plus/views/layers/CoordinatesGridHelper.java
```diff
@@ -0,0 +1,179 @@
...
+import net.osmand.core.jni.ZoomLevel;
+import net.osmand.plus.OsmandApplication;
+import net.osmand.plus.base.containers.Limits;
+import net.osmand.plus.settings.backend.ApplicationMode;
+import net.osmand.plus.settings.backend.OsmandSettings;
+import net.osmand.plus.settings.enums.GridFormat;
+import net.osmand.plus.settings.enums.GridLabelsPosition;
+import net.osmand.plus.views.OsmandMapTileView;
+
+public class CoordinatesGridHelper {
```
vshcherb, line 20: Again helper...
CoordinatesGridLayerSettings what I see 80% of job here

nazar-kutz: `CoordinatesGridHelper` is meant to be just a wrapper around the settings — to provide more centralized access from different parts of the code where these preferences are used.
It's created in each case where we need to interact with the grid settings.
It might make sense to rename it or even access the settings directly, but I thought having this centralized helper would be a cleaner and more convenient approach.

OsmAnd/src/net/osmand/plus/configmap/GridColorController.java
```diff
@@ -0,0 +1,113 @@
...
+
+	@Override
+	@ColorInt
+	protected int getSavedColor(boolean nightMode) {
+		return gridSettings.getGridColor(appMode, nightMode);
+	}
+
+	@Override
+	public void onResume() {
+		super.onResume();
```
Chumva, line 98: Could be removed

Issue #22045, opened 2025-03-10
What's here menu: multipe duplicated adresses on the 17th zoom (for a specific location)


Zirochkabila:
### Description
Multiple duplicated addresses appear in the "What's here" popup when tapping on a specific map location.

[Дублювання адреси.obf.zip](https://github.com/user-attachments/files/19166122/obf.zip)

### Steps to reproduce
1. Open OsmAnd (Android or iOS) > Go to places `52°13.803' N, 21°00.219' E` > Zoom in to 17 level > Tap on the map (Selected area)

<details><summary>Area</summary>
<img width="400" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/2463ac5d-42c4-4133-94af-5547e7ac976e">
</details> 

### Actual result

<details><summary>Duplicated addresses popup</summary>
<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/3a0edb9c-cfb9-4bad-9b16-a5fe7fee960c" width="300" height="auto"></td>
    <td><img src="https://github.com/user-attachments/assets/79765688-6db0-488f-8aff-7c8e078b7cc2" width="300" height="auto"></td>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/db34a427-297f-456c-8cc1-6855818915e0" width="300" height="auto"></td>
    <td><img src="https://github.com/user-attachments/assets/7392df4c-4ad8-4606-bb09-3014ae09ee5b" width="300" height="auto"></td>
  </tr>
</table>
</details>

### Expected result

- [ ] The popup should show unique entries for each location without duplication.

### Your Environment
- OsmAnd version: OsmAnd~ 5.0.0#4619m, Released: 2025-03-10 \ 4.9.5.4990

RZR-UA:
Proposal: do not display nameless RenderedObject(s).

Explanation.

Actual menuObject.nameStr is calculated from name-tags and iconRes.
Default Map Style renders some objects using different icon names, for example:
"barrier=bollard" at z17 has icon "barrier_small_red_2" which is not translated (nameStr="")
"barrier=bollard" at z18+ has icon "barrier_bollard" and translated using "poi_bollard" ("Bollard")

[...]

PR #22250, opened 2025-04-04
[5.0] Do not display nameless (probably useless) Rendered Objects in MapMultiSelectionMenu

RZR-UA:


RZR-UA:

PR #22263, opened 2025-04-07
Fixed crash on restore fragment and disappear toolbar button

Corwin-Kh:


OsmAnd/src/net/osmand/plus/exploreplaces/ExplorePlacesFragment.kt
```diff
@@ -55,8 +57,9 @@ class ExplorePlacesFragment : BaseOsmAndFragment(), NearbyItemClickListener,
 	private val singleThreadExecutor = Executors.newSingleThreadExecutor()
 	private var convertAmenitiesTask: ConvertAmenitiesTask? = null
 
-	private lateinit var poiUIFilter: PoiUIFilter
-	private lateinit var adapter: ExplorePlacesAdapter
+	private var poiUIFilter: PoiUIFilter? = null
+	private var poiUIFilterID: String? = null
```
Chumva, line 61: Do we really need separate field poiUIFilterID? why not get it from poiUIFilter & savedInstanceState?

OsmAnd/src/net/osmand/plus/exploreplaces/ExplorePlacesFragment.kt
```diff
@@ -126,11 +135,23 @@ class ExplorePlacesFragment : BaseOsmAndFragment(), NearbyItemClickListener,
...
+				override fun onFinish(init: AppInitializer) {
+					init.removeListener(this)
+					poiUIFilter = getPoiUIFilter()
+					adapter?.setPoiUIFilter(poiUIFilter)
+					poiUIFilter?.let {
+						app.poiFilters.replaceSelectedPoiFilters(it)
+					}
+				}
+			})
+		}
```
Chumva, line 154: Better move this part to onCreate, also check if fragment is added

OsmAnd/src/net/osmand/plus/search/dialogs/QuickSearchDialogFragment.java
```diff
@@ -715,10 +716,14 @@ public void saveCustomFilter() {
 
 	public void restoreToolbar() {
 		if (toolbarVisible) {
-			if (toolbarTitle != null) {
-				showToolbar(toolbarTitle);
-			} else {
-				showToolbar();
+			MapActivity mapActivity = getMapActivity();
+			TopToolbarController controller = mapActivity.getTopToolbarController(toolbarController.getType());
+			if (controller == null) {
```
Chumva, line 721: Do we need to try to update toolbar title if it exists?

Corwin-Kh: i don't think so. it's state is restored. looks like it works as expected

PR #22281, opened 2025-04-09
[5.0] Fix clicks on POI (Amenity) located on OSM routes

RZR-UA:


RZR-UA:

PR #22301, opened 2025-04-10
[5.0] Refactor Names and Icons for Rendered Objects (draft)

RZR-UA:


RZR-UA:

OsmAnd/src/net/osmand/plus/mapcontextmenu/controllers/RenderedObjectMenuController.java
```diff
@@ -26,12 +27,20 @@ public class RenderedObjectMenuController extends MenuController {
...
+	private MapPoiTypes mapPoiTypes = null;
+	private MapPoiTypes.PoiTranslator poiTranslator = null;
+
 	public RenderedObjectMenuController(@NonNull MapActivity mapActivity,
 	                                    @NonNull PointDescription pointDescription,
 	                                    @NonNull RenderedObject renderedObject) {
 		super(new RenderedObjectMenuBuilder(mapActivity, renderedObject), pointDescription, mapActivity);
 		builder.setShowNearestWiki(true);
 		setRenderedObject(renderedObject);
+		if (this.getMapActivity() != null) {
```
Chumva, line 40: we get nonnull mapActivity in constructor, do we need to use geter & check here?

Also app or poiTypes could be stored in local variable to avoid copy pasted calls

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/RenderedObjectMenuBuilder.java
```diff
@@ -135,6 +138,11 @@ private static Amenity getSyntheticAmenity(@NonNull MapActivity mapActivity, @No
 		return am;
 	}
 
+	@NonNull
+	public Amenity getAmenity() {
```
Chumva, line 142: could be moved to upper level

PR #22303, opened 2025-04-10
Shift grid labels when vertical widgets panel visible

nazar-kutz:


OsmAnd/src/net/osmand/plus/views/layers/CoordinatesGridLayer.java
```diff
@@ -52,13 +59,23 @@ public class CoordinatesGridLayer extends OsmandMapLayer {
 	private boolean cachedNightMode;
 	private StateChangedListener settingsListener;
 
+	private ViewHeightChangeListener widgetsPanelHeightListener;
+	private VerticalWidgetPanel topWidgetsPanel;
+	private VerticalWidgetPanel bottomWidgetsPanel;
```
Chumva, line 64: Do we need to set these values to null if mapActivity is null?

PR #22308, opened 2025-04-11
Fix hot change rtl

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/layers/MapInfoLayer.java
```diff
@@ -72,6 +73,7 @@ public class MapInfoLayer extends OsmandMapLayer implements ICoveredScreenRectPr
 
 	private DrawSettings drawSettings;
 	private int themeId = -1;
+	private boolean oldRtlOrientation;
```
Chumva, line 76: Why old ?

OsmAnd/src/net/osmand/plus/views/layers/MapInfoLayer.java
```diff
@@ -187,6 +191,37 @@ public boolean isTopToolbarViewVisible() {
 		return topToolbarView != null && topToolbarView.isTopToolbarViewVisible();
 	}
 
+	public void onResumeUpdate() {
```
Chumva, line 194: Looks complicated, do we need all this here?

PR #22326, opened 2025-04-13
Select Location on Map: fixes after the 1st UI Review

nazar-kutz:


OsmAnd/src/net/osmand/plus/dialogs/SelectLocationFragment.java
```diff
@@ -109,7 +109,7 @@ private void updateCoordinatesView(@NonNull View view) {
 
 	@Override
 	protected void applyChanges() {
-		controller.onApplySelection();
+		app.runInUIThread(controller::onApplySelection, 100);
```
Chumva, line 112: Why do we need this delay?

PR #22330, opened 2025-04-14
[5.0] Ignore click-on-map with ignoreClick in MapSymbol

RZR-UA:


RZR-UA:

PR #22336, opened 2025-04-15
add outline for complex\vertical simple widgets

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/OutlineTextView.java
```diff
@@ -0,0 +1,94 @@
...
+	public void showOutline(boolean show) {
+		drawOutline = show;
+	}
+
+	@Override
+	public void invalidate() {
+		if (isDrawing){
+			postInvalidate();
+			return;
+		}
```
Chumva, line 60: Will this lead to any issues?

Issue #22345, opened 2025-04-15
It is impossible to select a stop for river transport


Nine-Tailed:
### Description

When you click on the water transport stop nothing happens.
https://github.com/user-attachments/assets/97c61e13-c261-46c7-8d04-08043b5292df

### Steps to reproduce

Try select
https://www.openstreetmap.org/node/9631376187

### Actual result

Stop is not selected.

### Expected result

The stop must be select

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version:OsmAnd~ 5.1.0#5016m, Дата выпуска: 2025-04-15
Android version:
Device model:
Crash-Logs: ?
```

mic140:
Apparently this happens when the tags amenity=ferry_terminal and public_transport =stop_position are combined on one object

Nine-Tailed:
@mic140, in 4.9.10 the selection works, but in 5.0 it doesn't.

Zirochkabila:
More data is presented at [22348](https://github.com/osmandapp/OsmAnd/issues/22348)

yuriiurshuliak:
**River transport stops (e.g., ferry terminals) are not selectable by tapping on them in recent OsmAnd versions. This worked previously but is now broken on both Android and iOS, making it a regression**.


**Issue confirmed** on both platforms.

### Steps to reproduce:
1. Open the OsmAnd app.
2. Navigate to coordinates [55.754035, 37.563897](https://www.openstreetmap.org/node/9631376187#map=19/55.754034/37.563899&layers=N).
3. Tap on the water transport stop.
4. **Actual result:** The stop is not selected.
5. **Expected result:** The stop should be selected and details shown.

---

### 📱 OsmAnd version comparison:

**Android**

<table style="width:100%">
  </tr>
  <tr>
    <th>4.9.10</th>
    <th>5.1.0#5023m</th>
  </tr>
  <tr>
    <td> <video source src="https://github.com/user-attachments/assets/07af5287-33da-45ec-9d65-c745128895f5"  width="300" height="auto"/>   </td>
    <td><video source src="https://github.com/user-attachments/assets/e5ac0ef1-890c-404e-96e3-21b96053fd32"  width="300" height="auto"/>   </td>
  </tr>
</table>

**iOS**

<table style="width:100%">
  </tr>
  <tr>
    <th>4.9.3 (4945)</th>
[...]

PR #22352, opened 2025-04-16
[5.0] Fix click-on-Amenity with unsupported route tags

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22345

PR #22353, opened 2025-04-16
Fixed restoring SelectedPointBottomSheetDialogFragment

Corwin-Kh:


OsmAnd/src/net/osmand/plus/measurementtool/SelectedPointBottomSheetDialogFragment.java
```diff
@@ -44,6 +46,18 @@ public class SelectedPointBottomSheetDialogFragment extends MenuBottomSheetDialo
 	private static final Log LOG = PlatformUtil.getLog(SelectedPointBottomSheetDialogFragment.class);
 	private MeasurementEditingContext editingCtx;
 
+	@Nullable
+	@Override
+	public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup parent, Bundle savedInstanceState) {
+		MapActivity mapActivity = getMapActivity();
+		if (mapActivity == null || mapActivity.getMapLayers().getMeasurementToolLayer().getEditingCtx().getPoints().isEmpty()) {
```
Chumva, line 53: MapActivity should be available here, can use require, or app.getOsmandMap().getMapLayers(), also global fix for these types of data loss should be implemented [here](https://github.com/osmandapp/OsmAnd-Issues/issues/1495)

Issue #21489, opened 2024-11-29
Duplicate favorite point shouldn't be created


Rowin63:
### Describe the idea

Sometimes tracks contain waypoints.
It is possible to copy this waypoints to favorites.
But unfortunately it is not possible to select an existing favorite folder, only "user" is possible.

Example:
TET Turkey has 27 Tracks. Every track contains some waypoints/favorites.
It is not possible to copy all the WP from all 27 tracks into a user defined foder, because after selecting a track > menu "points" > 3dots > add to favorites > the only option is predifined "user" and no existing favorite group can be selected..

It would be really helpful at this point to select an already existing folder and copy all the wp from these 27 tracks into the same folder (of course still one by one track, but at least all into the same existing folder). 

![Screenshot_20241129_163838_OsmAnd+](https://github.com/user-attachments/assets/6f22dbb9-a5d8-462b-a3ea-0d74370d64fc)

![Screenshot_20241129_163847_OsmAnd+](https://github.com/user-attachments/assets/a2858259-b69c-4127-9758-963a7d5bea7f)

### Expected behaviour

pls see above.

### Alternatives you've considered

none

Rowin63:
doing so (copying track waypoints to another folder) brought me back to a very earlier issue:
Because osmand crashed at one point, I started over again with an earlier track. But I already copied these waypoints, so some double ones were created (no hint, they´re already existing...). 
Now I´m back to the point where I´m not able to delete this double ones (if I tap/mark one, the second one is marked automatically, too).

yuriiurshuliak:
The bug has been reproduced.  Here are the details:  

1. **Copying Waypoints to Favorites**:  
   - To reproduce, import [tracks](https://github.com/user-attachments/files/18208519/TR.zip) containing waypoints and select one.  
   - Open its waypoint menu (Menu > Waypoints > 3 dots > Copy to Favorites).  
   - **Observation**: Only the predefined "user" folder is available, with no option to select a user-defined favorites folder.  

2. **Duplicate Favorites Issue**:  
   - Try copying the waypoints from the same track multiple times to the Favorites folder.  
   - **Observation**: Duplicate waypoints are created without any warning or prompt indicating they already exist.  

OsmAnd~ 5.0.0#3971m, released: 2024-12-19

<table style="width:100%">
  </tr>
  <tr>
[...]

PR #22363, opened 2025-04-16
Fix "Favorites duplicates when copying from GPX wpts #21489"

nazar-kutz:


OsmAnd/src/net/osmand/plus/myplaces/favorites/FavoriteGroup.java
```diff
@@ -195,4 +195,15 @@ public static FavoriteGroup fromPointsGroup(@NonNull PointsGroup pointsGroup) {
 		}
 		return favoriteGroup;
 	}
+
+	public boolean containsPointByName(@NonNull String name) {
```
Chumva, line 199: Is it correct to check only by name?

PR #22373, opened 2025-04-17
[5.0] Show map object tags in original order, refactor typeStr/nameStr, add translation example

Andrewschenko:
Don't change the order of tags of map objects

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
@@ -335,9 +335,7 @@ public class MapSelectionHelper {
 							obfMapObject = null;
 						}
 						if (obfMapObject != null) {
-							Pair<Map<String, String>, List<Pair<String, String>>> pair = getTags(obfMapObject.getResolvedAttributesListPairs());
-							Map<String, String> tags = pair.first;
-							List<Pair<String, String>> tagList = pair.second;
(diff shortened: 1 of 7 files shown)


OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
```diff
@@ -706,16 +708,19 @@ private static List<String> getValues(@Nullable QStringStringHash set) {
...
 	@NonNull
-	private static Map<String, String> getTags(@Nullable QStringStringHash set) {
-		Map<String, String> res = new HashMap<>();
-		if (set != null) {
-			QStringList keys = set.keys();
-			for (int i = 0; i < keys.size(); i++) {
-				String key = keys.get(i);
-				res.put(key, set.get(key));
+	private static Pair<Map<String, String>, List<Pair<String, String>>> getTags(@Nullable QStringStringList tags) {
+		Map<String, String> map = new HashMap<>();
```
RZR-UA, line 712: List<> seems not required as a duplicate result here.

LinkedHashMap would be enough to keep the List order in the Map:

`		Map<String, String> map = new HashMap<>();`

=>

`		Map<String, String> map = new LinkedHashMap<>();`

Andrewschenko: Corrected

ivanPyrohivskyi (approved):

PR #22378, opened 2025-04-17
[5.0] Parse icon and promote iconName to AbstractPoiType

RZR-UA:


OsmAnd-java/src/main/java/net/osmand/osm/AbstractPoiType.java
```diff
@@ -47,8 +53,12 @@ public String getKeyName() {
 		return keyName;
 	}
 
+	public String getIconName() {
```
vshcherb, line 56: 2 functions always confusing, better to leave 1.

If you need it internal it should be called getInternalIconKeyName with private or package access modifier

RZR-UA: renamed

Issue #21952, opened 2025-02-22
Online photos often require pinch-zooming to appear


sonora:
See https://github.com/osmandapp/OsmAnd/pull/21848#issuecomment-2666652767

It's weird, often the first time I tap on an online photo in a POI's context menu, I receive a blank screen with just date, author and license information.

Closing the screen by tapping X, then subsequently tapping the same or another picture on the same context menu opens the pics reliably. Until I select another POI, when the issue starts again...

UPDATE: It's a scaling issue, pinch-zooming out always displays any image.

yuriiurshuliak:
The bug was not reproduced.I attempted to reproduce the behavior by selecting the center of a historical city on the map and opening various POIs with online images. In my tests, the images initially appear as a black screen with metadata at the bottom, then transition to a pixelated, blurry version before fully loading in good quality.  

While I agree that the image does not open instantly, I did not observe the specific issue you described (where the image remains blank until reopening it). 

OsmAnd~ 5.0.0#4471m, released: 2025-02-26

https://github.com/user-attachments/assets/4d281878-fe6b-42f6-a31a-d49a8316269e

sonora:
I have some additional information:
1. This commit https://github.com/osmandapp/OsmAnd/commit/5214bf3891ec1a51840cb49ca0cff4c981017248 has not changed the behavior.
2. Even on builds before  that, I have now found out that when the problem occurs, the picture is displayed when slightly pinch-zooming out. I.e. it seems a scaling issue, like the picture attempts to be displayed slightly too large. Pinch-zooming out a little shows it reliably. So it may well be hardware or screen resolution dependent.
3. Funny: Tapping the device's home button while the picture is now displayed puts the app in the background. Changing back to it again produces the 'empty' screen(!), requiring again the pinch zoom-out to display it.
4 The issue seems to _only_ affect the first picture opened for every POI, whichever you pick, and disappears if subsequently you open the same picture again or another one _from the same context card_.
Changing to a new POI starts the issue again.

DmitryAlexei:
Merged to 
- https://github.com/osmandapp/OsmAnd-Issues/issues/2662

Chumva:
@sonora does this still happen in newer nightly builds? we migrated from picasso to coin it could influence your issue. 
If it still happens, could you provide video for reproducing?

sonora:
@Chumva Yes, I have just re-built from today's master, and the situation seems unchanged.

[...]

PR #22383, opened 2025-04-18
fix scale type

0xRe1nk0:


sonora:
@0xRe1nk0  Possibly something like this is slso needed to fix https://github.com/osmandapp/Osmand/issues/21952 ?

Issue #21993, opened 2025-03-03
Click on Relation POIs: issue with clicks on university POIs


vshcherb:
Backend - @xmd5a2 
https://www.openstreetmap.org/relation/5121502#map=17/50.448606/30.495488
- [x] Non informative name ```<string name="poi_landuse_education">Education</string>``` - Should be "Educational area"
- [x]  `education: university [ man_made='education'  ]`man_made=eductation  - non present in poi_types.xml - man_made is category for poi_landuse_education - check binary inspector (wrong output?) ( [Fix](https://github.com/osmandapp/OsmAnd/pull/22208) )
```
education: university Національний педагогічний університет ім. Михайла Драгоманова Lat 50.448776 Lon 30.495537 osmid=2204267673600  [ man_made='education'  ]  poi_additional:[ official_name='Національний педагогічний університет імені Михайла Петровича Драгоманова' website='https://npu.edu.ua/' opening_hours='Mo-Sa 10:00-20:00' wikipedia='http://uk.wikipedia.org/wiki/Національний педагогічний університет імені М. П. Драгоманова' wikidata='Q4315056'  ]  name:uk='Національний педагогічний університет ім. Михайла Драгоманова'  cities:[name:pt=... ]
man_made: landuse_education Національний педагогічний університет ім. Михайла Драгоманова Lat 50.448776 Lon 30.495537 osmid=2204267673600  [ education='university'  ]  poi_additional:[ official_name='Національний педагогічний університет імені Михайла Петровича Драгоманова' website='https://npu.edu.ua/' opening_hours='Mo-Sa 10:00-20:00' wikipedia='http://uk.wikipedia.org/wiki/Національний педагогічний університет імені М. П. Драгоманова' wikidata='Q4315056'  ]  name:.... ]
```
- [x] Click on icon https://www.openstreetmap.org/relation/5121502#map=19/50.448771/30.494996 near location - should prioritze University and display university and not Educational area.
- [x] Click on icon university - https://www.openstreetmap.org/way/344110485 - doesn't retrieve POI information 
- [x] Follow up task "Display section" - "Related" to display other members of relation - https://github.com/osmandapp/OsmAnd/issues/16808#issuecomment-2747425909

**Technical solution**: Remove old code and create poi for each multipolygon, also include bbox like in route to find other buildings.
Split POIs per outer ring so they could be displayed, each POIs should encode bbox where they should be searched

```
[...]

PR #22388, opened 2025-04-18
Fix lags on click, fix multypolygon POI

ivanPyrohivskyi:
To comment in issue https://github.com/osmandapp/OsmAnd/issues/21993#issuecomment-2815480777

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
@@ -450,7 +450,6 @@ public class MapSelectionHelper {
 	}
 
 	private Amenity getAmenity(LatLon latLon, ObfMapObject obfMapObject, Map<String, String> tags) {
-		Amenity amenity = null;
 		List<String> names = getValues(obfMapObject.getCaptionsInAllLanguages());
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
```diff
@@ -745,11 +736,13 @@ public static Amenity findAmenityByOsmId(@NonNull OsmandApplication app, @NonNul
 		QuadRect rect = MapUtils.calculateLatLonBbox(latLon.getLatitude(), latLon.getLongitude(), AMENITY_SEARCH_RADIUS);
 		List<Amenity> amenities = app.getResourceManager().searchAmenities(ACCEPT_ALL_POI_TYPE_FILTER, rect, true);
 
-		return findAmenityByOsmId(amenities, osmId);
+		return findAmenityByOsmId(amenities, osmId, latLon);
 	}
 
 	@Nullable
-	public static Amenity findAmenityByOsmId(@NonNull List<Amenity> amenities, long id) {
+	public static Amenity findAmenityByOsmId(@NonNull List<Amenity> amenities, long id, LatLon point) {
```
vshcherb, line 743: Here could be an issue that we need to sort amenity by distance, if there are duplicates

PR #22400, opened 2025-04-22
Explore places UI review fix

0xRe1nk0:


OsmAnd/src/net/osmand/plus/search/NearbyPlacesAdapter.kt
```diff
@@ -181,6 +187,26 @@ class NearbyPlacesAdapter(
 			}
 		}
 
+		private fun calculateMaxItemHeight(context: Context): Int {
```
Chumva, line 190: Is it correct to calculate maxHeight here? Can we fix it in xml?

PR #22402, opened 2025-04-22
[5.0] Hotfix ClickableWay for mtb:scale (add mtb:name)

RZR-UA:


RZR-UA:

PR #22406, opened 2025-04-22
[5.0] Hotfix try isClickableWay if isOsmRoute added nothing

RZR-UA:


RZR-UA:

PR #22435, opened 2025-04-25
[5.0] Hide useless "Others" category from Amenity card

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-iOS/issues/4524

Issue #21742, opened 2025-01-16
area:highway=steps is not rendered


demonbox73:
### Describe the idea

@xmd5a2

area:highway=steps is not render
for example this object
https://www.openstreetmap.org/way/295356617

It's kindly possible to fix this issue?

Thanks.

### Expected behaviour

@xmd5a2

This object rendered:
https://www.openstreetmap.org/way/295356617

### Alternatives you've considered

@xmd5a2

None

demonbox73:
@xmd5a2 

Instead:

area:highway=track

is correctly rendered.

xmd5a2:
Fixed. Place to test: https://www.openstreetmap.org/way/497031850
Map to test: **Ukraine_kyiv-city_europe** (will be updated tomorrow)
Screenshot from synthetic test location:

![Image](https://github.com/user-attachments/assets/f6575ff8-c57a-4209-9665-9b312ae1a452)

Zirochkabila:
# Testing 1
`OsmAnd~ 5.0.0#4216m, Released: 2025-01-24`
`4.9.5.4950`
Fixed

dmpr0:
<table style="width:100%">
  <tr>
    <th>Android</th>
    <th>iOS</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/c0155141-ff33-4d8a-8cbb-354820e74c3c"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/359e4ef6-6798-45ba-92af-c17ff39304f5"  width="300" height="auto"/> </td>
  </tr>
</table>   

* Example https://www.openstreetmap.org/way/1062633805#map=19/45.137592/8.448540

- [ ] Different set of fields, Name on iOS?
- [ ] Strings is missinng?

xmd5a2:
Looks like translation is fixed
![Image](https://github.com/user-attachments/assets/a9b5f36a-1e3f-4765-bafc-046584746bce)

xmd5a2:
`area:highway=steps` was fixed for POI. Now is is converted to already existing poi_type "highway_steps" so translations should be fixed.
Place to test: https://www.openstreetmap.org/way/497031850. Map *Ukraine_kyiv-city_europe* will be updated today.

Zirochkabila:
# Testing 2


1. Data in the POI is different

<details><summary>Different data</summary>
<table style="width:100%">
  <tr>
[...]

PR #22473, opened 2025-04-29
[5.0] Hide internal and/or non-translatable Amenity tags

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21742#issuecomment-2634483599

PR #22482, opened 2025-04-30
[5.0] Actualize image urls/width, add OsmAnd-shared MD5

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2960

PR #22518, opened 2025-05-06
fix lost intervals on toggle showOnMap, get/set correct split interva…

0xRe1nk0:
…l in SplitSegmentDialogFragment


OsmAnd/src/net/osmand/plus/myplaces/tracks/dialogs/SplitSegmentDialogFragment.java
```diff
@@ -244,6 +288,7 @@ public void trackSplittingFinished(boolean success) {
 					selectedGpxFile.setSplitGroups(groups, app);
 				}
 				updateContent();
+				app.getGpxDisplayHelper().removeSplitTrackTask(selectedGpxFile);
```
Chumva, line 291: Better move this removal to backend, UI could forget to use it, which will lead to new issues

PR #22530, opened 2025-05-07
[master] Optimize HH clearVisited (15x speedup for recalc)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-core-legacy/pull/270

PR #22531, opened 2025-05-07
Changed default sorting for root, rec and import folders

Corwin-Kh:


OsmAnd/src/net/osmand/plus/settings/enums/TracksSortMode.java
```diff
@@ -41,8 +42,13 @@ public int getIconId() {
...
 
 	@NonNull
-	public static TracksSortMode getDefaultSortMode() {
-		return LAST_MODIFIED;
+	public static TracksSortMode getDefaultSortMode(@Nullable String sortEntryId) {
+		if (Algorithms.isEmpty(sortEntryId) ||
+				Algorithms.stringsEqual("rec", sortEntryId) ||
+				Algorithms.stringsEqual("import", sortEntryId)) {
+			return LAST_MODIFIED;
+		}
```
Chumva, line 50: Do we need to use relativePath for it or only dir name?

Corwin-Kh: here shuld come sortEntryId in for like "new_folder/import"

OsmAnd/src/net/osmand/plus/configmap/tracks/TrackTab.java
```diff
@@ -34,11 +34,12 @@ public class TrackTab implements TracksGroup, ComparableTracksGroup {
 	public final SmartFolder smartFolder;
 	public final String initialName;
 
-	private TracksSortMode sortMode = TracksSortMode.getDefaultSortMode();
+	private TracksSortMode sortMode = TracksSortMode.getDefaultSortMode(null);
 	private TrackFolderAnalysis analysis = null;
 
 	public TrackTab(@NonNull Context context, @NonNull File directory) {
 		this(context, directory, null, FOLDER);
+		sortMode = TracksSortMode.getDefaultSortMode(directory.getName());
```
Chumva, line 42: getId() instead of directory.getName()?

PR #22548, opened 2025-05-09
[kotlin-ios] Fix ios GpxFormatter with usesGroupingSeparator=false

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-iOS/issues/4580

Issue #22545, opened 2025-05-09
[HH] Crash when computing route for bicycle


clementcontet:
### Description

When computing a route, OsmAnd is fine for car or pedestrian, but crashes when I choose bicycle.

### Steps to reproduce

Open OsmAnd, compute a bicycle route

### Actual result

Crash

### Expected result

No crash

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 5.0.5
Android version: 15
Device model: Pixel 9
Crash-Logs:

2025-05-09 13:02:43.924 23719-24011 net.osmand              net.osmand.plus                      D  CommandPlayer abandonAudioFocus {Thread-6906}
2025-05-09 13:02:43.926 23719-23990 net.osmand              net.osmand.plus                      I  RouteProvider Start finding route from Location[mProvider=fused,mTime=1746788563811,mLatitude=45.7765339,mLongitude=4.8250051,mHasAltitude=true,mAltitude=254.72002410888672,mHasSpeed=true,mSpeed=0.48821822,mHasBearing=true,mBearing=322.06818,mHasAccuracy=true,mAccuracy=26.008,mHasVerticalAccuracy=true,mVerticalAccuracy=2.0336251 to Lat 45.753323 Lon 4.861198 using OsmAnd (offline) {pool-7-thread-1}
2025-05-09 13:02:43.927 23719-24011 net.osmand              net.osmand.plus                      D  CommandPlayer abandonAudioFocus {Thread-6906}
2025-05-09 13:02:43.927 23719-23990 net.osmand              net.osmand.plus                      W  RouteProvider Use 405 MB Free 61.505676 of 146.57854 max 512.0 {pool-7-thread-1}
2025-05-09 13:02:43.927 23719-23990 net.osmand              net.osmand.plus                      W  RouteProvider Use 256 MB of native memory  {pool-7-thread-1}
2025-05-09 13:02:43.929 23719-23731 TextToSpeech            net.osmand.plus                      I  Disconnected from TTS engine
2025-05-09 13:02:43.939 23719-23719 MaterialProgressBar     net.osmand.plus                      W  Current drawable is not a MaterialProgressDrawable, you may want to set app:mpb_setBothDrawables
[...]

PR #22568, opened 2025-05-13
Mute/fallback p.dualPoint NPE for invalid HH files

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22545

PR #22581, opened 2025-05-15
add reader to brand search result

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/807

Issue #8319, opened 2020-01-25
Custom POI Search use "ref" tag to produce correct results


Fabi2607:
## Description

Emergency access points cannot be searched by their reference id, if they have a name. If they only have a describtion, it is possible to search for their reference id, if they have a description instead of a name.
This makes locating access points by their id difficult. It would be helpful if at least the reference number would be taken into account (or even shown) in the search. 

### Example 1

The node https://www.openstreetmap.org/node/2565610893 has the following properties:
name: Zwieselhütte
highway: emergency_access_point
ref: 	2078

Using a custom filter for "Emergency Access point" and entering "2078" yields no resuls.
Entering "Zwieselhütte" immediately yields the desired result.

### Example 2

The node https://www.openstreetmap.org/node/2565610671 has the following properties:
name: "Schrammsteinweg / Abzw. Zurückesteig"
highway: emergency_access_point
ref: 	2072

Using a custom filter for "Emergency Access point" and entering "2072" yields the correct result.
Entering "Schrammsteinweg ..." does not produce any results.

## Details
Version of OsmAnd: Osmand~ 3.5.5
Maps used: offline maps for "Germany Saxony"

vshcherb:
tag "ref" is not indexed for text search yet.

vshcherb:
The POI's today are only searchable by *name*, in case *name* is _empty_ than *ref* is used.
2-nd case is not going to be fixed cause "name" is in description tag which is not indexed.

We plan to append "ref" to the name  so it will be filterable as well. According to the data as of 10.02.2020it will look like
1. "Zwieselhütte (2078)"
2. "2072"

Fabi2607:
That sounds like a perfect solution to my specific issue.

For context: 
In mountain rescue we use the reference number to locate the rescue point, so searching for the description is not as important. The search can then be improved by just adjusting the OSM data.

PR #22586, opened 2025-05-15
Fix22089

vshcherb:


OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -1364,23 +1375,27 @@ public boolean publish(Amenity object) {
 						}
 					}
 					if (ns != null) {
+						// TODO why only ref
```
ivanPyrohivskyi, line 1378: https://github.com/osmandapp/OsmAnd/issues/8319

Issue #22197, opened 2025-03-31
Online search: Big cities listed as "Travel Articles"


scaidermern:
### Description

When using online search, large cities are shown as category "Travel Articles":

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/ef4cf47c-db40-4d9e-85ad-0d879290946c"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/c9909c00-c617-4227-a666-1fe94ba3a2ef"  width="300" height="auto"/> </th>
  </tr>
</table>

(The second Paris is located in Texas. I guess since it isn't described as Travel Article it isn't worth visiting 🙃)

### Steps to reproduce

Use online search to search for large cities.

### Actual result

Large cities have the wrong category.

### Expected result

The category should be something like "city".

### Your Environment (required)

```
OsmAnd Version: 5.0.2
Android version: 13
Device model: Fairphone 4
```

scaidermern:
> Please try updating to the beta version **5.0.3** or the latest **nightly build**, and let us know if the issue persists. Thanks!

The issue persists with 5.0.3. I haven't tried the nightly.

scaidermern:
Also exists in 5.0.4. Which means it hasn't been fixed for the 5.0.x branch.

scaidermern:
@DmitryAlexei Looks like this only happens if Europe Wikivoyage is downloaded.

@yuriiurshuliak Could you try again to reproduce this after installing Europe Wikivoyage?

yuriiurshuliak:
Thank you for your detailed observations. You are correct that the issue is reproducible on the **beta version** (5.0.3). However, we have verified that the issue does not occur on the **nightly build** (latest version).

To reproduce the issue on the beta version, please follow these steps:

1. Go to **Menu → Travel guides**.
2. Download **Europe Wikivoyage**.
3. Navigate to **Online search** and search for large cities.

In this case, large cities are incorrectly categorized as **"Travel Articles"** rather than **"City"**.

OsmAnd 5.0.3, released: 2025-04-03


<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/62f8471d-49ae-437c-a603-44eb5a17897a"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/08a3cd0c-7cd1-4b2a-b0e5-7ad5ce519679"  width="300" height="auto"/> </th>
[...]

PR #22591, opened 2025-05-16
[5.0] Hotifix - do not merge synthetic osmId (such as wiki)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22197

Issue #22579, opened 2025-05-15
Return the ability to add third-party tags to a file poi_types.xml


Prusaker:
### Describe the idea

In versions of the OsmAnd before May 2025, it was possible to add the ability to display tags in a file poi_types.xml.
For example: <poi_additional name="ref:belpost" tag="ref:belpost" type="text"/>

This method worked and I used it for many years.

Starting around May 2025, the program changed the way it displays tags from a file poi_types.xml: tags without translation are no longer displayed.

### Expected behaviour

Is it possible to return the ability to display additional tags without translation, as was the case in previous versions of the program?

### Alternatives you've considered

Unfortunately there are no alternatives.

RZR-UA:
Based on r5.0

- [ ] get back display defined but non-translateable POI types in the Amenity card @Chumva 
- [ ] hide internal tags such as Wikidata-related travel_elo, etc @RZR-UA

Zirochkabila:
# Testing 1
`OsmAnd~ 5.1.0#5273m, Released: 2025-05-26`

It was checked by creating (on the Android 12 emulator) the folder `/settings` in the repository `net.osmand.plus/files` and moving the [file](https://github.com/osmandapp/OsmAnd-resources/blob/r5.0/poi/poi_types.xml) to it - tags without translation are displayed
@Prusaker, please download the [nightly](https://osmand.net/docs/versions/nightly_versions) version and check the behavior on your device

Prusaker:
I confirm that in the night version, tags without translate are displayed correctly.
Thanks a lot!

PR #22612, opened 2025-05-19
[5.0] Fix NON-PoiType tags, improve isHidden() call

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22579

PR #22620, opened 2025-05-21
split track on uphill downhill intervals

0xRe1nk0:


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
```diff
@@ -457,6 +459,7 @@ class GpxTrackAnalysis {
 				}
 			}
 			processElevationDiff(s)
+			segmentSlopeType = s.segmentSlopeType
```
alex-dev-neo, line 462: Why last one applied?

Issue #22567, opened 2025-05-13
Trip Recording: Do Not Discard Filtered Points


sonora:
I had first posted here: #22370

**Suggestion:** With an increasing number of inexperienced users, would we not do everybody a favor by implementing that **points affected by filtering are nevertheless included in the gpx files and simply commented out**.

In this fashion
- they would be 'emergency-recoverable' and nobody loses 'important recordings' for good
- We could much more easily detect if a particular recording issue is due to filtering or other effects like power saving issues.

(Perhaps with the exception of the displacement filter, which may be used to keep the gpx file size small, and there you do not lose essential topology of your recording.)

yuriiurshuliak:
Your suggestion to include filtered points as comments in GPX files is interesting.

It's true that being able to see which points were filtered out could be very helpful for users trying to recover potentially important data, and for us when diagnosing whether recording gaps are due to filtering or other issues. It would offer more transparency into the recording process.

vshcherb:
Good suggestion, but we can't process it later as comments won't be read and displayed - they will be only stored

sonora:
Yes, that's why I call it an 'emergency recovery'. Users would have to process the file in Notepad++ or similar, but in theory could recover their complete recording.

vshcherb:
I don't really like implementation idea as it's too complicated and could be easily lost on resave of the track. However we can definitely save filtered points with flag.

Probably we can save them into special table and implement special export in Development for example. We can also keep all points for example for 1-4 weeks and then delete them.

vshcherb:
For example 1 button that will export all complete tracks into for last 30 days rec/bak-rec/

ezemskov:
Hi,
Keen to try adding the unfiltered gpx recording feature, but not sure what is the consensus/decision : 
 - add filtered points  as commented-out xml, that is `<!--trkpt> ....</trkpt-->`
 - export filtered `track.gpx` as before, but add an extra "export/write full/unfiltered gpx" feature and UI for it

Pls correct is I've misunderstood anything above

ezemskov:
Note inspired by the discussion in 
https://github.com/osmandapp/OsmAnd/pull/22654 : 
If we want a feature hard-to-disable, then it looks reasonabe to add some screen of hidden settings (easter-egg control like Android developer options, or smth a bit more obvious).

For desktop/PC applications, I normally put  such sparsely-used settings in a config file/whatever settings contaoner used (not adding them into GUI).
[...]

Issue #22595, opened 2025-05-17
Clearer UI for Routes from Arbitrary Start Points


sonora:
Carried over from #22542:

How to calculate and inspect a route from an arbitrary location (without slipping into turn-by-turn guidance from one"s own position) is a repeated question from users.

Perhaps we have a UI idea to make it clearer that
- a **route** is being calculated immediately, and
- the 'Start' button applies to starting the '**turn-by-turn guidance**, not the route calculation.

Also: The fact that you can pull up the route's From/To context dialog to inspect the calculated route's details is not obvious or indicated, and overlooked by some users. The dialog even has a 'Dteails' button, but it is invisible until you pull up the menu _twice_.

![Image](https://github.com/user-attachments/assets/dde29adb-d786-457e-999e-502b03f3f86e) 

![Image](https://github.com/user-attachments/assets/7bc33c7e-7fe6-4248-9782-84650de00f9a)

![Image](https://github.com/user-attachments/assets/e5e70bda-755b-41c8-af72-756b2a071e07)

Rowin63:
"Swiping up" and "details" is the solution...
Once you have that in mind, it's easy... 👍

ezemskov:
One more thought : 
I'm pretty happy with a "build a route, save it as a gpx track and display that track on the map" (when I want a route, not navigation along that route) - but it does involve a dozen of buttons pressed.

Maybe a user-friendly solution would be an extra button (in addition to Start/Stop), like "Display Route".
It might either keep the calculated route displayed (in default thick purple line style), or perform the sequence mentioned above.

sonora:
Perhaps it can be something like this:

 1. We display an **always visible status line** above the current route calculation progress bar, in the form of the current "Details" button. It says:
 -- **"No destination set"** - while this is still not done
  -- **"Calculating route preview..."** (perhaps with a spinner) - as soon as this is ongoing
  -- **"Route details"** - now it is activated to become tappable

2. We also enhance the "Start" button by giving it a **speaker** symbol instead of the current arrow symbol.

Rowin63:
A "calculation spinner" would probably prevent from tapping "start".
(yes, there is this thin progress line, but if calculating takes longer, then this line usually is stuck shortly before the end and if you don't notice this detail, the user may think they have to tap the noticeable and active (big! and blue!) "start" to continue)

itsbrody:
[...]

Issue #22555, opened 2025-05-11
UI Improvement to new routing direction widget


Guylby:
### Describe the idea

Hi,

I'm really liking the new direction widgets but there is a small thing that could be improved:
On the old widget, the exit number at roundabouts was shown in the middle of the turn widget, now it's on the opposite end of the widget which makes it harder to read especially when driving.

If you use the widget "Second next turn", you can see that the Exit information is located just after the roundabout turn indicator, which is way easier to read quickly. 

Would it be possible to move this "Exit" information left as close to the roundabout turn indicator on the "Next turn" widget for a more consistent experience and to be able to read it faster as shown below? 

<img src="https://github.com/user-attachments/assets/9bf8105a-c0da-43ac-932d-e746c11c45b1" width="50%">

### Expected behaviour

Exit number indicator is located as close as possible to the Turn widget for quicker information acquisition on the "Next turn" widget, same as the "Second next turn" widget 

### Alternatives you've considered

No alternatives found

Issue #22079, opened 2025-03-15
Add Filters to Prevent Spurious GPS Points and Improve Accuracy During Track Resumption


JohnCadei:
### Describe the idea

There is an issue of spurious random points being recorded when the mobile device is more or less stationary during a longish stop during a walk. The OSMAnd Support team advised to either pause the track recording at the beginning of the stop and resume it at the end of the stop, or to try setting in Track Recording Logging Accuracy Settings some filter parameters to prevent spurious points being recorded. In response to this I set the minimum speed filter to “>0”.

However, subsequently, when I stopped for a coffee break during a walk that was being recorded, I paused track recording, and resumed it when I got going again. But this produced a widely off-the-mark random location point being recorded. It looks as if the first location value that Osmand gets from Android after a period of inactivity could be an uninitialised random value. On another occasion after staying inside a building with track recording running, the recording included a large number if random points that appeared as radial spikes from the true position.

It see me s to me that to address this problem the following measures are needed:
1) In addition to lower bound filter criteria (minimum speed, minimum displacement, etc.), which cannot catch inordinately large values of speed, displacement, etc., upper bound filter criteria are also needed, e.g. maximum velocity, displacement, etc.
2) When track recording is resumed after a period of location polling inactivity, the app shouldn’t adopt the first location value it gets from Android, but should repeat getting values until they converge. Only then should it record the location.
3) A feature should be added whereby immediately before saving a track recording, the recording should be checked for the presence of random points that create spikes (deviations) which are not compatible with the best fit track geometry, and these should be deleted.

I reported this to the OSMSnd Support. It advised me that spurious or unwanted points can be deleted from a recorded track after it has been saved  by using the “GPS filter” option in the Options menu of the track view.

This advice enabled me to filter out the spurious point/points satisfactorily once the track has been saved. So thanks to the advice I now know how to clean up recorded tracks of spurious GPS points.

Having said that, OSMAnd would be a better app if it could filter out inappropriate GPS points during track recording, on the basis of some additional Track Recording “Logging accuracy” filter criteria, such as upper bounds on position, velocity, etc., as well as lower and upper bounds on GPS accuracy.

### Expected behaviour

The effect of the proposed additional logging accuracy criteria is that spurious GPS will be filtered out at source and not get recorded in the ongoing track recording.

### Alternatives you've considered

[...]

PR #22654, opened 2025-05-25
Trip Recording: Do Not Discard Filtered Points

ezemskov:
Optionally save all trackpoints (including ones filtered according to the recoding plugin settings) in a separate .gpx file.
Full tracklog is saved in `/rec/unfiltered/<track_filename>_unfiltered.gpx`
Saving of full tracklogs is default disabled with a new boolean flag added into the plugin settings.

Note : 
Automatic purge of full tracklogs after some time interval (as suggested in https://github.com/osmandapp/OsmAnd/issues/22567) is not implemented.
Size of full tracklogs shouldn't be an issue with modern amount of storage (single point is ~250 bytes, 24hrs of unfiltered tracklog recorded at 1Hz is ~21Mb), even if it's explicitly enabled by the user.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -4481,6 +4481,8 @@ Download tile maps directly, or copy them as SQLite database files to OsmAnd\'s 
     <string name="stop_navigation_service">Stop</string>
     <string name="confirm_every_run">Always ask</string>
     <string name="save_global_track_interval">General logging interval</string>
+    <string name="save_track_unfiltered_title">Save full track</string>
+    <string name="save_track_unfiltered_descr">If enabled, both full and filtered tracks will be saved</string>
     <string name="background_service_int">GPS Wake-up interval</string>
     <string name="enable_sleep_mode">Enable GPS background mode</string>
     <string name="save_track_to_gpx_globally_headline">On demand track logging</string>
--- a/OsmAnd/res/xml/monitoring_settings.xml
+++ b/OsmAnd/res/xml/monitoring_settings.xml
@@ -57,6 +57,13 @@
 		android:layout="@layout/preference_category_with_descr"
 		android:title="@string/save_track_logging_accuracy" />
 
+	<net.osmand.plus.settings.preferences.SwitchPreferenceEx
+		android:key="save_track_unfiltered"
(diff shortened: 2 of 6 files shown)


ezemskov:
> The intent of my original suggestion was **to mitigate the cases where users played with filters not knowing what they do, and then later lose valuable data**.
> 
> If we introduce one more setting to "_optionally save all trackpoints_", it only complicates the situation further, and this defies the purpose.

Ok, I 've understood the intent as "give user an option to troubleshoot the track filtering, analyzing/comparing the full tracklog".

That is - the idea is to save full tracklogs unconditionally, and then erase/purge them after some time period ?
This would require some configuration of full tracklog storage duration, anyway.
From all my experience, setting some hardcoded period (that would look extremely reasonable at the moment, like a month of tracklogs on a device with 64Gb of available storage) would eventually become a rake to step on.
Think of some industrial IMU, reporting position at 100Hz on a device with 4Gb of storage.

Can replace the "enable full tracklog saving" toggle switch with a "full tracklog storage duration" control (with an option to disable it).

ezemskov:
No worries
Happy to amend yhe PR if/whenever you guys decide how to implement it.
For the future - would be good to mark (at least in the issues marked as 'nice to have') wherher the scope/desired behaviour is agreed/approved, or it'a a discussion in progress.

I have some of following in mind to fix eventually - not sure of their status
https://github.com/osmandapp/OsmAnd/issues/22595
https://github.com/osmandapp/OsmAnd/issues/22555
https://github.com/osmandapp/OsmAnd/issues/22079
https://github.com/osmandapp/Osmand/issues/22062
https://github.com/osmandapp/OsmAnd/issues/21695

sonora:
Yes, I have the same question sometimes: Is our 'Nice to have' tag to be taken in the literal sense ("would be a nice enhancement"), or is the stress rather with "not a must-have, so let's rather not have this complication"

Probably we have examples of both...

sonora:
How about we simply put all filtered-out points in the file as commented out `<!-- FILTERED: ... --`>.

- This is a rather non-intrusive code change,
- in this fashion they are out of the way automatically for all app functionality and if the file is shared,
- but are in the file visible in support cases,
- and could even be recovered in an emergency?

Issue #21954, opened 2025-02-22
Tapping a POI is laggy (~0.5 - 3 sec, depending on POI)


sonora:
On the map, tapping the MyPosition icon, or a Favorite icon or a displayed GPX track produces a corresponding context menu instantly.

But tapping on a POI icon or a house number mostly produces a notable lag of 0.5 - 1 sec. before the context menu appears.

yuriiurshuliak:
The bug has been reproduced.

### **Steps to reproduce:**  
1. Open the OsmAnd app.  
2. Navigate to the map view.  
3. Tap on **My Position icon, a Favorite icon, or a displayed GPX track** → The context menu appears instantly.  
4. Tap on a **POI icon or a house number** → A delay of approximately **0.5 sec** occurs before the context menu appears.  

OsmAnd~ 5.0.0#4465m, released: 2025-02-25

<table style="width:100%">
  </tr>
  <tr>
    <th>Clicking on My Position, Favorite, or GPX track</th>
    <th>Clicking on a POI icon or a house number</th>
  </tr>
  <tr>
    <td> <video source src="https://github.com/user-attachments/assets/859cca71-9dd6-453b-b7c8-9d5e72388e0b"  width="300" height="auto"/>   </td>
    <td><video source src="https://github.com/user-attachments/assets/8d1a999d-dc8e-44b2-b7eb-36439613a475"  width="300" height="auto"/>   </td>
  </tr>
</table>

vshcherb:
That's probably delay for detecting address, so it's expected for now if it's less < 0.6-0.7 seconds

sonora:
Lag time varies a bit. Mostly I'd say it is 0.5 sec, but occasionally it feels more like 1 sec or slightly above.

In almost any case, even after the context menus appears, I still see _"Looking up address..."_ in the card's subtitle for perhaps another 1-3 sec before the actual address is displayed. So if the initial delay is really due to the address search, it seems not finished by the time the context card is first displayed.

vshcherb:
No it doesn't look it's address search but probably POI search and the radius should be decreased to speedup.

sonora:
Here's a funny experiment: Display one or several recorded tracks  which also cross your current position or a Favorite;
- Tap on MyPosition or the Favorite: The multi-selection menu with MyPosition, the Favorites and the track(s) pops up instantly.
[...]

PR #22678, opened 2025-05-28
Implement bbox inflate (Utils) and getAllPolygonsBounds (WorldRegion)

ivanPyrohivskyi:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21954

PR #22691, opened 2025-05-30
fix #22535

### LINKED ISSUE #22535: Deleted Travel track continues to be displayed on the map
### Description
A previously selected track from Travel routes (Cycle, Hiking, etc.) continues to be displayed on the map as loaded after reopening it + it cannot be deleted anymore (In fact, the track is deleted from the memory (it is not displayed in the Travel folder), but continues to be displayed on the map)

### Steps to reproduce

1. Configure map > Cycle routes > Enable > Go back to the map > Select any route > Download > Options > Delete > Select again = **No offer to download the route (displayed as downloaded)** > Options > Delete = **Nothing happening**

### Actual result

<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/954317c6-c7eb-41bb-b30a-0f5dbab42133
"></video>
</details> 

### Expected result

- [ ] 1. After each deletion, the route should be displayed as Unloaded and it should be possible to download the route

- [ ] 2. If the route is displayed on the map as loaded, it should be possible to delete the route (**most likely, this is a derivative of the 1st point**)

### Your Environment
OsmAnd Version: OsmAnd~ 5.1.0#5156m, Released: 2025-05-08

0xRe1nk0:
- deleted Travel track continues to be displayed on the map

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/NetworkRouteSelectionLayer.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/NetworkRouteSelectionLayer.java
@@ -209,7 +209,8 @@ public class NetworkRouteSelectionLayer extends OsmandMapLayer implements IConte
 
 	private RouteKey findKeyByValue(@NonNull GpxFile gpxFile) {
 		for (Map.Entry<RouteKey, GpxFile> entry : routesCache.entrySet()) {
-			if (entry.getValue() == gpxFile) {
+			if (entry.getValue() == gpxFile ||
+					Algorithms.stringsEqual(entry.getValue().getPath(), gpxFile.getPath())) {
 				return entry.getKey();
 			}
 		}


OsmAnd/src/net/osmand/plus/views/layers/NetworkRouteSelectionLayer.java
```diff
@@ -194,4 +194,25 @@ public void collectObjectsFromPoint(@NonNull MapSelectionResult result, boolean
...
+		}
+
+		if (routeKey != null) {
+			routesCache.remove(routeKey);
+		}
+	}
+
+	private RouteKey findKeyByValue(@NonNull GpxFile gpxFile) {
+		for (Map.Entry<RouteKey, GpxFile> entry : routesCache.entrySet()) {
+			if (entry.getValue() == gpxFile) {
```
Chumva: Is it a correct comparison? Could we check by path?

PR #22702, opened 2025-06-02
Fix retrieving geometry

ivanPyrohivskyi:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22708

PR #22743, opened 2025-06-09
Allow click on snowmobile trails in ClickableWay

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2894

PR #22747, opened 2025-06-10
Implement click test2

Corwin-Kh:


OsmAnd/test/assets/clicks.json
Chumva, line 1: Do we need this in the project, or will it be added programmatically before test?

OsmAnd/src/net/osmand/plus/utils/FileUtils.java
```diff
@@ -453,6 +476,28 @@ public static File getBackupFileForCustomAppMode(@NonNull OsmandApplication app,
...
+		if (outputFile.exists()) {
+			outputFile.delete();
+		}
+		FileOutputStream fos = null;
+		try {
+			fos = new FileOutputStream(outputFile);
+			fos.write(jsonString.getBytes());
+		} catch (IOException e) {
+			PlatformUtil.getLog(FileUtils.class).error(e);
+		} finally {
```
Chumva, line 490: We could use Algorithms.streamCopy & Algorithms.closeStream for cleaner code

OsmAnd/src/net/osmand/plus/mapcontextmenu/MenuBuilder.java
```diff
@@ -450,7 +450,7 @@ protected void buildWithinRow(ViewGroup viewGroup) {
 				buildDetailsRow(rowContainer, getRowIcon(R.drawable.ic_action_pin_location),
 						app.getString(R.string.transport_nearby_routes), null,
 						MenuObjectUtils.getMenuObjectsNamesByComma(menuObjects),
-						getWithinCollapsableView(menuObjects), true, null);
+						getWithinCollapsableView(menuObjects), true, null, "ic_action_pin_location");
```
Chumva, line 453: Looks somewhat harder to read/maintain, as we have to pass one more parameter, same as icon

PR #22778, opened 2025-06-13
fix #21456 ui freeze

### LINKED ISSUE #21456: Search History deletion has glitches
I recently noticed that I had over 1400 entries in my search history, at least that's the number the dialog claimed after long-tapping one, then tapping the 'Select all' button.

- [ ] I selected all, and hit delete. After a longer wait time and an 'app not responding'  OS warning, the list seemed finally empty. But after an app restart and re-entering the search, though, the history seemed to (still or again) have 156  entries. I repeated the procedure to delete these also 
- [ ] After re-start, there is again 1 entry in the search history (a rather old one from May, potentially the oldest I had). That entry seems undeletable, it always shows up after re-starts (no matter if deleted alone or together with other,new ones).
- [ ] I wonder if we should cap the number of entries kept in the search history and not let them increase indefinitely?

0xRe1nk0:
-async history deleting with progress dialog

### DIFF
--- a/OsmAnd/src/net/osmand/plus/settings/fragments/DeleteHistoryTask.java
+++ b/OsmAnd/src/net/osmand/plus/settings/fragments/DeleteHistoryTask.java
@@ -1,45 +1,103 @@
 package net.osmand.plus.settings.fragments;
 
-import android.os.AsyncTask;
-
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.fragment.app.FragmentActivity;
 
-public class DeleteHistoryTask extends AsyncTask<Void, Void, Void> {
+import net.osmand.plus.R;
+import net.osmand.plus.base.BaseLoadAsyncTask;
+import net.osmand.plus.helpers.SearchHistoryHelper;
+import net.osmand.plus.helpers.TargetPoint;
+import net.osmand.plus.mapmarkers.MapMarker;
+import net.osmand.search.core.SearchResult;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+public class DeleteHistoryTask extends BaseLoadAsyncTask<Void, Void, Void> {
+
+	private final Set<Object> selectedItems;
+	private final DeleteHistoryType historyType;
 	private final DeleteHistoryListener listener;
 
-	public DeleteHistoryTask(@Nullable DeleteHistoryListener listener) {
+	public DeleteHistoryTask(@NonNull FragmentActivity activity, @NonNull DeleteHistoryType historyType,
+	                         @NonNull Set<Object> selectedItems, @Nullable DeleteHistoryListener listener) {
+		super(activity);
+		this.selectedItems = selectedItems;
+		this.historyType = historyType;
 		this.listener = listener;
 	}
 
-	@Override
-	protected void onPreExecute() {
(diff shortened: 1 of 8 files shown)


OsmAnd/src/net/osmand/plus/settings/fragments/DeleteHistoryTask.java
```diff
@@ -0,0 +1,45 @@
+package net.osmand.plus.settings.fragments;
+
+import android.os.AsyncTask;
+
+import androidx.annotation.Nullable;
+
+public class DeleteHistoryTask extends AsyncTask<Void, Void, Void> {
```
Chumva: Better extend BaseLoadTask for progress dialog

OsmAnd/src/net/osmand/plus/settings/fragments/DeleteHistoryTask.java
```diff
@@ -0,0 +1,103 @@
...
+
+	private void deleteMarkerHistory() {
+		List<MapMarker> mapMarkers = new ArrayList<>();
+		for (Object item : selectedItems) {
+			if (item instanceof MapMarker) {
+				mapMarkers.add((MapMarker) item);
+			}
+		}
+		app.getMapMarkersHelper().removeMarkers(mapMarkers);
+	}
```
Chumva: Probably we can combine it into a single method with foreach & several instanceof

OsmAnd/src/net/osmand/plus/settings/fragments/HistoryItemsFragment.java
```diff
@@ -69,6 +73,8 @@ public abstract class HistoryItemsFragment extends BaseOsmAndDialogFragment impl
 	private boolean locationUpdateStarted;
 	private boolean compassUpdateAllowed = true;
 
+	private ProgressDialog progressDialog;
```
Chumva: Is it needed here?

OsmAnd/src/net/osmand/plus/settings/fragments/HistoryItemsFragment.java
```diff
@@ -264,7 +270,12 @@ public void onCategorySelected(List<Object> items, boolean selected) {
 
 	@Override
 	public void onDeletionConfirmed() {
-		deleteSelectedItems();
+		DeleteHistoryTask deleteHistoryTask = new DeleteHistoryTask(requireActivity(), getDeleteHistoryType(), selectedItems, this);
+		deleteHistoryTask.setShouldShowProgress(true);
```
Chumva: setShouldShowProgress could be set inside constructor

OsmAnd/src/net/osmand/plus/settings/fragments/HistoryItemsFragment.java
```diff
@@ -264,7 +270,12 @@ public void onCategorySelected(List<Object> items, boolean selected) {
 
 	@Override
 	public void onDeletionConfirmed() {
-		deleteSelectedItems();
+		DeleteHistoryTask deleteHistoryTask = new DeleteHistoryTask(requireActivity(), getDeleteHistoryType(), selectedItems, this);
```
Chumva: We added async deletion; won't it lead to some errors or crashes in SearchHistoryHelper?

PR #22795, opened 2025-06-16
Add thread-safe OsmandRegions.prepareFileAtomic() for tools

RZR-UA:


RZR-UA:
will be moved to tools

PR #22798, opened 2025-06-16
Unify fragment inheritance via shared base class

nazar-kutz:


Chumva:
Moved to [another request](https://github.com/osmandapp/OsmAnd/pull/23218)

PR #22809, opened 2025-06-18
Update Turkish

mikropsoft:
Both files were translated by syncing with the English file, so there should be no missing lines. The strange thing is that, despite no change in the number of lines next to it, it appeared split in the commit, which makes it look complicated.

Ekran Görüntüsü (155)
Ekran Görüntüsü (156)

If approved, I’d appreciate it if you could transfer this to Weblate. Since I translated it from scratch, I had to open a PR.


Chumva:
@mikropsoft It is better to use Weblate for translations

mikropsoft:
> @mikropsoft It is better to use Weblate for translations

It’s not important whether you accept the PR. I’d be grateful if you could upload it yourself. Since I prefer Crowdin, I’m not very familiar with Weblate, and the last time I used it, I couldn’t find an upload button for files in the projects.

PR #22822, opened 2025-06-20
Updated tiles and symbols debug info

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd-Issues/issues/3023

PR #22826, opened 2025-06-20
Try to fix tunnel issues (SimulationProvider)

RZR-UA:


RZR-UA:

OsmAnd/src/net/osmand/plus/simulation/SimulationProvider.java
```diff
@@ -47,15 +49,20 @@ public void startSimulation() {
...
+			return -1;
+		}
 		for (int i = currentRoad; i < roads.size(); i++) {
 			RouteSegmentResult road = roads.get(i);
 			boolean firstRoad = i == currentRoad;
 			boolean plus = road.getStartPointIndex() < road.getEndPointIndex();
-			for (int j = firstRoad ? currentSegment : road.getStartPointIndex() + 1; j <= road.getEndPointIndex(); ) {
+			for (int j = firstRoad ? currentSegment : (road.getStartPointIndex() + (plus ? +1 : -1));
+			     plus ? j <= road.getEndPointIndex() : j >= road.getEndPointIndex();
+			     j += plus ? +1 : -1) {
```
mwerle, line 62: Nitpick : I would have done:

```
int inc = plus ? +1 : -1
...
someVar + inc;
```

instead of all the `someVar + (plus ? +1 : -1)`

More concise, easier to read, and more performant.

Further down I see it's inverted, but that could also be easily covered with `someVar - inc` instead which would also more closely match the original `- 1` logic.

RZR-UA: The whole `SimulationProvider.java` class should be refactored soon.

PR #22828, opened 2025-06-20
Avoid SIMULATED_PROVIDER_TUNNEL for routing recalc (tunnels issue)

RZR-UA:


RZR-UA:

PR #22846, opened 2025-06-23
Optimize SimulationProvider.java for performance

sonora:


RZR-UA:

sonora:
- Removes operations which need not be in the loop to before the loop
- Minimizes object calls
- Minimize arithmetics

sonora:
@RZR-UA I guess optimizing the Simulation Provider is not really a priority, so we can close this PR unmerged?

RZR-UA:
@sonora, the algorithm in the SimulationProvider class is somewhat too fragile to changes. I'm sure this is why this request was passed over for months. In addition, Victor's modifications to the algorithm cause unresolved conflicts in your request. As for me, I don't think that kind of road.getObject() (get reference) could improve performance significantly.

We should have implemented unit tests for `SimulationProvider.java`; previously, it took long hours of driving in real tunnels to catch bugs such as crashes and map coordinates jumping to 0, 0 (x, y).

If we implemented unit tests, we would additionally improve the algorithm's readability and optimization.

sonora:
@RZR-UA Agree - thanks! Closing this to move on to other optimization priorities.  :wink:

PR #22854, opened 2025-06-24
Fixed crash getting content limit

Corwin-Kh:


OsmAnd/src/net/osmand/plus/auto/screens/SearchResultsScreen.java
```diff
@@ -37,6 +37,7 @@ public final class SearchResultsScreen extends BaseSearchScreen implements Defau
 	private boolean loading;
 	private boolean destroyed;
 	private boolean showResult;
+	private boolean firstTimeGetTemplate = true;
```
Chumva: We have the same variable in three different screens; it could probably be moved to the upper level?

OsmAnd/src/net/osmand/plus/auto/screens/BaseAndroidAutoScreen.kt
```diff
@@ -35,19 +37,26 @@ abstract class BaseAndroidAutoScreen(carContext: CarContext) : Screen(carContext
...
+			contentLimit = DEFAULT_CONTENT_LIMIT
+		}
 	}
 
+	final override fun onGetTemplate(): Template {
+		initContentLimit()
+		return getTemplate()
+	}
+
+	abstract fun getTemplate(): Template
```
alex-dev-neo, line 58: Drop it, call super at onGetTemplate()

OsmAnd/src/net/osmand/plus/auto/screens/SearchResultsScreen.java
alex-dev-neo: Introduce onFirstGetTemplate() method at base class

Issue #21573, opened 2024-12-16
Bad suggestion with accentuated POI type


luigifab:
### Description

When you add a new POI, type bike shop (Magasin de vélos).

if you write: magasin de v
you get: 'magasin de vélos' in the list

if you write: magasin de ve
you get: nothing

if you write: magasin de vé
you get: 'magasin de vélos' in the list

it's ridiculous :)

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/c8243eb8-0157-4eb7-ba61-a3f16407520a"  width="300" height="auto"/> </th>
  </tr>
</table>

### Steps to reproduce

.

### Actual result

.

### Expected result

.

### Your Environment (required)

```
OsmAnd Version: 4.9.10
Android/iOS version: 11
Device model: wiko
```

luigifab:
There is the same problem when you use the search:

![Capture d'écrans_20241216-110342](https://github.com/user-attachments/assets/349a57b4-4de9-46eb-9783-351aa095811b)

![Capture d'écrans_20241216-110354](https://github.com/user-attachments/assets/4e10fa22-bb60-4a08-ba0e-59f2925a6a56)

![Capture d'écrans_20241216-110402](https://github.com/user-attachments/assets/56628f95-0876-4696-a4f5-7d7f476b8b46)

vshcherb:
- [x] Problem with search in poi types confirmed. Looks like Collator is not used when search in types.
P.S. General search doesn't have this problem just type "velo". However general search has problem with "de" but that's another issue to be reported.

Chumva:
[Looks like poi type retrieved by key](https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/plugins/osmedit/data/EditPoiData.java#L118-L126)

dmpr0:
<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/5b43ed19-7443-4a4b-9b41-a2fb57fa0a0c"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/2d6ef6b5-4d75-4de1-b8b5-3603cc8c48d4"  width="300" height="auto"/> </td>
  </tr>
</table>   

Fixed

PR #22899, opened 2025-06-30
#21573: Implemented collator filtering

Corwin-Kh:


OsmAnd/src/net/osmand/plus/utils/CollatorFilteredAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+			@Override
+			protected FilterResults performFiltering(CharSequence constraint) {
+				FilterResults filterResults = new FilterResults();
+				List<String> suggestions = new ArrayList<>();
+
+				if (Algorithms.isEmpty(constraint)) {
+					suggestions.addAll(originalCollection);
+				} else {
+					String filterPattern = constraint.toString().toLowerCase().trim();
+					CollatorStringMatcher matcher = new CollatorStringMatcher(filterPattern, CHECK_STARTS_FROM_SPACE);
```
Chumva: Several other Filters use [NameStringMatcher(namePart, CHECK_CONTAINS)](https://github.com/osmandapp/Osmand/blob/dfac98bc21d4bf713c9fe1f7ca298f3c09ac4151/OsmAnd/src/net/osmand/plus/download/local/dialogs/LocalSearchFilter.java#L43), probably here we need the same one, if not specified otherwise

PR #22903, opened 2025-07-02
Added PoiType suggestions test

Corwin-Kh:


OsmAnd/test/java/net/osmand/test/common/AppSettings.java
```diff
@@ -21,4 +24,13 @@ public static void showWikiOnMap(@NonNull OsmandApplication app) {
 			helper.addSelectedPoiFilter(filter);
 		}
 	}
+
+	public static void setLocale(@NonNull OsmandApplication app, String language, String country) {
```
Chumva: Should we use localeHelper here?

PR #22904, opened 2025-07-02
fix crop text, updating container on new text

0xRe1nk0:


Chumva:
It should already be fixed [here](https://github.com/osmandapp/OsmAnd/pull/23014)

Issue #22515, opened 2025-05-05
Problem navigating from Prague airport to old town and reverse


andrixnet:
### Routing engine

- [x] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile

Public transport. 
Upper limit on changes = 2
consider temporary limitations = yes
avoid transport types = none selected

### Start and end points

Start = Vaclav havel airport, Prague (later enhanced to terminal 2 public transport stop; same result)
End = Hotel Ibis Praha old town

### Actual route

route 1 = walk + AE + 911 + walk
route 2 = walk + AE + 908 + walk
route 3 = walk + AE + 907 + walk
route 4 = walk + AE + 905 + walk
route 6 = walk + AE + walk
route 7 = 100 + 851 + walk
route 8 = 100 + X913 + walk
route 9 = 100 + X913 + walk (apparent duplicate)
route 10 = 100 + 915 + walk
route 11 = 907 + walk
route 12 = 910 + walk
route 13 = 191 + walk

Note1: all public transport selections are buses.
Note2: there is no distinction displayed by OsmAnd about night-only public transport

### Expected route

Best route (one we actually took based on local info): 
START = Termina 2 - take bus 59 (never offered by OsmAnd, though if i clicked on map, on station, it showed as having stop there)
CHANGE_1 = Divoka Sarka - change to tram 26
END = Namesti Republiky + walk 1 min to hotel

### Maps you used (online or offline)

- [x] Offline maps offered within the OsmAnd app for download. 
czech-republic-20250423.osm.pbf manually converted with the Java converter using full options (all 5 checkboxes checked in UI)

### Your Environment

```
OsmAnd Version: 4.9.10
Android version: 9
Device model: Samsung Galaxy S9+

```

### Anything else?

The same was true for the reverse trip (hotel to airport). We used tram 26 + bus 59. Nothing like what OsmAnd offered.

yuriiurshuliak:
**OsmAnd fails to suggest the correct public transport route between Prague Airport and Ibis Praha Old Town — specifically the combination of **bus 59 + tram 26**, which is available in data but not used in routing.**

**The bug has been reproduced.**
To reproduce, select a **public transport profile**, load the **Prague map**, and build a route starting from **Václav Havel Airport Prague (PRG)** to **Ibis Praha Old Town**. Observe that **bus 59 and tram 26 are not suggested** despite being present in the data.

[...]

PR #22908, opened 2025-07-02
Fix public transport search

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/22515

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
@@ -2,11 +2,15 @@ package net.osmand.router;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -54,40 +58,57 @@ public List<TransportRouteResult> buildRoute(TransportRoutingContext ctx, LatLon
...
 		while (!queue.isEmpty()) {
 			long beginMs = MEASURE_TIME ? System.currentTimeMillis() : 0;
 			if (ctx.calculationProgress != null && ctx.calculationProgress.isCancelled) {
 				return null;
 			}
 			TransportRouteSegment segment = queue.poll();
 			TransportRouteSegment ex = ctx.visitedSegments.get(segment.getId());
+			TransportRouteSegment finish = null;
 			if(ex != null) {
+				TransportRouteSegment finalized = finalizedSegments.get(segment.getId());
```
vshcherb, line 82: Technically it's not correct cause segment could be finalized later and this event is independent from that current point is visited

PR #22909, opened 2025-07-03
Cleared vector lines after text size changed

ihoryakovenko:


ihoryakovenko:

Issue #22821, opened 2025-06-20
Add "Show icon" option for L / R panel widgets


dmpr0:
## Description
This was somehow missed in the original task (https://github.com/osmandapp/OsmAnd/issues/21814), but the ability to hide the widget icon for the **Left** and **Right** panels was lost.

## ToDo

- [x] Add "**Show icon**" option
- - [x] For the **Left** / **Right** panels
- - [x] Only for **Medium** and **Large** sizes

Zirochkabila:
# Testing 1
`OsmAnd~ 5.1.0#5541m, Released: 2025-07-07`

Checked the ability to add a show/hide icon for widgets of the right/left panels + display after profile export - **Added**

dmpr0:
## UI Review

- [x] Show "Show icon" parameter for small widgets in Top / Bottom panels.

PR #22911, opened 2025-07-03
#22821: Implemented show icon setting for left-right widgets

Corwin-Kh:


OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/settings/BaseSimpleWidgetInfoFragment.java
```diff
@@ -55,13 +66,11 @@ private void updateShowIcon(boolean shouldShowIcon, SwitchCompat switchCompat) {
...
 	@Override
 	protected void applySettings() {
-		if (isVerticalPanel) {
-			shouldShowIconPref.set(showIcon);
-			if (widgetInfo != null) {
-				if (widgetInfo.widget instanceof SimpleWidget simpleWidget) {
-					simpleWidget.showIcon(shouldShowIconPref.get());
-					//simpleWidget.recreateView();
-				}
+		showIcon = showIcon || isSmallHeight();
```
Chumva: Do we need to check for isSmallHeight() & change preference because of it?

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/SimpleWidget.java
```diff
@@ -288,7 +288,6 @@ protected View.OnClickListener getOnClickListener() {
 	}
 
 	public void showIcon(boolean showIcon) {
-		AndroidUiHelper.updateVisibility(imageView, showIcon);
```
Chumva, line 291: Looks strange, setter that only calls invalidate?

Issue #22526, opened 2025-05-07
Tracks disappears after re-connecting to AA


zipav:
### Description

Our user left a complaint that after reconnecting to Android Auto (AA), all visible tracks disappear from the map screen when vehicle screen switch from AA to Rearview camera and back.

### Steps to reproduce

* Connect to  AA
* Switch to rearview camera
* Switch back to OsmAnd on AA

### Actual result

* Previously visible tracks are disappearing

### Expected result

* All settings and visible layers remain unchanged

Nakedsnake88:
I hope it will be solved as soon as possible

Zirochkabila:
`OsmAnd~ 5.1.0#5156m, Released: 2025-05-08`

Steps:

1. Ask me user's file > Import files > Leave active only the profiles shown in the screenshot > Display imported tracks on the map > Switch between YouTube Music and OsmAnd


<details><summary>Details Profile</summary>
<img width="400" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/5748f44c-54ac-4bf7-95a9-22d7fcbc537c">
</details> 

<details><summary>Details result</summary>
<video src="https://github.com/user-attachments/assets/00b4c5b0-9242-45d9-8932-3be06b83ef0c
"></video>
</details>

Nakedsnake88:
I’m currently using version 4.9, which doesn’t have the issue of track visibility when switching between apps or to the vehicle’s rear camera. The video that was posted shows exactly what I want to avoid — having to re-enable the tracks every time OsmAnd in Android Auto goes into the background.

Zirochkabila:
# Testing 1
`OsmAnd~ 5.1.0#5498m, Released: 2025-06-30`

Checked the display of tracks on the AA map after reconnection - **Fixed**
@Nakedsnake88, please install the [nightly](https://osmand.net/docs/versions/nightly_versions#nightly-builds) version and try it on your own device 

_____________

- [x] Waiting for AT

PR #22919, opened 2025-07-04
atest for #22526

0xRe1nk0:


OsmAnd/test/java/net/osmand/test/ui/map/CleanGpxLayerResourcesTest.java
```diff
@@ -0,0 +1,155 @@
...
+			});
+		} catch (IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	@Test
+	public void test() throws Throwable {
+		skipAppStartDialogs(app);
+		moveMapToGpx(importedGpxFile);
```
Chumva: This part could fail if the import is not finished

Issue #22178, opened 2025-03-27
Solve the HH Routing Map Version Quagmire


wookey:
### Description

I have all of the UK's 'Standard Map's downloaded from 1st January. I had updated my own area (East of England) to 1st March. At this point I found that any longer route (crossing between standard maps), would give a long dotted line to northern France, a route through Normandy and a long dotted line back to the UK destination (Oxfordshire in my example). Or if I try to route to Newcastle (North-east England), I get a dotted line to somewhere in the North Sea, then a route in to Newcastle. I did get messages saying 'Missing or outdated maps', but my maps are only one or two months old which is not really 'outdated' at all. And they are certainly not missing. The bit that seems definitely to be not working is that when I click on 'use downloaded maps' it does not use the downloaded 'East of England (March)' or London or Southern England (Jan) maps, but _does_ use the (Feb, Normandy) map. Why does it use the 1-month old France maps, with a 150-mile 'leap' but not the adjacent 'London' or 'Southern England' 2-month old maps? Was there some kind of discontinuity/recalculation between these dates such that ways cannot be lined up?

This is very unhelpful behaviour. I have expensive data so never want to download maps over a phone connection after I've set off, only on wifi at home.   Using whatever maps are available should work if 'use downloaded maps' is clicked. If there is some kind of fundamental incompatibility between different maps then a message about that to make clear what it going on would be better than 'leaping' to France and back. 

I do not expect to be told that a 2 or 3 month old map is 'too old' for routing. Maybe a year old would make sense. But maybe that 'too old' message actually means 'formats have changed and these maps are incompatible for routing purposes'? 

I can still route _within_ an older (Jan or Feb 2025) standard map area. And I can route between Wales, (Jan) and 'West Midlands', (Jan), which adds weight to the idea that there was some kind of incompatible change between Jan 2025 and Feb 2025, (given that routes including Normandy (Feb) and East of England (March) route successfully.  Is that the actual issue here?

I hope this report is useful (I was very confused for a while but eventually worked out what was going on). My version of OSMand is 4.9.10, fdroid build. I checked and that still seems to be current. 

I will now update all my maps to have consistent dates and make this go away so I can't easily go back and reproduce. Sorry. 

### Steps to reproduce

1) Have all UK 'Standard Map's from 1st January 2025 downloaded. Have Normandy and Brittany and Hauts de France Standard Maps from 1st Feb 2025 downloaded. Download East of England from 1st March 2025.
[...]

PR #22920, opened 2025-07-04
Improve HH containsStartEnd detection (Java part)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22178

OsmAnd-java/src/main/java/net/osmand/router/RoutingContext.java
```diff
@@ -75,7 +75,8 @@ public class RoutingContext {
 	public int dijkstraMode;
 	public boolean publicTransport;
 	public HashSet<BinaryMapIndexReader> mapIndexReaderFilter = new HashSet<>();
-	
+	public String[] regionsWithAllRoutePoints = new String[0];
```
vshcherb: Map<String, Integer> regionsContainsRoutePoints  // value number of start / end points is connected

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -718,6 +718,18 @@ public boolean contains(LatLon p) throws IOException {
 			}
 			return false;
 		}
+
+		public boolean containsRegion(List<String> regionsWithAllRoutePoints) {
+			for (BinaryMapIndexReader reader : readers) {
```
vshcherb, line 726: Needs to check section I believe not file names 
regionsContainsRoutePoints.get()

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -718,6 +718,18 @@ public boolean contains(LatLon p) throws IOException {
 			}
 			return false;
 		}
+
+		public boolean containsRegion(List<String> regionsWithAllRoutePoints) {
```
vshcherb: public int getMaxRegionContainsCounter(Map<....>)
return int // max value per section

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -740,8 +752,9 @@ private HHRoutingContext<T> selectBestRoutingFiles(LatLon start, LatLon end, HHR
 				}
 			}
 		}
+		List<String> regionsWithAllRoutePoints = Arrays.asList(hctx.rctx.regionsWithAllRoutePoints);
 		for (HHRouteRegionsGroup<T> g : groups) {
-			g.containsStartEnd = g.contains(start) && g.contains(end);
+			g.containsStartEnd = g.contains(start) && g.contains(end) && g.containsRegion(regionsWithAllRoutePoints);
```
vshcherb: int regionCnt = getMaxRegionContainsCounter();

&& (regionCnt == 0 || regionCnt > 1); // take into account that user can have different maps with uknown names or region names could change between releases

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
RZR-UA: d6e396bd98

OsmAnd-java/src/main/java/net/osmand/router/RoutingContext.java
RZR-UA: It would be overcomplicated, especially with JNI.

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
RZR-UA: The simplest way to clarify containsStartEnd is to identify the regions that truly contain start and end points, using ray tracing based on their polygons. This is actually already done. In rare cases of user-defined maps and/or renamed regions with old maps, we lose containsStartEnd, but still have sumIntersects based on region bbox intersection with start-end bbox.

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
RZR-UA: see next comment plz

Issue #22910, opened 2025-07-03
Regression: Map Center off Screen Center


sonora:
Not sure when this started, but there are now several scenarios when the map center (the point on the map screen where your position is displayed after tapping the My Position widget) is gets shifted off the screen screen center, though set to be in the middle

**Configuration:** Device and Osmand is in Portrait orientation. V1 Renderer (V2 not tested). Use the Browse map profile in North is up (fixed). My Position display set to be centered. Configure screen is set to show the Route information widget at the bottom (only when a route is shown, of course).

**Reproduce Example Scenario A:**
1. Tap My Position: Shows in screen center
2. Tap the navigation button and calculate a route to anywhere (do not enter turn-by-turn mode).
3. Once the route is calculated, tap into the map to hide the route config dialog.
4. Tap My Position. It will now show roughly in the middle of the screen minus the height of the Route information widget, i.e. a little higher than before (probably ok in this view)
5. But now tap the navigation button and stop the route.
6. My Position on the screen will remain in the higher position, even after panning and re-centering, until the next app start or a profile change.
 
**Scenario B, even more blatant:**
1. Repeat 1-4 above.
5. Zoom out the map to a view port where you know it will contain 1 or a few search hits of the following:
6. Tap the search button
7. Type a brand, or select a category like "filling station"
8 Tap 'Show all on map'
9. Back on the map screen, the screen center is now shifted significantly higher (on my device to the upper quarter of the screen!). Like the centering takes into account an (invisible), half expanded POI selection dialog or context menu at the bottom

Step 2 is crucial for the reproduction of Scenario B. As long as you display a map section where zero search hits are located, things are ok.

Zirochkabila:
`OsmAnd~ 5.1.0#5516m, Released: 2025-07-03`

**Configuration:** 

1. Device and Osmand is in Portrait orientation. (V1 Renderer or V2 - it's doesn't matter)
2. Use the Browse map profile in North is up (fixed)
3. My Position display set to be centered
4. Configure screen is set to show the **Route information widget at the bottom** (only when a route is shown, of course).
5. Take note of where the center point of the map is located when you click on My Location

**Steps to reproduce - Plan A:** 

1. Navigation > Build the route > Back to map = **Center of the map is little bit higher (thats expected)**
2. Navigation > Stop = **Center of the map is still a little bit higher (not as expected)**
3. Reopen OsmAnd = **Center of map return to previous position (as Confiiguration)**

[...]

PR #22923, opened 2025-07-04
Fix "Map Center off Screen Center #22910"

nazar-kutz:


OsmAnd/src/net/osmand/plus/utils/AndroidUtils.java
```diff
@@ -888,26 +888,34 @@ public static void switchNavBarVisibility(@NonNull Activity activity) {
 
 	@NonNull
 	public static Rect getViewBoundOnScreen(@NonNull View view) {
-		int[] pixel = getLocationOnScreen(view);
-		int left = pixel[0];
-		int top = pixel[1];
-		return new Rect(left, top, left + view.getWidth(), top + view.getHeight());
+		if (view.getVisibility() == View.VISIBLE) {
```
Chumva: Do we need to get the invisible views rect? Probably better check != GONE

PR #22933, opened 2025-07-07
Used ruler by tap text size settings for v2 distance marker

ihoryakovenko:


ihoryakovenko:

Issue #22074, opened 2025-03-14
Inconsistent result between Search and  `Show on map` (Android / iOS)


Zirochkabila:
### Description

[Probably](https://github.com/osmandapp/OsmAnd/issues/22045) related

Adjective `St` displays a lot of POIs on the map after clicking the `Show on map` button
A good example is the search for `St. Volodymyr's Cathedral`
1. When searching for `Volodymyr's Cathedral`, only one POI is displayed (the one that should be displayed)
2. When searching for `St.`, many POI around the world are displayed

[OSM data](https://www.openstreetmap.org/way/43383147#map=18/50.444883/30.509242)

### Steps to reproduce
1. Go to the Kyiv map
2. Search > `St. Volodymyr's Cathedral` > Show on map button = Many different POIs will displayed on the map
3. Search > `St.` > Show on map button = Many different POIs will displayed on the map (**Same**)
4. Search > `Volodymyr's Cathedral` > Show on map = Only one POI will showed

### Actual result

<details><summary>2nd step - St. Volodymyr's Cathedral</summary>
<table style="width:100%">
  <tr>
    <th>Android result "St. Volodymyr's Cathedral"</th>
    <th>iOS result "St. Volodymyr's Cathedral"</th>
  </tr>
  <tr>
    <td><video source src="https://github.com/user-attachments/assets/57090b45-7146-425f-9a95-9f35a2718cab
"  width="300" height="auto"/></td>
    <td><video source src="https://github.com/user-attachments/assets/8e6cb972-fb79-4ec7-8343-b8c13c624109
"  width="300" height="auto"/> </td>
  </tr>
</table> </details> 

<details><summary>3rd step - St.</summary>
<table style="width:100%">
  <tr>
    <th>Android result "St."</th>
    <th>iOS result "St."</th>
  </tr>
  <tr>
    <td><video source src="https://github.com/user-attachments/assets/7822eabe-2e1a-4978-8c57-7588af2c61dd
"  width="300" height="auto"/></td>
    <td><video source src="https://github.com/user-attachments/assets/977eadf3-8b97-4e34-af7f-6b74d430a8cb
"  width="300" height="auto"/> </td>
  </tr>
</table> </details> 

### Expected result

- [ ] As expected, only one (correct) POI is displayed

<details><summary>4th step - Volodymyr's Cathedral</summary>
<table style="width:100%">
  <tr>
    <th>Android result "Volodymyr's Cathedral"</th>
[...]

Issue #22089, opened 2025-03-17
Additionally consider alt_name in the search of objects


gorgonz:
### Describe the idea

It is about finding objects in the app.
The search algorythm only considers the tag name and ignores alt_name. This results in a lot of objects, that are not found. Though the sense of it is obvious, I will add an example:
In Nürnberg is a (museum) tower called "turmdersinne", which is an intended unique style of writing it. That is why it is the tagged value of "name". If you don't know this, You will search for "Turm der Sinne", which is the value of tag "alt_name". The search for the latter will not find any results.

### Expected behaviour

Searching a prompt in OSMAnd should also consider alt_name to find results

### Alternatives you've considered

I don't know any

yuriiurshuliak:
**Objects with an `alt_name` tag are not found via search in OsmAnd, only when the exact `name` tag is used. This limits discoverability of places known by alternate or natural names.**

**The bug has been reproduced.**

To reproduce:  
- Go to coordinates: **49.452220, 11.069498**  
- Search for: **Turm der Sinne** → ❌ No results  
- Search for: **turmdersinne** → ✅ Object found  

🔗 OSM object: [https://www.openstreetmap.org/node/1539266304](https://www.openstreetmap.org/node/1539266304)  
🔍 Tag info (`alt_name`): [https://taginfo.openstreetmap.org/keys/alt_name](https://taginfo.openstreetmap.org/keys/alt_name)


OsmAnd~ 5.1.0#5007m, released: 2025-04-15


<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/4fe991ac-d591-4b43-92bd-d8f3be51157b"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/5392f31d-f7e9-4cc1-ac27-dbf118c71952"  width="300" height="auto"/> </th>
  </tr>
</table>

vshcherb:
Please double check the search, it might be an issue only with show on map functionality (Android / iOS)

yuriiurshuliak:
**Tested on both Android and iOS.**  
Confirmed that the issue is with the **search logic itself**, not just with showing results.

- Search: **"Turm der Sinne"** → ❌ No result  
- Search: **"turmdersinne"** → ✅ Returns the correct object

This confirms that `alt_name` is currently **not considered in the search algorithm** on either platform. The object is not found at all when using the `alt_name` value.


[...]

PR #22986, opened 2025-07-11
Fix Show on map

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/22089

Perhaps also will fix: https://github.com/osmandapp/OsmAnd/issues/22074

### DIFF
--- a/OsmAnd/src/net/osmand/plus/search/SearchUtils.java
+++ b/OsmAnd/src/net/osmand/plus/search/SearchUtils.java
@@ -51,8 +51,8 @@ public class SearchUtils {
 					}
 				} else {
 					filter = app.getPoiFilters().getSearchByNamePOIFilter();
-					if (!Algorithms.isEmpty(searchPhrase.getFirstUnknownSearchWord())) {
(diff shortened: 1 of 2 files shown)


OsmAnd/src/net/osmand/plus/search/SearchUtils.java
```diff
@@ -51,8 +51,8 @@ public static PoiUIFilter getShowOnMapFilter(@NonNull OsmandApplication app, @No
 					}
 				} else {
 					filter = app.getPoiFilters().getSearchByNamePOIFilter();
```
vshcherb, line 53: What SearchUtils is about? Feels wrong class design

ivanPyrohivskyi: Used only in UI, as add-ons to QuickSearchDialogFragment.
Perhaps not relevant name
Знімок екрана 2025-07-11 о 17 10 17

vshcherb: SearchUtils should be refactored no static methods combined / split with PoiUIFilter differently

Issue #22824, opened 2025-06-20
Search POI - wrong category depends on specific maps to be loaded


DmitryAlexei:
Follow-up for https://github.com/osmandapp/OsmAnd/issues/22265

> **Steps to reproduce**
> 
> **1. Example**
> - Load map Deutschland Nordrhein-Westfalen Münster (Germany North Rhine-Westfalia Münster)
> - To be sure to reproduce deactivate all other maps, force close the app, restart and give osmand some time to reindex the search
> - Switch language to german
> - Go to the area of the downloaded map
> - Search for "Spielplatz" and then you found categorie "Spielplatz-Gastronomie" instead of "Spielplatz-Freizeit"
> - Typing on the categorie "Spielplatz-Gastronomie"
> - If nothing is found press "Suchradius erweitern" (extend search radius)
> - You should find then "Sitzplätze im Freien, Spielplatz" (OSM-Way 757041132)
> - By checking the osm-way (757041132) you can find the tagging "leisure=outdoor_seating;playground"
> - I think osmand is mixing up the poi-categories because of wrong indexing this kind of semicolon separated pois
> 
> - If you set the language to english the categorie seemed to be ok (Playground-Leisure)
> - But typing on one of the result and then inside of the poi menu select "Nearby POI Playground"
> - The only playground that get find is the osm-way 757041132 (maybee you have to extend the radius (find more)
> 
> **2. Example**
> - Load map Deutschland Rheinland-Pfalz (Germany Rhineland-Palatinate)
> - To be sure to reproduce deactivate all other maps, force close the app, restart and give osmand some time to reindex the search
> - Switch language to german
> - Go to the area of the downloaded map
> - Search for "Spielplatz" and then you found categorie "Spielplatz-Tourismus" instead of "Spielplatz-Freizeit"
> - Typing on the categorie "Spielplatz-Tourismus"
> - If nothing is found press "Suchradius erweitern" (extend search radius)
> - You should find then "Campingplatz Nahemühle" (OSM-Way 33629405)
> - By checking the osm-way (33629405) you can find the tagging "leisure=firepit;playground"
> 
> - This part of the issue should also be reproducible in english
> - Search for playground and then you found categorie "Playground-Tourism" instead of "Playground-Leisure"
> - Typing on the categorie "Playground-Tourism" should lead you then again to the osm-way 33629405
> 
> I tink the importand point to reproduce is the used map, and if there are pois with semicolon separated categories.
[...]

PR #23026, opened 2025-07-15
[5.1] Optimize with getStandardPoiTypesKeyNames() cache

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22824

Issue #23001, opened 2025-07-13
5.0.5: Crash/App does not respond, after switching from Version 1 to Version 2 (OpenGL)


ProBackup-nl:
### Description

OsmAnd Plus 5.0.5 can not longer be started, after switching the map rendering from "Version 1" to "Version 2 (OpenGL)".

Sometimes a message "the app is not responding" appears.

Regression:

1. Deleted /Android/data/net.osmand.plus/files/roads/Netherlands....obf = no difference.
2. Battery > Device Performance > Energy saving -> Maximum performance = no difference.
3. Android Settings > System > Advanced > Developer Options > Preferences for display drivers: standard -> display driver for system = no difference
4. Enable WiFi = no difference
5. ... Developer settings: Force 4xMSAA (in OpenGL ES 2.0-apps) = no difference

### Steps to reproduce

1. Use Samsung Galaxy S4 value (GT-I9515)
2. Install LineageOS 18
3. Install OsmdAnd using F-Droid
4. Download maps
5. Switch from default "Version 2 (OpenGL)" to "Version 1"
6. Restart OsmAnd
7. Switch from default "Version 1" back to "Version 2 (OpenGL)"
8. Start OsmAnd
9. Wait approximately 15 seconds

### Actual result

Crash / App does not respond, approximately 15 seconds after starting.

Sometimes (approximately 1 out of 4), the map button interface is shown (without map elements) and then crashes.

Three times, out of approximately 20-30 tries, I was able to open the "Settings" (3 horizontal bars button) menu, however I can't scroll down to settings.

### Expected result

No crash, because initially map display worked great when "Version 2 (OpenGL)" was set.

### Your Environment (required)

```
OsmAnd Version: 5.0.5
Android version: [11 (lineage-18.1-20250203)](https://download.lineage.microg.org/jfvelte/)
Device model: GT-I9515
Crash-Logs: not sure what to extract from `adb bugreport`
```
PS Is there and `adb` terminal command to switch back to "Version 1"?

sonora:
@Chumva `.close() `or proper error handling for inaccessible paths may effectively be missing in this method, if it is called from a thread that might be destroyed before completion, like probably during app start? https://github.com/osmandapp/OsmAnd/blob/8188d0d4ebc91a1e34703c2cdf764dee3ac413ec/OsmAnd/src/net/osmand/plus/render/RendererRegistry.java#L251 ?

sonora:
[...]

PR #23029, opened 2025-07-16
Try fix or improve #23001 (untested!)

sonora:


sonora:
Revisiting this, it seems the PR would only patch the symptoms in one instance of the issue (this particular thread destroyed while running), but does not address the root cause of short-lived /unexpectedly killed threads.

If you feel it's not worth merging, feel free to cancel.

Issue #22687, opened 2025-05-30
Shortest Walking route not found


martinstingl:
### Routing engine

- [x] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile

Walking

### Start and end points

Start: 47.75985 N, 11.55955 E
End: 47.76056 N, 11.56088 E

### Actual route

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/0cddef56-c48a-448d-924e-2a05949558a6"  width="400" height="auto"/> </th>
  </tr>
</table>

### Expected route

I have highlighted the expected shorter route in red.
https://www.openstreetmap.org/way/1221853379 should be part of the route.

### Maps you used (online or offline)

- [x ] Offline maps offered within the OsmAnd app for download.  May 1, 2025

### Your Environment

```
OsmAnd Version: 5.0.5
Android version: 15
```

vshcherb:
@xmd5a2 Needs to be tested why
https://osmand.net/map/navigate/?start=47.759850,11.559550&end=47.760246,11.560431&profile=pedestrian#20/47.76020/11.55968 - cost 205 
vs 
https://osmand.net/map/navigate/?start=47.759850,11.559550&end=47.760314,11.560478&profile=pedestrian#20/47.76020/11.55968 - only cost 186 

Because of height, stairs penalty?

xmd5a2:
Reproduced but reason is unknown.

martinstingl:
The first junction seems to be the problem:
When making left turn at the first junction (before the "Pfarrhaus") it costs 39 (end point 1m after the junction). But turning here right costs 85 after 1m.

RZR-UA:
Notes.

Connected Issue: https://github.com/osmandapp/Osmand/issues/7369

In the current case, the possible reason is a high `penalty_transition`:

```
			<!-- give highway penalty to use mostly footways if available -->
			<select value="90"/>
```

RZR-UA:
Proposal:

1. Set penalty_transition=0 for ways with [service=alley](https://wiki.openstreetmap.org/wiki/Tag:service%3Dalley)
2. Decrease penalty_transition from 90 to 20 for highway=service
3. Increase penalty_transition from 0 to 10 for highway=residential

RZR-UA:
[...]

PR #23036, opened 2025-07-16
Get RouteTestingTest back (was ignored by mistake)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22687

PR #23044, opened 2025-07-17
Added MSAA option for 3d models

ihoryakovenko:
https://github.com/osmandapp/OsmAnd-Issues/issues/3021


Chumva:
@ihoryakovenko Do we still need this pr?

PR #23062, opened 2025-07-18
Added MSAA option to development plugin

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd-Issues/issues/3021

PR #23078, opened 2025-07-21
Revert tiles debug info to original version

ihoryakovenko:


ihoryakovenko:

PR #23080, opened 2025-07-21
Removed show symbols debug info from development plugin

ihoryakovenko:


ihoryakovenko:

PR #23091, opened 2025-07-23
fix #2878 use new track screen instead dialog

### LINKED ISSUE #2878: Search UI: Support Basic Search Syntax
Looks like in our new search UI the BLANK separator (" ") functions as an OR operator, which is fine.

I guess it would not be very complicated now to also support some basic useful other search syntax, like 
- AND: to force an AND operation for search terms
- the quotation marks, as in "best western" instead of best western, to force OsmAnd to search for exactly this occurrence of words. (Use case example: So far searching for best western returns all sorts of hits with occurrence of the term "best" alone, but it is hard to e.g. find only hits for the hotel chain "Best Western")

0xRe1nk0:


OsmAnd/src/net/osmand/plus/mapcontextmenu/MapContextMenu.java
```diff
@@ -1256,22 +1257,16 @@ public void addNewWptToGPXFile(@Nullable String title, @Nullable Amenity amenity
 
 	private void addNewWptToGPXFileImpl(@NonNull MapActivity mapActivity,
 	                                    @Nullable String title, @Nullable Amenity amenity) {
-		GpxUiHelper.selectSingleGPXFile(mapActivity, true, result -> {
+		SelectTrackTabsFragment.GpxFileSelectionListener gpxFileSelectionListener = gpxFile -> {
```
Chumva, line 1254: Do we need to replace other calls to GpxUiHelper.selectSingleGPXFile()?

Issue #22697, opened 2025-06-01
Unable to save changes (on some elements)


mnalis:
### Description

I was unable to modify  [way 35866855](https://www.openstreetmap.org/way/35866855)  and add a tag, which usually works for me without any problem. I tried multiple times without success.

Here is a video:

https://github.com/user-attachments/assets/b4195af4-1bfb-421c-a383-60494918df18

### Steps to reproduce

- open this [way 35866855](https://www.openstreetmap.org/way/35866855) 
- click on Actions / Modify POI
- click on Advanced tab and "Add tag"
- specifty `bicycle=no`
- click "SAVE"  button

### Actual result

nothing happens! I can click "save" multiple types, and still it is like I haven't clicked anything

### Expected result

the added tag will be saved, form would close, and map would indicate a change waiting upload. (like it happens to any other POI I tried editing in the past!)

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: OsmAnd~ 5.0.5
Android version: Android 14
Device model: Samsung Galaxy S23+
Crash-Logs: yes, logcat
```

See attached logs at about 19:05 when I was trying it, it seems there are several `StrictMode`-related errors logged.

[logcat.zip](https://github.com/user-attachments/files/20536664/logcat.zip)

yuriiurshuliak:
**Unable to save an added tag (`bicycle=no`) when editing **way 35866855 (Zagreb Botanical Garden)**. After tapping **Save**, nothing happens — the tag is not saved, and the form remains open.**

We have successfully reproduced the issue using the following steps:

1. Go to coordinates **45.8047265, 15.9726443** (Zagreb Botanical Garden).
2. Tap the entire garden polygon and choose **Actions → Modify POI**.
3. Open the **Advanced** tab and tap **Add tag**.
4. Enter: `bicycle=no`.
5. Tap **Save**.

###  Actual Result

* The **Save** button does nothing.
* The tag is not applied.
* The edit form stays open.
* No pending change is shown on the map.


OsmAnd 5.0.5, released: 2025-04-17

https://github.com/user-attachments/assets/bfb9f062-b854-437f-9e35-c0891f6611ff

Corwin-Kh:
[...]

PR #23130, opened 2025-07-28
#22697: Enabled saving non-editable poiType if type is not changed

Corwin-Kh:


OsmAnd/src/net/osmand/plus/plugins/osmedit/data/EditPoiData.java
```diff
@@ -35,7 +35,7 @@ public class EditPoiData {
 
 	private final Set<String> changedTags = Collections.synchronizedSet(new HashSet<>());
 
-	public EditPoiData(Entity entity, OsmandApplication app) {
+	public EditPoiData(Entity entity, OsmandApplication app, boolean isNew) {
```
Chumva: Do we need this new flag?

OsmAnd/src/net/osmand/plus/plugins/osmedit/dialogs/EditPoiDialogFragment.java
```diff
@@ -363,6 +365,8 @@ public void trySave() {
 			onSaveButtonClickListener.onSaveButtonClick();
 		}
 		String tagWithExceedingValue = isTextLengthInRange();
+		boolean poiTypeChanged = editPoiData.getChangedTags().contains(POI_TYPE_TAG)
```
Chumva: Could be moved in separate method in editPoiData

PR #23150, opened 2025-07-31
add possibility to en/disable map rotation modes

0xRe1nk0:


OsmAnd/src/net/osmand/plus/settings/enums/CompassMode.java
```diff
@@ -12,43 +14,53 @@ public enum CompassMode {
...
-			new ThemedIconId(R.drawable.ic_compass_niu, R.drawable.ic_compass_niu_white)
+			new ThemedIconId(R.drawable.ic_compass_niu, R.drawable.ic_compass_niu_white),
+			"north_is_up"
 	);
 
 	@StringRes
 	private final int titleId;
 	private final ThemedIconId themedIconId;
 	private final int value;
+	private final String key;
```
Chumva: Do we need this variable? Can we just use something like 	

public String getKey() {
		return name().toLowerCase();
} ?

OsmAnd/src/net/osmand/plus/quickaction/actions/ChangeMapOrientationAction.java
```diff
@@ -54,11 +83,190 @@ public int getIconRes(Context context) {
...
+			if (compassMode == null) {
+				it.remove();
+			}
+		}
+
+		return list;
+	}
+
+	@Nullable
+	public CompassMode getModeForKey(String key) {
```
Chumva: Could be moved to enum valueOf()

OsmAnd/src/net/osmand/plus/quickaction/SelectMapViewQuickActionsBottomSheet.java
```diff
@@ -106,6 +110,14 @@ public void createMenuItems(Bundle savedInstanceState) {
 		populateItemsList();
 	}
 
+	@LayoutRes
+	private int getLayoutId() {
+		if (action instanceof ChangeMapOrientationAction) {
+			return R.layout.bottom_sheet_item_with_descr_and_radio_btn;
```
Chumva, line 116: Why do we use layout with description & then hide it via AndroidUiHelper.updateVisibility(descriptionTv, false); ?

OsmAnd/src/net/osmand/plus/quickaction/SelectMultipleOrientationsBottomSheet.java
```diff
@@ -0,0 +1,167 @@
...
+	@Override
+	public void createMenuItems(Bundle savedInstanceState) {
+		items.add(new TitleItem(getString(R.string.rotate_map_to)));
+
+		for (CompassMode compassMode : CompassMode.values()){
+			addProfileItem(compassMode);
+		}
+	}
+
+	private void addProfileItem(CompassMode compassMode) {
```
Chumva: AddCompassModeItem/addItem? looks like copy-pasted code

Issue #23099, opened 2025-07-23
[Android Auto] Route preview moved to the right


Yaper:
### Description

In Android Auto after selecting navigation target, route preview is displayed but it is moved far way to the right. Not showing whole route on screen.
![Image](https://github.com/user-attachments/assets/d0ecb36c-3f76-44fe-b948-e68393da1a59)

It has been broken n 5.1.7 probably. It was working correctly few versions back.

### Steps to reproduce

In Android Auto please plan a route and observe route preview.

### Actual result

Route preview is moved far way to the right. Not showing whole route on screen.

### Expected result

Route preview should show whole route on the screen.

### Your Environment (required)

OsmAnd Version: 5.1.7

Zirochkabila:
`OsmAnd~ 5.2.0#5682m, Released: 2025-07-24`

- [x] Confirmed - Route preview moved to the right (**this may be due to the map's centering**)

`nb-2025-05-16` - the latest build where it works
`nb-2025-05-19` - the first build where the route is shifted to the right

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
    <th>Nightly 2025-05-16</th>
    <th>Current nightly & 2025-05-19</th>
  </tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/9cd82ac1-3fa7-4cbd-ae3d-b0ec9a24d50f"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/30fc9c9e-ffc3-4721-aef9-4cec786794a6"  width="300" height="auto"/> </th>
  </tr>
</table> 
</details> 

________

`OsmAnd~ 5.2.0#5721m, Released: 2025-07-31`

- [ ] V1 > Connect to AA > Build a route > Upper part of the route is cut off

<details><summary>Details</summary>
<img width="400" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/1ef47598-6f73-4241-8b37-9af0d095b0a0">
</details>

Zirochkabila:
# Testing 1
`OsmAnd~ 5.2.0#5805m, Released: 2025-08-11`

V2 — The route is displayed in full - 🔵 **Done**

<details><summary>Details</summary>
<img width="400" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/057e083b-5c1b-4450-818a-73c2323c5aca">
</details>

[...]

PR #23153, opened 2025-07-31
Fixed route preview moved to the right for v2 engine

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd/issues/23099#issuecomment-3112279825

PR #23159, opened 2025-08-01
Increase precision of coordinates in live monitoring by using doubles instead of floats

kiddikai:
LiveMonitoringData receives location coordinates as doubles, but converts them to floats, significantly reducing precision. You can spot this easily by comparing a recorded GPX track to the live locations sent during the same ride.

Green track: GPX trip recording
Red track: Live monitoring locations

image

Removing the float cast should allow the live locations to be as precise as the GPX track.


sonora:
Yes, but wherever things are stored (or transmitted) as a string, like in an actual GPX file, it probably makes sense to limit this to 8 or so decimal places (~ worst case / equator precision of 1.1 meters), no need to increase files sizes by including a full double with fake precision in the sub-micrometer range.

kiddikai:
Good point! I added calls to `GpxFormatter`'s `formatLatLon` method.

PR #23179, opened 2025-08-05
add hr analyze

0xRe1nk0:


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
```diff
@@ -123,6 +123,8 @@ class GpxTrackAnalysis {
 		get() = getGpxParameter(GpxParameter.MAX_SENSOR_HEART_RATE) as Int
 		set(value) = setGpxParameter(GpxParameter.MAX_SENSOR_HEART_RATE, value)
 
+	var minSensorHr: Int = 0
```
Chumva: Do we need it in the DB?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
```diff
@@ -441,6 +443,11 @@ class GpxTrackAnalysis {
 
 				if (attribute.heartRate > 0) {
 					maxSensorHr = maxOf(attribute.heartRate.toInt(), maxSensorHr)
+					minSensorHr = if (minSensorHr == 0) {
+						attribute.heartRate.toInt()
+					} else {
+						minOf(attribute.heartRate.toInt(), minSensorHr)
+					}
```
Chumva: could be a single line, without if-else

Issue #23053, opened 2025-07-18
Text search results are inconsistently displayed on the map


sonora:
As noted as a side finding in https://github.com/osmandapp/OsmAnd/issues/22855#issuecomment-3085282011: Do a **text** search and tap "Show all XXX on the map". (The issue does not occur when you simply use a category search.)

The orange disks marking the various search hits on the map disappear when zooming in to zoom 19 or closer, and only re-appear when zooming out to 17 or further. Tested on both Touring view and OsmAnd renderer, seem to behave alike.

DmitryAlexei:
Search results (orange POIs) disappear at the 20th zoom level and when zooming out only reappear at the 17th zoom level.

https://github.com/user-attachments/assets/260f2655-3804-4910-a0db-334fa5a6dd5b

isso-git:
Some more findings regarding this issue: 
Look like the zoom on wich the search result disappear and re-appear depend on where the search result is located.
On my nightly osmand it is as follow:

* Norma-1 49.80286° N, 9.94501° E (from the video above) disappear on zoom 19 or closer, re-appear on Zoom 17 or further.
* Norma-2  49.99631° N, 10.19777° E disappear on zoom 20 or closer, re-appear on Zoom 18 or further.
* Norma-3  50.14458° N, 10.13092° E disappear on zoom 21 or closer, re-appear on Zoom 19 or further.
* Norma-4  50.02142° N, 10.27833° E disappear on zoom 22 or closer, re-appear on Zoom 20 or further.

Maybee the search result is hidden by some other osm-nodes rendered by osmand but not visible?
By the way I would mention that the issue #22979 looks thomehow equal, but the other way round regarding visibility.

I hope this help, greetings.

sonora:
Good observation connecting to https://github.com/osmandapp/OsmAnd/issues/22979.

Also worth noting that in the situations when the hits are not displayed on the map, they also disappear from the search context list. (Tapping the list button in the displayed top bar on the map falsely produces "No places in visible region" then.) So it looks like perhaps not a mere rendering issue. And it affects both the V1 and V2 rendering engine.

Zirochkabila:
# Testing 1
`OsmAnd~ 5.2.0#5833m, Released: 2025-08-13`
`5.2.1.5209`

Checked the display of search results (orange infrastructure objects) at the 20th zoom level, and when zooming out, they appear only at the 17th zoom level - **Fixed, they are displayed at all zoom levels.**



[...]

PR #23189, opened 2025-08-05
Overlap POI name index searching

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/23053


OsmAnd-java/src/main/java/net/osmand/util/MapUtils.java
```diff
@@ -931,4 +931,23 @@ public static void inflateBBoxLatLon(QuadRect bbox, double dx, double dy) {
 		bbox.top = Math.max(-90.0, Math.min(90.0, bbox.top));
 		bbox.bottom = Math.max(-90.0, Math.min(90.0, bbox.bottom));
 	}
+
+	public static QuadRect roundBoundingBox31(QuadRect bbox31, int zoom)
```
ivanPyrohivskyi: From C++

PR #23218, opened 2025-08-08
Improve fragments inheritance

Chumva:


Chumva:
Should we implement IOsmAndFragment in BaseSettingsFragment?

OsmAnd/src/net/osmand/plus/dialogs/XMasDialogFragment.java
```diff
@@ -65,48 +70,51 @@ private static Date createDateInCurrentYear(int month, int date, int hour, int m
...
+			negativeButton.setTextColor(mapActivity.getColor(R.color.card_and_list_background_light));
+			negativeButton.invalidate();
 		});
 		return dialog;
 	}
+
+	@Override
+	protected void updateNightMode() {
+		this.themedInflater = LayoutInflater.from(getThemedContext());
+	}
```
Chumva: Why do we not use parent method?

OsmAnd/src/net/osmand/plus/base/dialog/IOsmAndFragment.java
```diff
@@ -0,0 +1,205 @@
...
+	// === Layout inflation ===
+
+	@NonNull
+	default View inflate(@LayoutRes int layoutResId) {
+		return inflate(layoutResId, null);
+	}
+
+	@NonNull
+	default View inflate(@LayoutRes int layoutResId, @Nullable ViewGroup parent) {
+		return inflate(layoutResId, parent, false);
```
Chumva: Will this lead to bugs in future because of inconsistency with the android api?

OsmAnd/src/net/osmand/plus/base/dialog/IOsmAndFragment.java
```diff
@@ -0,0 +1,208 @@
...
+		ApplicationMode appMode = getAppMode();
+		ThemeUsageContext usageContext = getThemeUsageContext();
+		return getApp().getDaynightHelper().isNightMode(appMode, usageContext);
+	}
+
+	// === Layout inflation ===
+
+	@NonNull
+	default View inflate(@LayoutRes int layoutResId) {
+		return inflate(layoutResId);
```
Chumva: Recursive call? Possible exception?

nazar-kutz: Yes, it fixed here: https://github.com/osmandapp/OsmAnd/pull/23263/commits/f6ce99178d1dc4dcd90c1ef9504c481c8942ea6c

OsmAnd/src/net/osmand/SecondSplashScreenFragment.java
```diff
@@ -34,18 +33,14 @@ public class SecondSplashScreenFragment extends BaseOsmAndFragment {
...
-
 	private int getNavigationBarWidth() {
 		if (!AndroidUtils.hasNavBar(getContext()) && !AndroidUtils.isNavBarVisible(getMapActivity()))
 			return 0;
 		int orientation = getResources().getConfiguration().orientation;
 		if (orientation == Configuration.ORIENTATION_LANDSCAPE && !AndroidUiHelper.isTablet(getContext())) {
 			int id = getResources().getIdentifier("navigation_bar_width", "dimen", "android");
 			if (id > 0)
-				return getResources().getDimensionPixelSize(id);
+				return (id);
```
Chumva: Possible bug?

sonora:
Looks like there is some issue still? [exception.log.txt](https://github.com/user-attachments/files/21862666/exception.log.txt)

Think it is caused by `(editable).isEmoty()` not existing on Android 11. Probably should use `(editable).length()==0` instead.

nazar-kutz:
Should be fixed here: https://github.com/osmandapp/OsmAnd/pull/23296

sonora:
Confirmed working! Good job, thank you, @nazar-kutz !

PR #23239, opened 2025-08-12
Update links from creator to builder

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/3077

PR #23246, opened 2025-08-12
Revert Draw ruler line while moving map

ihoryakovenko:


ihoryakovenko:

PR #23248, opened 2025-08-12
Update Qt-5.15.4 jar files from builder workspace (hotfix broken v2 AAR/JAR crash with updateNativeActivity)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/3077

PR #23279, opened 2025-08-18
improve add to favorite

0xRe1nk0:


OsmAnd/src/net/osmand/plus/myplaces/favorites/FavouritesHelper.java
```diff
@@ -369,6 +374,42 @@ public void setSpecialPoint(@NonNull LatLon latLon, SpecialPointType specialType
 		}
 	}
 
+	public void copyToFavorites(@NonNull GpxDisplayGroup displayGroup, @NonNull String groupName, @Nullable OnGroupNameChangeListener listener) {
```
Chumva: Do we need this listener here?

PR #23292, opened 2025-08-19
new dialog for axle load, max weight full load properties

0xRe1nk0:


OsmAnd/src/net/osmand/plus/settings/vehiclesize/VehicleSizes.java
```diff
@@ -51,7 +51,7 @@ public boolean verifyValue(@NonNull Context ctx, @NonNull SizeType type,
 	}
 
 	public int getMetricStringId(@NonNull SizeType type, @NonNull Metric metric) {
-		if (type != SizeType.WEIGHT) {
+		if (type != SizeType.WEIGHT && type != SizeType.AXLE_LOAD && type != SizeType.WEIGHT_FULL_LOAD) {
```
Chumva: We could create a method in an enum to check if it is some weight type

Issue #23093, opened 2025-07-23
Map loses location lock with double-tap slide zooming


sonora:
Not sure if this is a regression or a big issue, I just notice:
- Tap the My Position button so the map gets centered to where you are. The otherwise blue button turns white, indicating the map is linked to (will follow) your position.
- If you zoom using the +/- buttons, the map remains linked.
- But if you zoom with the "double-tap + swipe up/down" gesture, the My Position button turns blue, indicating we unlink it from the position lock. This may be a nuisance or  issue if you use the gesture while moving, e.h. on a bike.

The issue occurs for me independently of whether the Dev plugin's 'Use transparent status bar' setting is on or off.

sonora:
PS: It seems similar to #10108, but these days on current versions of OsmAnd I do not see the map center visibly 'drift off' anymore with zooming in. Instead, the map seems to remain precisely in place as it should be, only the Position button turns blue.

So it may not be an issue of using the wrong "screen center" any more in some constellations, but simply some other bug which has now crept in.

Zirochkabila:
`OsmAnd~ 5.2.0#5682m, Released: 2025-07-24`

And it's similar to #10108 and at the same time not, because when zooming with two fingers, the map does shift the center, which seems quite logical, **but when “double-tap + swipe up/down” the center of the map remains unchanged, although the location loses its anchor**

<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/113de088-2a02-4c7b-8475-a17c002ca349"></video>
</details>

vshcherb:
Yes, center position shouldn't be reset as this gesture purely to change zoom

sonora:
Confirmed fixed for me, thank you all!

Zirochkabila:
# Testing 1
`OsmAnd~ 5.2.0#5915m, Released: 2025-08-27`

When “double tap + swipe up/down”, the center of the map remains unchanged and the location remains fixed to the position.

<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/c373598c-4db7-424f-ba6f-1f9607a79bb5"></video>
</details>

PR #23324, opened 2025-08-23
#23093: Fixed map location loc on double_tap_scale zoom

Corwin-Kh:


OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
```diff
@@ -2268,7 +2269,10 @@ public void onZoomStarted(PointF centerPoint) {
 					initialMultiTouchCenterPoint.x, initialMultiTouchCenterPoint.y);
 			startRotating = false;
 			startZooming = false;
-			notifyLocationListeners(getLatitude(), getLongitude());
+			Location myLocation = app.getLocationProvider().getLastKnownLocation();
+			if (myLocation == null || myLocation.distanceTo(new Location("", initialCenterLatLon.getLatitude(), initialCenterLatLon.getLongitude())) > 1) {
```
Chumva: Do we need to ignore for all listeners? Also, 1m better would be to move to constant; 

We could also use MapUtils.getDistance() or MapUtils.areLatLonEqual()

PR #23345, opened 2025-08-26
Fix bug with mouse scroll position - zoom in/out

nazar-kutz:


OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
```diff
@@ -1874,11 +1874,15 @@ public boolean onGenericMotionEvent(MotionEvent event) {
...
 				event.getAxisValue(MotionEvent.AXIS_VSCROLL) != 0) {
 			RotatedTileBox tb = getCurrentRotatedTileBox();
-			double lat = tb.getLatFromPixel(event.getX(), event.getY());
-			double lon = tb.getLonFromPixel(event.getX(), event.getY());
+			LatLon latlon = NativeUtilities.getLatLonFromElevatedPixel(mapRenderer, tb, event.getX(), event.getY());
 			int zoomDir = event.getAxisValue(MotionEvent.AXIS_VSCROLL) < 0 ? -1 : 1;
 			int endZoom = normalizeZoomWithLimits(getZoom() + zoomDir);
-			getAnimatedDraggingThread().startMoving(lat, lon, endZoom);
+			float zoomFloatPart = getZoomFloatPart();
+			if (hasMapRenderer()) {
```
Chumva, line 1881: Do we need to use different methods for V1/V2 renderings?

PR #23365, opened 2025-08-29
Made Carly scanner discoverable

Corwin-Kh:


Chumva:
It should already be deprecated

PR #23376, opened 2025-09-02
23302 fix aa radius ruler

Corwin-Kh:


OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
```diff
@@ -212,6 +214,21 @@ public void onDraw(Canvas canvas, RotatedTileBox tb, DrawSettings drawSettings)
 			circleAttrsAlt.paint2.setStyle(Style.FILL);
 
 			QuadPoint center = tb.getCenterPixelPoint();
+			if (app.getOsmandMap().getMapView().isCarView()) {
```
Chumva, line 217: Check V1 rendering

PR #23395, opened 2025-09-03
Not showing Context Menu on note stop recording (if using keyboard)

nazar-kutz:


OsmAnd/src/net/osmand/plus/plugins/audionotes/AudioVideoNotesPlugin.java
```diff
@@ -814,7 +815,7 @@ protected void stopCamera() {
 		}
 	}
 
-	private boolean stopMediaRecording(boolean restart) {
+	private boolean stopMediaRecording(boolean restart, boolean displayContextMenuOnFinish) {
```
Chumva, line 818: Can we use runnable instead of displayContextMenuOnFinish?

Issue #20091, opened 2024-06-13
Bicycle navigation voice volume too low


speleoluc:
### Describe the idea (required)

When navigating a bicycle route, the sound volume of the voice is too low to be heard from the handlebar. Earphones being forbidden while riding, Osmand should offer to boost sound like Google Maps does.

### Tell us about the expected behaviour (required)

Add a Loud Sound option in Settings

### Tell us about alternatives you've considered (required)

Using a battery-consuming Bluetooth speaker.

mikehgentry:
While not a solution to the problem, the loud beep modes are a lot easier to hear.

https://osmand.net/docs/user/navigation/guidance/voice-navigation#beep-modes

speleoluc:
Most of the time, I ride with the screen off. Voice commands with street names are enough, and I'd rather not look at the screen while moving.

Increasing the sound volume above maximum must be programmatically simple if Maps did it.

Themanwithoutaplan:
Glad it's not just me! It would definitely be very useful to have an additional controller for TTS in OsmAnd. When I cycling I use Sony earphones with external drivers with the sound turned down a lot so that I can hear what's going on around me. This is fine when listening to music but voice navigation is easily drowned out, sort of defeating the point. The issue is extremely easy to verify but difficult to measure, I guess.

Issue #17497, opened 2023-06-25
Voice prompts during navigation


GuentherGredy:
# 🚀

mikehgentry:
If you look in Navigation -> Settings -> Voice prompts (settings) -> Language -> Recorded

You'll find exactly that!

We hammered it out in #10501 so if you have feedback probably post it there.

Cheers!

mikehgentry:
I think you raise a valid point. Maybe "Voice prompts" could be changed to "Audio prompts", and "Language" could arguably also be changed to "Type / Language" (since that better reflects the choice between TTS and recorded voice anyway).

It also wouldn't do any harm in my opinion mentioning the beep options in the documentation...

https://osmand.net/docs/user/navigation/guidance/voice-navigation/

I suppose I'll try to write a PR if I can figure out how to do that

GuentherGredy:
If you change "Voice prompts" to "Audio prompts" then you could add and entry "Beeps" or something like that above "Type / Language". Both then get a switch control to enable / disable it (mutually excluded).
What is actually set could be shown in smaller letters beneath the entry. This kind of UI would ideally fit to many of that kind in the application.
Please, as an idea, refer to attached picture quickly drawn in MS Visio.
Thanks for you attention.
BR Guenther
![Suggestion](https://github.com/osmandapp/OsmAnd/assets/85982948/a7de8cc3-631e-4e11-8c9c-da4f1bda3ade)

sonora:
I tend to agree that the user experience could likely benefit from restructuring the menu hierarchy. 

Today the user finds Navigation settings > Voice prompts > TTS/Recorded > Language

The beeps are not easiliy found. And also I have a feeling too many users still select "Recorded" over the preferable TTS, perhaps because TTS simply sounds unfamiliar. So if restructuring plus recording somehow mitigates these two issues, it may be worth it. Having said that, I believe that the vast majority of users should simply get (perhaps preconfigured) their device's system language as arTTS, so that option should not be buried deep in the hierarchy.

Perhaps something like Navigation settings > Audio guidance > languages (which would mean TTS), and there we would offer Beeps and (fallback) Recorded voices under "further options"?

GuentherGredy:
[...]

Issue #16347, opened 2023-01-29
Guidance volume too low to be usable


Bleuzen:
### Description

Hi, most of the time when driving I listen to music. When osmand has something to say, the music volume decreases during that, so audio ducking works. However, even when the music is ducked, the voice volume of osmand is so quiet, it is still really hard to understand.

In Google Maps for example, there are these options:
![grafik](https://user-images.githubusercontent.com/12885163/215347715-7b2a8657-6e48-48c3-9580-9b53150e2645.png)

I guess the volume of OsmAnd voice is comparable to somewhere between the "Softer" and "Normal" setting in Google Maps.
And this is too quiet, which sadly makes this app somewhat useless for car navigation.

### How to reproduce?

1.) Listen to music (optionally with some background noise (engine, other stuff from outside)) while using OsmAnd for navigation
2.) Try to understand the guidance voice

### Your Environment
OsmAnd Version: 4.3.5
Android/iOS version: Android 13
Device model: Oneplus 9 Pro

**Maps used (online or offline):**   
offline

### Additional

I know about the `Voice guidance output` option:
![grafik](https://user-images.githubusercontent.com/12885163/215347953-487480ca-cac8-40c2-9bf1-e7e513989f6b.png)

I saw in another issue that one solution could be to run osmand on another audio channel/category. This way, the volume (music vs. navigation voice) could be controlled independently.
However:
- `Notification audio` is not an option for me because I have notifications muted, especially while driving. Unmuting it would play all notifications, also from other apps, which is distracting and annoying.
- `Phone call audio` does play the guidance voice on my phones earpiece speaker which is much too quiet. Also it interrupts the music, which I don't want (music should continue playing, just get quieter during osmand is playing voice). It is buggy too. It sometimes switches between the phones earpiece to my cars stereo while the guidance voice is still playing, or interrupts the music too early, which is irritating.

So the first option `Media/navigation audio` is still the best. Just too quiet osmand output.

Also I found this in my system settings:
![Screenshot_2023-01-29-19-42-17-62_a6ef3d3be36e3bff34e5a5ac2e5be79d](https://user-images.githubusercontent.com/12885163/215348627-689b1ea5-83e4-409e-a858-85aaf877fd78.jpg)

I activated the "Amplify speech volume" option for TTS already, however this had no effect on the speech volume in osmand.

ossilator:
related to #17828.

PR #23416, opened 2025-09-08
tts: implement coarse volume amplification for TTS navigation

### LINKED ISSUE #17828: Volume of voice prompts should be lounder, preferably self-adjusting
# 🚀 feature request

### Description

First I thought my proposal is similar to issue #17497, but the focus drifted somewhat from the original proposal, so I add my own:
In general I don't like my phone being very loud as I mostly use it inside quiet rooms.
However when driving, the ambient sound is much louder, so the default volume for voice prompts is *way* too low.
Most of the tiome you don't even hear that it's talking.
As I also own a dedicated navigation device ("TomTom"), I'm proposing some improvement here:

First I wonder whether there can be a per-application setting for the default volume; if so I could manually adjust that (once implemented). Obviously such adjustments should be per map mode (map style) (car, walk, boat, etc.)

As my TomTom can do it, I also wonder whether the app could use the microphone to measure the ambient noise, and adjust the volume accordingly. It does not have to do it all the time, but maybe just before making a voice announcement (or at some regular intervals).

(My TomTom also recognizes some voice commands ("Speak & Go") to increase the volume, but that would bee demanding too much I guess. The image below is taken from the manual (https://download.tomtom.com/open/manuals/new_GO/html/en-gb/index.htm#Thevoicecontrolscreen-MASTEREU.htm)).

grafik

Raising the volume while the app is active manually (if you don't forget to do it in-time); if you forgot, doing it while driving is not the best option.

tdfischer:
When navigating by bicycle, the default TTS volume can be rather difficult to hear even if any background music has its volume ducked.

This implements a new per-profile setting for TTS amplification using the android DynamicsProcessing API.

A single-band compressor is applied to the audio stream, with some sensible defaults to allow user-configurable post-processing amplification.

In turn, this changes the 'default' audio volume to include a +10db amplification which, after comression, sounds generally identical to the unaltered stream. Additionally a 'quieter' and 'louder' setting adjusts the amplification to +5db and +15db, respectively.

This is my first PR against OsmAnd and first work on an android project since many years, so I'm not entirely sure if this is the right approach. In particular, some of the strings are not yet translatable and I can't figure out how to move the new settings entry to be grouped along with the existing voice options.

Fixes osmandapp/OsmAnd#17828

For comparison, here are some audio samples recorded using the android studio emulator:

- Original audio without this patch: https://nextcloud.malloc.hackerbots.net/nextcloud/index.php/s/6AxGHKEnT28rQmK
- 5db "Quieter" setting: https://nextcloud.malloc.hackerbots.net/nextcloud/index.php/s/e5zS5APnZq537Ys
- 10db "Normal" setting: https://nextcloud.malloc.hackerbots.net/nextcloud/index.php/s/yTPCs5kk6W5cgCo
- 15db "Louder" setting: https://nextcloud.malloc.hackerbots.net/nextcloud/index.php/s/PGXPJeZpqBBczdG

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -6247,4 +6247,5 @@ Download tile maps directly, or copy them as SQLite database files to OsmAnd\'s 
     <string name="routing_attr_freeride_policy_name">Off-piste</string>
     <string name="routing_attr_freeride_policy_description">\'Freeride\' and \'Off-piste\' are unofficial routes and passages. Typically ungroomed, unmaintained and not checked in the evening. Enter at your own risk.</string>
     <string name="voice_prompts_timetable">Voice prompts times</string>
+    <string name="voice_volume_title">Voice Volume</string>
 </resources>
--- a/OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java
+++ b/OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java
@@ -1964,6 +1964,9 @@ public class OsmandSettings {
 		AUDIO_USAGE[5] = new IntPreference(this, "audio_usage_5", 5).makeGlobal().makeShared().cache(); /*AudioManager.STREAM_NOTIFICATION -> AudioAttributes.USAGE_NOTIFICATION*/
 	}
 
+	// "low" is +5db, "normal" is +10db, "loud" is +15db
+	public final OsmandPreference<Float> VOICE_AMP_VOLUME = new FloatPreference(this, "voice_amp_volume", 10.0f).makeProfile().cache();
+
 	// For now this can be changed only in TestVoiceActivity
 	public final OsmandPreference<Integer>[] VOICE_PROMPT_DELAY = new IntPreference[10];
 
--- a/OsmAnd/src/net/osmand/plus/settings/fragments/voice/VoiceAnnouncesFragment.java
+++ b/OsmAnd/src/net/osmand/plus/settings/fragments/voice/VoiceAnnouncesFragment.java
@@ -97,6 +97,7 @@ public class VoiceAnnouncesFragment extends BaseSettingsFragment {
 
 		setupKeepInformingPref();
 		setupArrivalAnnouncementPref();
+		setupVoiceVolumePref();
 
 		setupAudioStreamGuidancePref();
 		setupInterruptMusicPref();
@@ -153,6 +154,18 @@ public class VoiceAnnouncesFragment extends BaseSettingsFragment {
 		ListPreferenceEx keepInforming = findPreference(settings.KEEP_INFORMING.getId());
 		keepInforming.setEntries(keepInformingNames);
 		keepInforming.setEntryValues(keepInformingValues);
+	}
+
+	private void setupVoiceVolumePref() {
+		Float[] volumeValues = {5.0f, 10.0f, 15.0f};
+		String[] volumeNames = {
+				"Quieter",
+				"Normal",
+				"Louder"
+		};
+
+		ListPreferenceEx volumePref = createListPreferenceEx(settings.VOICE_AMP_VOLUME.getId(), volumeNames, volumeValues, "Voice guidance volume", R.layout.preference_with_descr);
+		getPreferenceScreen().addPreference(volumePref);
 	}
 
 	private void setupArrivalAnnouncementPref() {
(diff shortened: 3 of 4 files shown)


sonora:
I actually fail to understand the issue you are trying to solve here., and am not convinced an app-setting centric solution is best practice.

On Android, music playback and navigation prompts are handled as different audio streams, and the way to set their volumes is to adjust each stream while it’s actively producing sound.

E.g.Music apps (Spotify, YouTube Music, etc.) use the STREAM_MUSIC channel,while navigation prompts use the STREAM_NOTIFICATION or STREAM_ASSISTANT channel, depending on app and  setup.

To adjust each to your liking, you simply have to use the volume buttons during  the respective playback. The system will keep track of this and handle it for you then.

tdfischer:
The problem I am trying to solve is that the TTS volume is still too low to be heard while music is being played and I am riding my bicycle.

Maxing out the volume on my phone means I have music that is way too loud for a safe bicycle ride. There is actually no UI for adjusting "navigation" streams separate from my  music playback, and I cannot use the notification volume control because my phone is always in silent/vibrate.

Google Maps implements their TTS navigation in a similar way with a 3 way selector of quiet, normal, and loud. And it works great for me, but with OsmAnd it is actually trying to get me to have a very unsafe situation where I need to have extremely loud music just so I can hear the navigation instructions.

To reproduce, get a pair of earbuds, hop on a bike, and start playing some Ratatat. Then when the audio ducking kicks in as you come up to an intersection, you will struggle to hear the TTS audio while the wind goes past.

Here are a few other open tickets that describe exactly this use case and the issue:

- https://github.com/osmandapp/OsmAnd/issues/20091
- https://github.com/osmandapp/OsmAnd/issues/17497
- https://github.com/osmandapp/OsmAnd/issues/16347

I'm also not sure what you mean with `STREAM_ASSISTANT`. That does not exist in Android. The only useful streams are:

- STREAM_MUSIC
- STREAM_NOTIFICATION
- STREAM_VOICE_CALL

The second one is useless when my phone is always on vibrate. The third one causes significantly lower quality bluetooth audio codecs on most phones.

sonora:
You're right, my wording was imprecise. I'm on the road and talking from memory: We're using streams STREAM_MUSIC, STREAM_NOTIFICATION and STREAM_VOICE_CALL, and tagging with respective AudioAttributes USAGE_ASSISTANCE_NAVIGATION_GUIDANCE, USAGE_NOTIFICATION, USAGE_VOICE_COMMUNICATION.

And yes, in terms of volume, many systems seem to actually combine by stream and do not keep track of volume to higher (usage or app) granularity. Hence choosing a separate _stream_ for the navigation sound is the primary solution for separate volume control...

All I'm saying is that it's not very common for Android apps to have their own built-in volume settings, hence my initial reservation if this is really considered good practice....

tdfischer:
Sure, apps should not be expected to implement a full audio mixing interface, and most don't but Google Maps /does/ do this. This is simply applying a vocal compressor to a vocal stream that is difficult to hear in the busy environment of riding a bicycle where accessibility is also critical for safety, a far cry from being a full audio mixer interface

I would like to continue using OsmAnd for my everyday commute and riding around town, but I do not feel safe doing so if it means I need to have music at an unsafe level just to hear turning instructions.

If there was a way to adjust the volume of just the TTS using the existing system tools I would have already done that and would be instead submitting an MR to update the documentation.

sonora:
Right... And (other than using KEY_PARAM_VOLUME) it seems your PR would now introduce DynamicsProcessing for all TTS utterances. Have you considered and tested system load and latency effects on low-end hardware, or if users have multiple audio effects chained (e.g., EQ + compressor + limiter)?

tdfischer:
No, I only have one device available to me, it is not possible for me to test on lower end hardware. But a single-band compressor is extremely low overhead in any situation.  The DynamicsProcessing API has been in Android since 2018, it is extremely old and optimized by any measure. If this was using multiple bands and processing stages, sure, performance could be relevant. But this is a very trivial processing "chain" of exactly one stage. My earbuds have a more complicated EQ pipeline than this.

There also is not any way that I know of for a user to chain multiple audio effects onto an arbitrary media session, but is that really something to be worried about supporting? This feels like an awful lot of pushback for a very simple accessibility feature coming from a new contributor.

sonora:
Please don't get me wrong, I'm not pushing back on this. Neither will I be making the acceptance decision at all.

I am only trying to shine some light on aspects like what is this trying to achieve, how common are  user problems and use cases this intends to fix, is this the least-impact solution and have other solutions been probed, to what extent has it been tested, what's the cost of supporting /maintaining this going forward, etc.

ItsRogueRen:
I'm in favor of this exact thing. On my Pixel, music and navigation both use the same volume slider so its inpossible to make OSMand louder than the music amd even when ducked the music is louder than the TTS voice. Adjusting the volume in my TTS app doesn't actually make it louder from OSMand, and instead it just starts clipping and becoming even harder to hear. 

The only other solution would be to use another audio stream that base Android doesn't merge together like Notifications, but then that has issues for people who keep their phone on silent or vibrate. 

Some OEMs allow per-app volume control but as far as I'm aware that is OEM specific and not a feature of Android normally

PR #23417, opened 2025-09-08
Enabled weather forecast source switch

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd-Issues/issues/2613

Issue #15034, opened 2022-08-08
OpenGL: Pinch zoom to magnify text / icons


sonora:
It may not be easy to fix, but let me trigger the investgation anyway: Using legacy rendering, if in bad light conditions like direct sunlight, and trying to read text on the map, you can use "pinch zoom and hold", i.e. pinch-zooming the map and leaving your fingers in position, to temporarily magnify a map area (including all text) without it getting re-rendered, this e.g. facititates reading small or tilted map text.

Using OpenGL rendering, this trick fails, because map text immediately disappears during the maneuver, thus is never magnified.

Any chance to fix OpenGL to support that trick again?

vshcherb:
This wouldn't work with Intermediate zoom feature which we plan to fix. So text should be increased only with a setting of Text size.

vshcherb:
#15009

sonora:
Let me perhaps reopen this to see if we can find a solution.

This issue is about (with the OpenGL engine) being able to use pinch zoom as an on-the-fly  screen magnifier, like the V1 engine allows.

We had interim linked it to #15009, but unfortunately the fix for that did not solve this issue here.

vshcherb:
I still don't think we should change anything, I don't think it's important feature but let's keep it open

Gregor0000:
I think this feature is quite important and it is one of the major reasons, I don't use OpenGL rendering yet.
It is so convenient to be able to increase the text size with pinch&zoom in situations where the readability of the screen is bad (bad light - too dark, too bright, when wearing sunglasses). For me, those situations happen all the time during my hikes. Increasing the overall text size is not a good solution: It clutters the map view and makes a good overview more difficult.
I would very much appreciate if such a functionality could be implemented.

sonora:
I wonder if the feature requested in this issue could not be achieved by wrapping the content of this method https://github.com/osmandapp/OsmAnd-core/blob/a37dfcced5fb0ce9f0783ee3707ea7aa6cf25545/wrappers/android/src/net/osmand/core/android/MapRendererView.java#L1734 in a condition like
```
public void onDrawFrame(GL10 gl) {
	if (!(OsmandMapTileView).isPinchZoomingOrRotating()) {

		(...)

	}
}
```
which would halt any re-rendering until the fingers are taken off the screen, just like the legacy renderer handles things...

vshcherb:
[...]

PR #23432, opened 2025-09-10
Change pinch zoom magnification without enabling accessibility mode

ihoryakovenko:


ihoryakovenko:

OsmAnd/src/net/osmand/plus/settings/backend/preferences/GlobalAccessibilityPreference.java
```diff
@@ -0,0 +1,31 @@
+package net.osmand.plus.settings.backend.preferences;
+
+import net.osmand.plus.settings.backend.ApplicationMode;
+import net.osmand.plus.settings.backend.OsmandSettings;
+
+public class GlobalAccessibilityPreference extends BooleanPreference {
```
Chumva: Do we really need this new class?

ihoryakovenko: Fixed

PR #23441, opened 2025-09-11
Proper camera movement during zoom magnification

ihoryakovenko:


ihoryakovenko:

Issue #22226, opened 2025-04-03
Search Routes: display altitude and uphill stats


vshcherb:
**OSM routes Elevation data**
- [5.3] Automatically retrieve and display elevation data (could be async)

**Docs**
- [ ] Refresh documentation with details about TravelGpx and ClickableWay

**ClickableWay UI**
- [x] Show difficulty-defined-color shields in Search results
- [x] Deduplicate ClickableWay MAP<->POI in results
- [x] Implement click on ClickableWay POI in Search results
- [x] Show Elevation metrics for ClickableWay objects (in Search and Multi-select)

**ClickableWay Generation**
- [x] Move ClickableWay tags processing to OsmAnd-Java
- [x] Generate Elevation metrics tags for ClickableWay objects

**Generation**
- [x] Precalculate uphill/downhill and other ele-stats (generation)

**UI: route context menu, search results**
- [x] Implement methods to fetch formatted route metrics
- [x] Search Results: modify second line `Hiking • 17,6 km, ↘︎ 13 m | ↗︎ 5 m • Kyiv`
- [x] Route/Track context menu: add elevation stats in the header, nearby distance
- [x] Do not show uphill/downhill below a certain threshold in Search results (**% of distance**)

**Clicks / Search**
- [x] Fix multi-select menu ("Track" is shown instead of "Route+type+distance")
- [x] Check and finalize click-on-map for new OSM routes (resources branch `new-routes`)
- [x] Disable click on OSM routes v1 (remove old code - android and tools)

**Test**

- [x] check for regressions [osmand_poi_lat_lon](https://github.com/osmandapp/OsmAnd/issues/8159) tag has just been removed

RZR-UA:
To Test (Android):

OSM routes (**Android only**):

- [x] distance, uphill, and downhill should be shown in search results
- [x] these values should be shown in the OSM route Track card (non-clickable)

OSM routes (**iOS**):

- [x] The new style for OSM routes should be displayed
- [x] click-on-map to route shield or name should work and load route
- [x] as iOS uses a special "new-to-old" workaround, the elevation chart should be shown in the Track card

ClickableWay (ski trails, mtb trails, etc) - **Android only**:

- [x] use `Winter and ski` style to get Ski trails displayed
- [x] test Ski trails (14A, 22C, etc) in Bukovel: do search (by name, by category), test map/search clicks
- [x] ski piste icon (in search list and track card) should correspond to difficulty (green, blue, red, black)
[...]

PR #23446, opened 2025-09-11
Display OSM routes tag-defined elevation metrics in Search and Track Context Menu

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22226


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxFile.kt
```diff
@@ -306,11 +307,32 @@ class GpxFile : GpxExtensions {
 			analysis.setGpxParameters(parameters)
 		}
 
+		analysis.hasElevationMetricsInTags = parseElevationMetricsTags(analysis)
```
alex-dev-neo: `hasElevationMetricsInGpx`

RZR-UA: done

PR #23447, opened 2025-09-11
Prepared transport route planning for covering with Unit Test

LVladymyr:
Actualization: as functionality of searching route with transport is complex and sometimes [works unexpectedly](https://github.com/osmandapp/OsmAnd/issues/22515), it required some sort of stabilization. In this pull-request, I made the first step in that direction that described below and start discussion about which use-cases should be covered. Such use cases could be provided with changes in `TransportRouteBuilder` and related unit tests.

I properly understand that according to OsmAnd solution, such changes must be done in central native library. While it required more time to implement and prepare test in C++ infrastructure, so probably the rule of [Three Fazes of Software Development](https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast) is becoming applicable here as we focused on fase "make right".

Current path finding implementation would works only with `SAFE_MODE=true` (or more direct `PT_SAFE_MODE=true`).

You could see in attached pictures that with or without patches it works in the same way, so it shows that patch not introduce change in logic.
<table>
<tr><th>with patch</th><th>without</th><th>native</th></td>
    <tr>
        <td>with patch</td>
        <td>without pathc</td>
        <td>
Screenshot_2025-09-12-10-28-17_27146(native)
</td>
    </tr>
</table>

During change preparation, I moved some internal classes into his own file. This preparation creates more files with name `TransportRouting*` in package `net.osmand.router` than it should be, so I moved them into their own package, `net.osmand.router.transport`. To make clear what changes happens in `TransportRoutePlanner.java` it is needed to make custom diff view command: 
```bash
git diff -M01 master add-unit-test-for-transport-routing -- OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java OsmAnd-java/src/main/java/net/osmand/router/transport/TransportRoutePlanner.java
```
or trust me and check this patch file [TransportRoutePlanner.java.patch](https://github.com/user-attachments/files/22279646/TransportRoutePlanner.java.patch)

Prepared abstractions for unit test to cover transport route planning:
- Added `DistanceProvider` and `IDistanceProvider` interfaces to abstract distance calculations.
- Introduced `ITransportRoutingContext` and `ITransportRoutingConfiguration` interfaces for better testability and extensibility.
- Added unit tests for `TransportRoutePlanner`.


LVladymyr:
Hi guys, would happy to hear any feedback from you @vshcherb @ivanPyrohivskyi @Chumva @MadWasp79

Chumva:
Thanks for your contribution! @ivanPyrohivskyi will review this pull request next week

ivanPyrohivskyi:
1) What is main goal of class [TransportRoutePlannerTest.java](https://github.com/osmandapp/OsmAnd/pull/23447/files#diff-d5d51a113963ac4a7766a15b0b2ae1588199ada6a2b8687d6521060cd95cfd24) ?
Usually, we make test for [real cases](https://github.com/osmandapp/OsmAnd-resources/tree/master/test-resources/routing), see [routing tests](https://github.com/osmandapp/OsmAnd/tree/master/OsmAnd-java/src/test/java/net/osmand/router).
For each compilation/build in Android, we run them as unit tests.

2. Refactoring of TransportRoutePlanner is fine. But usually we divide: refactoring, maintenance, fix bugs, logic changes, and tests. So refactoring as is must be in a separate request! At this moment, it is hard to check & merge

3. Generally, we planned to change algorithm of TransportRoutePlanner. E.g. for task [Problem navigating from Prague airport](https://github.com/osmandapp/OsmAnd/issues/22515) I already make some [PR](https://github.com/osmandapp/OsmAnd/pull/22908) that still is not accepted, because it breaks classical Dijkstra algorithm.

LVladymyr:
Hi @ivanPyrohivskyi,

Thank you for the detailed feedback and for being so open to contributions. I appreciate you taking the time to explain the project's workflow.

Let me address your points:

> 1. What is main goal of class TransportRoutePlannerTest.java? Usually, we make test for real cases...

You are right, the integration tests with real-world data are essential for end-to-end validation. My goal with TransportRoutePlannerTest.java was to supplement those with unit tests. I found the logic within TransportRoutePlanner to be quite complex and have many edge cases (like the Prague Airport issue).

My intention was to create a set of fast-running unit tests that isolate this specific class. This creates a "safety net," making it easier and safer to refactor or change the internal logic in the future, while being confident that existing cases are not broken.

> 2. Refactoring of TransportRoutePlanner is fine. But usually we divide: refactoring, maintenance, fix bugs, logic changes, and tests. So refactoring as is must be in a separate request!

Thank you for clarifying this. I was not aware of this rule, but I completely understand why you have it. Combining refactoring and new tests in one PR makes it very difficult to review. My apologies for that.

I agree that splitting the work is the best path forward. I propose the following plan:

    PR 1: Refactoring. Introduce the abstractions (IDistanceProvider, etc.), move classes, and adapt TransportRoutePlanner to use them. The goal of this PR will be zero logic changes, only structural improvements to enable testability.

    PR 2: Unit Tests. Once the first PR is merged, I will submit a second one that adds the new TransportRoutePlannerTest.java, which will use mock implementations of the new interfaces to test the logic.

Does this two-PR approach sound good to you?

> 3. Generally, we planned to change algorithm of TransportRoutePlanner... for task #22515 I already make some #22908...

That's great to know. This confirms my feeling that this is a critical and complex part of the codebase. My hope is that the unit tests from my proposed PR 2 would be a valuable asset for this future algorithm change. They could help validate that any new algorithm (like your work in #22908) still correctly handles all the existing corner cases before testing it with the full "in vivo" routing tests.

### Next Steps

I am very interested in helping with this task. If the two-PR plan I outlined above is acceptable, I will close this current pull request and create a new, smaller one focused only on the initial refactoring (PR 1).

Please let me know your thoughts. I am happy to adjust the plan to best fit your team's workflow.

Issue #23491, opened 2025-09-17
OpenGL Pinch zoom magnifier: Slight anchor misalignment on gesture end


sonora:
We have recently introduced also in the V2 engine a **transient Pinch zoom magnification** functionality (exactly while the two fingers are on the screen). This aligns the UX / behavior with how it has always been the V1 engine (where it is built-in, i.e. 'always on'), but for V2 it currently requires explicit activation in the Accessibility plugin.

There is still a small quality glitch upon the gesture end, reproducible as follows (https://github.com/osmandapp/OsmAnd/issues/15034#issuecomment-3301577552):

- Use the V2 engine, with pinch zoom magnification enabled:
- Pinch-zoom in some, then leave the fingers static on the screen for a moment and watch the map viewport displayed, or some prominent point on the map.
 - Now lift the fingers, and (with the map re-rendering now) observe a small viewport "jump" (the map jumps a little in one direction). The direction and amount of the shift varies, perhaps with where you happen to be on a tile. The effect is rather small in 2D, but can be more apparent in 3D.

I am convinced this is a small anchor mismatch between releasing the temporary viewport scaling (in `onZoomOrRotationEnded(...)`) 
https://github.com/osmandapp/OsmAnd/blob/e69b31ac240c37e9696999570e8ffcd4201c835e/OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java#L2201-L2205
and the successive map re-render commit, probably in `onTouchEvent(...) / ACTION_UP` https://github.com/osmandapp/OsmAnd/blob/e69b31ac240c37e9696999570e8ffcd4201c835e/OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java#L1969

Would be nice to catch this, because there seems no reason why the new functionality could not simply be made 'always on' like in V1, or at least enabled by default, because most users never discover it in the Accessibility plugin.

sonora:
Seems to me that this fixes it: https://github.com/osmandapp/OsmAnd/pull/23499

sonora:
I have committed it so we can test. @Zirochkabila , can you perhaps repeat your videos from https://github.com/osmandapp/OsmAnd/issues/15034#issuecomment-3301577552. It looks fixed now on all my test devices.

vshcherb:
Merged back to https://github.com/osmandapp/OsmAnd/issues/15034

sonora:
[...]

PR #23501, opened 2025-09-17
#15034 Pixel drift

vshcherb:

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
+++ b/OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
@@ -1081,8 +1081,8 @@ public class OsmandMapTileView implements IMapDownloaderCallback {
 			return;
 		}
 		PointF ratio = mapViewTrackingUtilities.getMapDisplayPositionManager().getMapRatio();
-		int cy = (int) (ratio.y * view.getHeight());
-		int cx = (int) (ratio.x * view.getWidth());
+		int cy = Math.round(ratio.y * view.getHeight());
+		int cx = Math.round(ratio.x * view.getWidth());
 		boolean updateMapRenderer = false;
 		MapRendererView mapRenderer = getMapRenderer();
 		if (mapRenderer != null) {
@@ -1481,8 +1481,8 @@ public class OsmandMapTileView implements IMapDownloaderCallback {
 	}
 
 	private void setLatLonImpl(double latitude, double longitude, float ratiox, float ratioy) {
-		int cx = (int) (ratiox * view.getWidth());
-		int cy = (int) (ratioy * view.getHeight());
+		int cx = Math.round(ratiox * view.getWidth());
+		int cy = Math.round(ratioy * view.getHeight());
 		if (currentViewport.getCenterPixelY() == cy && currentViewport.getCenterPixelX() == cx) {
 			setLatLonImpl(latitude, longitude);
 		} else {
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
```diff
@@ -1081,8 +1081,8 @@ private void refreshMapInternal(@NonNull DrawSettings drawSettings) {
 			return;
 		}
 		PointF ratio = mapViewTrackingUtilities.getMapDisplayPositionManager().getMapRatio();
-		int cy = (int) (ratio.y * view.getHeight());
-		int cx = (int) (ratio.x * view.getWidth());
+		int cy = Math.round(ratio.y * view.getHeight());
```
ihoryakovenko: Do we need all those Math.round operations? They caused this bug.
https://github.com/osmandapp/OsmAnd/issues/15034#issuecomment-3298035234

OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
```diff
@@ -2317,8 +2328,21 @@ public void onZoomStarted(PointF centerPoint) {
 			initialMultiTouchCenterPoint = centerPoint;
 			initialViewport = getRotatedTileBox();
 			MapRendererView mapRenderer = getMapRenderer();
+			// Remember the tile31 under the pinch center so we can re-center later
+			if (mapRenderer != null) {
```
ihoryakovenko, line 2332: I think we need to add if(isPinchZoomMagnificationEnabled) to understand that all the next new operations are related only to the new feature.

OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
sonora: In theory we should have it, but it had no beneficial impact on the issue at hand, so I suggest to leave this as it was originally.

OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
sonora: Agreed.

sonora:
@ihoryakovenko 

**Patch 1:** I have repackaged the previous PR here https://github.com/osmandapp/OsmAnd/pull/23513. I think this is safe, and fixes both the initial jump and the panning behavior.

**Patch 2 (still needed):** After this, the only issue still I am aware of with the new functionality is #23491. So we need a Patch 2 for it. The fix I had suggested for that is solely:
 - [Pinch zoom magnifier: Convert pinch to real zoom/rotation and re-anch…](https://github.com/osmandapp/OsmAnd/pull/23501/commits/d27b56da45d787f43585da4a09cea28d338f2b26)
 - [Update OsmandMapTileView.java](https://github.com/osmandapp/OsmAnd/pull/23501/commits/65d781b593e4f1ec7d6ebf619f30814c389e099e) - which is just a code formatting to the above

But this fix after testing turns out to have two issues:
1. It fixes #23491 reliably only if you wait a second or two before lifting the fingers after the end of your pinch-zoom gesture. This indicates we have a timing issue, lifting the fingers too early still produces the small map jump. 
 2. The fix is bad if you pinch-zooom while in motion, because it resets the screen anchor to an outdated point.

So I think we need to see if we find a better fix for #23491, perhaps based on the above idea, or even completely different?

sonora:
With **"Patch 1"** https://github.com/osmandapp/OsmAnd/pull/23513 merged now, I recommend we close this PR now unmerged.

We still need a **"Patch 2"** as noted in my comment above to fix the residual map jump at gesture end, but the solution I found [Pinch zoom magnifier: Convert pinch to real zoom/rotation and re-anch…](https://github.com/osmandapp/OsmAnd/pull/23501/commits/d27b56da45d787f43585da4a09cea28d338f2b26) has the above described 2 flaws (timing issue and false caching while in motion) and we need a more refined approach.

PR #23513, opened 2025-09-18
Patch 1: Fix Pinch Zoom Magnifier Overshoot

sonora:


OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
```diff
@@ -2317,8 +2317,21 @@ public void onZoomStarted(PointF centerPoint) {
 			initialMultiTouchCenterPoint = centerPoint;
 			initialViewport = getRotatedTileBox();
 			MapRendererView mapRenderer = getMapRenderer();
+			// Remember the tile31 under the pinch center so we can re-center later
+			if (mapRenderer != null) {
```
ihoryakovenko: Please add isPinchZoomMagnificationEnabled check as we discussed before

ihoryakovenko (approved):

PR #23524, opened 2025-09-21
Plan Route: show Context Menu to add point when clicking on map object

nazar-kutz:


OsmAnd/src/net/osmand/plus/mapcontextmenu/other/MapMultiSelectionMenuFragment.java
```diff
@@ -139,11 +141,25 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+		super.onViewCreated(view, savedInstanceState);
+		backPressedCallback = new OnBackPressedCallback(true) {
+			@Override
+			public void handleOnBackPressed() {
+				if (menu != null) {
+					menu.hide();
+				}
+			}
+		};
+		view.post(() -> requireActivity().getOnBackPressedDispatcher().addCallback(getViewLifecycleOwner(), backPressedCallback));
```
Chumva, line 162: Why do we need post() here?

OsmAnd/src/net/osmand/plus/mapcontextmenu/AddPlanRoutePointButtonController.java
```diff
@@ -0,0 +1,29 @@
...
+
+	@Override
+	public void buttonPressed() {
+		MapActivity mapActivity = controller.getMapActivity();
+		if (mapActivity != null) {
+			MapContextMenu menu = mapActivity.getContextMenu();
+			LatLon latLon = controller.getLatLon();
+			MeasurementToolFragment fragment = mapActivity.getFragmentsHelper().getMeasurementToolFragment();
+			if (fragment != null && latLon != null) {
+				fragment.addPoint(latLon);
```
Chumva, line 24: Probably we can do it via editing context instead of fragment

PR #23525, opened 2025-09-22
Add a chime feature into the trip recording plugin

IsaacHorvath:
This PR provides a feature to allow the trip recording plugin to make a simple chime noise at a specified interval. The plugin has a new setting to enable the feature and set the interval, and the list of available intervals matches the "split track" options in the analyze track by intervals feature.

I first expressed interest for this feature in [this discussion](https://github.com/osmandapp/OsmAnd/discussions/23466). I decided to try implementing it - let me know if I need to make any adjustments to the code, pick a new sound, etc.


vshcherb:
This looks like very specific feature, so chime effect could be applicable for all important for user statuses - trip recording, gps signal, navigation status. So this request feels very niche.

I think we won't accept this pull request due to the product organization.

PR #23545, opened 2025-09-25
search-test-android-v2.3

Gateway07:
Adding getId() to calculate actual_place

https://github.com/osmandapp/OsmAnd-Issues/issues/2905#issuecomment-3322482710


vshcherb:
All this changes should be part of search tools not osmand project

PR #23549, opened 2025-09-25
22665: Fxed show MissingMaps screen on different activities

Corwin-Kh:


OsmAnd/src/net/osmand/plus/activities/OsmandInAppPurchaseActivity.java
```diff
@@ -44,6 +45,7 @@ public class OsmandInAppPurchaseActivity extends AppCompatActivity implements In
 	private boolean activityDestroyed;
 	private boolean activityHiddenForTalkback = false;
 	private FragmentLifecycleCallbacks lifecycleCallbacks = TalkbackUtils.getLifecycleCallbacks(this);
+	private final AndroidAutoActionReceiver androidAutoReceiver = new AndroidAutoActionReceiver();
```
Chumva: Will it work properly in other activities? We have this [check](https://github.com/osmandapp/Osmand/blob/0cd34946bb892ae6b1d166573a58fe81cd314e50/OsmAnd/src/net/osmand/plus/routepreparationmenu/RequiredMapsFragment.java#L284) in fragment

Issue #23320, opened 2025-08-22
Search for building '30-Б' is not found when the query '30б' omits the hyphen


dmpr0:
## Use cases 
- [x] **Вадима Гетьмана 30-Б** vs **Вадима Гетьмана 30Б**
- [x] 'Київська вулиця 10-Д' (ровно) - doesn't list 'Київська вулиця 10'
- [x] Related `18B Holmby Road, Cheltenham, Victoria, Australia` (18B or 18 B or 18-B) addr:unit to add to house number with space - https://github.com/osmandapp/OsmAnd/issues/21479
- [x] Test tool Add functions to generate variants for 'X-Y' -> ["X Y", "X-Y"], "X'Y" -> ["XY", "X'Y"]

## Description
The search functionality fails to find buildings when the query omits a hyphen that is part of the official building number. For example, a search for `30б` does not find the building designated as `30-Б`.

### How to reproduce
* Navigate to the **Kyiv** region on the map.
* Open the **Search** menu.
* Search for the address `30б вадима`.

### Actual result

<table style="width:100%">
  <tr>
    <th>Search for "30б"</th>
    <th>Search for "30-б"</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/279f8a52-5246-4a59-85a4-f13a09d221db"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/7effe0e6-6768-438e-9e6c-c96c257db21e"  width="300" height="auto"/> </td>
  </tr>
</table>   

* Correct building, **Вадима Гетьмана** 30-Б, is not found. The search returns no relevant results.

### Expected result

* The search should find the building at **Вадима Гетьмана** 30-Б, treating the query `30б` as equivalent to `30-б`. 

**Environment**
OsmAnd Version: OsmAnd~ 5.2.0#5887m, released: 2025-08-22
Android: 15
Device model:  Google Pixel 6A

Gateway07:
<img width="1468" height="292" alt="Image" src="https://github.com/user-attachments/assets/7b867939-3ec2-4e83-8dea-fc4fef0e37b2" />

DmitryAlexei:
- [ ] not only buildings, but also streets

Zirochkabila:
# Testing 1

> OsmAnd~ 5.3.0#6541m, Released: 2025-11-19
> iOS – 5.3.0.5302
> Web – prod

### **Severity: 🟠 Major**

[...]

PR #23561, opened 2025-09-29
search-hyphen

Gateway07:
https://github.com/osmandapp/OsmAnd/issues/23320#issue-3345890301

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
+++ b/OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
@@ -786,11 +786,11 @@ public class SearchPhrase {
     }
 
 	public static class NameStringMatcher implements StringMatcher {
-
+		private static final Pattern DIGIT_SEP_LETTER = Pattern.compile("(\\d)\\s*[-–—/\\\\.]?\\s*(\\p{L})");
 		private CollatorStringMatcher sm;
 
 		public NameStringMatcher(String namePart, StringMatcherMode mode) {
-			sm = new CollatorStringMatcher(namePart, mode);
+			sm = new CollatorStringMatcher(normalizeHouseNumber(namePart), mode);
 		}
 		
 		public boolean matches(Collection<String> map) {
@@ -807,14 +807,22 @@ public class SearchPhrase {
 
 		@Override
 		public boolean matches(String name) {
-			if (name == null || name.length() == 0) {
+			if (name == null || name.isEmpty()) {
 				return false;
 			}
-			return sm.matches(name);
+			return sm.matches(normalizeHouseNumber(name));
 		}
-		
+
+		private static String normalizeHouseNumber(String s) {
+			if (s == null || s.isEmpty()) {
+				return s;
+			}
+
+			// Collapse a single separator between trailing digits and an immediately following letter
+			return DIGIT_SEP_LETTER.matcher(s).replaceAll("$1$2");
+		}
 	}
-	
+
 	public int countUnknownWordsMatchMainResult(SearchResult sr) {
 		return countUnknownWordsMatchInternal(sr, null, 0);
 	}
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
```diff
@@ -807,14 +806,39 @@ public boolean matches(Collection<String> map) {
...
 				return false;
 			}
 			return sm.matches(name);
 		}
-		
 	}
-	
+
+	public static class BuldingNameStringMatcher extends NameStringMatcher {
+		private static final Pattern pattern = Pattern.compile("(\\d)\\s*[-–—/\\\\.]?\\s*(\\p{L})");
```
vshcherb, line 817: we don't use regex in search, only for-loops and clean checks

vshcherb: Can we double check all addr:housenumber (probably in overpass somehow) and see all special symbols and add which of them to ignore or do normalization.

OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
```diff
@@ -807,14 +806,39 @@ public boolean matches(Collection<String> map) {
...
+
+		public BuldingNameStringMatcher(String namePart, boolean complete) {
+			super(normalize(namePart), complete ?
+					StringMatcherMode.CHECK_EQUALS_FROM_SPACE :
+					StringMatcherMode.CHECK_STARTS_FROM_SPACE);
+		}
+
+		@Override
+		public boolean matches(String name) {
+			if (name == null || name.isEmpty()) {
```
vshcherb, line 827: double check for null not needed

OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
```diff
@@ -908,13 +932,15 @@ private int getUnknownWordToSearchBuildingInd() {
 		} 
 		return 0;
 	}
-	
+
 	public NameStringMatcher getUnknownWordToSearchBuildingNameMatcher() {
 		int ind = getUnknownWordToSearchBuildingInd();
```
vshcherb, line 937: No caches comparing to firstUnknownNameStringMatcher, ... probably not needed of here

OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -1586,15 +1586,15 @@ public int compare(Building o1, Building o2) {
 				}
 				String lw = phrase.getUnknownWordToSearchBuilding();
 				NameStringMatcher buildingMatch = phrase.getUnknownWordToSearchBuildingNameMatcher();
-				NameStringMatcher startMatch = new NameStringMatcher(lw, StringMatcherMode.CHECK_ONLY_STARTS_WITH);
```
vshcherb, line 1589: Looks like methods countUnknownWordsMatchMainResult, ResultCompareStep also could be affected by this change so overview is missing

OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
vshcherb: Also we have only predefined list of chars to filter we can do a simple loop over chars if(ch == '-' || ch == '/') normalization is needed. However different chars could require different normalization '58/43' != '5843' however '58/43' = '58-43'

OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
Gateway07: Approve proposed 7 separators: - (U+002D), – (U+2013), — (U+2014), − (U+2212), /, \, .

OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
Gateway07: Approve proposed rules:
- Collapses a proposed separator between trailing DIGITS and following LETTERS (“15-A” → “15A”).
- Normalizes any separator between DIGITS only to “-” (“58/43” → “58-43”).

OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
Gateway07: Not sure I understand your comment. What does it mean specifically?

OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
Gateway07: As I see no code changes needed for countUnknownWordsMatchMainResult() or ResultCompareStep.

vshcherb:
Another fix is done

PR #23572, opened 2025-09-30
Fixed single finger tap after two-finger distance ruler

Corwin-Kh:


Chumva:
It should be fixed [here](https://github.com/osmandapp/OsmAnd/pull/23614)

Issue #22059, opened 2025-03-12
Improve search for US addresses (TIger)


vshcherb:
Based on analysis follow up was created
According to test I suggest next technical improvements:
- [ ] Create abbreviations for states. E.g. "Pennsylvania": ["Pa.", "PA", "Penn", "Penn.", "Penna.", "US-PA"]. Full list [here](https://en.wikipedia.org/wiki/List_of_U.S._state_and_territory_abbreviations).
- [ ] To give preference some *.obf files during searching if was found state name. For USA users only.
- [ ] Create abbreviations of common words, full list [here](https://pe.usps.com/text/pub28/28apc_002.htm).
- [ ] Improve [Tiger generation](https://creator.osmand.net:8080/view/OSM%20Planet/job/Planet_GenerateTiger/). Preselect admin_centre name instead of county.
- [ ] Check interpolation in address search (bug).
- [ ] For full address search (copy/paste or more 3 words) preselect location according to region (search in Worlds_basemap_mini map section).

DmitryAlexei:
May be related: plus code + **city** doesn't find the correct address
- https://github.com/osmandapp/OsmAnd/issues/7022#issuecomment-2764690014

Zirochkabila:
Feedback from the user https://github.com/osmandapp/OsmAnd/issues/18924#issuecomment-2956548004

> Recently switched to osmand and have become a map contributor. I see this issue moving forward as one of the biggest problems for adoption. Majority of addresses I attempt to find fail to return results. Basic US address formats.
> 
> XXXth St E, Puyallup, WA 98374 for example does not return the street at all. The result attempts to return a nearby street, XXXth St Ct E.
> 
> Alexander Ave SE XXX, Auburn, WA 98092 does not return an address despite being in the preferred format nor does it return it when having the house numbers ahead of the street.
> 
> All of these resolve under major map provider apps.

sonora:
As a long distance goal, we will probably have to look into a pattern matching approach and algorithm to identify if something is meant as an address and which one, like e.g. outlined here https://wiki.discoverydataservice.org/index.php?title=UPRN_address_matching_algorithm. 

[...]

PR #23573, opened 2025-10-01
Sort address search. Precise vs interpolated

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/22059#issuecomment-3345962865


OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
```diff
@@ -1219,6 +1230,17 @@ public int compare(SearchResult o1, SearchResult o2, SearchResultComparator c) {
 				}
 				break;
 			}
+			case COMPARE_INTERPOLATED:
+				if (o1.object instanceof Building building1 &&
+						o2.object instanceof Building building2) {
+					boolean interpolated1 = building1.getInterpolationType() != null;
```
vshcherb, line 1236: add sort by id (to have osm earlier if interpolation)

Issue #23306, opened 2025-08-20
Display local names as 2nd line (iOS regression)


vshcherb:
That's a regression on iOS, on Android it's a missing feature. First we need to fix on iOS 

Feature is visible here
https://docs.osmand.net/docs/user/map/vector-maps/?_highlight=map&_highlight=lan&current-os=ios#map-language

vshcherb:
https://github.com/osmandapp/web/issues/952

ihoryakovenko:
qa:
1) IOS check if local names are displayed, and if names do not intersect each other
2) Check if the feature works properly on Android

dmpr0:
## UI Review

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/c49c564c-7b15-4d5f-a17a-e12239bb12dc"  width="300" height="auto"/> </td>
  </tr>
</table>   


- [x] Wikidata ID is shown below continents names

Zirochkabila:
# Testing 1

`OsmAnd~ 5.2.0#5961m, Released: 2025-09-04`
`5.2.1.5229`

### **Severity**: 🟡 Minor / **Priority**: 🟠 Normal

- [ ] 1. Some names are still intertwined with each other, while others (Kaposvár) have a large space between them.

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/b9004fbb-f2cc-4fa6-ad44-1a10f17d4b02"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/2cbdfb2a-6486-442e-85bb-db419037c0fe"  width="300" height="auto"/> </th>
  </tr>
</table>  
</details>

ihoryakovenko:
QA: 

- [x] Check names positions on IOS and Android with different Text size.

Zirochkabila:
# Testing 2
`5.2.1.5248`
`OsmAnd~ 5.2.0#6225m, Released: 2025-10-08`

Text display with the following text sizes (75%, 100%, 200%, 300%) has been verified - **Fixed** (no overlapping in name position or large gaps observed).

Zirochkabila:
- [x] Basemap should use the phone's language
`map language only affects detailed maps from 9 zoom level`

<table style="width:100%">
  <tr>
[...]

PR #23582, opened 2025-10-03
Fixed map language preference

ihoryakovenko:


ihoryakovenko:

PR #23649, opened 2025-10-17
3102: Change open wiki article url

Corwin-Kh:


OsmAnd/src/net/osmand/plus/wikipedia/WikipediaArticleWikiLinkFragment.java
```diff
@@ -37,49 +79,149 @@ public void createMenuItems(Bundle savedInstanceState) {
...
-						ChoosePlanFragment.showInstance(activity, OsmAndFeature.WIKIVOYAGE);
+	@Override
+	public void downloadHasFinished() {
+		DownloadIndexesThread.DownloadEvents.super.downloadHasFinished();
+		MapActivity mapActivity = getMapActivity();
+		if (mapActivity != null) {
+			MapContextMenu contextMenu = mapActivity.getContextMenu();
+			Object menuObject = mapActivity.getContextMenu().getObject();
+			if (menuObject instanceof BaseDetailsObject baseDetailsObject) {
+				List<NativeLibrary.RenderedObject> renderedObjects = baseDetailsObject.getRenderedObjects();
```
Chumva, line 177: Do we need a specifically renderedObject? Will it not work with amenity?

PR #23650, opened 2025-10-17
Insets maphud fixes

0xRe1nk0:


OsmAnd/src/net/osmand/plus/utils/InsetsUtils.java
```diff
@@ -390,6 +389,11 @@ private static void applyClipPadding(View view) {
 		}
 	}
 
+	public static boolean isLandscape(Context context) {
```
Chumva, line 395: We already have AndroidUiHelper.isOrientationPortrait(), could we use it?

PR #23667, opened 2025-10-19
Filter non-point objects in point-only selection mode

nazar-kutz:
This change updates both `selectObjectsFromNative()` and `selectObjectsFromOpenGl()` to skip non-point rendered objects when `rules.isOnlyPoints()` is true.
Also improves consistency between OpenGL and Native selection methods, and delays `setObjectLatLon` assignment to avoid side effects.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
@@ -214,9 +214,6 @@ public class MapSelectionHelper {
 					log.warn("selectObjectsFromNative(v1) got duplicate: " + renderedObject);
 					continue;
 				}
-				if (rules.isOnlyPoints() && !renderedObject.isPoint()) {
-					continue;
-				}
 				LatLon objectLatLon = null;
 				Map<String, String> tags = renderedObject.getTags();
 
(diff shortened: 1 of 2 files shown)


OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
```diff
@@ -212,14 +214,22 @@ private void selectObjectsFromNative(@NonNull MapSelectionResult result,
 					log.warn("selectObjectsFromNative(v1) got duplicate: " + renderedObject);
 					continue;
 				}
+				if (rules.isOnlyPoints() && !renderedObject.isPoint()) {
```
Chumva: Do we really need to filter these objects?

OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
```diff
@@ -212,14 +214,19 @@ private void selectObjectsFromNative(@NonNull MapSelectionResult result,
...
 				boolean isClickableWay = app.getClickableWayHelper().isClickableWay(renderedObject);
 
-				if (!isClickableWay && !isTravelGpx && !isOldOsmRoute && (renderedObject.getId() == null
-						|| !renderedObject.isVisible() || renderedObject.isDrawOnPath())) {
+				String routeId = tags.get(ROUTE_ID);
+				boolean isNewOsmRoute = isNewOsmRoute(routeId, isTravelGpx);
+				boolean shouldFilter = renderedObject.getId() == null || !renderedObject.isVisible() || renderedObject.isDrawOnPath();
+				boolean isSpecial = isOldOsmRoute || isNewOsmRoute || isTravelGpx || isClickableWay;
+
+				if ((isSpecial && rules.isOnlyPoints()) || (!isSpecial && shouldFilter)) {
```
Chumva, line 229: Is this a correct logic change? Earlier, we did not check for isNewOsmRoute here

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22226 (ios-port-list for OSM routes and related code)

Issue #23701, opened 2025-10-23
Crash 5.2.6


Zirochkabila:
**5.2.5-5.2.6**

> 5.2.5 @Corwin-Kh 

- [x] IllegalStateException at MapContextMenuFragment https://github.com/osmandapp/OsmAnd/issues/23730
- [x] net.osmand.plus.auto.NavigationCarAppService.startForegroundWithPermission 
https://play.google.com/console/u/0/developers/8483587772816822023/app/4972741958039669892/vitals/crashes/d09630d50361db1cf661149c30f185d3/details?days=60&versionCode=5205 ((android 16)When I first launch the application, I only have permission for notifications upon map loading. I start the AA (Android Auto) emulator. The screen in the emulator shows a green button to 'Grant Location Permission'. I tap on it, and nothing happens on the screen, but a crash stack trace is being dumped into the logs. this also must be fixed)

- [x] net.osmand.plus.download.SelectIndexesHelper.showSrtmMultipleSelectionDialog https://play.google.com/console/u/0/developers/8483587772816822023/app/4972741958039669892/vitals/crashes/94e127eb5c556d984c388799c95e3c59/details?days=60&versionCode=5205

- [x] net.osmand.plus.plugins.osmedit.dialogs.BugBottomSheetDialog.createMenuItems https://play.google.com/console/u/0/developers/8483587772816822023/app/4972717699233382173/vitals/crashes/3fa6b29c15c96449e5b4917101afd4c5/details?days=60&versionCode=5206%2C5205 (enable osmedit plugin. turn show osm notes layer in configure map. click on note, click close button -> close note dialog will be shown. press home button to hide osmand app. rotate screen. show osmand app from history list -> crash )

- [x] net.osmand.plus.mapcontextmenu.other.MapMultiSelectionMenuFragment.updateNightMode https://play.google.com/console/u/0/developers/8483587772816822023/app/4972717699233382173/vitals/crashes/ddf161df4b356557db7ae2ec89beae0d/details?days=60&versionCode=5206%2C5205

-------------------------

**5.2.5**

Not reproduced - No action !
> Routing and native rendering @ivanPyrohivskyi 

[...]

PR #23706, opened 2025-10-23
Dont call native lib in ais layer constructor

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd/issues/23701

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerLayer.java
```diff
@@ -57,6 +58,11 @@ public AisTrackerLayer(@NonNull Context context) {
 		this.bitmapPaint.setFilterBitmap(true);
 		this.bitmapPaint.setStrokeWidth(4);
 		this.bitmapPaint.setColor(Color.DKGRAY);
+	}
+
+	@Override
+	public void initLayer(@NonNull OsmandMapTileView view) {
```
Chumva, line 60: Will it fix the crash? It happens in V1 rendering. Does initLayer work only in V2?

ihoryakovenko: Removing NativeUtilities fixes it

PR #23739, opened 2025-10-27
Updated zoom delegate to set propper map language

ihoryakovenko:


ihoryakovenko:

PR #23754, opened 2025-10-29
Clear mapPrimitivesProvider cache when obfMapSymbolsProvider is removed

ihoryakovenko:


ihoryakovenko:

Issue #23675, opened 2025-10-20
Add shields for nameless routes/trails/pistes (without ref/name/symbol)


dmpr0:
## ToDo
We need to add shields for all supported routes without a name, so they can be tappable from the map. We did part of that in the previous #19617 issue, but it seems we missed some types, for example `piste:type=downhill`.

**Example**
- [x] `piste:type=downhill` – https://www.openstreetmap.org/way/158661968#map=18/48.247480/24.231875
- [x] `route=fitness_trail` – https://www.openstreetmap.org/relation/13454700#map=18/48.828099/12.954445
- [x] `route=running` – https://www.openstreetmap.org/relation/1480236#map=15/50.02742/14.47235
- [x] `route=canoe` – https://www.openstreetmap.org/relation/10513361
- ??? -- Canoe launch and take-out points are enabled through Configure Map > Routes > Whitewater sports, but can be found through Search > Routes > Filtered by Whitewater sports.

<table style="width:100%">
  <tr>
    <th>piste:type=downhill</th>
    <th>route=fitness_trail</th>
    <th>route=running</th>
    <th colspan=2>route=canoe</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/51989126-1001-4740-9951-244010a57def"  width="200" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/f0234d65-41b8-472c-bc9f-56425ce6e069"  width="200" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/9f8c538c-4aa5-4fae-972b-ad5d57c9bff4"  width="200" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/5bd55526-091f-4f16-97e1-339746c2a93b"  width="200" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/a9e3258a-4ad4-4a27-a05f-4b2e3aeecca1"  width="200" height="auto"/> </td>
  </tr>
</table>

RZR-UA:
To Do

OSM routes
- [x] for nameless routes (no ref, no name, no symbol) generate shield_fg according to activity (i.e., mm_route_activity_type_trail_running) and without shield_bg
- [x] for fitness_trails, generate activity shield_fg even if name exists (but do not if ref exists)
[...]

PR #23765, opened 2025-10-30
[5.2] Display shields and translated activity type in ClickableWay multi-selection menu

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/23675
https://github.com/osmandapp/OsmAnd/issues/22226 (ios-port-list)


OsmAnd/src/net/osmand/data/PointDescription.java
```diff
@@ -31,6 +32,7 @@ public class PointDescription {
 	private String name = "";
 	private String typeName;
 	private String iconName;
+	private Drawable iconDrawable;
```
Chumva, line 35: Won`t we lose it during serialisation?

RZR-UA: It doesn't need to be serialized because it's used only once. Currently, ClickableWays are not saved in History, so we do not need serialization of PointDescription with a Drawable icon.

PR #23796, opened 2025-11-03
Fix POI without name in search list

ivanPyrohivskyi:


OsmAnd/src/net/osmand/plus/search/listitems/QuickSearchListItem.java
```diff
@@ -112,6 +112,18 @@ public static String getName(OsmandApplication app, SearchResult searchResult) {
 			case LOCATION:
 				LatLon latLon = searchResult.location;
 				return PointDescription.getLocationNamePlain(app, latLon.getLatitude(), latLon.getLongitude());
+			case POI:
+				String lang = app.getLanguage();
```
Chumva: Should we get lang here via searchResult.requiredSearchPhrase.getSettings().getLang()?

Issue #23622, opened 2025-10-11
Incorrect lane indication TSLL


ludodv:
### Routing engine

- [x] OsmAnd's in-app offline routing
- [x] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile

Car

### Start and end points

https://osmand.net/map/navigate/?start=51.796478,4.676810&end=51.794849,4.677002&profile=car#19/51.79567/4.67802

### Actual route

Turn right (TL|+TSLL|TSLL|C|C|TR) onto Laan der Verenigde Naties   and go 0.1 km

### Expected route

Turn right (TL|TSLL|TSLL|C|C|+TR) onto Laan der Verenigde Naties   and go 0.1 km

### Maps you used (online or offline)

- [x] Offline maps offered within the OsmAnd app for download. 
- [x] Online (tile / raster) maps OsmAnd.net Online map

### Your Environment

```
OsmAnd Version:5.2.0#6262m
Android version:15
```

### Anything else?

It says to turn right (correct), but the lane indication says to turn slightly left.

yuriiurshuliak:
**OsmAnd shows incorrect active lane indication during navigation — it suggests a slight-left turn (TSLL) instead of highlighting the right-turn lane (TR), even though the voice instruction correctly says “Turn right.”**

Similar behavior can indeed be reproduced.

**To reproduce:**

1. Open OsmAnd with **offline routing** enabled.
2. Set profile to **Car**.
3. Enter the route:

   * Start: **51.796478, 4.676810**
   * End: **51.794849, 4.677002**
4. Generate the route.
5. Observe the lane guidance at the intersection with **Laan der Verenigde Naties**.
   [Web version link](https://osmand.net/map/navigate/?start=51.796478,4.676810&end=51.794849,4.677002&profile=car#18/51.79569/4.67854)

**Observed behavior:**

* Voice instruction: “Turn right” — correct.
* Lane display: highlights **slight-left lanes (TSLL)** — incorrect.

**Expected behavior:**
Lane display should highlight the **right-turn lane (TR)** instead of the **slight-left lanes (TSLL)**.

**Actual result:**

```
TL | +TSLL | TSLL | C | C | TR
```

(suggesting two slight-left lanes before the correct right-turn lane)

**Expected result:**

```
TL | TSLL | TSLL | C | C | +TR
```

(the right-turn lane marked as active)



OsmAnd~ 5.2.0#6267m, released: 2025-10-13

[...]

PR #23802, opened 2025-11-03
Make startDirection less greedy in findActiveIndex

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/23622#event-20661358093

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
@@ -2302,10 +2302,12 @@ public class RouteResultPreparation {
 		if (rs.roadsOnLeft + rs.roadsOnRight < directions.length) {
 			int activeTurn = directions[rs.roadsOnLeft];
 			int startDirection = activeTurn; // classic beginning
-			boolean possibleTurnsDetected = possiblyLeft || possiblyRight;
-			boolean hasLeftRoadsWithLanes = rs.roadsOnLeft >0 && rs.leftLanes > 0;
+
+			boolean potentialTurnsDetected = possiblyLeft || possiblyRight;
+			boolean hasLeftRoadsWithLanes = rs.roadsOnLeft > 0 && rs.leftLanes > 0;
 			boolean hasRightRoadsWithLanes = rs.roadsOnRight > 0 && rs.rightLanes > 0;
-			if (possibleTurnsDetected && (hasLeftRoadsWithLanes || hasRightRoadsWithLanes)) {
+			if (potentialTurnsDetected && (hasLeftRoadsWithLanes || hasRightRoadsWithLanes)) {
+				// find more suitable and less greedy start direction if potential turns exist
 				for (int i = rs.roadsOnLeft; i < directions.length - rs.roadsOnRight; i++) {
 					if (possiblyLeft && TurnType.isLeftTurn(directions[i])) {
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -2299,15 +2300,31 @@ private int[] findActiveIndex(RouteSegmentResult prevSegm, RouteSegmentResult cu
 		}
 		int[] directions = getUniqTurnTypes(turnLanes);
 		if (rs.roadsOnLeft + rs.roadsOnRight < directions.length) {
-			int startDirection = directions[rs.roadsOnLeft];
+			int activeTurn = directions[rs.roadsOnLeft];
+			int startDirection = activeTurn; // classic beginning
+			boolean possibleTurnsDetected = possiblyLeft || possiblyRight;
+			boolean hasLeftRoadsWithLanes = rs.roadsOnLeft >0 && rs.leftLanes > 0;
+			boolean hasRightRoadsWithLanes = rs.roadsOnRight > 0 && rs.rightLanes > 0;
+			if (possibleTurnsDetected && (hasLeftRoadsWithLanes || hasRightRoadsWithLanes)) {
```
ivanPyrohivskyi: For what you make check (hasLeftRoadsWithLanes || hasRightRoadsWithLanes) ?
If your algorithm is right it must work without it. If - not, then you just decrease possible bugs by this check.

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -2299,15 +2300,31 @@ private int[] findActiveIndex(RouteSegmentResult prevSegm, RouteSegmentResult cu
...
+			}
 			int endDirection = directions[directions.length - rs.roadsOnRight - 1];
 			for (int i = 0; i < rawLanes.length; i++) {
 				int p = TurnType.getPrimaryTurn(rawLanes[i]);
 				int s = TurnType.getSecondaryTurn(rawLanes[i]);
 				int t = TurnType.getTertiaryTurn(rawLanes[i]);
 				if (p == startDirection || s == startDirection || t == startDirection) {
 					pair[0] = i;
-					pair[2] = startDirection;
+					pair[2] = activeTurn;
```
ivanPyrohivskyi, line 2330: Hmm. Why here activeTurn ?

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
RZR-UA: We need to cut down on "greediness" only when there are potential turns and attached roads with lanes exist.

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
RZR-UA: It should logically be refreshed startDirection, but in fact, startDirection here makes 2 tests fail:

Test 35
```
Segment 84975049
Expected :TR:TL|+TR,C
Actual   :[MUTE] C:TL|+C,TR
```

Test 36
```
Segment 35361252
Expected :TR:TL|+TR,TL,C
Actual   :C:TL|+C,TR,TL
```

RZR-UA:
https://github.com/osmandapp/OsmAnd/pull/23884

PR #23848, opened 2025-11-07
Chumva fixes

Chumva:


sonora:
@Chumva  Vitaliy, did you see this? https://github.com/osmandapp/OsmAnd/pull/23848#discussion_r2504383679 Not sure it's true, but might be?

PR #23859, opened 2025-11-09
Update phrases.xml - reorder groups

aceman444:
Group some historic,military,barrier,hazard strings together for better translation overview.


aceman444:
Thanks. This just groups strings together, it does not add any new POIs or remove duplicates. The Copilot is wrong.

PR #23860, opened 2025-11-09
Update strings.xml - group lang_

aceman444:
Group lang_* strings together for better translation overview.


aceman444:
The indentation problems are weird, I just copied the strings around using the github webui.

aceman444:
Also the Copilot review is wrong, this does not add any new language entries. That would be pointless.

PR #23870, opened 2025-11-10
Fixed crash on click update map

Corwin-Kh:


OsmAnd/src/net/osmand/plus/download/ui/UpdatesIndexFragment.java
```diff
@@ -431,21 +448,47 @@ public DownloadActivity getMyActivity() {
 		return (DownloadActivity) getActivity();
 	}
 
-	private class UpdateIndexAdapter extends ArrayAdapter<DownloadItem> implements LocalIndexInfoAdapter {
+	static class LocalIndexItem {
+		private boolean isBanner;
+		private boolean isDeletedMaps;
+		private boolean isDownloadItem;
```
Chumva: Can we avoid using multiple flags?

PR #23893, opened 2025-11-12
20740: Implemented show Destination reached AA screen

Corwin-Kh:


OsmAnd/src/net/osmand/plus/OsmandApplication.java
```diff
@@ -770,6 +771,7 @@ public NavigationSession getCarNavigationSession() {
 
 	public void onCarNavigationSessionStart(@NonNull NavigationSession carNavigationSession) {
 		androidAutoInForeground = true;
+		routingHelper.onCarNavigationStart();
 	}
 
 	public void onCarNavigationSessionStop(@NonNull NavigationSession carNavigationSession) {
```
vshcherb, line 777: Should it be check pause navigation?

PR #23901, opened 2025-11-13
fix rotation icon color

0xRe1nk0:


OsmAnd/src/net/osmand/plus/quickaction/actions/ChangeMapOrientationAction.java
```diff
@@ -78,7 +78,7 @@ public int getIconRes(Context context) {
 		OsmandApplication app = (OsmandApplication) context.getApplicationContext();
 		OsmandSettings settings = app.getSettings();
 		CompassMode compassMode = settings.getCompassMode();
-		boolean nightMode = app.getDaynightHelper().isNightMode(ThemeUsageContext.APP);
+		boolean nightMode = app.getDaynightHelper().isNightMode(ThemeUsageContext.MAP);
```
Chumva, line 81: Will it return the correct icon in different UI screens?

PR #23902, opened 2025-11-13
Added string resources for sample collection

Corwin-Kh:


OsmAnd/res/values/strings.xml
```diff
@@ -10,6 +10,7 @@
 	- For wording and consistency, please note https://docs.osmand.net/docs/technical/contributions/translating-osmand
 	Thx - Hardy
 -->
+    <string name="poi_sample_collection">Sample collection</string>
```
Chumva: Probably we should add it via phrases.xml

PR #23930, opened 2025-11-16
Allow including deactivated maps into updates list

nazar-kutz:


OsmAnd/src/net/osmand/plus/download/DownloadIndexesThread.java
```diff
@@ -546,14 +552,22 @@ protected String doInBackground(IndexItem... filesToDownload) {
 							if (bf.exists()) {
 								Algorithms.removeAllFiles(bf);
 							}
+							if (backuped) {
```
Chumva, line 555: Will it have any issues during reindexFiles?

nazar-kutz: The LocalOperationHelper class does not start any new threads — it only contains shared logic that can be safely reused across different async tasks (as in our case). Therefore, it should not introduce any multithreading issues during reindexFiles.

Similarly, the backup flag simply reflects the file’s location in another directory, so it should not negatively affect the reindexFiles logic either.

Issue #23713, opened 2025-10-23
Search POI issues (mix with address by distance)


vshcherb:
This is a task to wrap up cases & datasets where OsmAnd full text search could be improved. Mainly separate issues will be created and fixed separately but here major updates will be posted

**Fixed**
- [x] https://github.com/osmandapp/OsmAnd/issues/7412 (Think about duplicate house numbers Tiger)
- [x] https://github.com/osmandapp/OsmAnd/issues/15726
- [x] https://github.com/osmandapp/OsmAnd/issues/19059
- [x] https://github.com/osmandapp/OsmAnd/issues/23708

**In progress**
https://github.com/osmandapp/OsmAnd/pull/23770
- [x] Push objects by rating higher
- [x] Mix order POI / address for 1-2 common word search "Rossio" / "Apple" (to do find issues related to brand as attempt to fix it)
- [x] Guy's https://github.com/osmandapp/OsmAnd/issues/4491)
- [x]  https://github.com/osmandapp/OsmAnd/issues/23320 https://github.com/osmandapp/OsmAnd/issues/21479 
Related if brand is not registered - https://github.com/osmandapp/OsmAnd/issues/16397

**Search tool**
- [ ] Create cases No-code - dipslay number filtered (1. Filter Where 2. Regex replace), fix size for case
- [ ] **To think** Create test map from search tool
- [ ] **Postponed**: Add flag to find OSM objects in OBF file by tags - add column "obf_present" - example value ```<CITY_OR_POSTCODE>/<STREET_OR_GROUP>/<BUILDING>,.... ```. Node could be found inside the maps by proper tags `addr:city`, `addr:place`, `addr:suburb`, `addr:housenumber`, `addr:postcode`, ... according to documentation - could be separate utility based on osm file to run continuous separately
- [ ] **Postponed ObfChecker**: Partially complete (no idea exact check yet) Create test for check maps to check that same building name/number on same street / city is < 5 km

-----------
### Other issues

**Space issue / split words**
- [ ] https://github.com/osmandapp/OsmAnd/issues/23709, https://github.com/osmandapp/OsmAnd/issues/4923

**Ready to do**
- [ ]  https://github.com/osmandapp/OsmAnd/issues/23697
- [ ] Merge streets with Suburbs -   https://github.com/osmandapp/OsmAnd/issues/11345
- [ ] https://github.com/osmandapp/OsmAnd/issues/16127

**POI & Address mixed**
[...]

PR #23975, opened 2025-11-20
Deduplicate transport in search

ivanPyrohivskyi:
To issue: _"Deduplicate if search is by name public transport stops ( Quedlinburg ) as on screenshot - only station should be displayed and merge transpost_stop_position, platform."_
https://github.com/osmandapp/OsmAnd/issues/23713#issuecomment-3481299688


OsmAnd-java/src/main/java/net/osmand/data/Amenity.java
```diff
@@ -1034,4 +1046,11 @@ public Map<String, String> getOsmTags() {
 
 		return result;
 	}
+
+	public boolean isTransport() {
+		if (type != null) {
+			return "transportation".equals(type.getKeyName());
```
alex-dev-neo, line 1053: "transportation" to constant

OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
```diff
@@ -437,6 +437,8 @@ public boolean sameSearchResult(SearchResult r1, SearchResult r2) {
 
 						if (type1.equals("natural")) {
 							similarityRadius = 50000;
+						} else if (type1.equals("transportation")) {
```
vshcherb, line 440: Same code 
private static final Set<String> FILTER_DUPLICATE_POI_SUBTYPE = new TreeSet<String>(
			Arrays.asList("building", "internet_access_yes"));

OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
```diff
@@ -831,6 +831,7 @@ private Amenity readPoiPoint(int left31, int right31, int top31, int bottom31,
 					}
 				}
 				am.setRegionName(region.getName());
+				am.initOrder();
```
vshcherb, line 834: why do we change in this request?

OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
```diff
@@ -1263,6 +1266,18 @@ public int compare(SearchResult o1, SearchResult o2, SearchResultComparator c) {
 				}
 				break;
 			}
+			case TRANSPORTATION_ORDER:
```
vshcherb, line 1269: Already exists concept for POI deduplication
```
} else if (ObjectType.isAddress(r1.objectType) && ObjectType.isAddress(r2.objectType)) {
						if (interpolated) {
							similarityRadius = 1000;
						} else {
							similarityRadius = 100;
						}
```

OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapPoiReaderAdapter.java
ivanPyrohivskyi: We use amenity order only in map generation. So for same location we can have several amenities in same order.
But here we need order on client side, because public_transport_station, public_transport_stop, public_transport_platform are not in the same location.

OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
ivanPyrohivskyi: Not same!
Really we need ``<poi_filter name="public_transport" top="true">``
POI_SUBTYPE ``public_transport_station`` != ``public_transport_stop``

OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
ivanPyrohivskyi: TRANSPORTATION_ORDER need for set order like (for same name="Quedlinburg"):
1. public_transport_station 
2. public_transport_platform
3. public_transport_stop

After order we can "glue" public_transport_stop and public_transport_platform into **public_transport_station** . Otherwise deduplication will depend from COMPARE_BY_DISTANCE

OsmAnd-java/src/main/java/net/osmand/data/Amenity.java
ivanPyrohivskyi, line 1053: Need change to <poi_filter name="public_transport" top="true">

OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
```diff
@@ -1163,6 +1165,7 @@ public JSONObject createTestJSON(SearchResultCollection searchResult) {
 	private enum ResultCompareStep {
 		TOP_VISIBLE,
 		FOUND_WORD_COUNT, // more is better (top)
+		TRANSPORTATION_ORDER,
 		OBF_RESOURCE,
 		UNKNOWN_PHRASE_MATCH_WEIGHT, // more is better (top)
```
vshcherb, line 1170: Try to increase weight for stations (railway station, aeroport, subway station).

Issue #23957, opened 2025-11-19
POI Type is not shown in What's here menu — Regression


dmpr0:
### Actual result
<table style="width:100%">
  <tr>
    <th>4.9.10</th>
    <th>5.2</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/3c780867-6948-4704-9009-610c22ddeea5"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/ed15dcca-4d0e-4ed0-9a5d-29322e9c1bb5"  width="300" height="auto"/> </td>
  </tr>
</table>  
 
* POI Type is not shown starting 5.1 release

### Expected result

* POI Type • Address 
* Additionally check what shows for Sample collection POI type

**Environment**
OsmAnd Version: OsmAnd~ 5.3.0#6541m, released: 2025-11-19
Android: 16
Device model: Google Pixel 6A

RZR-UA:
@Zirochkabila, please test build # 6559

Zirochkabila:
# Testing 1

> 5.2.5.5286
> OsmAnd~ 5.3.0#6559m, Released: 2025-11-21

**Fixed** for Android - needs to be implemented for iOS - @dmpr0  migration may be required

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/20d9d47a-e9c4-49e8-9f8d-e4c630ed0c63"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/22fdd2e6-42f1-4b0f-b36b-45de2799638e"  width="300" height="auto"/> </th>
  </tr>
</table>    
</details> 

________

Sample collection POI type - various information is displayed (it seems to depend on the number of objects in the What's here menu) - probably related to [23414](https://github.com/osmandapp/OsmAnd/issues/23414#issuecomment-3548332288)


<details><summary>Details</summary>
<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/0573f44c-b2a0-497c-a57b-54333ea96a9c"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/3769c540-2c88-443f-81a4-90ca4e4edf19"  width="300" height="auto"/> </th>
[...]

PR #23977, opened 2025-11-21
[5.2] Refactor typeStr and nameStr in RenderedObject menu, fix 5.2 build after switch from master

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/23957


OsmAnd-java/build.gradle
```diff
@@ -51,6 +51,7 @@ tasks.register('collectTestPhrasesResources', Copy) {
 }
 
 tasks.register('collectTestResources', Copy) {
+	delete fileTree("src/test/resources/search")
```
Chumva, line 54: Do we need similar clean-up for other collect methods?

RZR-UA: Nope. If the build problem between branches happens for other folders, we will fix it.

PR #23986, opened 2025-11-23
update of AIS tracker plugin

f-n:
After testing at a holiday trip on a boat I had some improvements and ideas that I like to offer as an update:

- new: ability to show the shape of the vessel if dimensions are available
- new: decode some missing AIS ship types
- new: ability to hide the own AIS object in case that the own vessel has an AIS transmitter
- shorten the direction lines of moving vessels
- fix: no need to cleanup resources on restart of network listener
- new setting: AIS message receiver may stay active if OsmAnd gets paused (in my opinion this should be always active, but after code refactoring, the network listener always was stopped in case of mapActivityPause)

Best Regards
Falk


gradle.properties
```diff
@@ -1,29 +1,20 @@
-## Project-wide Gradle settings.
-#
-# For more details on how to configure your build environment visit
+## For more details on how to configure your build environment visit
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 #
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-# Default value: -Xmx10248m -XX:MaxPermSize=256m
+# Default value: -Xmx1024m -XX:MaxPermSize=256m
```
Chumva, line 6: These changes should be local, not committed to pr

gradle.properties
```diff
@@ -1,29 +1,20 @@
-## Project-wide Gradle settings.
```
alex-dev-neo, line 1: Do not change this file

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerPlugin.java
```diff
@@ -315,15 +327,17 @@ public void mapActivityResume(@NonNull MapActivity activity) {
 
 	@Override
 	public void mapActivityPause(@NonNull MapActivity activity) {
-		stopAisListener();
+        if (!AIS_RECEIVE_IN_BACKGROUND.get()) {
```
vshcherb, line 330: Does it work correctly without service?

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisObjectDrawable.java
```diff
@@ -64,7 +66,7 @@ private void invalidateBitmap() {
 	}
 
 	public void set(@NonNull AisObject ais) {
-		this.ais.set(ais);
+		// this.ais.set(ais);
```
alex-dev-neo, line 69: Why commented?

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerLayer.java
```diff
@@ -86,12 +89,27 @@ public void cleanupResources() {
...
 	@Override
 	public void onAisObjectReceived(@NonNull AisObject ais) {
-		AisObjectDrawable drawable = objectDrawables.get(ais.getMmsi());
+		int mmsi = ais.getMmsi();
+		boolean showOwnObject = plugin.AIS_DISPLAY_OWN_POSITION.get();
+		boolean own = (mmsi == plugin.AIS_OWN_MMSI.get());
+		AisObjectDrawable drawable = objectDrawables.get(mmsi);
 		if (drawable == null) {
+			if ((own) && (!showOwnObject)) {
+				AisObjectDrawable.setOwnObject(null);
```
vshcherb, line 98: To do refactor to avoid static methods, here the flag AisObjectDrawable.setOwnLocation(true/false)

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerPlugin.java
alex-dev-neo, line 330: Looks like `OsmandNotification` and foreground service is needed to work in background

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisObjectDrawable.java
```diff
@@ -230,32 +235,88 @@ private void drawCircle(float locationX, float locationY,
 		canvas.drawCircle(locationX, locationY, 18.0f, localPaint);
 	}
 
+	private void drawShape(float locationX, float locationY, @NonNull RotatedTileBox tileBox,
```
vshcherb, line 238: @Chumva what case will we display svg?

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisObjectDrawable.java
f-n: This one I remember: The call seems to be redundant, since the ais object is already initialized in function AisTrackerPlugin.AisDataManager.onAisObjectReceived()

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerPlugin.java
f-n: at least in my tests it was working.
in my opinion, the listener should always stay running if the app is in background/paused, so this method may stay empty.

Chumva:
Thank you for the contribution and for the AIS plugin improvements

The proposed changes have been reviewed and migrated to the [KMP-based implementation](https://github.com/osmandapp/OsmAnd/pull/25313), and the functionality is now covered in another PR. Because of that, I’m closing this PR as superseded

PR #23996, opened 2025-11-24
Sync TransportRoutePlanner with C++ (Java improvements)

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22515

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
@@ -316,14 +316,6 @@ public class TransportRoutePlanner {
 					System.out.printf("ptLimitResultsByNumber (%d) reached\n", limitByNumber);
 					break;
 				}
-				int limitByPercent = ctx.cfg.ptLimitResultsByPercent;
-				if (!lst.isEmpty() && limitByPercent > 0) {
-					double timeLimit = lst.get(0).getRouteTime() * (1.0 + (double) limitByPercent / 100.0);
-					if (route.getRouteTime() > timeLimit) {
-						System.out.printf("ptLimitResultsByPercent (%d%%) reached\n", limitByPercent);
-						break;
-					}
-				}
 				System.out.println(route);
 				lst.add(route);
 			}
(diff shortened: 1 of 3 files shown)


OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -314,28 +311,39 @@ private List<TransportRouteResult> prepareResults(TransportRoutingContext ctx, L
...
+				}
+				int limitByPercent = ctx.cfg.ptLimitResultsByPercent;
+				if (!lst.isEmpty() && limitByPercent > 0) {
+					double timeLimit = lst.get(0).getRouteTime() * (1.0 + (double) limitByPercent / 100.0);
+					if (route.getRouteTime() > timeLimit) {
+						System.out.printf("ptLimitResultsByPercent (%d%%) reached\n", limitByPercent);
+						break;
+					}
+				}
+				System.out.println(route);
```
vshcherb, line 319: ptLimitResults not needed we already have 2 constants

public double increaseForAlternativesRoutes = 2.0;
public double increaseForAltRoutesWalking = 3.0;

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -314,28 +311,39 @@ private List<TransportRouteResult> prepareResults(TransportRoutingContext ctx, L
 				}
 			}
 			if (!exclude) {
+				int limitByNumber = ctx.cfg.ptLimitResultsByNumber;
```
vshcherb, line 314: create constant instead of parameter TransportRoutingConfiguration

RZR-UA: done (const already exists)

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
RZR-UA: ptLimitResults only limits final results and do not affect algorithm. algorithm now seems good for both small and big cities. pt_exclude removed. pt_limit disabled and set = 50 by default const. tested ok.

PR #24004, opened 2025-11-26
common-word-in-street-name-fix

Gateway07:
- **Remove common word from street name**
- **Add Polish street terms and abbreviations**

https://github.com/osmandapp/OsmAnd/issues/23713#issuecomment-3481299688

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
+++ b/OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
@@ -208,8 +208,15 @@ public class SearchResult {
 		} else {
 			localResultNames = SearchPhrase.splitWords(name, new ArrayList<String>(), SearchPhrase.ALLDELIMITERS);
 		}
-		if (object instanceof Street)
-			localResultNames = localResultNames.stream().filter(x -> !CommonWords.isStaticCommon(x.toLowerCase())).toList();
+		if (object instanceof Street) {
+			List<String> filteredNames = new ArrayList<>();
+			for (String localResultName : localResultNames) {
+				if (!CommonWords.isStaticCommon(localResultName.toLowerCase())) {
+					filteredNames.add(localResultName);
+				}
+			}
+			localResultNames = filteredNames;
+		}
 
 		boolean wordMatched;
 		if (searchPhraseNames.isEmpty()) {


OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
```diff
@@ -207,7 +208,9 @@ private boolean allWordsMatched(String name, SearchResult exactResult, CheckWord
 		} else {
 			localResultNames = SearchPhrase.splitWords(name, new ArrayList<String>(), SearchPhrase.ALLDELIMITERS);
 		}
-		
+		if (object instanceof Street)
+			localResultNames = localResultNames.stream().filter(x -> CommonWords.getCommon(x.toLowerCase()) < 0).toList();
```
vshcherb: region names  should be excluded from common words

OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
```diff
@@ -1360,6 +1360,16 @@ private static void parseRegionNames(WorldRegion region, Set<String> result) {
 		addCommon("trafficway");
 		addCommon("plaine");
 
+		// Polish street terms and abbreviations
```
vshcherb: ? why it's here and only polish names? "ulica" - is not a frequently used

OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
Gateway07: How to exclude region names from common words? Everything is in single commonWordsDictionary map.

OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
Gateway07: addCommon("ulica"); was already added. I've added common words related to Poland that were not before..

OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
```diff
@@ -207,7 +208,9 @@ private boolean allWordsMatched(String name, SearchResult exactResult, CheckWord
 		} else {
 			localResultNames = SearchPhrase.splitWords(name, new ArrayList<String>(), SearchPhrase.ALLDELIMITERS);
 		}
-		
+		if (object instanceof Street)
```
vshcherb: 1. better write normal loop
2. Missing braces for if

OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
vshcherb: Let's remove  it as 
1. Not generic just 1 language
2. ul. which means ulican is not a frequently used but a common word
3. I didn't find any streets in OSM in Warsaw with "ul."

OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
```diff
@@ -207,7 +208,16 @@ private boolean allWordsMatched(String name, SearchResult exactResult, CheckWord
 		} else {
 			localResultNames = SearchPhrase.splitWords(name, new ArrayList<String>(), SearchPhrase.ALLDELIMITERS);
 		}
-		
+		if (object instanceof Street) {
```
vshcherb, line 211: bad change breaks the tests, couldn't be fixed. https://github.com/osmandapp/OsmAnd/commit/361e42368571f9b0f06f21d9de733caeb819a9e0

PR #24006, opened 2025-11-26
paremtSearchResult-set-null-fix

Gateway07:
Wrong parameters position fix


vshcherb:
Fix doesn't. look correct

PR #24007, opened 2025-11-26
Filter duplicated amenities in "Nearby popular places" list

nazar-kutz:


OsmAnd/src/net/osmand/plus/mapcontextmenu/SearchAmenitiesTask.java
```diff
@@ -46,16 +50,24 @@ protected List<Amenity> doInBackground(Void... params) {
...
-			amenities = getAmenities(rect);
-			amenities.remove(amenity);
+
+			amenities = collectAmenities(rect);
 			radius *= NEARBY_POI_SEARCH_FACTOR;
+
+			// If enough POIs collected OR search radius limit reached — merge duplicates.
+			if (amenities.size() >= NEARBY_MAX_POI_COUNT || radius > NEARBY_POI_MAX_RADIUS) {
+				if (currentAmenity != null) {
+					amenities.add(currentAmenity);
```
Chumva, line 60: Do we need to remove existing amenity from list?

nazar-kutz: It will be filtered inside the amenityObjectsMerger.merge call

nazar-kutz: it is needed to filter not only simple amenity but the full cluster of the amenities those are overlapped with current one

PR #24018, opened 2025-11-27
Share favorite: include favorite name by adding "label" to geo URI

nazar-kutz:


OsmAnd-java/src/main/java/net/osmand/util/MapUtils.java
```diff
@@ -426,9 +426,62 @@ public int compare(MapObject o1, MapObject o2) {
...
+	public static String buildGeoUrl(String latitude, String longitude, int zoom, String label) {
+		String url = "geo:" + latitude + "," + longitude + "?z=" + zoom;
+		if (label != null && !label.isEmpty()) {
+			String encodedLabel = encodeLabel(label);
+			return url + "&q=" + latitude + "," + longitude + "(" + encodedLabel + ")";
+		}
+		return url;
+	}
+
+	public static String encodeLabel(String label) {
```
Chumva: URLEncoder.encode()? Or we could use GeoParsedPoint.getGeoUriString()

nazar-kutz: You are absolutely right. I checked GeoParsedPoint and it handles the label generation perfectly.

I’ve refactored the code to use GeoParsedPoint.getGeoUriString() and reverted the custom implementation in MapUtils.

OsmAnd/src/net/osmand/plus/mapcontextmenu/other/ShareMenu.java
```diff
@@ -169,15 +169,16 @@ private void setupSharingFields(@NonNull MapActivity activity) {
...
-			geoUrl = MapUtils.buildGeoUrl(lat, lon, zoom);
-			link = buildOsmandPoiUri(title, typeStr,
-					lat, lon,
-					zoom, lat, lon);
+
+			GeoParsedPoint parsedPoint = new GeoParsedPoint(latVal, lonVal, zoom, title);
+			geoUrl = parsedPoint.getGeoUriString();
+
+			String latStr = String.format(Locale.US, "%.6f", latVal);
+			String lonStr = String.format(Locale.US, "%.6f", lonVal);
```
Chumva: Why we replaced LocationConvert.convertLatitude() with String.format(Locale.US, "%.6f"...)?

Issue #23715, opened 2025-10-23
Notify & suggest migration for deprecated maps  (2025 Germany_nordrhein-westfalen)


valynor:
### Description

Since October 2025 one of the Germany maps is missing: 

https://download.osmand.net/download?standard=yes&file=Germany_nordrhein-westfalen_europe_2.obf.zip
is not present anymore. 

This is the main map of Nordrhein-Westfalen combining several sub-maps. 

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/3545f3c1-8d45-4718-a28e-e91fe9046cc1" />

sonora:
And in fact (as per user omplaints) if you have an older Nordrhein-Westfalen Standard map like from July on your device, and try a short route entirely in e.g. Cologne (Nordrhein-Westfalen), the app produces the 'outdated maps' HH routing errror message.

Why??

Zirochkabila:
@valynor As far as I can see, the North Rhine-Westphalia map has not disappeared—it has been updated and is now available as a “Standard map.” This change is part of a global update to the map loading system that was implemented in issue [#22438](https://github.com/osmandapp/OsmAnd/issues/22438). 

<details><summary>Details</summary>
<img width="400" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/f59da091-715d-44b3-a602-7e3cdbba567f">
</details>

scaidermern:
It is not available as a standard map, it has been split into 5 regions. The full region as a single map doesn't exist anymore.

Looks like there is no migration available from the previous single standard map to the new 5 regions. Why?
For users it isn't obvious that the old map is deprecated and must be replaced by individual regions. The old map is just visualized in green, meaning it is up-to-date. But instead it is outdated and doesn't work for routing if newer maps are installed.

Please introduce a migration mechanism for these kind of changes. Don't let users have to figure this out on their own.

scaidermern:
Looks like Germany_bayern_europe.pbf (Bavaria), Germany_baden-wuerttemberg_europe.pbf, Germany_hessen_europe.pbf, Germany_niedersachsen_europe.pbf all have the same problems and aren't available anymore as a single map.

scaidermern:
https://download.osmand.net/list still shows these full maps and another user reports that they are still available with OsmAnd 5.1.9. So maybe this is a bug in OsmAnd 5.2. If not, please introduce a migration mechanism.

sonora:
[...]

PR #24024, opened 2025-11-28
Fixed trimming map version if no version present

Corwin-Kh:


Zirochkabila:
https://github.com/osmandapp/OsmAnd/issues/23715

PR #24038, opened 2025-12-02
improve intent parser for new links

0xRe1nk0:


OsmAnd-java/src/main/java/net/osmand/search/AmenitySearcher.java
```diff
@@ -269,6 +274,9 @@ public List<Amenity> filterAmenities(Collection<Amenity> amenities, Request requ
                 if (amenity != null) {
                     filtered = filterByOsmIdOrWikidata(
                             amenities, amenity.getOsmId(), amenity.getLocation(), amenity.getWikidata());
+                    if (!Algorithms.isEmpty(request.amenityCategoryKeyName)) {
```
Chumva: Can it break anything in search? We could use SearchPoiTypeFilter in amenitySearcher.searchAmenities() for this

PR #24051, opened 2025-12-04
Implement initDisposableRegionsFromResources() (useful for tools)

RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2951


OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
```diff
@@ -104,17 +105,31 @@ public void setTranslator(RegionTranslation translator) {
...
+			InputStream is = OsmandRegions.class.getResourceAsStream(REGIONS_OCBF);
 			FileOutputStream fous = new FileOutputStream(regions);
 			Algorithms.streamCopy(is, fous);
 			fous.close();
 //		}
 		return prepareFile(regions.getAbsolutePath());
 	}
-	
+
+	public void initDisposableRegionsFromResources() throws IOException {
```
ivanPyrohivskyi: Add comment or rename. For parallel runs.

RZR-UA: PlatformUtil.initOsmandRegionsAsTempFile();
osmandRegions.prepareRegionsFromResourcesAsTempFile();

Issue #24087, opened 2025-12-12
Search could ignore common words mismatch during sorting


vshcherb:
While search in Warsaw
"18 Bohaterów Września"

**Expected** result is on 3rd place

<img width="593" height="208" alt="Image" src="https://github.com/user-attachments/assets/b7445c70-b75f-45e5-8cb5-3088834aa88e" />

If we could somehow ignore that "Aleja" is a common word and result could be put to top place

Gateway07:
With 52.213997, 20.961227 and "18 Bohaterów Września" I see result on 5th place:
<img width="656" height="432" alt="Image" src="https://github.com/user-attachments/assets/e0f7348c-efb5-4192-9931-31de792741d9" />

ivanPyrohivskyi:
Fixed in PR 
https://github.com/osmandapp/OsmAnd/pull/24360
https://github.com/osmandapp/OsmAnd-resources/pull/1335
https://github.com/osmandapp/OsmAnd-iOS/pull/5175

Zirochkabila:
# Testing 1

> Web - prod
> OsmAnd Maps 5339, released 20.02.26
> OsmAnd~ 5.3.0#7214m, Released: 2026-02-20

Search for the address “18 Bohaterów Września” in Warsaw has been verified - the exact result is now displayed in the first place, the word “Aleja” no longer overrides the priority of partial matches - **Fixed**

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/2f44e78d-3213-4a51-bd7e-7529b0bd83fa"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/226e8f8b-ddad-4082-ae0b-71801febf247"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/46fd7428-3d8e-4c35-af90-e93873a86136"  width="300" height="auto"/> </th>
  </tr>
</table>  
</details> 

__________


Search for the address “18 Bohaterów Września” in Warsaw on **Android and iOS**: at first, the result with ‘Aleja’ is displayed for a moment, then the list is updated and the exact address “18 Bohaterów Września” takes first place - **fixed works, but there is a momentary flicker with incorrect priority**.


<details><summary>Details</summary>
<table style="width:100%">
  <tr>
    <th>Web - as expected</th>
    <th>Android</th>
    <th>iOS</th>
  </tr>
[...]

PR #24118, opened 2025-12-16
common-word-in-street-fix

Gateway07:
Add check of common word in street

https://github.com/osmandapp/OsmAnd/issues/24087

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
+++ b/OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
@@ -29,6 +29,7 @@ public class CommonWords {
 	public static boolean isCommon(String name) {
 		return commonWordsDictionary.containsKey(name) || isNumber2Letters(name);
 	}
+
 	public static int getCommon(String name) {
 		if (isNumber2Letters(name)) {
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
```diff
@@ -23,14 +24,18 @@ private static void addFrequentlyUsed(String string) {
 		frequentlyUsedWordsDictionary.put(string, frequentlyUsedWordsDictionary.size());
 	}
 
+	public static boolean isStaticCommon(String name) {
```
vshcherb: public static boolean isCommon(String name) {
already exists

OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
```diff
@@ -113,6 +117,8 @@ private static void addAbbrevations() {
 	}
 	
 	private static void addRegionNames() {
+		staticCommonWords.addAll(commonWordsDictionary.keySet());
```
vshcherb: on purpose separated should not be combined !

PR #24121, opened 2025-12-16
Fix search items layout

Corwin-Kh:


OsmAnd/src/net/osmand/plus/mapcontextmenu/other/DelegateDrawTextView.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.graphics.Canvas
+import android.util.AttributeSet
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import androidx.appcompat.widget.AppCompatTextView
+import androidx.core.view.children
+import net.osmand.plus.mapcontextmenu.controllers.NetworkRouteDrawable
+
+class DelegateDrawTextView @JvmOverloads constructor(
```
Chumva, line 13: Do we really need these custom views?

PR #24163, opened 2025-12-22
Added 3d building transperency and details settings

ihoryakovenko:
https://github.com/osmandapp/OsmAnd-Issues/issues/2529


OsmAnd/src/net/osmand/plus/plugins/srtm/Buildings3DFragment.java
```diff
@@ -69,6 +82,19 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
 		setupContent(view);
 
 		updateUiMode();
+
+		if (plugin.ENABLE_3D_MAP_OBJECTS.get()) {
+			float alpha = plugin.BUILDINGS_3D_ALPHA.get();
+			if (alpha <= 0f || alpha > 1f) {
+				alpha = 0.7f;
```
Chumva, line 89: By default, preference has `0.8f`, here `0.7f`, which one is correct?

PR #24166, opened 2025-12-22
Implement organize by

Corwin-Kh:


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/data/ComparableTracksGroup.kt
```diff
@@ -7,4 +7,5 @@ interface ComparableTracksGroup {
 	fun getDirName(includingSubdirs: Boolean): String
 	fun lastModified(): Long
 	fun getDefaultOrder(): Int = -1
+	fun getSortValue(): Double = 0.0
```
alex-dev-neo: is it order? SortValue - bad name. Why double? is Int enough?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/api/OsmAndContext.kt
```diff
@@ -27,6 +28,7 @@ interface OsmAndContext {
 	fun getSelectedFileByPath(path: String): GpxFile?
 	fun getNameStringMatcher(name: String, mode: KStringMatcherMode): KStringMatcher
 	fun getTrackPointsAnalyser(): TrackPointsAnalyser?
+	fun getOrganizeTracksResourceMapper(): OrganizeTracksResourceMapper
```
alex-dev-neo: OrganizeTracksResourceMapper should be moved to commonMain.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/organization/enums/OrganizeByType.kt
```diff
@@ -0,0 +1,228 @@
...
+		strategy = OrganizeByRangeStrategy
+	);
+
+	fun getName() = Localization.getString(nameResId)
+
+	fun getIconName() = iconResId
+
+	fun getTrackSortScope() = strategy.getTrackSortScope()
+
+	fun getGpxParameter() = filterType.property!!
```
alex-dev-neo: !! ?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/organization/enums/OrganizeByType.kt
```diff
@@ -0,0 +1,228 @@
...
+
+	fun getDisplayUnits() = filterType.measureUnitType.getUnit()
+
+	fun getDefaultStepInBaseUnits() = getDisplayUnits().toBase(stepRange!!.getMidpoint())
```
alex-dev-neo: !! ?

OsmAnd/src/net/osmand/plus/configmap/tracks/TrackSortModesHelper.java
```diff
@@ -8,20 +8,24 @@
...
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class TrackSortModesHelper {
 
 	private static final String ROOT_FOLDER_ID = "";
 	private static final String SEPARATOR = ",,";
+	private static final String SCOPE_SEPARATOR = "::";
```
alex-dev-neo, line 28: may be ;; ?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/data/ComparableTracksGroup.kt
nazar-kutz: Renamed to getComparisonValue() to clearer reflect its purpose.

We use Double as a unified type for two main reasons:

1. It represents a metric, not an index. This value holds the underlying data used to group items (e.g., the exact timestamp or the start of a distance range), not a pre-defined list index.

2. Type unification.

2.1. For Ranges (Distance/Altitude): We need floating-point precision (e.g., when converting display units like miles to base units in meters).

2.2. For Timestamps: We need a type large enough to hold epoch milliseconds (which exceed Int.MAX_VALUE). Since we need to support both floating-point distances and large timestamps in a single interface method, Double covers both cases without precision loss for our sorting needs.

OsmAnd/src/net/osmand/plus/configmap/tracks/TrackSortModesHelper.java
nazar-kutz: We already use ;; as a main separator in the preference:

https://github.com/osmandapp/OsmAnd/blame/3fcb5b9f9c410d8a72b813725efa4de1f41b5d3a/OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java#L2136

nazar-kutz: public final ListStringPreference TRACKS_TABS_SORT_MODES = (ListStringPreference) new ListStringPreference(this, "tracks_tabs_sort_modes", null, ";;").makeGlobal().makeShared().cache();

PR #24199, opened 2025-12-28
SQLiteTileSource.java: Open RW only when needed

sonora:
Originally designed to reduce app startup time and increase hillshade sqlite performance during rendering. No big such effect measurable, but perhaps still worth having, may save some resources?


sonora:
This doe not fix slow startup on affected devices and has some risk, let me close it unmerged.

PR #24212, opened 2025-12-31
Fix GPX DB batch loading consistency and limit IO concurrency

sonora:


sonora:
With major portions now already cherry-picked, let me close this.

I think the remainder should as needed be split and re-submitted in these portions:
1. **Skip DB access for irrelevant files**, including readability cleanups query --> cursor
2. **Pagination and count consistency**: Count using the same logical dataset as the fetch query, and stable ordering improvement to avoid page drift
3. **Concurrency / performance tuning**: Semaphore introduction, bounded async batch fetching

Issue #24213, opened 2026-01-01
GPX: Light analysis mode for gpx db creation/maintenance


sonora:
As noted in https://github.com/osmandapp/OsmAnd/issues/22506#issuecomment-3702692638, the gpx database creation and maintenance seems to primarily suffer from over-analyzing all GPX files.

I believe we should for the purpose of creating and maintaining the gpx db create a LIGHT mode in GpxTrackAnalysis, like drafted below.

Full analytics would be deferred until the user actually opens the track.

Basic idea for **GpxTrackAnalysis.kt:**

<details>
  <summary>Click to expand</summary>

  ```
enum class AnalysisMode { LIGHT, FULL }

class GpxTrackAnalysis(
    private val gpx: GpxFile,
    private val mode: AnalysisMode = AnalysisMode.FULL
 ) {
     // existing fields...

    data class QuickStats(
        val lengthMeters: Double,
        val pointsCount: Int,
        val bboxN: Double, val bboxS: Double,
        val bboxE: Double, val bboxW: Double,
        val timeStartMillis: Long?, val timeEndMillis: Long?
    )

    private fun computeQuickStats(): QuickStats {
        var len = 0.0
        var prevLat = Double.NaN; var prevLon = Double.NaN
        var n = -90.0; var s = 90.0; var e = -180.0; var w = 180.0
        var firstTime: Long? = null; var lastTime: Long? = null
        var points = 0

        gpx.trackPoints().forEach { pt ->
            val lat = pt.lat; val lon = pt.lon
            if (!prevLat.isNaN() && !prevLon.isNaN()) {
                len += haversine(prevLat, prevLon, lat, lon)
            }
            prevLat = lat; prevLon = lon
            if (lat > n) n = lat; if (lat < s) s = lat
            if (lon > e) e = lon; if (lon < w) w = lon
            points++
            pt.timeMillis?.let { t ->
                if (firstTime == null) firstTime = t
                lastTime = t
            }
        }
        return QuickStats(len, points, n, s, e, w, firstTime, lastTime)
    }

    private fun toLightAnalysis(q: QuickStats): Analysis {
        val a = Analysis()
        a.totalDistance = q.lengthMeters
        a.pointsCount = q.pointsCount
        a.bboxNorth = q.bboxN; a.bboxSouth = q.bboxS
        a.bboxEast = q.bboxE;  a.bboxWest  = q.bboxW
[...]

PR #21358, opened 2024-11-16
[Kotlin] Fix GpxTrackAnalysis, port getEllipsoidDistance, fix squareDist31TileMetric, add GetDistanceTest

RZR-UA:

PR #24222, opened 2026-01-03
Optimize GPX point processing, and gate OBD analysis

sonora:
EDIT: I have limited this now to just
- avoiding per‑point allocations for PointAttributes,
- and gating void OBD analysis,
- method output is unchanged. 

Upon request I have removed optimizing distance calculation via Haversine and leave ellipsoidal in place. But please note
- that Haversine (spherical) is essentially as precise as ellipsoidal calculation unless for long arcs on the earth surface. And in GPX track analysis we sum total distance (even for long arcs) from many short segments between neighboring points, where even 2D geometry would not make a noticeable difference.
- (We could likely even simplify further by just using equirectangular calculation with no practical impact on accuracy.)

### DIFF
--- a/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
+++ b/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxTrackAnalysis.kt
@@ -528,9 +528,7 @@ class GpxTrackAnalysis {
 				if (j > 0) {
 					val prev = s[j - 1]
 					if (distance < 0f) {
-						//distance = KMapUtils.getEllipsoidDistance(prev.lat, prev.lon, point.lat, point.lon).toFloat()
-						// Use Haversine, inaccuracy only 0.1-0.5% for 2-4x faster
-						distance = KMapUtils.getDistance(prev.lat, prev.lon, point.lat, point.lon).toFloat()
+						distance = KMapUtils.getEllipsoidDistance(prev.lat, prev.lon, point.lat, point.lon).toFloat()
 					}
 					if (distance > maxDistanceBetweenPoints) {
 						maxDistanceBetweenPoints = distance


RZR-UA:
@sonora, I have some objections against getDistance() in GpxAnalysis. Some time ago, we tried to optimize this section using different distance algorithms. Yes, it is much faster, but there is a difference of several kilometers on real tracks with a length of ~1000km, compared to Garmin (Devices, Web) and a couple of other GIS applications. We should show the track distance as accurately as possible.

sonora:
Thanks for commenting! I have removed the dist calculation optimization from this PR.

But It would be worth investigating, there may be some other flaw involved then.

Because the worst case difference between elliptical and spherical computation for a 1000 km stretch on earth is <4 km. And that only if you compute the entire 1000 km stretch as one single great‑circle segment with one start point, one end point and no intermediate points.

In a GPX track analysis we are clearly in another situation, with many intermediate points.

Elliptical calculation for the usual  track point spacings there would propel you from the cm accuracy regime (spherical) to the mm accuracy regime. Both are actually overkill vs. the involved consumer type GPS jitter. The regime of elliptical computation is normally professional land surveying with special equipment.

Here's a good sanity test to check if I am right: Let's go through our track collections in OsmAnd and see if we find any difference in reported track stats for a build with and without this PR. If we find such a track, it would be worth looking into (and it would prove me wrong, unless we have some other formula or rounding problem 🤔).

sonora:
Perhaps a quick recap on how this all works:

- **Equirectangular** calculates distance as a straight line (line of sight) - no earth curvature, etc. For track point spacings like 10 m for hiking or 200 m for driving this should clearly be enough.

- **Spherical** computation (Haversine) treats the earth like a sphere. Good enough where some curvature of the earth surface comes into play (basically everything beyond horizontal line of sight).

- **Ellipsoidal** (Vincenty) takes additionally into account the deviation of the earth from a perfect sphere (polar radius is some 20 km less than equatorial radius). Only worth it in high precision surveying or for very long arcs without intermediate points. 

**Differences**, let's take 2 points about 100 m apart:
- Equirectangular vs. spherical: Typically 1 mm, worst case (high latitude, pure east-west direction) a few millimeters at most. For segment lengths like 100 m, the curvature is so small that the planar approximation is almost exact.
- Spherical vs  ellipsoidal (e.g. Haversine vs Vincenty/WGS‑84): Typical difference (100 m segment, mid‑latitudes) well under 1 cm, usually just a few millimeters. Worst case (100 m at latitudes where curvature difference is maximal): on the order of 1–2 cm at most. Again, this is orders of magnitude smaller than GPS noise, multipath, sampling, or device smoothing.

alex-dev-neo:
@sonora Thank you for your improvements! However I agree with @RZR-UA - we cannot just revert calculations to Haversine, since it was a bug and it was [fixed](https://github.com/osmandapp/OsmAnd/pull/21358) with ellipsoidal calculation. I don't rule out that the problem may not be only in Haversine, but there was definitely an error in calculations. I suggest investigate with real data (long tracks). Also it would be good to have measurements to see if Haversine is really 2-4x faster.

sonora:
I have run some tests and agree that this PR does not address the main culprit. While PRs https://github.com/osmandapp/OsmAnd/pull/24212 and https://github.com/osmandapp/OsmAnd/pull/24222 combined provide some 15% improvement (and especially https://github.com/osmandapp/OsmAnd/pull/24212 should probably be merged for stability), the true performance killer is illustrated here: https://github.com/osmandapp/OsmAnd/issues/24213#issuecomment-3707903087

**I suggest we prioritize doing something about that urgently?** The distance computation question is truly secondary in light of these findings: The true killer seems that part of GPX analysis correlating tracks to maps, not the distance calculation churn. Sorry for only discovering after raising this PR here... :wink:

Issue #24229, opened 2026-01-05
Overly slow address search on the default Search screen


sonora:
Address search is mainstream feature OsmAnd is used for.

Today, if a user on the map screen taps the Search button, and then simply enters an address snippet (e.g. a street name), this gets searched for on the "Explore" tab, and can be unbearably slow. On my setup (this probably depends on how many obf maps you have) when I simply type a street name in my current neighborhood, it mostly takes **>2 minutes** until the correct hit appears at the top of the search hits. (Prior to that, all sorts of very far away non-address hits start populating the hit list.)

This can of course be worked around by searching for the same search string in the "Address" tab of the search screen, where the same operation succeeds in typically ~10 sec.

But many users are not aware they have to take that extra step, and get frustrated with OsmAnd. Also, once you started typing on the default (Explore) tab, you cannot simply change tabs any more or somehow mark that you mean to search for an address. We should clearly improve here somehow.

Zirochkabila:
> OsmAnd~ 5.3.0#6828m, Released: 2026-01-05

@sonora I tried it on a device with more than 15 GB of maps downloaded - as you can see in the video, the search works quite fast.

Please provide more detailed information:
1. How many maps you have downloaded
2. The coordinates from which the search is performed (we will try to recreate an artificial location)
3. The name of the street you are looking for

<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/63990e91-058e-4bf6-8e10-737315edc136"></video>
</details>

sonora:
@Zirochkabila Yes, my results vary, too, with different locations. I have about 30 GB of OBF standard maps, ~80 files. Data folder is on external SD-storage (perhaps that has an impact).

Let's try it on this example:

Center your map around village _Poikam_ (https://www.openstreetmap.org/node/61836301), and search for a local street named _Lindenweg_, by typing just that and waiting. When I do this on the "Address" tab, the correct result is shown within ~2 seconds. But when I do the same on the default (Explore) tab, it takes just ~1 minute...

scaidermern:
> Let's try it on this example:
> 
> Center your map around village _Poikam_ (https://www.openstreetmap.org/node/61836301), and search for a local street named _Lindenweg_, by typing just that and waiting. When I do this on the "Address" tab, the correct result is shown within ~2 seconds. But when I do the same on the default (Explore) tab, it takes just ~1 minute...

[...]

PR #24253, opened 2026-01-07
Priroitize address search to speed up overall search UX (Issue #24229)

sonora:


vshcherb:
Closing as it looks like a wrong approach

PR #24256, opened 2026-01-07
Search timing: De-prioritize WPT and Track search API, big search speed-up on setups with many tracks (#24229)

sonora:


sonora:
Just timing with no overall impact on search results. But address search hits now appear pretty instantly, whereas before in some configurations (many maps, many tracks) they took >1 minute to finally show up.

Issue #22506, opened 2025-05-04
Track db sometimes updates, impacting performance


sonora:
See screenshot below, not sure what triggers it. Perhaps using 'Show all on map' on a folder containing a few dozen tracks (although displaying them itself works fine).

- The spinner comes on on the Tracks screen, and stays on for about 10 min if staying on the screen, only then finally stopping.
- Leaving the screen before that, or even re-starting the app will simply recreate the situation next time that screen is entered.
- During the process and until it's finished, often one or several folder stats figures remain at 0m (often 'downhill' like in the screenshot below). Perhaps this indicates that the stats are re-calculated during this (though I have not recently changed, added or removed  any track).
- To resolve the entire situation, I need to stay on the Tracks screen with the app in the foreground for about 10 min.
- Not entirely sure, but it seems app performance is overall negatively affected until this phenomenon finishes.

![Image](https://github.com/user-attachments/assets/4f51f7a8-9911-4979-bc51-8aef4f933f1f)

Rowin63:
Had some similar behavior right now.
I added a track folder including three tracks (copied it into Osmands tracks folder by a file manager).
After this I could not rename this folder in "my places". So I started a refresh by swiping down. After five or more minutes the app crashed (sent a logfile at 12:14 to the support address).

yuriiurshuliak:
**When using the "Show all on map" feature for a folder with multiple tracks, the **spinner stays active for about 10 minutes**. This seems to impact performance and causes issues like folder stats showing zero values, even though no tracks have been added or changed.**

I tried to reproduce the issue several times by following the steps you mentioned. Here’s what I did:

1. Placed a folder with several GPX tracks (a few dozen) in OsmAnd's tracks folder.
2. Opened OsmAnd and went to **My Places → Tracks**.
3. Used **"Show all on map"** for the folder.
4. Returned to the **Tracks screen** and tried pulling down to refresh the page.

**Results:**
I did not experience the spinner staying active endlessly. The statistics at the bottom of the page remained stable, and I didn’t observe any performance degradation. Everything worked as expected.



OsmAnd~ 5.1.0#5162m, released: 2025-05-09

https://github.com/user-attachments/assets/f0a001cf-9c3a-4b82-93af-5504ecb7f8fc

sonora:
[...]

PR #24257, opened 2026-01-07
Skip GPX nearest city tagging: 50x faster GPX database setup, resolves app stall (#22506, #24213)

sonora:


sonora:
Superseded by #24510.

The question if tagging GPX files with "nearest city" has a tangible user benefit (https://github.com/osmandapp/OsmAnd/issues/24213#issuecomment-3758655772) needs separate doscussion.

PR #24259, opened 2026-01-07
Optimize PT algorithm with optimalLimitOfResults

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22515

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
@@ -208,8 +208,8 @@ public class TransportRoutePlanner {
 						(finish.distFromStart < maxTravelTimeCmpToWalk || results.size() == 0)) {
 					results.add(finish);
 					// Stop when results reached range [1000 min, 2500 (for default limit * changes), 5000 max]
-					int dynamicResultsLimit = 25 * ctx.cfg.ptLimitResultsByNumber * ctx.cfg.maxNumberOfChanges;
-					if (results.size() > Math.min(Math.max(1000, dynamicResultsLimit), 5000)) {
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -165,6 +166,11 @@ public List<TransportRouteResult> buildRoute(TransportRoutingContext ctx, LatLon
 								ctx.cfg.getChangeTime(segment.road.getType(), sgm.road.getType());
 						nextSegment.distFromStart = segment.distFromStart + travelTime + walkTime;
 						nextSegment.nonce = nonce++;
+						double dynamicDecrease = Math.min(1.2, (1 - 0.2 * (segmentDepth - 1))); // -20%
```
vshcherb: Try Simpler limit first

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -165,6 +166,11 @@ public List<TransportRouteResult> buildRoute(TransportRoutingContext ctx, LatLon
...
+						double dynamicQueueLimit = ctx.cfg.increaseForAlternativesRoutes * dynamicDecrease;
```
vshcherb: Much unclear variables, limit should be directly related to calculation speed and calculation speed is related to number of visited segments, requires investigation first how much time 1k visited segments takes

RZR-UA: 82ffe746fb

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
RZR-UA: 82ffe746fb

Issue #24206, opened 2025-12-29
Map inertia is overly throttled in v2 renderer


sonora:
"Flinging" is the gesture where a user drags the map (e.g. with their thumb) and then releases it so it keeps moving with inertia.

In our v1 engine we support this rather well, allowing to shift the viewport by a good measure with a single such gesture. But in our v2 engine the map motion comes to a grinding halt much sooner, making the gesture almost worthless.

Can we adapt the v1 behavior also for v2?

yuriiurshuliak:
In Renderer v2, the map “fling” inertia is noticeably weaker than in Renderer v1 — after a fast drag-and-release, the map slows down and stops much sooner, so the viewport shifts much less.

<details><summary>Steps to reproduce</summary>

We can reproduce a similar behavior.

**Steps to reproduce**

1. Open OsmAnd → go to the **map screen**.
2. Enable **Renderer v2**.
3. Perform a **fling**: drag the map quickly (short, fast swipe) and **release**.
4. **Actual (v2):** the map continues only briefly and stops quickly, moving the viewport by a small amount.
5. Switch to **Renderer v1**.
6. Repeat the same fling gesture with similar speed/distance.
7. **Comparison:** in v1, the map typically travels farther and inertia feels longer.

</details>

<details><summary>Expected vs Actual</summary>

**Expected vs Actual**

* **Expected:** v2 matches v1-like behavior — a single fling moves the map a noticeable distance with smooth inertia.
* **Actual:** in v2, inertia is throttled and the map comes to a quick stop.

</details>

<details><summary>Videos</summary>


OsmAnd~ 5.3.0#6805m, released: 2026-01-02

<table style="width:100%">
  </tr>
  <tr>
    <th>V2</th>
    <th>V1</th>
  </tr>
  <tr>
    <td> <video source src="https://github.com/user-attachments/assets/07cc3d08-1c85-4181-a5be-eb11ab42cc53"  width="300" height="auto"/>   </td>
    <td><video source src="https://github.com/user-attachments/assets/dba7e0e7-b225-4803-82e6-462c3800a2af"  width="300" height="auto"/>   </td>
  </tr>
</table>

</details>

vshcherb:
@yuriiurshuliak 

- [ ] needs to be checked for regression
- [ ] compare with iOS

yuriiurshuliak:
Regression check: **not a recent regression**.
iOS check: not reproduced

<details><summary>Android</summary>

[...]

PR #24282, opened 2026-01-12
Incresed map inertia with v2 engine

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd/issues/24206

PR #24299, opened 2026-01-14
Sort matched strings in 30km zone. Remove likelyAddressSearch

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/23713#issuecomment-3610715878

Tests https://github.com/osmandapp/OsmAnd-resources/pull/1331

Знімок екрана 2026-01-14 о 15 58 46


OsmAnd-java/src/main/java/net/osmand/search/core/SearchResult.java
```diff
@@ -264,6 +265,10 @@ private boolean allWordsMatched(String name, SearchResult exactResult, CheckWord
 		if (searchPhraseNames.size() == localResultNames.size()) {
 			cnt.allWordsEqual = true;
 		}
```
vshcherb, line 267: not correct

sonora:
Optimizing for performance, the following could be done here:
1. We could avoid division in the loop via pre-calculated multipliers:
```
static final double INV_NEAREST_LIMIT = 1.0 / NEAREST_METERS_LIMIT;
...
(int)((NEAREST_METERS_LIMIT - dist) * INV_NEAREST_LIMIT * 15)
```

2. Calculating distance is the heaviest math involved here (particularly since we use ellipsoidal?). We could likely short-circuit expensive math for poor matches, instead of computing `MapUtils.getDistance(...)` and only afterwards deciding whether the distance matters, we could guard like 

```
if (completeMatchRes.allWordsEqual || objectType != ObjectType.POI) {
    // THEN compute distance
}
```
?

vshcherb:
@sonora main performance issue still reading files (not sorting results)

sonora:
> @sonora main performance issue still reading files (not sorting results)

@vshcherb Absolutely true! I am just advising obvious (and very cheap) optimizations where they happen to jump at me here. Going forward, it seems easier to implement at commit time, we may never go back to it later. And small accumulative effects like this may make the app snappier over time. :wink:

PR #24302, opened 2026-01-14
Fix incorrect Intermediate point caption size

nazar-kutz:
Use Configure Map -> "Text size" preference to choose correct intermediate point caption size

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/PointNavigationLayer.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/PointNavigationLayer.java
@@ -42,6 +42,8 @@ import java.util.List;
 public class PointNavigationLayer extends OsmandMapLayer implements
 		IContextMenuProvider, IMoveObjectProvider {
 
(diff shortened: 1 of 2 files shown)


OsmAnd/src/net/osmand/plus/views/layers/PointNavigationLayer.java
```diff
@@ -210,15 +210,18 @@ private void updateBitmaps(boolean forceUpdate) {
...
 
 	private void updateTextSize() {
-		mTextPaint.setTextSize(18f * Resources.getSystem().getDisplayMetrics().scaledDensity
-				* getApplication().getOsmandMap().getCarDensityScaleCoef());
+		float density = view.getDensity();
+		float textSize = 18f * textScale * density;
+
+		if (textSize != mTextPaint.getTextSize()) {
+			mTextPaint.setTextSize(textSize);
+			captionStyle = null;
```
Chumva: Will it update OpenGL correctly?

Issue #21618, opened 2024-12-25
Duplicate shield 1 road in the street name iOS / Android


Zirochkabila:
### Description

During navigating on “Salzburger Straße”, the data in the Lanes widget on Android differs from iOS
[OSM](https://www.openstreetmap.org/way/217461026#map=18/48.152867/13.986030)
[OsmAnd](https://osmand.net/map/navigate/?start=48.152902,13.985472&end=48.153428,13.990391&profile=car&pin=50.437668,30.516562#18/48.15297/13.98620)

### Steps to reproduce

1. Navigation > From `48.15290, 13.98547` > To `48.15343, 13.99039` > Look at the Line widget data

### Actual result

<table style="width:100%">
  <tr>
    <th>iOS</th>
    <th>Android</th>
  </tr>
  <tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/ed12b464-1832-4069-b549-2700b1b3b5eb"  width="600" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/ed2aacb0-f78c-48d6-a6c4-0aee48b1e003"  width="300" height="auto"/> </th>
  </tr>
  </tr>
</table> 

### Expected result

@vshcherb Perhaps it is worth displaying “B1” - as it is displayed when reaching the destination and as indicated in “ref B1”

<table style="width:100%">
  <tr>
    <th>iOS</th>
    <th>Android</th>
  </tr>
  <tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/f9ba70da-7b36-46bb-b640-1f574e29241f"  width="600" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/fc49772c-a754-4c00-8796-23f7363116b5"  width="300" height="auto"/> </th>
  </tr>
  </tr>
</table> 

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 4.9.2.4929 / OsmAnd~ 5.0.0#3994m, Released: 2024-12-25
Android/iOS version:
Device model:
Crash-Logs: ?
```

yuriiurshuliak:
The bug has been reproduced. To reproduce the issue:
1. Build the navigation from **48.15290, 13.98547** to **48.15343, 13.99039**.
2. Enable **navigation simulation**.
[...]

PR #24312, opened 2026-01-16
don't show same route shields in widgets

nnngrach:
[issue](https://github.com/osmandapp/OsmAnd/issues/21618)

Test points:
A 48.15290, 13.98547 
B 48.15343, 13.99039
C 48.19288 13.92583
D 48.18890 13.92183

A -> B (before/after):  no doubling

Screenshot 2026-01-16 at 15 30 50   Screenshot 2026-01-15 at 13 52 32

C -> D (before/after):  no changes

Screenshot 2026-01-17 at 01 25 45    Screenshot 2026-01-17 at 01 25 54

---

video before:

https://github.com/user-attachments/assets/1ca02fa5-a76e-4ef0-ad74-c714725ab69f

video after:

https://github.com/user-attachments/assets/8c795076-50b9-4ba6-bbd6-314cc1efe3cd

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/StreetNameWidget.java
+++ b/OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/StreetNameWidget.java
@@ -347,37 +347,14 @@ public class StreetNameWidget extends MapWidget {
 	private static boolean isSameShieldAdded(RoadShield currentShield, List<RoadShield> previousShields) {
 		if (!previousShields.isEmpty()) {
 			String currentLabel = currentShield.getValue();
-			Map<String, String> currentTags = getShieldTagsMap(currentShield.getAdditional().toString());
-			// "route_road_1_ref"  ->  "route_road_1_shield_color"
-			String currentColorKey = currentShield.getTag().replace("_ref", "_shield_color");
-			String currentColor = currentTags.get(currentColorKey);
-
 			for (RoadShield previousShield : previousShields) {
 				String previousLabel = previousShield.getValue();
-				Map<String, String> previousTags = getShieldTagsMap(previousShield.getAdditional().toString());
-				String previousColorKey = previousShield.getTag().replace("_ref", "_shield_color");
-				String previousColor = previousTags.get(previousColorKey);
-
-				if (Objects.equals(currentLabel, previousLabel) && Objects.equals(currentColor, previousColor)) {
+				if (Objects.equals(currentLabel, previousLabel)) {
 					return true;
 				}
 			}
 		}
 		return false;
-	}
-
-	private static Map<String, String> getShieldTagsMap(String tagsString) {
-		// convert String "tag1=val1;tag2=val2;"  to  Map {tag1: val1, tag2: val2}
-		Map<String, String> map = new LinkedHashMap<>();
-		for (String tagValue : tagsString.split(";")) {
-			int index = tagValue.indexOf('=');
-			if (index > 0) {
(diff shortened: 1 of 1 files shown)


nnngrach:
Debug output:

C -> C (different names in data)

Screenshot 2026-01-17 at 01 32 42

---

A -> B  (same names in data)

Screenshot 2026-01-17 at 01 34 23

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/StreetNameWidget.java
```diff
@@ -330,9 +340,46 @@ public static boolean setShieldImage(@NonNull RoadShield shield,
...
+		for (String tagValue : tagsString.split(";")) {
+			int index = tagValue.indexOf('=');
+			if (index > 0) {
+				String key = tagValue.substring(0, index);
+				String value = tagValue.substring(index + 1);
+				map.put(key, value);
+			}
+		}
+		return map;
+	}
```
RZR-UA: Replace this method with a simple additionalMap in RoadShield.java and fill this additionalMap together with `additional`. It is not a good idea to convert a Map to a String and then convert back again every time.

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/StreetNameWidget.java
```diff
@@ -330,9 +340,46 @@ public static boolean setShieldImage(@NonNull RoadShield shield,
...
+		addedShields.add(shield);
 		return true;
 	}
 
+	private static boolean isSameShieldAdded(RoadShield currentShield, List<RoadShield> previousShields) {
+		if (!previousShields.isEmpty()) {
+			String currentLabel = currentShield.getValue();
+			Map<String, String> currentTags = getShieldTagsMap(currentShield.getAdditional().toString());
+			// "route_road_1_ref"  ->  "route_road_1_shield_color"
+			String currentColorKey = currentShield.getTag().replace("_ref", "_shield_color");
```
RZR-UA: This `replace()` seems too unclear. It is hard to understand what you want to do - always use shield_color instead of ref color, or something else?

Could you show at least 1 object that requires a color match and 1 object that requires an additional `_ref -> _shield_color` match?

The best code is no code (it is great if the Issue is solved without the color check).

If you still need to compare colors, make 2 variables currentRefColor and currentShieldColor and compare them with `||`.

OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/StreetNameWidget.java
```diff
@@ -330,9 +340,23 @@ public static boolean setShieldImage(@NonNull RoadShield shield,
...
+		addedShields.add(shield);
 		return true;
 	}
 
+	private static boolean isSameShieldAdded(RoadShield currentShield, List<RoadShield> previousShields) {
+		if (!previousShields.isEmpty()) {
+			String currentLabel = currentShield.getValue();
+			for (RoadShield previousShield : previousShields) {
+				String previousLabel = previousShield.getValue();
+				if (Objects.equals(currentLabel, previousLabel)) {
```
Chumva, line 352: Do we need to check only value, or also other shield parameters? (tag/additional)

RZR-UA: It seems, no need to check other parameters to resolve the bug reported in the Issue.

Issue #24274, opened 2026-01-11
Corrupted pinch zoom gesture shifts map center until next app start


sonora:
Observed in v2 engine with 'Pinch zoom magnification' enabled:

There are cases where a pinch zoom gesture becomes corrupted and this leads to an unwanted shift of the map center which persists until the next app start.

One example how to reproduce is this: Start a pinch zoom gesture with one finger in the map area but the second accidentally slightly off the left screen edge, so that the drawer gets partly dragged into the screen. Pinch in or pinch out for at least one zoom level and apply some rotation change also at this time while a portion of the drawer is partly visible,  then abandon the gesture. 

Now pan the map a little and tap the 'My Position' button to re-center it. Your position will from now on center at (I believe) the anchor point of this former pinch gesture. This persists until next app start.

Similar situations can be observed if a pinch gesture ends by one finger briefly activating a drawer appearing, or perhaps by moving off the map area onto some other screen UI element or screen area.

I believe this results from a corrupted pinch gesture where at some point during a pinch gesture one finger is off the actual map area and the UI perceives a one finger pan mixed into the pinch gesture.

A fix will probably have to guard the pinch gesture for starting or ending abnormally, or by generically resetting the map center in rhe end?

sonora:
It's a hassle with Android multi touch handling. By my testing, this guards it well: https://github.com/osmandapp/OsmAnd/commit/2cf1b9fa51ffda6308fcc2d316c603b47a2e80af

I will not make it a PR yet because I think the existing PR for the anchor misalignment correction (end-of-gesture jumps) is more significant and should have priority.

Zirochkabila:
@yuriiurshuliak 

- [ ] 1. Regression
- [ ] 2. Compare iOS

dmpr0:
<table style="width:100%">
  <tr>
    <td><video src="https://github.com/user-attachments/assets/1316f31e-505f-497e-ab59-53b527f18d37"  width="300" height="auto"/> </td>
  </tr>
</table>

Andrewschenko:
[Fixed](https://github.com/osmandapp/OsmAnd/pull/24294)

There is one more thing: you cannot drag the map if you touch it with two fingers (with synchronous finger movements).

sonora:
@Andrewschenko

[...]

PR #24319, opened 2026-01-17
Remove some surplus end-of-gesture map redraw and reduce flicker

sonora:
(I am aware that via MAP_REFRESH_MESSAGE we throttle map refresh trigger, but probably we can even save the overhead here?)


sonora:
Let me close this unmerged.

We need a deeper investigation of what's described in https://github.com/osmandapp/OsmAnd/issues/24274#issuecomment-3753614603 point (5) **"End of gesture flicker / incomplete update"**

PR #24322, opened 2026-01-18
Disable Drawer action during multi-touch

sonora:


OsmAnd/src/net/osmand/plus/activities/MapActivity.java
```diff
@@ -1454,6 +1454,28 @@ public boolean dispatchTouchEvent(MotionEvent event) {
 				}
 			}
 		}
+
+		DrawerLayout drawerLayout = findViewById(R.id.drawer_layout);
+		if (drawerLayout != null && !drawerDisabled) {
```
alex-dev-neo: `drawerLayout` is already accessible and non-null.
It is safer to use `drawerDisabled` via `isDrawerDisabled()`

OsmAnd/src/net/osmand/plus/activities/MapActivity.java
```diff
@@ -1454,6 +1454,28 @@ public boolean dispatchTouchEvent(MotionEvent event) {
...
+			switch (event.getActionMasked()) {
+				case MotionEvent.ACTION_POINTER_DOWN:
+				case MotionEvent.ACTION_MOVE:
+					// 2+ fingers on screen: disable drawer swipe
+					if (event.getPointerCount() >= 2) {
+						drawerLayout.setDrawerLockMode(DrawerLayout.LOCK_MODE_LOCKED_CLOSED, GravityCompat.START);
```
Chumva: Better use setDrawerLockMode(DrawerLayout.LOCK_MODE_LOCKED_CLOSED/LOCK_MODE_UNLOCKED);, same as disableDrawer()/enableDrawer() to avoid any issues with gravity

OsmAnd/src/net/osmand/plus/activities/MapActivity.java
sonora: Good catch, thanks!! Updated.

OsmAnd/src/net/osmand/plus/activities/MapActivity.java
sonora: Right, thanks! We have no END drawer in the layout, so removing `, GravityCompat.START` for code consistency will cause no harm.

OsmAnd/src/net/osmand/plus/activities/MapActivity.java
```diff
@@ -1454,6 +1454,27 @@ public boolean dispatchTouchEvent(MotionEvent event) {
...
 			}
 		}
+
+		if (!isDrawerDisabled()) {
+			switch (event.getActionMasked()) {
+				case MotionEvent.ACTION_POINTER_DOWN:
+				case MotionEvent.ACTION_MOVE:
+					// 2+ fingers on screen: disable drawer swipe
+					if (event.getPointerCount() >= 2) {
+						drawerLayout.setDrawerLockMode(DrawerLayout.LOCK_MODE_LOCKED_CLOSED);
```
alex-dev-neo, line 1463: Also it would be good to not setDrawerLockMode on every move event.

sonora: Agreed - I've just tested without the ACTION_MOVE case and it seems good enough to achieve the purpose in common situations.

PR #24388, opened 2026-01-29
Apply approximated height data to OSM routes using GpxElevationTransfer (geometry-loseless)

RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/3168
https://github.com/osmandapp/OsmAnd/issues/22226 (ios-port-list)


OsmAnd/src/net/osmand/plus/wikivoyage/data/TravelObfGpxFileReader.java
```diff
@@ -127,6 +156,24 @@ protected void onPostExecute(@Nullable GpxFile gpxFile) {
         hideProgress();
     }
 
+    private synchronized void acquireGpxFileHeightData(@NonNull GpxFile targetGpxFile) {
```
Chumva, line 159: synchronized for asynctask method?

RZR-UA: Yes, to be sure not to read the same OBF files in parallel threads.

PR #24393, opened 2026-01-30
Fixed nearest cities disappear after return from city search with back

Corwin-Kh:


OsmAnd/src/net/osmand/plus/search/dialogs/QuickSearchDialogFragment.java
```diff
@@ -650,6 +650,7 @@ private void onBackButtonPressed() {
 			hideKeyboard();
 			searchEditText.setText("");
 			updateTabBarVisibility(true);
+			reloadCities();
```
Chumva: It will fix, but with this, we have an inconsistent flag citiesLoaded. Also, it will lead to running additional cities search on every click on back button, even if we are searching categories or something else. 

Probably we could replace citiesLoaded with cached list of [cities](https://github.com/osmandapp/Osmand/blob/8a89ebcf872451fa6bd28cbd15a2c87b973b70e4/OsmAnd/src/net/osmand/plus/search/dialogs/QuickSearchDialogFragment.java#L1397-L1403)

PR #24397, opened 2026-02-01
Fix speedometer animation

Corwin-Kh:


OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/SpeedometerWidget.java
```diff
@@ -456,7 +462,7 @@ private void updateSpeedLimitVisibility(boolean show) {
 						speedLimitContainer.getVisibility() == (show ? View.VISIBLE : View.GONE)) {
 					return;
 				}
-				if (show) {
+				if (show || speedLimitContainer.getAlpha() > 0) {
```
Chumva, line 465: Looks strange, will it lead to bugs if show is false?

Issue #24243, opened 2026-01-06
Map of Prague is not included in the list of multi-download regions of the Czechia


Zirochkabila:
### Description

The map of Prague is not included in the list of multi-download regions of the Czech Republic.

### Steps to reproduce

Menu > Maps & Resources > Europe > Czechia

### Actual result

There are 7 regions in the list (no map of Prague available), and 8 maps available for download.

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/a90ec71f-b047-4ebf-a71e-1a665bcd643c"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/0c9cdf0c-ec3b-4d0e-a4b7-1b1eac8ed1fa"  width="300" height="auto"/> </th>
  </tr>
</table>  
</details> 

### Expected result

- [ ] Map of Prague should be included in the list of multi-download regions 

### Your Environment
OsmAnd Version: OsmAnd~ 5.3.0#6845m, Released: 2026-01-06
Android/iOS version:  12

RZR-UA:
Dev notes

- [x] Map of Prague should be included in the list of multi-download regions
- [x] Map of Prague should be included for (and between) Start-End points in Missing Maps Calculator
- [x] After the fix of Prague, verify multi-download for Berlin, Kyiv (both are included in the parent regions)
- [x] check/fix download of Prague by clicking on the map when europe_czech-republic_stredni-cechy is downloaded
- [x] add Unit-tests for a couple of regions with excluded-polygons (Prague, Berlin, etc)
- [x] to think: refactor Java code to keep only 1 set of polygons
- [x] to test: FirstUsageWizardFragment with Prague location
- [x] refactor addPolygonToRegion()
- [x] refactor containsPoint()
- [x] fix map click in Austria-Tirol (with downloaded Tirol map, click on the smaller region, suggest to download inexisting map of the whole Austria)
- [x] fix broken MissingMapCalculator in the additional polygon of Tyrol (android fixed, iOS ok)
- [x] port changes from Android to iOS, check Web (Regions)

vshcherb:
Algorithm need to be adjusted, if center point of region Prague is not inside poly of the other polygon that means they don't overlap so it should be suggested

RZR-UA:
The list of fixed regions:

**EU: Central Bohemia ≠ Prague — capital city is a separate unit, not part of the surrounding region.**

[...]

PR #24402, opened 2026-02-02
Fix Prague region with exclusion polygons, refactor WorldRegion/OsmandRegions

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/24243

### DIFF
--- a/OsmAnd-java/src/test/java/net/osmand/map/OsmandRegionsTest.java
+++ b/OsmAnd-java/src/test/java/net/osmand/map/OsmandRegionsTest.java
@@ -57,6 +57,9 @@ public class OsmandRegionsTest {
         testIncludedExcluded(-35.308056, 149.124444,
                 "australia-oceania_australian-capital-territory_australia-oceania",
                 "australia-oceania_new-south-wales_australia-oceania");
+
+        // Additional subregion query must return the main region name (Lienz, Tirol, Austria)
+        testIncludedExcluded(46.82987, 12.76812, "austria_tyrol_europe", "");
     }
 
     private void testIncludedExcluded(double lat, double lon, String included, String excluded) {


OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
```diff
@@ -249,9 +250,9 @@ public boolean containsRegion(WorldRegion another) {
 		}
 
 		// Finally check inner point
-		boolean isInnerPoint = another.containsPoint(another.regionCenter);
+		boolean isInnerPoint = another.containsPointInAllPolygons(another.regionCenter);
```
vshcherb: why do we need to rename at all?

RZR-UA: The name containsPoints does not seem clear (intersects with a lot of other kinds of containPoints in other classes). In addition, we need to use containsPoints in 2 different ways. Please wait for the final request.

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
```diff
@@ -1008,14 +1013,46 @@ private void addPolygonToRegionIfValid(BinaryMapDataObject mapObject, WorldRegio
...
+			worldRegion.excludingPolygons.add(polygon); // marked with ! in *.poly files
 		}
 	}
+
+	public List<BinaryMapDataObject> filterQueryResultByPoint(List<BinaryMapDataObject> objects, int x, int y) {
+		List<BinaryMapDataObject> filtered = new ArrayList<>();
+		Map<String, Integer> intersectionCounter = new HashMap<>();
+		for (BinaryMapDataObject o : objects) {
+			if (OsmandRegions.contain(o, x, y)) {
+				updateIntersectionCounter(o, intersectionCounter, true);
```
vshcherb: oneliner
```
intersectionCounter.compute("key", (tag, count) -> count == null ? 1 : count + 1);
```

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
```diff
@@ -1008,14 +1013,46 @@ private void addPolygonToRegionIfValid(BinaryMapDataObject mapObject, WorldRegio
...
+			}
+		}
+		for (BinaryMapDataObject o : objects) {
```
vshcherb, line 1017: ?
if(!o.getName().isEmpty() && intersectionCounter.get(getDownloadName(o)) % 2 == 1)

OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
vshcherb: There should be only 1 method WorldRegions.containsPoint - means exactly contains point or not. All other methods are unclear

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
```diff
@@ -1008,14 +1013,46 @@ private void addPolygonToRegionIfValid(BinaryMapDataObject mapObject, WorldRegio
 
 		boolean outside = true;
 		for (LatLon point : polygon) {
-			if (worldRegion.containsPoint(point)) {
```
vshcherb, line 1011: Rename addPolygonToRegionIfValid, all polygons are valid!

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
```diff
@@ -1008,14 +1013,46 @@ private void addPolygonToRegionIfValid(BinaryMapDataObject mapObject, WorldRegio
...
+			if (worldRegion.containsPointInMainPolygon(point)) {
 				outside = false;
```
vshcherb: 1. This is incorrect check for excluding polygons in general, so in that case excludingPolygons should be renamed to insideMainPolygons. 
2. No need to check all point it's enough to check one point or a center point

OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
```diff
@@ -268,10 +271,28 @@ private boolean containsPolygon(List<LatLon> another) {
 				Algorithms.isFirstPolygonInsideSecond(another, polygon);
 	}
 
-	public boolean containsPoint(LatLon latLon) {
+	protected boolean containsPointInMainPolygon(LatLon latLon) {
 		return polygon != null && Algorithms.isPointInsidePolygon(latLon, polygon);
 	}
 
+	public boolean containsPointInAllPolygons(LatLon latLon) {
+		for (List<LatLon> excluding : excludingPolygons) {
```
vshcherb: This loop should be inside for 
```
		if (polygon != null && Algorithms.isPointInsidePolygon(latLon, polygon)) {
                for (List<LatLon> excluding : excludingPolygons) {
                     if (Algorithms.isPointInsidePolygon(latLon, excluding)) {
                             return false;
                     }
                 }
                  return true;
         }
```

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
RZR-UA: updateIntersectionCounter removed

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
RZR-UA: updateIntersectionCounter removed

OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
RZR-UA: right, ok

OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
RZR-UA: Renamed back to WorldRegions.containsPoint()

OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
RZR-UA: That is NOT correct, actually! The point could be **outside** the main polygon but **within** additionalPolygons. In this case, we have to check all excluding polygons regardless of their inclusion in the main polygon. Got back this loop to the very beginning.

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
RZR-UA: addPolygonToRegion()

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
RZR-UA: Refactored and optimized. Saved ~250-1000 ms at startup time. Results of addPolygonToRegion is the same as previosly. Commit 0a1f9cb30b

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
```diff
@@ -1006,16 +999,36 @@ private void addPolygonToRegionIfValid(BinaryMapDataObject mapObject, WorldRegio
 			polygon.add(new LatLon(lat, lon));
 		}
 
-		boolean outside = true;
-		for (LatLon point : polygon) {
-			if (worldRegion.containsPoint(point)) {
-				outside = false;
-				break;
-			}
+		if (worldRegion.containsPoint(polygon.get(polygon.size() / 2))) {
```
vshcherb: Can't use polygons until they are read ! Chicken and egg problem

OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
```diff
@@ -269,7 +270,21 @@ private boolean containsPolygon(List<LatLon> another) {
 	}
 
 	public boolean containsPoint(LatLon latLon) {
-		return polygon != null && Algorithms.isPointInsidePolygon(latLon, polygon);
+		for (List<LatLon> excluding : exclusionPolygons) {
```
vshcherb: Exclusion should be checked after points are checked for main polygons! That's not effective codde

OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
RZR-UA: Yes. You are right. Better approach required.

Complicated regions:

```
asia/ocean-asia/philippines/central-visayas.poly-1
asia/china/qinghai/haixi.poly-海西蒙古族藏族自治州
asia/east-asia/india/gujarat.poly-12
asia/east-asia/india/gujarat.poly-13
asia/east-asia/india/gujarat.poly-14
russia/far-eastern-federal-district/kamchatka.poly-3
russia/siberian-federal-district/krasnoyarsk.poly-1
russia/siberian-federal-district/krasnoyarsk.poly-2
```

RZR-UA: a142e9f11f

OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
RZR-UA: a142e9f11f

OsmAnd-java/src/main/java/net/osmand/map/WorldRegion.java
RZR-UA: a142e9f11f

PR #24406, opened 2026-02-03
add new astro configure view screen

0xRe1nk0:


OsmAnd/src/net/osmand/plus/plugins/astro/AstroConfigureViewBottomSheet.kt
```diff
@@ -0,0 +1,451 @@
...
+import net.osmand.plus.utils.ColorUtilities
+import net.osmand.plus.utils.InsetTarget
+import net.osmand.plus.utils.InsetTargetsCollection
+import net.osmand.plus.utils.InsetsUtils
+import net.osmand.plus.utils.UiUtilities
+
+class AstroConfigureViewBottomSheet :
+    BaseMaterialBottomSheetDialogFragment() {
+
+    private lateinit var configScreenListener: AstroConfigScreenListener
```
alex-dev-neo: Use StarMapFragment directly instead of listener

PR #24485, opened 2026-02-17
Fixed starting navigation in aa, when created on device

Corwin-Kh:


OsmAnd/src/net/osmand/plus/auto/screens/RoutePreviewScreen.java
```diff
@@ -245,6 +245,7 @@ private void onRoutesVisible(int startIndex, int endIndex) {
 	}
 
 	private void onNavigate() {
+		getApp().getOsmandMap().getMapActions().startNavigation();
```
Chumva, line 248: With this, we will call startNavigation a second time in [BaseAndroidAutoScreen](https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/auto/screens/BaseAndroidAutoScreen.kt#L87-L97)

PR #24486, opened 2026-02-17
Speed up indexed string table traversal

sonora:
- No per-subtable copies and rescans
- In-place query filtering + restoration (no ArrayList copy per subtable)
- Compact matched index list for VAL writes
- Reusable buffers to reduce allocations
- No behavior change


sonora:
I am closing for now to concentrate on more decisive bottlenecks first.

PR #24500, opened 2026-02-19
Fix Tracks Organization feature after 1st UI Review

nazar-kutz:


OsmAnd/res/values/sizes.xml
```diff
@@ -286,6 +286,7 @@
 	<dimen name="content_padding_double">32dp</dimen>
 	<dimen name="content_padding_small">12dp</dimen>
 	<dimen name="content_padding_half">8dp</dimen>
+	<dimen name="content_padding_half_inverted">-8dp</dimen>
```
Chumva, line 289: Do we really need this inverted value? Slider height cannot be changed?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/data/SmartFolder.kt
```diff
@@ -55,10 +55,24 @@ class SmartFolder(@Serializable var folderName: String) : TracksGroup, Comparabl
...
+	 */
 	fun addTrackItem(trackItem: TrackItem) {
-		if (!getTrackItems().contains(trackItem)) {
-			trackItems = KCollectionUtils.addToList(getTrackItems(), trackItem)
-			folderAnalysis = null
+		val currentItems = getTrackItems()
+		if (!currentItems.contains(trackItem)) {
+			trackItems = KCollectionUtils.addToList(currentItems, trackItem)
+		}
+		invalidateCache()
```
Chumva: Why we call invalidateCache() even if trackItem can already be inside currentItems? Logic is different from removeTrackItem

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/api/SettingsAPI.kt
```diff
@@ -12,4 +14,10 @@ interface SettingsAPI {
 
 	fun getFloatPreference(name: String): Float?
 	fun setFloatPreference(name: String, value: Float)
+
+	fun <T : Enum<T>> registerEnumPreference(name: String, defValue: T, values: Array<T>,
+	                                         clazz: KClass<T>, global: Boolean, shared: Boolean)
+	fun <T : Enum<T>> addEnumPreferenceListener(name: String, listener: KStateChangedListener<T>)
+	fun <T : Enum<T>> getEnumPreference(name: String): T?
+	fun <T : Enum<T>> setEnumPreference(name: String, value: T)
```
Chumva, line 22: Do we need all these new methods for every preference type?

Issue #23857, opened 2025-11-09
Missing search hits in search radius


sonora:
Even after the latitude correction of the search bounding box (https://github.com/osmandapp/OsmAnd/commit/2255c72e54fbc8b5f6eca40230cebe5dd29cbe0a), it seems there is still a flaw as follows.

Reproduce in the far North, like on map Norway_finnmark_europe.obf. 
- Center the map on or near the hamlet Nordkapp.
- Go to search and tap e.g. the Category 'Accomodation', or 'Cafe and Restaurants' 

In both cases will the app report 'No search hits found within 10 km', but after you tap 'Extend the search radius to 20 km' it displays results only 7.x km away.

yuriiurshuliak:
**OsmAnd’s local search near Nordkapp does not return POIs within a 10 km radius, even though several POIs ~7 km away exist and only appear after extending the radius to 20 km.**

This issue can be reproduced.

To reproduce:

1. Open OsmAnd and move the map to Nordkapp
   [https://osmand.net/map/search/result/?query=71.169923%2C25.785853#15/71.1688/25.7816](https://osmand.net/map/search/result/?query=71.169923%2C25.785853#15/71.1688/25.7816)

2. Open **Search**.

3. Select a category such as **Accommodation** or **Café and Restaurants**.

4. Observe:
   **“No search hits found within 10 km.”**

5. Tap **“Extend the search radius to 20 km.”**

6. Results now appear, including POIs located **~7 km away**, which should have been included in the initial 10 km search radius.

This indicates that the search radius filtering still behaves incorrectly for this region.



OsmAnd~ 5.3.0#6541m, released: 2025-11-19

<table style="width:100%">
  </tr>
  <tr>
    <td> <video source src="https://github.com/user-attachments/assets/d5e8c930-0ff7-451b-b0ed-be7c323c1c19"  width="300" height="auto"/>   </td>
 <th><img src="https://github.com/user-attachments/assets/9a8064b5-d20c-4c1c-b48e-5b6c060c2aee"  width="300" height="auto"/> </th>
  </tr>
</table>

vshcherb:
It looks like this issue is related only to high latitudes, I'm not sure it's worth fixing cause I'm not sure how complex it is. Due to Merkaator projection 1 unit is equal 15 km bbox on Equator and ~5 km on 60 latitude.

ivanPyrohivskyi:
Reproduced:
[...]

PR #24508, opened 2026-02-20
Fix bbox search

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/23857

By default was [latitude 30.0](https://github.com/osmandapp/OsmAnd/blob/2255c72e54fbc8b5f6eca40230cebe5dd29cbe0a/OsmAnd-java/src/main/java/net/osmand/util/MapUtils.java#L364)

Example 1. From task 
Purple line ~6.4 km
Green square - wrong bbox (current)
Red circle - 10km radius
Yellow square - if use ``Math.max(Math.abs(l.getLatitude()), 30.0);`` 
**Blue square - result bbox** ``Math.max(Math.abs(l.getLatitude()) - 4, 30.0);``
Знімок екрана 2026-02-20 о 18 05 04

Example 2. (From test poi_biergarten.json)
Yellow square - current
Red circle - 10km radius
**Blue square - result bbox** ``Math.max(Math.abs(l.getLatitude()) - 4, 30.0);``
Green square - if use ``Math.max(Math.abs(l.getLatitude()), 30.0);`` (fail in test, change order of 1 item)
Purple dot and circle - ``"Biergarten Villa Else [[1, POI, 1.000, 13.84 km]]"`` from test
Знімок екрана 2026-02-20 о 18 03 57

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
+++ b/OsmAnd-java/src/main/java/net/osmand/search/core/SearchPhrase.java
@@ -386,17 +386,13 @@ public class SearchPhrase {
 	}
 
 	public static QuadRect calculateBbox(int radiusMeters, LatLon l) {
-		double lat = Math.max(Math.abs(l.getLatitude()) - 4, 30.0);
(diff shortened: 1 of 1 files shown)


ivanPyrohivskyi:
Green - wrong
Blue - fixed
Знімок екрана 2026-02-20 о 19 20 36

PR #24519, opened 2026-02-23
review fix

0xRe1nk0:


OsmAnd/src/net/osmand/plus/myplaces/MyPlacesActivity.java
```diff
@@ -140,6 +162,113 @@ private void setTabs(@NonNull List<TabItem> tabItems) {
...
+		appBar.setExpanded(true, true);
+
+		mSlidingTabLayout.setClickable(!hideTabs);
+		mSlidingTabLayout.setFocusable(!hideTabs);
+
+		updateToolbar();
+	}
+
+	private boolean isDisableAnimations() {
+		return app.getSettings().DO_NOT_USE_ANIMATIONS.getModeValue(app.getSettings().getApplicationMode());
```
Chumva: For current appMode, we can use simple `.get()`

OsmAnd/src/net/osmand/plus/utils/AndroidUtils.java
```diff
@@ -615,6 +615,26 @@ public static int getTextMaxWidth(float textSize, List<String> titles) {
 		return width;
 	}
 
+	public static int getTextWidth(
```
Chumva: We already have [getTextWidth()](https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/utils/AndroidUtils.java#L660), at least reuse code to avoid copy-paste

OsmAnd/src/net/osmand/plus/utils/AndroidUtils.java
```diff
@@ -1533,4 +1553,15 @@ public static int getBatteryLevel(@NonNull Context context) {
 		}
 		return 0;
 	}
+
+	public static String trimTextToMax(@Nullable String text, int maxSymbolNumber) {
```
Chumva: `trim` could be misleading, standard trim works for spaces - "Returns a string whose value is this string, with all leading and trailing space removed". Probably we could use something like `truncateWithEllipsis`

PR #24541, opened 2026-02-27
Fixed after review

Corwin-Kh:


OsmAnd/src/net/osmand/plus/plugins/OsmandPlugin.java
```diff
@@ -447,8 +447,14 @@ public void onMapActivityExternalResult(int requestCode, int resultCode, Intent
 	}
 
 	protected CommonPreference<Boolean> registerBooleanPreference(@NonNull String prefId, boolean defValue) {
+		return registerBooleanPreference(prefId, defValue, true);
+	}
+
+	protected CommonPreference<Boolean> registerBooleanPreference(@NonNull String prefId, boolean defValue, boolean setRelatedPlugin) {
 		CommonPreference<Boolean> preference = settings.registerBooleanPreference(prefId, defValue);
-		preference.setRelatedPlugin(this);
+		if (setRelatedPlugin) {
```
Chumva, line 455: Do we really need this check?

Corwin-Kh: yes. used here ENABLE_3D_MAP_OBJECTS = registerBooleanPreference("enable_3d_map_objects", false, false).makeProfile().cache(); to be able to change preference even if plugin is not active (e.g. free user)

PR #24544, opened 2026-02-28
Fix: Empty row for untagged buildings in Context Menu "Within" list #24233

### LINKED ISSUE #24233: Missing translation - so empty row by default
### Description

Trying to modify some map elements as OSM contributor, I've noticed that, when selecting a building on the map, a void description is returned for it, instead the element type.

Procedure:

I'll make some examples, tryin to be clearer:

1. Element @ 44.98132 N, 9.02663 E (swimming pool) -> The _Within_ menu shows it properly ("Swimming pool)

Image

2. Element @ 44.98137 N, 9.02654 E ("isolated" building) -> The _Within_ menu shows an empty field: it looks like there's nothing to select, but if I tap the empty area then the building is selected.

Image

3. Element @ 44.98312 N, 9.02654 E (building inside a landuse area) -> The _Within_ menu shows an empty field for the building and a "Farmyard" item for the external area. Also, the preview description indicates  " , Farmyard".

Image

4. Element @ 44.99618 N, 9.04223 E (labeled building inside a landuse area) -> In this case, the _Within_ menu shows the building label ("Decatlhon") and a "Retail landuse" item for the external area.

5. Element @ 44.99418 N, 9.01046 E (Church tagged with _amenity=place_of_worship_ and _name=Chiesa di San sebastiano_) -> The _Within_ menu shows a _Place of worship - Chiesa di San Sebastiano_ entry.

Image

### Steps to reproduce

- With the OSM editin plugin enabled, long tap any map element;
- Tap the small grey pinpoint in the upper/right side of the context menu;
- Expand the "Within" section (see attached picture)
- Check the listed items; notice that the one corresponding to the building is empty unless it has some additional tag.

Image

See the example cases mentioned in the _Description_ section above.

### Actual result

If a building is selected that has no additional tags, a blank entry is displayed in the Within menu.

That's quite confusing, since it looks like there's nothing selectable/editable in the pointed location.

### Expected result

There should be an element description also for generic, unlabeled buildings: as it happens, i.e. in Case 5 above (Place of worship - Chiesa di San Sebastiano), I would expect to see "Building - [Building label, if any]" 

```
OsmAnd Version: OsmAnd+ v. 5.2.13
Android version: Android 16
Device model: Motorola Edge 60 Pro
Crash-Logs: None
```

nazar-kutz:
### Description
Fixes #24233

**Root Cause:**
As noted in the issue discussion, the empty row is caused by a missing translation. The `building` tags was intentionally removed from `poi_types.xml` in one of the previous commits to optimize OBF file sizes and POI search performance. Because it is no longer a registered POI type, `poiTranslator` fails to resolve the `building_yes` or `building` keys, returning an empty string for generic buildings in `RenderedObjectMenuController`.

**Solution:**
Added a targeted fallback in `RenderedObjectMenuController`. If the standard `poiTranslator` fails to find a translation for an object's tags, we explicitly check for the `building` key in the object's additional info and fetch the `poi_building` string directly from Android resources. 

This restores the "Building" label in the UI without reverting the POI index optimization and without bloating the map data.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/mapcontextmenu/controllers/RenderedObjectMenuController.java
+++ b/OsmAnd/src/net/osmand/plus/mapcontextmenu/controllers/RenderedObjectMenuController.java
@@ -18,6 +18,7 @@ import net.osmand.plus.render.RenderingIcons;
 import net.osmand.plus.utils.AndroidUtils;
 import net.osmand.util.Algorithms;
 
+import java.util.Collection;
 import java.util.Map;
 
 public class RenderedObjectMenuController extends MenuController {
@@ -163,7 +164,9 @@ public class RenderedObjectMenuController extends MenuController {
 		}
 
 		if (Algorithms.isEmpty(typeStr) && renderedObject != null && mapPoiTypes != null) {
-			typeStr = searchObjectNameByRawTags(mapPoiTypes, renderedObject.getTags());
+			Amenity amenity = builder != null ? builder.getAmenity() : null;
+			Collection<String> additionalInfoKeys = amenity != null ? amenity.getAdditionalInfoKeys() : null;
+			typeStr = searchObjectNameByRawTags(mapPoiTypes, renderedObject.getTags(), additionalInfoKeys);
 		}
 
 		return typeStr != null ? typeStr : super.getTypeStr();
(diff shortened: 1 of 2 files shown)


OsmAnd/src/net/osmand/plus/mapcontextmenu/controllers/RenderedObjectMenuController.java
```diff
@@ -162,9 +162,53 @@ public String getTypeStr() {
...
+			String key = entry.getKey();
+			String value = entry.getValue();
+
+			// Skip common metadata and name tags that shouldn't represent the object type
+			if (key.equals("name") || key.startsWith("name:") || key.equals("ele")
+					|| key.equals("height") || key.equals("min_height")
+					|| key.equals("levels") || key.equals("layer")
+					|| key.equals("type") || key.equals("osmwiki")) {
+				continue;
+			}
```
Chumva: This check looks error-prone; there may be many other tags that shouldn't represent the object type

OsmAnd/src/net/osmand/plus/mapcontextmenu/controllers/RenderedObjectMenuController.java
```diff
@@ -162,9 +162,53 @@ public String getTypeStr() {
...
+
+			String translation = null;
+
+			// Try to translate the key-value pair first
+			if (!Algorithms.isEmpty(value) && !value.equals("yes")) {
```
Chumva: "yes" we skip, but "no" is ok?)

OsmAnd/src/net/osmand/plus/mapcontextmenu/controllers/RenderedObjectMenuController.java
nazar-kutz: Fixed in https://github.com/osmandapp/OsmAnd/pull/24544/changes/8cbe91a00eed49551f83d29c802bb6820d7e1421

OsmAnd/src/net/osmand/plus/mapcontextmenu/controllers/RenderedObjectMenuController.java
nazar-kutz: Fixed in https://github.com/osmandapp/OsmAnd/pull/24544/changes/8cbe91a00eed49551f83d29c802bb6820d7e1421

Issue #23853, opened 2025-11-08
In case start / finish are different HH routing editions immediately switch to A* routing


traut21:
### Description

I like when osmand gets updates and fixes.

But it's an absolute no-go that osmand auto-updates itself - and when using it the next time you will find out that all your downloaded maps are useless now and can't be used for routing any more - which is even worse if you want to go to a country where roaming is so expensive that you have to turn it off.

Happened in summer, happened today again. 

And it's even worse when the app claims: all offline maps of the older app version will be supported in the new version. That's what osmand claims when copying maps from the internal to the external storage.

### Steps to reproduce

open the app and use routing - using after 2025-11-01 with maps earlier than 2025-11-01

### Actual result

routing is not possible, maps are claimed to be outdated

### Expected result

routing with old maps would be required (backwards compatibility) or app updates should not happen with outdated maps, but should suggest to update the app AND all maps when using it.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 5.2.7
Android version:
Device model:
Crash-Logs: ?
```

traut21:
![Image](https://github.com/user-attachments/assets/aae56434-24de-4442-bf24-c5f34194ce16) 
screenshot from move storage operation

sonora:
Can you provide a detailed step-by-step what you did to get inro the situation you experienxe?

I am not aware of a general restriction or problem ike _"using after 2025-11-01 with maps earlier than 2025-11-01"_. 

But there are in deed some issues like
- #23715 (affecting single maps, in v5.2.9 at least mitigated by a notification on the update screen), or
- #22178, which can be worked around by simply switching to one of the olderr 'A-star' routing mechanisms. Unfortunately an automatic fallback seems not in place, this may have to be done manually.

traut21:
All what Osmand hat told me is that it could not route with the outdated maps, which were from 2025-09. It offered me to use online routing instead or to navigate by hand - the maps were visible.

The same had happened to me in 2025-09, with maps from 2025-06

It's just annoying to open the app, without knowing that the app had been auto updated and to see that routing then is not possible any more.

It's not related to the German NRW problem where the larger federal map has been replace by four regional maps, although this topic somwhere does mention the "Missing or outdated maps" problem. I don't know how to change the routing algorithm - but if there is a workaround like that, it could be offered directly after reporting so called outdated maps.


[...]

PR #24569, opened 2026-03-05
Improve MissingMaps strings for better UX clarity

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/23853

1. "**Calculate online**" (users think that it calculates **route** online) -> "**Determine missing maps online**" (Hardy's) -- clarification about maps

2. missing_maps_ignore_details = "**Downloaded maps cover your route? You can try navigating with them, but route calculation may be slow.**" (Hardy's) -- clarification about slow down

3. missing_maps_ignore = "**Use downloaded maps (slow)**" -- clarification (empasize "slow" factor)


RZR-UA:
@sonora, actualize_maps_online replaces the ambiguous "Calculate online" button.

Please revert your change, or make actualize_maps_online more relevant based on the screenshot.

mm

sonora:
@RZR-UA: Ah, now I understand better, thank you!

Regarding the first string, I suggest "_Determine missing maps online_".

For the second case, I am still unclear about the use case and consequence: Why would we ask users if their downloaded maps cover their route? They would not normally know, right? And what's the point advising them to use current maps only if they know the area? Isn't that a weird advice?

Perhaps my knowledge gap here is that in my understanding
- if a device has all necessary maps but with varying time stamps, routing will work perfectly with A\* (which is slower). - No dialog needed, just a toast?
- And if there is an actual gap in the maps needed, routing is not possible in any case. - Also no dialog needed other than the advice to download the needed maps?

But maybe I misunderstand? Please don't hesiate to revert to your original string - we can always fine-tune it once I understand the situation. 😀

PR #24582, opened 2026-03-09
Images description

0xRe1nk0:


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/wiki/WikiCoreHelper.kt
```diff
@@ -268,21 +279,72 @@ object WikiCoreHelper {
...
+
+	private fun buildDescriptionLanguageCandidates(preferredLanguage: String?): List<String> {
+		val languages = linkedSetOf<String>()
+		preferredLanguage?.trim()?.lowercase()?.takeIf { it.isNotEmpty() }?.let { language ->
+			languages.add(language)
+			val normalizedLanguage = language.replace('_', '-')
+			languages.add(normalizedLanguage)
+			normalizedLanguage.substringBefore('-').takeIf { it.isNotEmpty() }?.let { languages.add(it) }
+		}
+		languages.add("en")
```
Chumva: We always have "en"? Also, we create this list multiple times inside the cycle, even if preferredLanguage the same

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/wiki/WikiCoreHelper.kt
```diff
@@ -268,21 +279,72 @@ object WikiCoreHelper {
...
+	): WikiImage? {
 		val imageUrl = image.getString("image") ?: return null
 		val wikiImage = parseImageDataFromFile(imageUrl) ?: return null
 
 		val metadata = wikiImage.metadata
 		image.getString("date")?.let { metadata.date = it }
 		image.getString("author")?.let { metadata.author = it }
 		image.getString("license")?.let { metadata.license = it }
+		image.getString("description")?.let {
+			metadata.description = getLocalizedDescription(it, preferredLanguage)
```
Chumva: Should we store descriptions map? Instead of a single language?

OsmAnd/src/net/osmand/plus/mapcontextmenu/gallery/tasks/GetImageCardsTask.java
```diff
@@ -92,7 +92,7 @@ protected ImageCardsHolder doInBackground(Void... voids) {
 				httpPms.put("lang", preferredLang);
 			}
 
-			List<WikiImage> wikimediaImageList = WikiCoreHelper.INSTANCE.getWikiImageList(params, networkResponseListener);
+			List<WikiImage> wikimediaImageList = WikiCoreHelper.INSTANCE.getWikiImageList(params, app.getLanguage(), networkResponseListener);
```
Chumva: We already have preferredLang [above](https://github.com/osmandapp/OsmAnd/pull/24582/changes#diff-5be5756e27603d2d1129c533a8f2e3e851708156e67a567311ef69612d861db3R87)

PR #24638, opened 2026-03-19
Astro plugin puchase

0xRe1nk0:


OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java
```diff
@@ -1519,6 +1519,7 @@ public boolean usingEnglishNames() {
 	public final OsmandPreference<Boolean> FULL_VERSION_PURCHASED = new BooleanPreference(this, "billing_full_version_purchased", false).makeGlobal();
 	public final OsmandPreference<Boolean> DEPTH_CONTOURS_PURCHASED = new BooleanPreference(this, "billing_sea_depth_purchased", false).makeGlobal();
 	public final OsmandPreference<Boolean> CONTOUR_LINES_PURCHASED = new BooleanPreference(this, "billing_srtm_purchased", false).makeGlobal();
+	public final OsmandPreference<Boolean> ASTRONOMY_PURCHASED = new BooleanPreference(this, "billing_astronomy_purchased", false).makeGlobal();
```
Chumva, line 1522: Do we really need a separate preference for this plugin? Or is it part of a pro subscription?

Chumva:
Plugin is a part of the pro subscription; we should not add separate logic to purchase it

PR #24663, opened 2026-03-25
Disabled automatic connection to BLE devices, if bluetooth is not on

Corwin-Kh:


OsmAnd/src/net/osmand/plus/plugins/externalsensors/devices/sensors/SensorTextWidget.java
```diff
@@ -300,7 +300,9 @@ public int getIconId(boolean nightMode) {
 		if (isConnected) {
 			return isShowSensorData() ? nightMode ? fieldType.nightIconId : fieldType.dayIconId : nightMode ? fieldType.nightBatteryIconId : fieldType.dayBatteryIconId;
 		} else {
-			return isShowSensorData() ? fieldType.disconnectedIconId : fieldType.disconnectedBatteryIconId;
+			//temporary, until new icons be ready
+			return isShowSensorData() ? nightMode ? fieldType.nightIconId : fieldType.dayIconId : nightMode ? fieldType.nightBatteryIconId : fieldType.dayBatteryIconId;
```
Chumva, line 304: Nested ternary operators are usually harder to read, at least use `"(...)"`, or split into two local variables

PR #24668, opened 2026-03-25
Fix 3D Buildings custom color issue when switching map theme

nazar-kutz:


OsmAnd/src/net/osmand/core/android/MapRendererContext.java
```diff
@@ -532,17 +532,18 @@ public void recreate3DObjectsProvider() {
...
 			SRTMPlugin srtmPlugin = PluginsHelper.getPlugin(SRTMPlugin.class);
 			if (srtmPlugin != null && srtmPlugin.ENABLE_3D_MAP_OBJECTS.get()) {
-				Buildings3DColorType buildings3DColorType = srtmPlugin.get3DBuildingsColorStyle();
-				int buildings3DCustomColor = srtmPlugin.getBuildings3dColor();
-				map3DObjectsProvider = new Map3DObjectsTiledProvider(mapPrimitivesProvider, mapPresentationEnvironment, buildings3DColorType == Buildings3DColorType.CUSTOM, NativeUtilities.createFColorRGB(buildings3DCustomColor));
+				Buildings3DColorType colorType = srtmPlugin.get3DBuildingsColorStyle();
+				boolean useCustomColor = colorType == Buildings3DColorType.CUSTOM;
+				int customColor = useCustomColor
+						? srtmPlugin.getBuildings3dCustomColor(nightMode)
+						: BUILDINGS_3D_DEFAULT_COLOR;
```
Chumva, line 541: Should we first try to use the default colour from the style "base3DBuildingsColor"? If we fail to get colour from style, use BUILDINGS_3D_DEFAULT_COLOR. [Example for acquiring colour from style](https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/buttons/MapButtonCard.java#L81-L83)

PR #24671, opened 2026-03-25
Gradient Editor: use appropriate display units guided by the app settings

nazar-kutz:


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/filters/MeasureUnitType.kt
```diff
@@ -1,29 +1,32 @@
...
 import net.osmand.shared.units.PowerUnits
 import net.osmand.shared.units.RotationUnits
 import net.osmand.shared.units.SpeedUnits
-import net.osmand.shared.units.TemperatureUnits
 import net.osmand.shared.units.TimeUnits
 import net.osmand.shared.util.PlatformUtil
 
 enum class MeasureUnitType {
 	TIME_DURATION,
-	DATE,
```
Chumva, line 19: Why remove it? Is it really not needed for android or iOS?

PR #24672, opened 2026-03-25
Fix search issues

ivanPyrohivskyi:
Big Apple - bug in interpolation, set "Big" into house name
New York - increase basemap weight
Велика Фонтанна і Велика Васильківська - decrease street_intersection weight

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/search/core/RegionPriorityProvider.java
+++ b/OsmAnd-java/src/main/java/net/osmand/search/core/RegionPriorityProvider.java
@@ -1,6 +1,5 @@
 package net.osmand.search.core;
 
-import net.osmand.IndexConstants;
 import net.osmand.binary.BinaryMapIndexReader;
 import net.osmand.data.LatLon;
 import net.osmand.data.QuadRect;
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/search/core/RegionPriorityProvider.java
```diff
@@ -92,12 +93,19 @@ private void initPriorityMap(SearchPhrase phrase) {
 
 
     private int calculatePriorityValue(BinaryMapIndexReader region) {
+        int lowPriority = BBOX_MAX / BBOX_STEP + 1;
```
vshcherb: Questionable why do we deprioritise wiki maps ?

ivanPyrohivskyi: According to old logic in search results sorting
```
case OBF_RESOURCE:
				boolean fp1 = o1.isFullPhraseEqualLocaleName();
				boolean fp2 = o2.isFullPhraseEqualLocaleName();
				// sort order: DETAILED|BASEMAP, WIKIPEDIA, TRAVEL
```

Issue #24667, opened 2026-03-25
Prevent Travel points from appearing in Search until exact match


dmpr0:
## Description
When searching for a specific location, **Travel** points from the associated article are displayed in the search results.

### How to reproduce
* Download Travel articles for Europe
* Open **Menu** - **Search** – Type `Amstelveen`

### Actual result

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/ea8814eb-5151-4901-b285-e59b0530d41f"  width="300" height="auto"/> </td>
  </tr>
</table>   

* **Travel** points from the **Amstelveen** article are unexpectedly visible

### Expected result

* The system should display the **Travel** article itself.
* Do not show **Travel** points until there is a full search match.

**Environment**
OsmAnd Version: OsmAnd~ 5.3.0#7421m, released: 2026-03-25
Android: 17
Device model: Google Pixel 6A

Zirochkabila:
# Testing 1

> OsmAnd~ 5.4.0#7552m, released: 2026-04-10
> OsmAnd Maps 5359, released 10.04.26

### **Severity**: 🟠 Major / **Priority**: 🟠 Normal

* [ ] 1. Instead of displaying the Travel article itself, Android shows Travel points, while iOS shows nothing

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
    <th><img src="https://github.com/user-attachments/assets/64e8e7ce-8256-4636-8b8b-808c699dd0e4" width="300" height="auto"/></th>
    <th><img src="https://github.com/user-attachments/assets/2e0f65c7-ac06-4062-b860-12b3ca387855" width="300" height="auto"/></th>
  </tr>
</table>
</details>

dmpr0:
The main issue was fixed, but searching for Amsterdam over Kyiv doesn't produce any results besides those from Travel.

<table style="width:100%">
  <tr>
    <th>With Travel map</th>
    <th>Without</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/20d9b329-630f-4f8c-aa23-0fad47b4c96a"  width="200" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/fc076bc1-90c4-496c-ad36-5c1b0b72f525"  width="200" height="auto"/> </td>
  </tr>
</table>

vshcherb:
<table style="width:100%">
  <tr>
[...]

PR #24676, opened 2026-03-26
Avoid matching parent route for route_article_point route_track_point in searching

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/24667


OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -829,6 +829,9 @@ public boolean publish(Amenity object) {
 								phrase.getSettings().isTransliterate());
 					}
 					if (!nm.matches(sr.localeName) && !nm.matches(sr.otherNames)) {
```
vshcherb, line 831: Route points should be searchable

Issue #24673, opened 2026-03-25
Missing all UI for region select from map


Guylby:
### Description

Hi,

I've recently updated Osmand+ to 5.3.1 and noticed that I can't interact with the map anymore to select or download any region on the map. I can't bring the menu by tapping the map anywhere with the map zoomed out enough, nor the popup to add the missing map when zooming in enough.

I've tried to do all this actions on Osmand 5.2.13 and all menus show up without a problem. 

### Steps to reproduce

Open Osmand
Navigate the map to a region not downloaded
Dezoom far enough
Try to tap the map to open the window allowing map download selection
Zoom enough to bring the popup to download region data
Nothing shows up

### Actual result

<details>
<summary>Osmand+ 5.3.1 with no map interactions showing the map select menus</summary>

https://github.com/user-attachments/assets/cfdcbe4e-b796-4ada-9bac-ea537eaeeb62

</details>

### Expected result

<details>
<summary>Osmand 5.2.13 showing map interactions bringing the map select menus as normal</summary>

https://github.com/user-attachments/assets/cfd6f5d1-2ccd-4c95-84b5-d82c1aca5064

</details>

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 5.3.1
Android version: 13
Device model: Xperia 10 III
Crash-Logs: ?
```

Zirochkabila:
Load the World overview map > Go to `48.86622° N, 6.68688° E` > Zoom level 7 > Click on the map


<details><summary>Details</summary>
<table style="width:100%">
  <tr>
    <th>5.2.13</th>
    <th>5.3.0</th>
  </tr>
  <tr>
    <td><video source src="https://github.com/user-attachments/assets/9e3c5a3c-5ed3-43b5-9e2c-5558fb0b6456"  width="300" height="auto"/></td>
    <td><video source src="https://github.com/user-attachments/assets/c4485185-7a33-4751-b630-98f555c76c11"  width="300" height="auto"/> </td>
  </tr>
</table> 
</details>

dmpr0:
It could happen because of an offline map that covers the entire world. If I remember correctly, this issue can occur with certain USA maps from Plugin. Please check your map list for any external maps.

Zirochkabila:
No external maps have been loaded

Zirochkabila:
# Testing 1

[...]

PR #24701, opened 2026-03-28
Fix "Missing all UI for region select from map #24673"

nazar-kutz:


OsmAnd/src/net/osmand/plus/views/layers/DownloadedRegionsLayer.java
RZR-UA, line 339: Please annotate the method with `@Nullable` and ensure no NPE in the caller.

nazar-kutz: Added `@Nullable` annotation. I also verified the caller: `queryData` is called in `calculateResult` where it's safely wrapped in a `Pair` (which allows nulls). The base `OsmandMapLayer` handles the null check in `onPostExecute` also this value is then assigned to `results`, which is allowed to be `null` and is already properly checked throughout the codebase (e.g., `if (data.getResults() != null))` before any usage. So there is no risk of an NPE here.

PR #24721, opened 2026-03-31
Try to reduce memory churn with widgets preferences

Chumva:


sonora:
Good job, @Chumva ! A notable improvement for low-end and mid-range devices!

Chumva:
@sonora, several places were fixed in this [issue](https://github.com/osmandapp/OsmAnd/issues/24723), I hope it'll help with memory churn

sonora:
Yes, I saw that! :heart: And it's much in line with what I did e.g. here [GpxTrackAnalysis memory/performance improvements](https://github.com/osmandapp/OsmAnd/pull/24222/commits/95e125d204eaa8c408147ed54a983497435185a1), _reusing instead of recreating allocations_.

We have a number of occurrences where in our code we create e.g. a lot of GC churn, which really hurts on lower end devices. We should continue to look out for optimizations, especially in our frequently traversed loops. We get the performance gain for free, making things a lot snappier. :wink:

PR #24731, opened 2026-04-01
Strip accents and diacritics

ivanPyrohivskyi:
To issue https://github.com/osmandapp/web/issues/1510 `Київська` ~ `Киівська`

To icu4j-49_1_patched.jar added symbols table, added new directory `com/ibm/icu/impl/data/icudt49b/translit` and copy files from official repository [icu4j-49.1.jar](https://repo1.maven.org/maven2/com/ibm/icu/icu4j/49.1/icu4j-49.1.jar)

According to iOS commit https://github.com/osmandapp/OsmAnd-core/pull/980


vshcherb:
It should be superefficient if we want to add on base level also it duplicates some functionality for collator or not ? some similar work has been done

PR #24743, opened 2026-04-02
Fix slow loading when simulating navigation of very long routes

Cuberootex:
When simulating navigation, if the route is extremely long, it takes a really long time before the simulation starts. The original code used a double loop, so for every road segment we had to iterate over a list of lat/lon points. Here we precompute the results and store them in a map, to eliminate the inner loop

### DIFF
--- a/OsmAnd/src/net/osmand/plus/simulation/LoadSimulatedLocationsTask.java
+++ b/OsmAnd/src/net/osmand/plus/simulation/LoadSimulatedLocationsTask.java
@@ -14,7 +14,9 @@ import net.osmand.plus.routing.RouteCalculationResult;
 import net.osmand.router.RouteSegmentResult;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/simulation/LoadSimulatedLocationsTask.java
```diff
@@ -59,23 +68,22 @@ private List<SimulatedLocation> getSimulatedLocationsForRoute() {
...
 			LatLon point = segmentResult.getPoint(startPointIndex);
-			for (SimulatedLocation location : locations) {
-				LatLon latLon = new LatLon(location.getLatitude(), location.getLongitude());
-				if (latLon.equals(point)) {
-					location.setHighwayType(segmentResult.getObject().getHighway());
-					location.setSpeedLimit(segmentResult.getObject().getMaximumSpeed(true));
-					if (segmentResult.getObject().hasTrafficLightAt(startPointIndex)) {
-						location.setTrafficLight(true);
-					}
+			SimulatedLocation location = locationMap.get(point);
```
Chumva, line 78: Will it work correctly if there are similar points or almost duplicates?

PR #24807, opened 2026-04-09
Add SliderEx and RangeSliderEx to capture UI crash hierarchy

Chumva:


Chumva:
Possible fix was added [here](https://github.com/osmandapp/OsmAnd/commit/fb23b5998d4003d524a8690d1f16e77f70bcf6aa), will close for now

Issue #18668, opened 2023-12-04
Searching for place should display city name, and street.


dbojan:
### Description

When searching for an object by name, only name of the object is displayed, sometimes city (in about 15% or results or less) in the results list. Street is not displayed in the results list.

Type of place is displayed, which is not useful. If I am searching for say a church, I know the result is the church. I do not know where the church is located (city, street).
If I click on the item, it is opened on the map, with city and street displayed in the text below the map.

I think street and city should be displayed on the result list. So people can know which object is the correct one.
Or at least city name should be included.

### Steps to reproduce

Click on lens to search
We want to search for 'Church of Apostles' in the city of 'Dubica'
Localised, it would be written like this: Crkva Apostola Dubica
enter: apostola, dubica

### Actual result

You get ten churches with no information of street, or even the city name of the object.
You do get information of the type of object (church), which is not very usefull to us. I know it is the church I am looking for, but Cannot make a distinction between 10+ list items, which is the one item I need.
You do get information of the distance, which is good.

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/1901895/b85502b6-a3da-47c8-a181-36f967d44528"  width="300" height="auto"/> </td>
  </tr>
</table>

and the next screen, more results ... 
<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/1901895/0a6587e6-adc9-4a14-8f7a-3fa550bcd708"  width="300" height="auto"/> </td>
  </tr>
</table>

### Expected result

**I think street and city should be displayed on the result list. So people can know which object is the correct one.
Or at least city name should be included.**

Here is how it is handled in Magic Earth app (uses OSM data):
**Street and the city** is displayed. Street is marked with red line (by me), and the city is marked with orange line.

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/1901895/e6b96ba7-ea4e-43d9-b43a-a82c3837a4f3"  width="300" height="auto"/> </td>
  </tr>
</table>

Here is how it is handled in Organic Maps (open source, uses OSM data):
**City** is displayed, but the street is not. Even just knowing the city is very helpfull.
The city is marked with orange line by me

<table style="width:100%">
  <tr>
[...]

PR #24818, opened 2026-04-10
Fix deduplication

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/18668#issuecomment-4214340434

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
+++ b/OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
@@ -268,26 +268,43 @@ public class SearchUICore {
 		}
 
 		private void filterSearchDuplicateResults(List<SearchResult> lst) {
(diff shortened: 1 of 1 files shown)


OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java
```diff
@@ -268,26 +268,43 @@ public void filterSearchDuplicateResults() {
...
+			int size = lst.size();
+
+			while (i < size) {
+				SearchResult current = lst.get(i);
+				List<SearchResult> sameList = new ArrayList<>();
+				sameList.add(current);
+
+				int j = i + 1;
+				while (j < size && sameSearchResult(current, lst.get(j))) {
+					sameList.add(lst.get(j));
```
vshcherb, line 282: if (maxWeight ... without cycle and list

Issue #16530, opened 2023-02-23
Add Dutch RD coordinate system


Zirochkabila:
"Dutch RD" (Rijksdriehoekscoördinaten) is a geodetic coordinate system used in the Netherlands. It is a projected coordinate system that uses a grid-based system of coordinates to identify locations on the earth's surface.

The Dutch RD coordinate system is based on the "Rijksdriehoeksmeting," which is a triangulation system that was established in the Netherlands in the 19th century. The system divides the Netherlands into a grid of squares, each of which is assigned a unique coordinate based on its position relative to a reference point. The reference point is located in the city of Amersfoort, and its coordinates are defined as (0,0) in the Dutch RD system.

To obtain the Dutch RD coordinates for a specific location in the Netherlands, you would need to determine the coordinates of that location relative to the reference point in Amersfoort. This can be done using specialized software or online tools that are designed to work with the Dutch RD coordinate system.

It is worth noting that the Dutch RD system is primarily used by professionals in surveying, mapping, and other geospatial fields. For most everyday purposes, such as finding directions or locating an address, the more common WGS84 or GPS coordinates are used.

mrngm:
Perhaps good to mention EPSG 28992 ([src](https://epsg.io/28992)) for future reference, as there are a [few](https://github.com/osmandapp/OsmAnd/issues/16156) [other](https://github.com/osmandapp/OsmAnd/issues/17311) issues also requesting support for country specific coordinate systems.

vshcherb:
Yes https://epsg.io/28992 we need to add all coordinate systems to allow choice. Merged to https://github.com/osmandapp/OsmAnd/issues/17311

PR #24833, opened 2026-04-13
Add Dutch RD coordinate format support

Joelsuperstar:
Added the Dutch RD coordinates system.

As also suggested in for example #16530


vshcherb:
Will be generalized using proj4.db project.

PR #24863, opened 2026-04-17
add all icons

### LINKED ISSUE #22508: New profile icon: Running
### Describe the idea

IMHO the profiles are missing icons/logos for more activities.  I am sure there are a number of other activities, but what I can think of right now:

- Running 🏃
- Sailing ⛵ 
- Treasure hunt / geocaching 📫 
- Roller skates and/or skateboard 🛹

### Expected behaviour

Symbols for the above mentioned activities.

- I currently use another symbol in this case (such as the UFO, which is funny and conveys that I intend something different), but that leads to secret knowledge which is required to use my profiles.
- One could add the possibility to add own icons, but that is probably way more complicated than just adding a few new symbols.

### Context

_No response_

0xRe1nk0:


OsmAnd/src/net/osmand/plus/mapcontextmenu/editors/icon/ProfileEditorIconController.java
```diff
@@ -0,0 +1,104 @@
...
+
+	@Override
+	protected void sortCategories() {
+		categories.sort((c1, c2) -> {
+			int order1 = getCategoryOrder(c1.getKey());
+			int order2 = getCategoryOrder(c2.getKey());
+			if (order1 != order2) {
+				return Integer.compare(order1, order2);
+			}
+			return c1.getTranslation().compareTo(c2.getTranslation());
```
Chumva: Should we use collator here?

Issue #6827, opened 2019-04-14
Time problem on bus routes (public transport navigation)


OpenSourceForEver:
Sometimes, time on bus routes is wrong. I think problem is with motorways and trunk roads. 

Examples:
From https://www.openstreetmap.org/#map=19/40.63313/-4.00485 to https://www.openstreetmap.org/#map=18/40.43505/-3.71884 there are some bus routes that takes about 30 minutes, (by car it takes 25 minutes according to Osmand with a similar route) however Osmand shows that these bus routes takes 66 minutes.
Bus routes mentioned are: https://www.openstreetmap.org/relation/4422724 https://www.openstreetmap.org/relation/2684575 https://www.openstreetmap.org/relation/5581608 https://www.openstreetmap.org/relation/5595416#map=11/40.5646/-3.8813

From: https://www.openstreetmap.org/#map=19/40.64132/-4.00552 to https://www.openstreetmap.org/#map=19/40.70233/-4.00507 there is a bus route that makes the same route as a car but Osmand calculates that car takes 6 minutes and bus takes 14 minutes. Mentioned bus route is: https://www.openstreetmap.org/relation/5580303

I think time on other type of roads (streets, tertiary...) is ok because time on other sections of the mentioned routes are ok.

vshcherb:
the bus speed is averaged to https://github.com/osmandapp/OsmAnd-resources/blob/master/routing/routing.xml#L2002 -  30kmh which is slow but it gives very good estimation in cities. Now we can't distinguish if it is a speed bus or it drives by motorway so there is no possibility to change it. Also bus is usually 20% + stop times slower than car.

OpenSourceForEver:
"Also bus is usually 20% + stop times slower than car." 
Yes but examples I mentioned bus and car follow the same route and there is no bus stops and the difference is bigger than 20%.

I think the problem is, as you mentioned, bus speed that should be different depending on road type

OpenSourceForEver:
Maybe the solution would be to establish bus speed depending on speed limit of the road. Bus speed would be 80% of road speed limit if this limit is 90 km/h or above  and 100% of speed limit if the limit is smaller than 90 km/h.

vshcherb:
[...]

PR #24891, opened 2026-04-21
Implement "bus-max" speed in TransportRoutePlanner

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/6827


vshcherb (changes requested):
minSegmentDistToIncreaseSpeed - strange assumptions

vshcherb:
Overcomplication on client side that wouldn't solve all other problems related to incorrect travel time of the bus.

PR #24908, opened 2026-04-22
fix internet check

0xRe1nk0:


OsmAnd/src/net/osmand/plus/plugins/astronomy/views/contextmenu/AstroArticleDialogFragment.kt
```diff
@@ -33,7 +33,10 @@ class AstroArticleDialogFragment : WikiArticleBaseDialogFragment() {
 
 		private const val ARG_WIKIDATA_ID = "wikidataId"
 		private const val ARG_LANG = "lang"
-		private const val DEFAULT_BASE_URL = "https://wikipedia.org/"
+		private val BODY_CONTENT_REGEX = Regex(
+			"<body[^>]*>([\\s\\S]*?)</body>",
```
Chumva, line 37: Will this work correctly?

PR #24910, opened 2026-04-22
100426 1

vshcherb:


OsmAnd-java/src/main/java/net/osmand/CollatorStringMatcher.java
```diff
@@ -212,7 +213,7 @@ public static String alignChars(String fullText) {
 		while ((i = fullText.indexOf('ß')) != -1) {
 			fullText = fullText.substring(0, i) + "ss" + fullText.substring(i + 1);
 		}
-		return fullText;
+		return UnicodeDiacritics.getInstance().stripDiacritics(fullText);
```
vshcherb, line 216: Strip is unnecessary overhead for normalized strings in index

OsmAnd-java/src/main/java/net/osmand/CollatorStringMatcher.java
vshcherb, line 85: Should be tested with commented out

PR #24922, opened 2026-04-24
add track description in multiselection menu

### LINKED ISSUE #24608: Display track info in What's here menu
### Describe the idea

Frequently I´m using working copies of existing tracks, to create new routes (split them up and merge the sections to a new route). After I created the "copy" (it´s not possible to just copy a track inside osmand, so I "share" it to osmand and get the copy into the import folder).
When tapping the track on the map (and now there is the original and the copy, and maybe others too), it´s not possible to differ between the original and the copy.

And here´s the feature request: displaying the containing folder name "instead of / beside" the displayed "track" would be helpful to know, which one is the one to work with.

Nothing important, but helpful. 
Thanks for considering!

<table style="width:100%">
  <tr>
<th> </th>
  </tr>
</table>  

### Expected behaviour

display the containing folder name

no alternative

### Context

_No response_

0xRe1nk0:


OsmAnd/src/net/osmand/plus/track/helpers/GpxUiHelper.java
```diff
@@ -92,6 +92,41 @@ public static String getColorValue(String clr, String value) {
 		return getColorValue(clr, value, true);
 	}
 
+
+
+	@NonNull
+	public static String getTrackShortDescription(@NonNull Context context,
```
Chumva, line 97: Is this copy pasted part from [TrackViewHolder](https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/src/net/osmand/plus/configmap/tracks/viewholders/TrackViewHolder.java#L198)? Can we somehow reuse it instead of adding similar code?

Issue #21912, opened 2025-02-16
Bad routing in roundabout when route length is approx > 27km


wolfy1339:
### Routing engine

- [x] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile

* car profile
* allow private access = enabled
* consider temporary limitiations = enabled
* fuel-efficient way = enabled
* No roads selected to be avoided

### Start and end points

Start: 46.13367, -70.65624
End: 46.32927, -70.41533

### Actual route

The route tells us to take the 4th exit on the roundabout which would mean going around the whole roundabout

### Expected route

The routing should not go around the whole roundabout but take the first exit to stay on 127e Rue

### Maps you used (online or offline)

- [x] Offline maps offered within the OsmAnd app for download.  February 2025 + Live updates.
- [ ] Online (tile / raster) maps 

### Your Environment

```
OsmAnd Version: 4.9.10
Android version: 15
Device model: Google Pixel 7 Pro

```

### Anything else?

It seems that the router only gets confused when routes have a length of around 27km or more. when the route is less, the routing guidance is just fine
![Image](https://github.com/user-attachments/assets/f015ce72-27b2-431a-ba33-a1acbf2896f0)

yuriiurshuliak:
The bug was **reproduce without live updates** by following these steps:  

1. Use **OsmAnd offline routing** with the **car profile**.  
2. Enable:  
   - **Allow private access**  
   - **Consider temporary limitations**  
   - **Fuel-efficient way**  
3. Set the following coordinates:  
   - **Start:** `46.13367, -70.65624`  
   - **End:** `46.32927, -70.41533`  
4. Observe the navigation instructions at the roundabout.  

**Issue:** The route incorrectly suggests taking the **4th exit**, causing a full circle around the roundabout instead of taking the **1st exit** to stay on **127e Rue**.  

**Possible Cause:** This behavior is likely influenced by the **"Short way"** option being enabled.  

For reference, here is the **Web version of the route**:  
[Web version of the route](https://osmand.net/map/navigate/?start=46.133670,-70.656240&end=46.329270,-70.415330&profile=car&params=car,short_way#17/46.13282/-70.64889)  

OsmAnd~ 5.0.0#4400m, released: 2025-02-18

<table style="width:100%">
  <tr>
[...]

PR #24924, opened 2026-04-24
Improve HH cost calculations for default parameters

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21912


OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
```diff
@@ -221,6 +209,22 @@ public int compare(RouteSegmentPoint o1, RouteSegmentPoint o2) {
 		return null;
 	}
 
+	public static RouteSegmentPoint calcPreciseRouteSegmentPoint(RouteDataObject r, int x, int y) {
```
vshcherb: non static

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -982,6 +982,29 @@ private TLongObjectHashMap<T> initStart(HHRoutingContext<T> hctx, RouteSegmentPo
 			System.out.println("  " + hctx.rctx.calculationProgress.getInfo(null));
 		}
 		if (frs != null) {
+			for (FinalRouteSegment o : frs.all) {
+				int startSegment = reverse ? o.getSegmentEnd() : o.getSegmentStart();
+				int endSegment = reverse ? o.getSegmentStart() : o.getSegmentEnd();
+				long pntId = calculateRoutePointInternalId(o.getRoad().getId(), startSegment, endSegment);
+				if (!hctx.pointsByGeo.containsKey(pntId)) {
```
vshcherb: - why it's not part of the cycle below?
- comment missing what's going on at least oneine

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -982,6 +982,29 @@ private TLongObjectHashMap<T> initStart(HHRoutingContext<T> hctx, RouteSegmentPo
...
+					int x = reverse ? hctx.startX : hctx.endX;
+					int y = reverse ? hctx.startY : hctx.endY;
+					RouteSegmentPoint road = RoutePlannerFrontEnd.calcPreciseRouteSegmentPoint(o.getRoad(), x, y);
+					if (road != null) {
+						if (reverse) {
+							hctx.roadStartX = road.preciseX;
```
vshcherb: is it always overwrite ? no checks if it is already set or not, simply random?

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
RZR-UA: 1. It is moved before the main cycle because we should get clarified `roadStart` points before going through `frs.all`.
2. `frs.all` is ordered, so as in the previous version (which is limited to 1st result using set+PNT_SHORT_ROUTE_START_END), it is enough to catch 1st non-HH point, catch precise points, and break to the main cycle

OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
RZR-UA: It is a one-off clarification. roadStart points are initially set to the real start/end, and here we just clarify it once and stop this cycle.

OsmAnd-java/src/main/java/net/osmand/router/RoutePlannerFrontEnd.java
RZR-UA: 1d0d5f5e07

Issue #23547, opened 2025-09-25
Place=allotments are not registered in the search index


Egorix2001:
### Description

There is an issue with how OsmAnd handles addresses for streets located within areas tagged as place=allotments (садоводческое товарищество). Instead of associating the street and house number with the allotment's name, OsmAnd associates it with a nearby, separate locality (e.g., a village).
This leads to two main problems:
Incorrect address display: When an object (like a building) inside an allotment is selected, the address shown at the bottom of the screen references a neighboring village, not the allotment itself.
Incorrect search behavior: It's impossible to search for an address using the allotment's name. The search only works if the incorrect, neighboring village name is used.
The community discussion suggests that OsmAnd may not fully support nested streets and addressing within the place=allotments tag, unlike administrative boundaries for towns or villages.

### Steps to reproduce

Navigate to the "Ивушка-91" (Ivushka-91) allotment, near the village of Бульково (Bulkovo) in Belarus. The correct coordinates are 52.10920, 23.91849.
Select the building with house number 68 on the street "4-я лінія" (4th Line).
Observe the address displayed at the bottom of the screen.
Now, use the search function.
Try to search for the correct address: Ивушка-91, 4-я лінія, 68.
Try to search for the incorrect address that OsmAnd uses: Бульково, 4-я лінія, 68.

### Actual result

When the building is selected, the address is displayed as "4-я лінія 68, Булькова".
The search for Ивушка-91, 4-я лінія, 68 yields no results.
The search for Бульково, 4-я лінія, 68 successfully finds the location.

### Expected result

The address should be displayed correctly as "4-я лінія 68, Ивушка-91".
The search for the address using the allotment's name (Ивушка-91, 4-я лінія, 68) should work correctly.
The address should not be associated with the separate, nearby village of Бульково.

### Your Environment (required)

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/6dea3107-b3ef-43f4-9709-56d448825466"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/6af1449b-1d56-4e66-a11b-786ac0ee1c22"  width="300" height="auto"/> </td>
  </tr>
</table>

OsmAnd Version: 5.2.0#6150m
Android version:13
Device model: Xiaomi Redmi Note 10S 
Crash-Logs: -

yuriiurshuliak:
[...]

Issue #16715, opened 2023-03-15
Incorrect display of the city in the name of the establishment (Android & iOS)


Zirochkabila:
### Description
Incorrect display of the city in the name of the establishment

- [ ] **Android**: Flowers > must be Clayton

- [ ] **iOS**: Wilson Mills > must be Clayton

Map

[OSM](https://www.openstreetmap.org/node/10220642830#map=19/35.64877/-78.39227)
[OSM](https://www.openstreetmap.org/node/6410706771#map=19/35.64877/-78.39276)

### Steps to reproduce
Find this place `35.64847° N, 78.39355° W` and `35.64872° N, 78.39230° W`

### Actual result

<details><summary>Android: Flowers</summary>
<video src="https://user-images.githubusercontent.com/104760013/225427838-2aac8055-4168-4ed5-8fdf-7beab6af536b.mp4"></video>
</details> 

<details><summary>iOS: Wilson Mills</summary>
<video src="https://user-images.githubusercontent.com/104760013/225428275-3656acac-6042-422e-84fb-bb7cf34ba478.mov"></video>
</details> 

### Expected result

Clayton - like that `35.64916° N, 78.39373° W`

<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
<th><img src="https://user-images.githubusercontent.com/104760013/225430319-a62f8650-9f16-4cd0-8d90-0fba98399546.jpeg"  width="240" height="auto"/> </th>
  <th><img src="https://user-images.githubusercontent.com/104760013/225429277-10ea2424-ea91-41d7-adc5-5634f0a0f8d0.png"  width="350" height="auto"/> </th>
  </tr>
</table> 

### Your Environment
OsmAnd Version: OsmAnd~ 4.4.0#14239mqta, Released: 2023-03-15 / 4.4.6.6
Android/iOS version: Android 12 / iOS 15.7.3
Device model: Poco x3 nfc / iPad mini 4

dmpr0:
https://www.openstreetmap.org/relation/179024#map=16/35.6487/-78.3976
![Screenshot 2023-03-22 at 17 16 04](https://user-images.githubusercontent.com/9365308/226951142-6eafa1c5-cb2c-4a23-b194-8107bed7cc28.png)

xmd5a2:
OsmAndMapCreator -> Where am I?
[...]

PR #24961, opened 2026-04-29
Improve justifyReverseGeocodingSearch(), support place=allotments entity

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/23547
https://github.com/osmandapp/OsmAnd/issues/16715

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
+++ b/OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
@@ -35,8 +35,8 @@ public class CommonWords {
 		}
 	}
 
-	private static boolean isNumber2Letters(String name) {
-		return Character.isDigit(name.charAt(0)) && letters(name) < 2;
+	public static boolean isNumber2Letters(String name) {
+		return !name.isEmpty() && Character.isDigit(name.charAt(0)) && letters(name) < 2;
 	}
 
 	public static boolean isCommon(String name) {
@@ -96,8 +96,8 @@ public class CommonWords {
 		if (i != null) {
 			return i.intValue();
 		}
-		if (regionNames.contains(name)) {
-			return commonWordsDictionary.size();
+		if (name.length() > 2 && regionNames.contains(name)) {
+			return commonWordsDictionary.size(); // length > 2 is for NC 42 ("NC" and "42" exist in regionNames)
 		}
 		return -1;
 	}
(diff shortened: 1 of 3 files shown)


OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
```diff
@@ -96,8 +96,8 @@ public static int getCommonGeocoding(String name) {
 		if (i != null) {
 			return i.intValue();
 		}
-		if (regionNames.contains(name)) {
-			return commonWordsDictionary.size();
+		if (name.length() > 2 && regionNames.contains(name)) {
```
ivanPyrohivskyi, line 99: Controversial decision. What about PA (Pensilvania) or TX (Texas).
I checked data (for street name only) and context where these are calling. I think is redundant for geocoding and need to remove:
```
		if (regionNames.contains(name)) {
			return commonWordsDictionary.size();
		}
```

OsmAnd-java/src/main/java/net/osmand/binary/GeocodingUtilities.java
```diff
@@ -44,7 +46,7 @@ public class GeocodingUtilities {
 	public static final float STOP_SEARCHING_STREET_WITHOUT_MULTIPLIER_RADIUS = 400;
 
 	public static final int DISTANCE_STREET_NAME_PROXIMITY_BY_NAME = 15000;
-	public static final float DISTANCE_STREET_FROM_CLOSEST_WITH_SAME_NAME = 1000;
+	public static final float DISTANCE_STREET_FROM_CLOSEST_WITH_SAME_NAME = 7500;
```
ivanPyrohivskyi, line 49: Controversial decision

RZR-UA: What controversy? The result is better (NC-42 and allotments "1-я центральная" fixed), and performance is not slowed; it stays under 150-180 ms per 1 reverse geocoding before/after the patch.

OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
RZR-UA: Remark "What about PA (Pensilvania) or TX (Texas)" is irrelevant, because state names do not exist in road names. This method is about reverse street search by parts of road names from the routing section.

OsmAnd-java/src/main/java/net/osmand/binary/GeocodingUtilities.java
vshcherb: Constant should be deleted 7.5 km almost equal 15 km

PR #24972, opened 2026-05-01
fix wrong fav size

### LINKED ISSUE #24795: wrong huge "favs data amount"
### Description

From https://github.com/osmandapp/OsmAnd/issues/24654

I mentioned this before as a side note in another issue ("_italic_ fav group names"), but probably it has to get "it´s own" issue:

All my favorites (9660 favs in 68 groups) do show up as 752MB in my places / favorites.
Each single favorite group - no matter if it contains 1 or 1000 favs - in here is displayed as 11MB, so it sums up to 752MB.

This "11MB" in fact is the "real size" of ALL fav files (seen by a file manager), the "real" size of each fav file is small (KB) (even my largest group of 1300 favs only is 500KB "real" file size).

) each single fav group seems to be displayed with the "real overall size" 11MB
) the overall favs size of 752MB doesn´t seem to be correct (it´s the sum of all the 11MB-groups)

### Steps to reproduce

see above

### Actual result

large unreal count?

### Expected result

correct count?

nightly and playstore.

0xRe1nk0:


OsmAnd/src/net/osmand/plus/myplaces/favorites/FavouritesHelper.java
```diff
@@ -237,6 +237,9 @@ private boolean merge(@NonNull Map<String, FavoriteGroup> source, @NonNull Map<S
 				destinationGroup = new FavoriteGroup(sourceGroup);
 				destination.put(key, destinationGroup);
 			} else {
+				if (sourceGroup.getSize() > 0 || sourceGroup.getTimeModified() > 0) {
```
Chumva: We could move these checks to copyFileMetadata

Issue #24880, opened 2026-04-19
Navigation ends when switching away from app and back again


sjorford:
### Description

If I set a walking destination and start navigation, and then briefly switch away from the app, when I switch back it says "You have arrived" even though I am not yet at the destination. If I press "Recalculate route" it correctly continues from my current location.

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/8bd16168-6ef9-4c13-a6e8-98c4787b37e2"  width="300" height="auto"/> </th>
  </tr>
</table>  

### Steps to reproduce

.

### Actual result

.

### Expected result

.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 5.3.5
Android version: 16
Device model: Pixel 7
Crash-Logs: ?
```

scaidermern:
Sounds a little bit like #24383.

sonora:
Yes, sounds exactly like what I had experienced and reported in #24383.

So if you have this in v5.3.5, it seems not fixed yet. I had not been able to reproduce again yet - means we still seem to have no reliable way to reproduce.

sjorford:
Yes sounds the same - I have only experienced it since the upgrade to 5.3.5. I will try and reproduce on another device.

Zirochkabila:
Related to the https://github.com/osmandapp/OsmAnd/issues/24383

I tried to reproduce the issue by closing the app (while navigation was active) and minimizing the screen - it didn't reproduced
It looks like it's been fixed, or perhaps the steps to reproduce it are more complicated than we realize.

sjorford:
I can't reproduce this today. The issue may be related to map downloads - I will reopen if I can reproduce it reliably.

sonora:
- I had no change in my offline maps between when I had the issue and when it disappeared again.
- And since I had this (sporadically) with our nightly build of January 28 while it has now been reproduced on our very recent v5.3.5, it seems also to be present in quite a span of app versions.

Zirochkabila:
@sonora @sjorford @suntri-star We need your system and profile settings - please attach them (they will be made public at https://osmand.net/docs/user/personal/import-export?_highlight=exp) or send them to our support team at support@osmand.net. 
Also, please check what is set in Location source (the issue may be there)

sonora:
[...]

PR #24975, opened 2026-05-01
Add test logs for location providers

Chumva:


sonora:
Perhaps we can / should incorporate this https://github.com/osmandapp/OsmAnd/issues/24880#issuecomment-4358303367 here?

Chumva:
@sonora, you could add logs if necessary. Also, it is better to check for PluginsHelper.isDevelopment() to avoid spamming logs

This request was mostly for testing hmd location providers, not planned to merge all these logs

PR #24977, opened 2026-05-01
Fix color palette cache not invalidated after cloud sync

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/543

PR #25008, opened 2026-05-05
Implement MAX_INC_COST_CF_VIGILANT to start/end points

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21912


OsmAnd-java/src/main/java/net/osmand/router/HHRoutePlanner.java
```diff
@@ -1029,7 +1032,9 @@ private TLongObjectHashMap<T> initStart(HHRoutingContext<T> hctx, RouteSegmentPo
 					pnt.setDistanceToEnd(reverse,
 							pnt.index == PNT_SHORT_ROUTE_START_END ? 0 : hctx.distanceToEnd(reverse, pnt));
 					pnt.setDetailedParentRt(reverse, o);
+					pnt.isStartEnd = isStartEnd;
```
vshcherb: will be problems with caching etc it's better to check via parent?

RZR-UA: I tried but was unable to discover any parent links or other "parent-like" indicators in NetworkDBPoint. IMHO, the boolean isStartEnd is the simplest way to explain why another const is used. Ultimately, as we always set pnt.isStartEnd in the current calculation, there should be no caching issues.

RZR-UA: 89f2af054bf65d3ea4be687ced2d11c7ec96c492

PR #25019, opened 2026-05-06
add-german-street-to-frequently-used

Gateway07:
improve search performance by adding streets in frequently used words

https://github.com/osmandapp/OsmAnd/issues/24229#issuecomment-4344311823


OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
```diff
@@ -894,7 +894,24 @@ private static void parseRegionNames(WorldRegion region, Set<String> result) {
 		addFrequentlyUsed("verdi");
 		addFrequentlyUsed("augusta");
 		addFrequentlyUsed("neuburger");
-
+		addFrequentlyUsed("hauptstr");
+		addFrequentlyUsed("hauptstraße");
+		addFrequentlyUsed("hauptstrasse");
```
ivanPyrohivskyi: We already make SearchAlgorithms.replaceGermanSS(string) in addFrequentlyUsed

OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
```diff
@@ -894,7 +894,85 @@ private static void parseRegionNames(WorldRegion region, Set<String> result) {
 		addFrequentlyUsed("verdi");
 		addFrequentlyUsed("augusta");
 		addFrequentlyUsed("neuburger");
-
+		addFrequentlyUsed("hauptstr");
```
vshcherb, line 897: order is not correct

PR #25038, opened 2026-05-08
Fix diff display scale app aa

Corwin-Kh:


OsmAnd/src/net/osmand/plus/views/OsmandMapTileView.java
```diff
@@ -450,6 +450,7 @@ public void setView(@Nullable View view) {
 			if (Build.VERSION.SDK_INT >= 26) {
 				view.setDefaultFocusHighlightEnabled(false);
 			}
+			setComplexZoom(getZoom(), getSettingsMapDensity());
```
Chumva: Do we need to call here MapRendererContext.updateMapSettings(), similar to setPhoneMapDensity & refreshMapScale?

PR #25041, opened 2026-05-08
Implement a per-profile setting for notifications

andreasscherbaum:
Per default is the setting on for all profiles. If turned off, no notification is sent for route navigation. This avoids sending many notifications in car profiles.

Solves #24180


andreasscherbaum:
For now this PR creates the notification, but then does not send it if the setting is off. This allows using the notification text for other use cases, like debugging.

coreysnipes:
Very nice, thanks.

PR #25059, opened 2026-05-11
Speed up search for short street house

vshcherb:


OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -674,7 +705,7 @@ public boolean isCancelled() {
 								boolean match = matchAddressName(phrase, res, cityResult, true);
 								if (match) {
 									newParentSearchResult = cityResult;
-								} else {
+								} else if(hasNonNumericLeftUnknownSearchWord(res)) { // speed up
```
ivanPyrohivskyi, line 708: Київська 30Д
Husova 243/11
14-45, 31st Road

PR #25070, opened 2026-05-13
24873: Fixed restore draw track on return to AA

Corwin-Kh:


OsmAnd/src/net/osmand/plus/auto/NavigationSession.java
```diff
@@ -213,6 +217,8 @@ public void onStart(@NonNull LifecycleOwner owner) {
 			checkAppInitialization(new RestoreNavigationHelper(app, null));
 		}
 		app.getRoutingHelper().addCalculationProgressListener(this);
+		GPXLayer gpxLayer = app.getOsmandMap().getMapLayers().getGpxLayer();
+		gpxLayer.setInvalidated(true);
```
Chumva, line 221: Looks like a specific fix, it is ok for testing, but probably needs improvements for master/release

Chumva:
After testing, we have to clean up these logs/changes

PR #25084, opened 2026-05-15
Favorite subgroups

### LINKED ISSUE #9967: Favorites: Introduce subgroups
It would be nice to have subgroups for favorites, the same way as we already have them for tracks. Currently with many favorites and favorite groups it becomes quite a mess.

0xRe1nk0:


OsmAnd/src/net/osmand/plus/myplaces/favorites/FavouritesHelper.java
```diff
@@ -101,6 +101,21 @@ public List<FavoriteGroup> getFavoriteGroups() {
 		return favoriteGroups;
 	}
 
+	@NonNull
+	public FavoriteFolderTree getFavoriteFolderTree() {
+		return new FavoriteFolderTree(favoriteGroups);
```
Chumva: Won't this have performance issues? It is used during UI updates, but if there are several thousand points, it may lead to lags

OsmAnd/res/values/strings.xml
```diff
@@ -10,6 +10,12 @@
 	- For wording and consistency, please note https://docs.osmand.net/docs/technical/contributions/translating-osmand
 	Thx - Hardy
 -->
+	<string name="favorite_confirm_delete_folder">This will delete folder \"%1$s\", including nested folders and %2$d favorite point(s).</string>
+	<string name="favorite_folder_rename_conflict">A folder with this name already exists.</string>
+	<string name="favorite_folder_invalid_name">Folder name can\'t contain / or _%_</string>
```
Chumva, line 15: Will old favourite groups with "/" in names be automatically split?

OsmAnd/src/net/osmand/plus/mapcontextmenu/editors/SelectFavouriteGroupBottomSheet.java
```diff
@@ -13,14 +19,23 @@
...
 import net.osmand.plus.myplaces.favorites.FavouritesHelper;
 import net.osmand.plus.utils.AndroidUtils;
 
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 public class SelectFavouriteGroupBottomSheet extends SelectPointsCategoryBottomSheet {
 
+	private static final String PATH_ELLIPSIS = "...";
```
Chumva: We already have FavoriteFolderPath.PATH_ELLIPSIS, or should we for UI use `…` ?

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
```diff
@@ -1670,6 +1675,10 @@ object GpxUtilities {
 		return gpxFile
 	}
 
+	private fun getExtensionsTagName(tag: String): String {
+		return tag.lowercase().removePrefix(OSMAND_EXTENSIONS_PREFIX)
```
Chumva: Why do we remove prefix here?

OsmAnd/src/net/osmand/plus/myplaces/favorites/FavoriteFolderNode.java
```diff
@@ -0,0 +1,141 @@
...
+
+import net.osmand.data.FavouritePoint;
+
+import java.text.Collator;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+public class FavoriteFolderNode {
```
Chumva: FavoriteFolderNode->FavoriteFolder? (similar to TrackFolder?)

OsmAnd/src/net/osmand/plus/myplaces/favorites/FavoriteFolderTree.java
```diff
@@ -0,0 +1,117 @@
...
+import net.osmand.util.Algorithms;
+
+import java.text.Collator;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class FavoriteFolderTree {
```
Chumva: Looks like the current structure works, but it runs in parallel with the existing `FavoriteGroup` + `FavouritePoint` model, which makes the favourites code more complicated. Do we really need both `FavoriteFolderTree` and `FavoriteFolderNode`?

Could we try a smaller model similar to `TrackFolder`: one `FavoriteFolder` object with `subFolders`, optional backing `FavoriteGroup`, direct points, flattened helpers/stats. That should still support virtual parent folders without adding a separate tree abstraction

OsmAnd/src/net/osmand/plus/myplaces/favorites/FavoriteFolderPath.java
```diff
@@ -0,0 +1,228 @@
...
+		List<String> segments = split(fullPath);
+		if (segments.size() <= 1) {
+			return "";
+		}
+		return join(segments.subList(0, segments.size() - 1));
+	}
+
+	@NonNull
+	public static String lastSegment(@Nullable String fullPath) {
+		List<String> segments = split(fullPath);
```
Chumva: Probably, this `split()` if called very often from UI may lead to memory churn
We could use `lastIndexOf + substring`

OsmAnd/src/net/osmand/plus/myplaces/favorites/FavoriteFolderPath.java
```diff
@@ -0,0 +1,228 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import net.osmand.plus.utils.ColorUtilities;
+import net.osmand.util.Algorithms;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class FavoriteFolderPath {
```
Chumva, line 11: Could we split into two classes? One with domain/path logic and the other for UI styling? If we move the new logic to the shared library, UI methods will become an issue.
Something like FavoriteFolderFormatter with breadcrumb & other UI methods.

OsmAnd/src/net/osmand/plus/myplaces/favorites/FavoriteFolder.java
```diff
@@ -0,0 +1,141 @@
...
+	@NonNull
+	public List<FavouritePoint> getExactPoints() {
+		return group != null ? group.getPoints() : Collections.emptyList();
+	}
+
+	public int getExactPointsCount() {
+		return group != null ? group.getPoints().size() : 0;
+	}
+
+	public int getSubtreePointsCount() {
```
Chumva, line 117: Should we cache values from these methods for better performance, similar to trackFolder.getTotalTracksCount()?

It is used inside descriptions in view holders. If a user has many groups, it may lead to memory churn & lags

OsmAnd/src/net/osmand/plus/mapcontextmenu/editors/SelectFavouriteGroupBottomSheet.java
```diff
@@ -52,16 +79,121 @@ protected BaseBottomSheetItem createCategoriesListItem() {
...
+	protected String getCategoryDescription(@NonNull PointsGroup pointsGroup) {
+		FavoriteFolderNode node = favoriteFolderNodes.get(pointsGroup.getName());
+		if (node != null && node.isVirtual()) {
+			return String.valueOf(node.getSubtreePointsCount());
+		}
+		return super.getCategoryDescription(pointsGroup);
+	}
+
+	@NonNull
+	private String getMiddleTruncatedBreadcrumb(@Nullable String fullPath, @NonNull TextPaint textPaint, int availableWidth) {
```
Chumva: Should this also be part of FavoriteFolderPath/FavoriteFolderFormatter?

PR #25108, opened 2026-05-20
track labels click support

0xRe1nk0:


OsmAnd/src/net/osmand/plus/views/layers/MapSelectionHelper.java
```diff
@@ -309,6 +309,9 @@ private void selectObjectsFromOpenGl(@NonNull MapSelectionResult result, @NonNul
 				IBillboardMapSymbol billboardMapSymbol = getBillboardMapSymbol(mapSymbol);
 				if (billboardMapSymbol != null) {
 					objectLatLon = fetchBillboardSymbolLatLon(symbolInfo, billboardMapSymbol);
+					if (tryCollectGpxSplitLabel(result, billboardMapSymbol, objectLatLon)) {
```
Chumva: Why do we collect labels here? It looks very specific. Can we do something similar to collectObjectsFromLayers?

OsmAnd/src/net/osmand/plus/views/layers/GPXLayer.java
```diff
@@ -754,6 +762,38 @@ private void clearSelectedFilesSplits() {
...
+		return extraId;
+	}
+
+	private void collectSplitLabel(@NonNull MapSelectionResult result, @NonNull SelectedGpxPoint gpxPoint) {
+		for (SelectedMapObject selectedObject : result.getAllObjects()) {
+			if (selectedObject.object() == gpxPoint) {
+				return;
+			}
+		}
+		removeCollectedGpxTrackPoints(result);
```
Chumva, line 791: Will it work correctly with multiple labels?

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/SelectedGpxMenuBuilder.java
```diff
@@ -204,6 +211,20 @@ private GpxDisplayItem findDisplayItem() {
 		return null;
 	}
 
+	private boolean isSplitLabelSelection() {
```
Chumva: Should we move this to selectedGpxPoint class? Something like isSplitLabel?

Issue #24700, opened 2026-03-28
"pinned" status of a favorites folder is not synced correctly


Rowin63:
### Description

a "pinned" folder is synced from device A to the cloud, but not from the cloud to device B.
"unpinning" a folder again is not synced too, it remains "pinned" in the cloud.

Don´t know how to reproduce it. Doing one change on A and another change on B, both are synced to cloud - but not between the devices. It seems, only something like "the last change" is synced between the devices. Each device has one change, but the cloud has two.

It´s confusing. 
I think, it´s not possible to have changes done to different favorite folders on each device. Then one gets lost. 

Actual status:
Device A: one folder pinned
Cloud: this folder pinned
Device B: no folder pinned - but also no changes to sync 

Did one more change on device A (set an additional folder "pinned" > synced to cloud is correct, now two pinned folder > nothing to sync on device B (no cloud changes, still no pinned folder on device B)

### Steps to reproduce

.

### Actual result

.

### Expected result

.

### Your Environment (required)

A: ~5.3.0#7444
B: +5.3.1

Rowin63:
Next try. 
) I added a new favorite folder on device A and set it to "pinned" and "hidden" 
) synced A > cloud, all changes appeared correctly in the web 
) then I synced Cloud > B, cloud changes appeared to download. But after finishing this sync, no "pinned" folder appeared on B. Hidden folder appear correctly as hidden. 

BUT: the one previously created file from A, synced via cloud to B, appeared IMMEDIATELY after downloading from cloud to B as a new local change on B, AS DELETED! 

See the screenshot. It appears as uploaded, synced and deletet at the same time. 
I downloaded the file again from the cloud to solve this.

If it helps, this is the log after all this. 

![logcat.log.txt.txt](https://github.com/user-attachments/files/26323818/logcat.log.txt.txt)


<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/27c78c1c-e839-489d-aa02-25238f0644d2"  width="300" height="auto"/> </th>
  </tr>
</table>

Rowin63:
And the "last sync time" still is away from reality. But that issue was closed as "not planned".

yuriiurshuliak:
> OsmAnd~ 5.4.0#7482m, released: 2026-04-01

**The bug has been reproduced.**

[...]

PR #25138, opened 2026-05-26
Fix favorites group pin not synced on import

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/24700

PR #25149, opened 2026-05-29
Added extended version of Maidenhead-locator

dmpr0:
<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
  </tr>
</table>


OsmAnd-java/src/main/java/net/osmand/LocationConvert.java
```diff
@@ -18,6 +18,7 @@ public class LocationConvert {
 	public static final int MGRS_FORMAT = 5;
 	public static final int SWISS_GRID_FORMAT = 6;
 	public static final int SWISS_GRID_PLUS_FORMAT = 7;
+	public static final int MAIDENHEAD_FORMAT = 8;
```
Chumva, line 21: We should also add support for [CoordinatesBaseWidget.showFormattedCoordinates](https://github.com/osmandapp/Osmand/blob/4b325f26a8901904003cba29a06bbf10a804b73c/OsmAnd/src/net/osmand/plus/views/mapwidgets/widgets/CoordinatesBaseWidget.java#L149)

PR #25130, opened 2026-05-25
Turn direction is not the same as highlighted route #23787 (variant 1)

nnngrach:
[issue](https://github.com/osmandapp/OsmAnd/issues/23787)

[testing route](https://osmand.net/map/navigate/?start=48.156773,14.028524&end=48.169177,14.024692&profile=car#16/48.1630/14.0217)

---

Before: has turn left (wrong)

Screenshot 2026-05-22 at 11 59 58

---

After: moving forward (correct)

Screenshot 2026-05-25 at 12 09 26

---

All tests passed

Screenshot 2026-05-25 at 13 30 33

---

[related ios pr](https://github.com/osmandapp/OsmAnd-core-legacy/pull/328)

Issue #23787, opened 2025-11-02
Turn direction is not the same as highlighted route


kucai2000:
### Routing engine

- [x] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile

motorcycle, faster, no restrictions

### Start and end points

start 3.13384, 101.75615
end  3.13388, 101.76601
https://osmand.net/map/navigate/?start=3.078574,101.658836&profile=car#18/3.13379/101.756313

### Actual route

lane direction problem. See attached expected route image. It is instructing me to turn left on a straight route.  behavior also exists on osmand.net.  I've checked on OSM and everything seems to be correct.

### Expected route

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/a368536d-53fc-4436-af55-4c01ba22d531"  width="300" height="auto"/> </th>
  </tr>
</table>  

### Maps you used (online or offline)

- [x ] Offline maps offered within the OsmAnd app for download.  compiled map from mid September. 
- [ ] Online (tile / raster) maps 

### Your Environment

```
OsmAnd Version: 5.2.7 (i believe, or the pervious version. can't pinpoint the update time)
Android version: 10
Device model: Mi9

```

### Anything else?

For my delivery job, I have to make split second decisions just by glancing at the phone while riding. An inconsistent direction vs highlighted route could become dangerous really quick if I have to decipher it on the road. Anyways, always appreciate what you guys are doing. I'll be back with more possible issues with version 5.2.7 (currently it is ignoring slip ways at T junctions/roundabouts - i am investigating)

Zirochkabila:
Related to https://github.com/osmandapp/OsmAnd/issues/23622#issuecomment-3478073491

RZR-UA:
It may not be related to #23622 (not a duplicate)

https://test.osmand.net/map/navigate/?start=3.133840,101.756150&end=3.133880,101.766010&profile=car#17/3.13178/101.76388

Incorrect: Turn left `(+TL,C|C|C)` onto Jalan Pandan Indah and go 0.6 km. [Way 772479243](https://openstreetmap.org/way/772479243)

Possible reason: check the boolean leftSide (should be true for Singapore), and the case seems not connected with determineTurnsToMerge

Map used: Malaysia_asia_2.obf

To Check

- [ ] RouteResultPreparation.java#L1547

vshcherb:
[...]

PR #25152, opened 2026-05-29
 Turn-direction-is-not-the-same-as-highlighted-route-#23787 (variant 2)

nnngrach:
[issue](https://github.com/osmandapp/OsmAnd/issues/23787)
[old pr](https://github.com/osmandapp/OsmAnd/pull/25130)

[testing route](https://osmand.net/map/navigate/?start=48.156773,14.028524&end=48.169177,14.024692&profile=car#16/48.1630/14.0217)

---

Before: has turn left (wrong)

Screenshot 2026-05-22 at 11 59 58

---

After: moving forward (correct)

Screenshot 2026-05-29 at 19 30 45

---

All tests passed

Screenshot 2026-05-29 at 19 33 44

---

[related ios pr](https://github.com/osmandapp/OsmAnd-core-legacy/pull/328)


ivanPyrohivskyi:
I checked the situation.
As you can see in debug we got right direction if to consider [service](https://www.openstreetmap.org/way/471190175) and [primary](https://www.openstreetmap.org/way/792838957) as equivalent intersection. But is not !

PR #25179, opened 2026-06-02
Handle missing download notification before starting foreground service

dmvkosm:
## Summary
Stop DownloadService cleanly when the foreground notification cannot be built.

## Audit reference
Static code audit item **A6**.

## Test plan

### DIFF
--- a/OsmAnd/src/net/osmand/plus/download/DownloadService.java
+++ b/OsmAnd/src/net/osmand/plus/download/DownloadService.java
@@ -47,6 +47,12 @@ public class DownloadService extends Service {
 
 		NotificationHelper notificationHelper = app.getNotificationHelper();
 		Notification notification = notificationHelper.buildDownloadNotification();
+		if (notification == null) {
+			LOG.error("DownloadService could not be started because the notification is null.");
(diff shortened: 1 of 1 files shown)


OsmAnd/src/net/osmand/plus/download/DownloadService.java
```diff
@@ -47,6 +47,12 @@ public int onStartCommand(Intent intent, int flags, int startId) {
 
 		NotificationHelper notificationHelper = app.getNotificationHelper();
 		Notification notification = notificationHelper.buildDownloadNotification();
```
vshcherb, line 49: By source code NotificationCompatBuilder doesn't return null notification as I see
 Notification n = buildInternal();
        if (styleContentView != null) {
            n.contentView = styleContentView;
        } else if (mBuilderCompat.mContentView != null) {
            n.contentView = mBuilderCompat.mContentView;
        }

dmvkosm:
Thanks for the review — agreed, NotificationCompat.Builder does not return null here. Closing this approach; the audit item was based on a static false positive.

PR #25186, opened 2026-06-02
Normalize extreme coordinates without unbounded loops

dmvkosm:
## Summary
Use IEEEremainder-based longitude, latitude, and bearing normalization to avoid hangs on pathological coordinate values.

## Audit reference
Static code audit item **J2**.

## Test plan


Chumva:
Is this correct? Looks like we changed edge-case behaviour:
- `degreesDiff(0, 180)` used to return `180`, but now returns `-180`.
- `checkLatitude(270)` used to clamp to `MAX_LATITUDE`, but now clamps to `MIN_LATITUDE` (`-270` flips similarly).

PR #25188, opened 2026-06-02
Make GeneralRouter evalCache thread-safe

dmvkosm:
## Summary
Use ConcurrentHashMap for routing attribute caches to avoid corruption when routing runs concurrently.

## Audit reference
Static code audit item **R1**.

## Test plan


Chumva:
This improves the cache maps, but the lazy nested-map creation is still racy. Two threads can both see `rM == null`, create different maps, and one cache entry set can be lost. Since this is a cache it may only cause recomputation, but it should use `computeIfAbsent()`

dmvkosm:
Updated: nested eval cache maps now use computeIfAbsent().

PR #25195, opened 2026-06-03
Make weather offline forecast cache map thread-safe

dmvkosm:
## Summary
`OfflineForecastHelper.offlineForecastInfo` is a plain `HashMap` that is mutated from background `AsyncTask` work (cache size calculation, clearing, removing local forecasts) and also read/written on the UI thread. Concurrent access can throw `ConcurrentModificationException` or lose updates.

Switch it to `ConcurrentHashMap` and use `computeIfAbsent` so the get-or-create path is atomic.

## Audit reference
Static code audit item **W5**.

## Test plan
- [ ] Download/clear/remove offline weather forecasts for several regions while the cache-size calculation runs
- [ ] Confirm no crash and correct cache info

### DIFF
--- a/OsmAnd/src/net/osmand/plus/plugins/weather/OfflineForecastHelper.java
+++ b/OsmAnd/src/net/osmand/plus/plugins/weather/OfflineForecastHelper.java
@@ -695,8 +695,7 @@ public class OfflineForecastHelper implements ResetTotalWeatherCacheSizeListener
 		if (destinationTilesCount <= 0) {
 			return;
 		}
-		int downloadedTilesCount = getOfflineForecastProgressInfo(regionId);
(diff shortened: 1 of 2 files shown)


Chumva:
This fixes the outer `offlineForecastInfo` map creation race, but it does not make the cached info itself thread-safe. `OfflineForecastInfo` still wraps a plain `HashMap`, and progress updates still use a non-atomic read-modify-write (`getOfflineForecastProgressInfo()` then `setOfflineForecastProgressInfo(++value)`). So this helps, but the PR should either also make the per-region info/progress update atomic, or narrow the description to the outer cache map only

dmvkosm:
Updated: OfflineForecastInfo uses ConcurrentHashMap + AtomicInteger progress; download progress increments atomically.

PR #25196, opened 2026-06-03
Make WeatherWidget date formatters thread-safe and locale-stable

dmvkosm:
## Summary
`WeatherWidget` kept two `static SimpleDateFormat` instances that are shared between background obtain-value callbacks and the UI thread. `SimpleDateFormat` is not thread-safe, so this can corrupt formatting/parsing.

This change moves both formatters to `ThreadLocal`, pins the forecast DB file-name format to `Locale.US` so the generated file name always uses ASCII digits (avoiding mismatches in locales with non-Latin digits), and makes the displayed time format locale explicit.

## Audit reference
Static code audit item **W6**.

## Test plan
- [ ] Display weather widgets while forecasts are obtained on background threads
- [ ] Switch device to a locale with non-Latin digits and verify forecast DB file lookup still works

### DIFF
--- a/OsmAnd/src/net/osmand/plus/plugins/weather/widgets/WeatherWidget.java
+++ b/OsmAnd/src/net/osmand/plus/plugins/weather/widgets/WeatherWidget.java
@@ -36,6 +36,7 @@ import java.io.File;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.Locale;
 import java.util.TimeZone;
 
 public class WeatherWidget extends SimpleWidget {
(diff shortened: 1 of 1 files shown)


Chumva:
This feels more complex than needed for this file. The formatter use is limited to the dev toast/forecast DB file lookup path, and the async callback posts back to the UI thread before touching widget state, so `ThreadLocal` is probably unnecessary.

A smaller fix would be to create local `SimpleDateFormat` instances where needed: use `Locale.US` + UTC for the forecast DB filename, and create the display formatter at use time with the current locale. That avoids shared formatter thread-safety issues without keeping stale per-thread locale state

dmvkosm:
Updated: use per-call SimpleDateFormat helpers instead of ThreadLocal.

Issue #25076, opened 2026-05-15
Silent save failure when finishing track recording due to storage/permission blocks


onlynone:
### Description

Starting in October 2025 the app has been unable to stop and save the currently recording track when I click finish. I've contacted support about this multiple times, but their only solutions are to update the app, reboot my phone, etc. They finally said to post here. 

### Steps to reproduce

1. Start recording a track
2. Click 'Finish'

### Actual result

The track stays recording and doesn't finish.

### Expected result

The recording track should be finished and saved.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
Device : ASUS_AI2302
Brand : asus
Manufacturer : asus
Version : 15
App Version : OsmAnd+
Apk Version : 5.3.8 5308 
```

It happened starting on or before this version `5.2.13 5213`

Zirochkabila:
> OsmAnd~ 5.4.0#7766m, released: 2026-05-14

@onlynone Please attach a video showing the steps you took and the result - as you can see in the recording, everything is working correctly.
Also, your profile settings may be affecting this - please attach them here (they will be made public) or send them to our support team (if you wish to keep them private).

<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/1c9d5043-69e0-4381-8d90-634344c8d941"></video>
</details>

dmpr0:
Hi, @onlynone!
I have found your previous logs and screen recording. We are investigating it.

yuriiurshuliak:
> OsmAnd~ 5.4.0#7791m, released: 2026-05-19

Trip Recording does not finish after tapping **Finish**. The recording remains active and is still shown as **Currently recording track**, so the GPX track is not saved as expected.

Not reproduced with default settings.

We found the video and logcat you previously sent by email and reviewed them.

The video confirms the reported behavior: after tapping **Finish**, OsmAnd does not complete the active recording. In the logcat, we can see repeated UI/input-related warnings and service activity around the recording flow, but the main issue is that the **Finish** action does not change the recording state correctly.

We tested the same basic scenario on our side, but the recording stopped and saved normally with default settings.

<details><summary>Tested steps:</summary>

**Tested steps:**

1. Open OsmAnd.
2. Start recording a track using Trip Recording.
3. Open the Trip Recording control panel.
4. Tap **Finish**.
5. Check whether the recording is stopped and saved.

**Actual result:**
[...]

PR #25198, opened 2026-06-03
Fix SaveCurrentTrackTask GPX write and error reporting

dmvkosm:
## Summary
- Write each `GpxFile` from `collectRecordedData()` instead of always writing the task constructor instance
- Propagate write failures to `SaveGpxListener.onSaveGpxFinished(Exception)`
- Use `mkdirs()` for the target directory

## Related issues
#25076 (silent save failure when finishing track recording)

## Test plan
- [ ] Finish recording a track with storage available — file appears
- [ ] Simulate write failure (read-only dir) — listener receives error, path not cleared

### DIFF
--- a/OsmAnd/src/net/osmand/plus/myplaces/tracks/tasks/SaveCurrentTrackTask.java
+++ b/OsmAnd/src/net/osmand/plus/myplaces/tracks/tasks/SaveCurrentTrackTask.java
@@ -3,6 +3,7 @@ package net.osmand.plus.myplaces.tracks.tasks;
 import android.os.AsyncTask;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
 import net.osmand.IndexConstants;
 import net.osmand.plus.shared.SharedUtil;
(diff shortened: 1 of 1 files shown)


Chumva:
This fixes the core bug of writing the wrong `GpxFile`, but it breaks existing callback users. Several callers use the original `gpxFile.getPath()` after `onSaveGpxFinished()` to share/export the saved current track. With this change, only the collected per-file `gpxFile` gets its path set by `writeGpxFile()`, while the constructor `gpx` can remain empty, so callers may end up with `new File("")`.

The task should return/set the actual saved output file path for the callback flow, or the callers need to be updated to use the saved file(s). Also `if (error == null && gpx.getPath().isEmpty()) gpx.setPath("")` is a no-op now

dmvkosm:
Updated: SaveResult now sets gpx.setPath(savedPath) before onSaveGpxFinished(), then clears path only for cache-dir saves after the callback (share/export flow).

Chumva:
The error propagation is a good fix, but this version introduces a couple of regressions.

`collectRecordedData()` can return multiple date-split GPX files, while the task stores only the last `savedPath`. Export/share callers read only `gpx.getPath()`, so multi-day recordings may export only the last saved part.

Also, when the input `gpx` already has a path, this can replace it with `tracksDir/<date>.gpx`, changing the caller’s original path unexpectedly.

I think this task should either write the passed `GpxFile` once for export/share, or delegate full recorded-track persistence to `saveDataToGpx()` instead of partially duplicating it.

PR #25202, opened 2026-06-03
Defer live map updates on app startup to scheduled alarms

### LINKED ISSUE #25148: Live update starts automatically even if map updates are available and can't be stopped.
I just opened OsmAnd after connecting my device to the internet for the first time in a while, and it immediately started downloading live map updates.

---

First of all, I DO NOT want OsmAnd to automatically update maps or anything else ever. Live maps or otherwise. All updates or synchronisations should be explicit.

Sometimes I briefly connect my navigation device to the internet and if it then starts downloading and I end up disconnected again I end up with a partially updated map set. I've had it more than once that map versions end up out of sync and routing breaks across map boundaries.

Other times my internet connection for my nav device is via hotspotting on another phone, so mobile data. I do not want to incur the costs of downloading 100's of megabytes of data over a mobile connection..

Of course I understand other people do want automated updates and sync. So there should be an app-wide setting to disable or enable automated downloads/updates.

If this is already an option, I haven't been able to find it. . Please point it out.

---

Secondly, aren't live updates pointless if the underlying maps need to be updated? In this case, shouldn't the maps be updated first? Or at least, provide a way to interrupt the live update, so the base maps can be updated first, and then any live updates installed over the top.

Apologies if I'm misunderstanding the way this is supposed to work.

---

Thank you for a great product.

dmvkosm:
## Summary
- Stop calling `runLiveUpdate` from `AppInitializer.checkLiveUpdatesAlerts` on cold start
- Overdue maps only reschedule alarms; downloads run via `LiveUpdatesAlarmReceiver`

## Test plan
- [ ] Enable live updates for a map, restart app: no immediate download dialog/work
- [ ] Alarm still fires at scheduled time and update runs
- [ ] Manual "Update now" in Live Updates still works

Fixes #25148

### DIFF
--- a/OsmAnd/src/net/osmand/plus/AppInitializer.java
+++ b/OsmAnd/src/net/osmand/plus/AppInitializer.java
@@ -593,7 +593,7 @@ public class AppInitializer implements IProgress {
 				long lastCheck = preferenceLastSuccessfulUpdateCheck(fileName, settings).get();
 
 				if (System.currentTimeMillis() - lastCheck > updateFrequency.intervalMillis * 2) {
-					runLiveUpdate(app, fileName, false, null);
+					// Do not download on cold start; LiveUpdatesAlarmReceiver runs updates (#25148).
 					PendingIntent alarmIntent = getPendingIntent(app, fileName);
 					int timeOfDayOrd = preferenceTimeOfDayToUpdate(fileName, settings).get();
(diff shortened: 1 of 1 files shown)


dmvkosm:
Superseded by #25210 (reschedule alarms on startup without immediate download).

Issue #24279, opened 2026-01-11
OOM Tracks on display causes crashes when "creating a route"


Rowin63:
### Description

The longer the route becomes, the more jerky the app becomes - until crashing.
Five tries, five crashes in the last hour. Heres the last four logs.

[exception(4).log.txt](https://github.com/user-attachments/files/24554137/exception.4.log.txt)

[exception(5).log.txt](https://github.com/user-attachments/files/24554138/exception.5.log.txt)

[exception(6).log.txt](https://github.com/user-attachments/files/24554139/exception.6.log.txt)

[exception(7).log.txt](https://github.com/user-attachments/files/24554140/exception.7.log.txt)

### Steps to reproduce

create a route, at a certain lenght the app gets jerky and crashes

### Actual result

crash

### Expected result

no crash

### Your Environment (required)

Osmand ~5.3.0#6887

Samsung Tab S6 / A12

sonora:
Quick analysis: OsmAnd is using the entire available heap - these are hard out of memory errors. Likely primarily caused by
- huge GPX files,
- and GPX analysis.
- And like I proved in my PR "50 x faster GPX database setup..." the 'nearest ciyt' logic currently consumes the bulk of related resources. in that code.

Rowin63:
Created the same route on my other device (+5.2.13) (same content tracks /places / smart folder / tracks on display /... ) was no problem, but perhaps randomly. 

This route, about 260km:

![P Anergui to Kelaat 260km 2.gpx.txt](https://github.com/user-attachments/files/24554293/P.Anergui.to.Kelaat.260km.2.gpx.txt)

Rowin63:
> Quick analysis: OsmAnd is using the entire available heap - these are hard out of memory errors. Likely primarily caused by
> - huge GPX files,
> - and GPX analysis.
> - And like I proved in my PR "50 x faster GPX database setup..." the 'nearest ciyt' logic currently consumes the bulk of related resources. in that code.

I don't understand all of this, but you're probably right 😁

Rowin63:
Osmands behaviour isn't really predictable. One time it is possible to create a 200km route, the next time, a few minutes later, the app crashes already after 50 or 60km while creating a route.

Rowin63:
It seems, this crashes have been caused by the amount of "tracks on display". Taking them all "off display" makes route planning smooth again.

80 tracks on display: crashes.
00 tracks on display: everything's fine.

Solution: somewhere in between... 🤔

[...]

PR #25203, opened 2026-06-03
Skip eager general track merge when restoring huge GPX files

### LINKED ISSUE #24420: Crashing OOM Track 200K points
### Description

[exception (5).log](https://github.com/user-attachments/files/25099403/exception.5.log)
[exception (6).log](https://github.com/user-attachments/files/25099404/exception.6.log)
[exception (7).log](https://github.com/user-attachments/files/25099402/exception.7.log)

Nightly. 

### Steps to reproduce

. 

### Actual result

. 

### Expected result

. 

```
OsmAnd Version:
Android version:
Device model:
Crash-Logs: ?
```

dmvkosm:
## Summary
- Restore selected GPX with `loadGpxFile(..., addGeneralTrack=false)`
- Build general track only when total track points ≤ 100k

## Test plan
- [ ] App starts with a large GPX selected: no OOM
- [ ] Small multi-segment GPX still shows merged track on map

Fixes #24420, #24279

### DIFF
--- a/OsmAnd/src/net/osmand/plus/track/helpers/GpxSelectionHelper.java
+++ b/OsmAnd/src/net/osmand/plus/track/helpers/GpxSelectionHelper.java
@@ -37,6 +37,7 @@ import net.osmand.shared.gpx.GpxTrackAnalysis;
 import net.osmand.shared.gpx.GpxUtilities;
 import net.osmand.shared.gpx.GpxUtilities.PointsGroup;
 import net.osmand.shared.gpx.TrackItem;
+import net.osmand.shared.gpx.primitives.TrkSegment;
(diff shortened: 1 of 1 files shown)


Chumva:
This fixes the startup OOM path, but it looks incomplete: large GPX files can still build the general track lazily later.

`getGeneralTrack()` / `getGeneralSegment()` still call `buildGeneralSegment()`, so the same OOM can happen if join gaps is enabled, track options are opened, or analyze-by-intervals is used. For huge GPX files we should also avoid those lazy calls, e.g. use `hasGeneralTrack()` in UI checks and disable/skip join-general-track behavior when point count exceeds the threshold.

PR #25213, opened 2026-06-03
Handle missing download notification before starting DownloadService (A6)

dmvkosm:
## Summary
- Return START_NOT_STICKY when buildDownloadNotification() returns null
- Avoid ForegroundServiceDidNotStartInTimeException / NPE on FGS start

## Test plan
- [ ] Start map download with notifications disabled or broken channel
- [ ] Service stops cleanly instead of crashing

Made with [Cursor](https://cursor.com)

### DIFF
--- a/OsmAnd/src/net/osmand/plus/download/DownloadService.java
+++ b/OsmAnd/src/net/osmand/plus/download/DownloadService.java
@@ -47,6 +47,12 @@ public class DownloadService extends Service {
 
 		NotificationHelper notificationHelper = app.getNotificationHelper();
 		Notification notification = notificationHelper.buildDownloadNotification();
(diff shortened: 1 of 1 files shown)


Chumva:
I don’t think this branch is useful in the current code. `buildDownloadNotification()` is non-null and `DownloadNotification.buildNotification()` always returns a builder, so `notification == null` looks unreachable.

If the real issue is `startForeground()` failing, the existing `try/catch` already covers that. If notification construction can throw, this patch does not cover it because `buildDownloadNotification()` is still called before the `try`

dmvkosm:
Closing per review: buildDownloadNotification() is non-null in current code; startForeground failures are already handled by try/catch.

PR #25215, opened 2026-06-03
Mark NavigationService usage bitmask volatile

dmvkosm:
## Summary
- usedBy is updated from binder and service threads; volatile ensures visibility

## Test plan
- [ ] GPX recording + navigation concurrently; service stops only when all usages cleared


Chumva:
This improves cross-thread visibility, but it does not make the bitmask updates thread-safe.
If usedBy can be modified from multiple threads, please use AtomicInteger or synchronize the
state transitions instead of only marking the field volatile

dmvkosm:
Updated: replaced volatile int with AtomicInteger for bitmask updates.

PR #25216, opened 2026-06-03
Cap wiki POI online loading wait at 30 seconds

dmvkosm:
## Summary
- Avoid infinite wait when ExplorePlacesProvider stays in loading state

## Test plan
- [ ] Wiki POI layer with slow/stuck online load returns within ~30s


Chumva:
This avoids an infinite wait, but it is a blunt fix. The loop still checks global `explorePlacesProvider.isLoading()`, so the 30s timeout can be hit because unrelated wiki tiles are loading, not because this requested rect is stuck.

If we add a cap here, it should at least be rect-scoped via `isLoadingRect(rect)`, and we should make sure the UI refreshes when the provider eventually finishes. Otherwise, this can return empty/partial results while the relevant data is still loading.

dmvkosm:
Updated: wait loop now uses isLoadingRect(rect) instead of global isLoading().

PR #25217, opened 2026-06-03
Synchronize PoiUIFilter search state and guard index reads

dmvkosm:
## Summary
- Lock currentSearchResult and distanceInd across UI and background search
- Catch RuntimeException from offline amenity index reads in searchAmenitiesInternal

## Test plan
- [ ] POI search + filter rename while search runs
- [ ] Corrupt map index: empty results, no crash

### DIFF
--- a/OsmAnd/src/net/osmand/plus/poi/PoiUIFilter.java
+++ b/OsmAnd/src/net/osmand/plus/poi/PoiUIFilter.java
@@ -103,6 +103,7 @@ public class PoiUIFilter implements Comparable<PoiUIFilter>, CustomSearchPoiFilt
 	SearchPoiAdditionalFilter additionalFilter;
 
 	private final PoiUIFilterDataProvider dataProvider;
+	private final Object searchLock = new Object();
 
 	// constructor for standard filters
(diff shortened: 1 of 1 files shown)


Chumva:
The current lock only covers base PoiUIFilter, while subclasses still mutate currentSearchResult, so it doesn’t fully solve the race. A simpler copy-on-write approach may be enough for the cache: publish a fresh/volatile result list, never mutate it in place, and calculate distance index locally per search. Only serialize the actual index access if there is evidence that provider/search internals are not thread-safe

dmvkosm:
Updated: replaced synchronized lock with volatile copy-on-write cache publication.

Chumva:
This version is better, but I don't think this is fully safe to merge as "copy-on-write/thread-safe" yet. `currentSearchResult` is still published before callers sort/trim the same list; `SearchByNameFilter` and `NominatimPoiFilter` bypass the new base-method guard and publish mutable lists early; `distanceInd++` is still a non-atomic volatile RMW; and the broad `RuntimeException` catch can hide real POI search bugs as empty results.

Suggest publishing `currentSearchResult` only after sort/trim, giving the subclasses equivalent publication behavior, and narrowing the catch to the specific expected exception.

PR #25220, opened 2026-06-04
Coordinates UI

### LINKED ISSUE #17311: Coordinates widget: reuqest to add EPSG 2039 or 6991
# 🚀 feature request

### Description

Much publicly available local geographical data info, as well as the vast majority of historical data is available only in EPSG 2039 (ITM). Therefore, when navigating in a field trying to find something according to its ITM coordinates , It's sometimes easier to just have a printed map at hand. I would like to use OSMand for such navigation  (and recording new data, possibly)

Add the CRS: EPSG 2039 (ITM 1993) (or [EPSG 6991](https://epsg.io/6991) ) to the profile units configuration options.

There's a lame paid close source option which I'd rather not use. The other option is to to pre-transform data to the approproate UTM strips and the use that. The problem, in my case - is that it's often not exact enough

0xRe1nk0:


OsmAnd/src/net/osmand/plus/AppVersionUpgradeOnInit.java
```diff
@@ -1166,6 +1171,13 @@ private void migrateAudioVideoNotesToFavorites() {
 		}
 	}
 
+	private void migrateCoordinateFormatPreferences(@NonNull OsmandSettings settings) {
+		CoordinateFormatPreferences coordinateFormatPreferences = new CoordinateFormatPreferences(settings);
+		for (ApplicationMode appMode : ApplicationMode.allPossibleValues()) {
+			coordinateFormatPreferences.migrateFromLegacyFormat(appMode, settings.COORDINATES_FORMAT.getModeValue(appMode));
```
Chumva: migrateFromLegacyFormat & getLegacyPreferredIds could be part of AppVersionUpgradeOnInit, we don't use them in other places, only during migration

OsmAnd/src/net/osmand/plus/settings/coordinates/CoordinateFormatSelectorBottomSheet.kt
```diff
@@ -0,0 +1,296 @@
...
+
+		@JvmStatic
+		@JvmOverloads
+		fun showInstance(
+			fragmentManager: FragmentManager,
+			requestKey: String = DEFAULT_REQUEST_KEY,
+			selectedFormatId: String? = null,
+			showSelectOtherFormat: Boolean = true
+		) {
+			if (AndroidUtils.isFragmentCanBeAdded(fragmentManager, TAG)) {
```
Chumva, line 273: Should we pass here also appMode? CoordinatesWidgetInfoFragment could use appMode that [was passed to it](https://github.com/osmandapp/Osmand/blob/f8a0d4e8b90cc94cb63f3332b8b9349cd166567b/OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/settings/WidgetInfoBaseFragment.java#L427), it may be different from currentAppMode

OsmAnd/src/net/osmand/plus/settings/backend/OsmandSettings.java
```diff
@@ -3346,6 +3347,13 @@ public <T, P extends CommonPreference<T>> P getLayoutPreference(@NonNull P baseP
 	public final OsmandPreference<String> CONTRIBUTION_INSTALL_APP_DATE = new StringPreference(this, "CONTRIBUTION_INSTALL_APP_DATE", null).makeGlobal();
 
 	public final OsmandPreference<Integer> COORDINATES_FORMAT = new IntPreference(this, "coordinates_format", PointDescription.FORMAT_DEGREES).makeProfile();
+	public final ListStringPreference PREFERRED_COORDINATE_FORMAT_IDS = (ListStringPreference)
```
Chumva: Should we add migration from the old COORDINATES_FORMAT during import?

OsmAnd/src/net/osmand/plus/OsmandApplication.java
```diff
@@ -681,6 +683,14 @@ public WeatherHelper getWeatherHelper() {
 		return weatherHelper;
 	}
 
+	@NonNull
+	public synchronized CoordinateFormatHelper getCoordinateFormatHelper() {
```
Chumva: Should this be consistent with other plugins?

OsmAnd/src/net/osmand/plus/settings/coordinates/CoordinateFormatPreferences.kt
```diff
@@ -0,0 +1,108 @@
+package net.osmand.plus.settings.coordinates
+
+import net.osmand.plus.settings.backend.ApplicationMode
+import net.osmand.plus.settings.backend.OsmandSettings
+import java.util.Collections
+import java.util.LinkedHashSet
+
+class CoordinateFormatPreferences(private val settings: OsmandSettings) {
```
Chumva: Maybe CoordinateFormatSettingsStorage would fit better here? It could be similar to ImpassableRoadsStorage / SettingsMapPointsStorage: owned as a single instance by OsmandSettings, with PREFERRED_COORDINATE_FORMAT_IDS and RECENTLY_ADDED_COORDINATE_FORMAT_IDS created in its constructor

alex-dev-neo (approved):

Issue #25065, opened 2026-05-12
Common words review (region names), main word for houses


vshcherb:
From https://github.com/osmandapp/OsmAnd/issues/24229

We need to effectively select main word from user input to search in the file.

For example, "Dosse 1A hauptstrasse" - starting from "hauptstrasse" gives > 1000 results and 20 seconds search, starting from "dosse" gives < 100 resutls and 2-3 seconds search.

### New index
@Gateway07 
- [x] Change structure for suffixes:
    - [x]  Fix name index! Remove words 'neighbourhood', 'hamlet', 'village' and numbers in Boundary / name index!!!
    - [x] For frequent word add all possible suffixes
    - [x] Add suffix numbers for all words?
    - [ ] Add postcodes for frequent names
    - [x] Adapt search algorithm to use this heuristic for "Calle 12, School 53"
    - [ ] Smart algorithm to reduce longest suffixes array??
- [x] Create test cases in search test based on analysis to find **slow searches** and missing !
- [x] Index all common words if name doesn't have any "47 West Avenue" (index on west & avenue)
- [x] Add name stats while creating maps

### Incomplete frequent street
- [x] 1. Fix `Dosse 1A hauptstr` - 70 sec - we can check that `hauptstr` subword of `hauptstrasse` (frequent word), but `Dosse` is complete.

### House name
- [ ] Prefix for houses longer than street name - add building name to index! All in the world too many OTHERBLD (14,836,350),  however FullSearchStats.BUILD_NAME errors specific houses only < 10K.

* `11-nuon leons` 
* 

<details>
<summary>Examples</summary>
```
query, lat, lon, obf, address
10a-wsch leons,53.1483919,5.6802964,4,Leons Leons 10a-WSCH
11-nuon leons,53.1519822,5.6775284,4,Leons Leons 11-NUON
14-traf rogat,52.6885709,6.2644386,2,Rogat Rogat 14-TRAF
27-nuon húns,53.1600754,5.6661987,4,Húns Húns 27-NUON
fort 15-ts,52.656459,6.3773274,2,Veeningen Fort 15-ts
húns 39a-mln,53.1555851,5.6570148,4,Húns Húns 39a-MLN
9f-gara waver,52.2531851,4.9107814,11,Waver Waver 9F-GARA
if remove latter suffix then search is showing result.
```
</details>
  
  * `Pulverturm  Rottweil 48.16965,8.62789`
[...]

PR #25229, opened 2026-06-05
Regions search

ivanPyrohivskyi:
Issue https://github.com/osmandapp/OsmAnd/issues/25065
Related https://github.com/osmandapp/OsmAnd-resources/pull/1388

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
+++ b/OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
@@ -9,6 +9,7 @@ import static net.osmand.binary.ObfConstants.isTagIndexedForSearchAsName;
 import static net.osmand.data.Amenity.POPULATION;
 import static net.osmand.osm.MapPoiTypes.OSM_WIKI_CATEGORY;
 import static net.osmand.osm.MapPoiTypes.WIKI_PLACE;
+import static net.osmand.search.core.ObjectType.BOUNDARY;
 import static net.osmand.search.core.ObjectType.POI;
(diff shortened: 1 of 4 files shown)


OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -734,6 +736,29 @@ public boolean isCancelled() {
 					}
 					lastRegionPriority = regionPriority;
 				}
+
+				int cnt = resultMatcher.getCount() - lastResultCount;
+				if (cnt == 0 && !longDistance && !phrase.hasObjectType(BOUNDARY)) {
```
vshcherb, line 741: 1. Very limited to cnt= 0 - restriction is not clear 
2. Quite a lot specific code which is exactly for 
```
 else if (res.objectType == ObjectType.BOUNDARY) {
							// require exact matching to speed up
```

I Suggest to add to offlineIterator and test crner cases
```
int lastResultCount = resultMatcher.getCount();
				BinaryMapIndexReader regionsOcbf = longDistance ? ... : null;
				while (offlineIterator.hasNext() && wordToSearch.length() > 0) {
					BinaryMapIndexReader r = regionsOcbf != null ? regionsOcbf : offlineIterator.next();
					regionsOcbf = null;
```

Anyway question is in stats how many extra cycles will be for certain cases. What's the worst case that will be added.

vshcherb:
There is no sense to store all regions in Common words frequent as high as now
1. It's incorrect for double region words like "city", "valley"
2. Questionnable what is bad if state name like "Pennsylvania" will be selected first.

If we assume that region names  are negatively impacting performance or results it should be checked

PR #25251, opened 2026-06-11
Fix media favorite creation for new avnotes

Chumva:


sonora:
@Chumva There's a potential problem with this patch: With the new build I created a new favorite, tapped it and added a photo via the camera. After that, my existing favs (all classic, no media) where gone, and there was only the new one in a folder called Media.

Reverting back to my June 9 build, restoring the favs from backup, and repeating the experiment showed no such issue.

I can only suspect that perhaps attachedMediaDataHelper.convertRecordingsToFavorites(recordings) drops all existing favorites from the in-memory list under certain circumstances, perhaps if the favorites index is too slow or temporarily unavailable.

PR #25262, opened 2026-06-14
Fix #25010 – Deleted favorites reappear after app restart

### LINKED ISSUE #25010: Deleting "favorites" from map quickly causes them to reappear
### Description

I can't delete "_Favorites_" in OsmAnd.

### Steps to reproduce

1. Delete a favorite: tap > Edit favorite > Trash bin icon > "Yes" **OR**  tap > Edit favorite > Scroll down and click "Delete" > "Yes" 
2. Move the map around, then get back to the previous location **OR** close and reopen the app
3. the favorites are still there.

If I just move the map around, the favorites are not clickable, like ghosts
if I close and reopen the app, I can re-delete them, in an infinite loop

I tried editing the favorite, like editing the name, and then delete, but no results
I tried leaving the OsmAnd Cloud feature, but no results

https://github.com/user-attachments/assets/3d345dd3-b46a-4384-9dd3-2fc8824212d9

### Actual result

Favorites keep re-appearing everytime

### Expected result

Favorites should be gone forever

```
OsmAnd Version: OsmAnd+ 5.3.8, released: 2026-04-23
Android version: 16
Device model: Google Pixel 9
```

nazar-kutz:
## Problem
Deleting a favorite point and immediately closing the app caused 
the point to reappear after restart. This happened because deletion 
was applied to in-memory state immediately, but the actual GPX file 
save was performed asynchronously. If the process was killed before 
the async task completed, the GPX files still contained the deleted 
points. On next launch, loadFavorites() would merge them back from 
the external files.

The issue was especially noticeable with large favorites collections 
(10,000+ points) where the save task could take 30+ seconds.

## Root cause
saveCurrentPointsIntoFile(true) queues an AsyncTask on a 
singleThreadExecutor. Android can kill the process before this task 
completes, leaving GPX files in the old state.

## Fix
Introduced a tombstone mechanism via PendingFavoriteDeletions — 
a lightweight append-only file (favorites_deletions.tmp) that is 
written synchronously at the moment of deletion, before the async 
save is queued.

On the next loadFavorites() call, pending deletions are applied 
to both internal and external group maps before merge, ensuring 
deleted points and groups are filtered out regardless of whether 
the previous async save completed.

The tombstone file is cleared inside SaveFavoritesTask.saveAllGroups() 
only after all GPX writes succeed, so if a save fails mid-way the 
tombstone is preserved for the next launch.

## Changes
- PendingFavoriteDeletions.kt — new class, stores sets of deleted 
  point keys and group names, handles serialization
- FavouritesFileHelper.java — added tombstone file read/write/clear
- FavouritesHelper.java — apply pending deletions in loadFavorites(), 
  write tombstone in delete() and deleteFavourite()
- SaveFavoritesTask.java — clear tombstone after successful save
- IndexConstants.kt — added TMP_FILE_EXT constant

## Testing
Reproduced on OsmAnd+ 5.3.8 / 5.4.0 with 23,000+ favorites. 
Verified fix by deleting a favorite and force-stopping the app 
immediately — point no longer reappears after restart.


Chumva:
The tombstone journal is a reasonable short-term mitigation for deletes, but it does not address the root problem: every favourite edit still triggers a full GPX read/merge/rewrite/backup pipeline, so persistence remains slow and killable. It also only protects deletes, while adds/edits can still be lost before the async save completes.

PR #25273, opened 2026-06-17
Coordinates review fix

### LINKED ISSUE #17311: Coordinates widget: reuqest to add EPSG 2039 or 6991

0xRe1nk0:


OsmAnd/src/net/osmand/plus/mapcontextmenu/MenuBuilder.java
```diff
@@ -700,14 +700,19 @@ private void buildCoordinatesRow(@NonNull View view) {
 		double latitude = latLon.getLatitude();
 		double longitude = latLon.getLongitude();
 		List<FormattedCoordinate> coordinateRows = PointDescription.getPreferredLocationData(mapActivity, latitude, longitude);
+		Map<Integer, String> legacyLocationData = PointDescription.getLocationData(mapActivity, latitude, longitude, true);
```
Chumva: We load map only for two formats. Can we simplify/optimise this?

PR #25288, opened 2026-06-19
Fix widget copying for separate layouts

Zirochkabila:
## Summary
- fix Configure screen profile copy to preserve the active separate layout mode
- ensure left/right panel widgets are copied from portrait/landscape layout settings instead of shared layout settings
- add regression tests for portrait, landscape, and shared layout copy mode resolution

## Testing
- `git diff --check`
- Android compile was not rerun because current master is blocked by unrelated `EpsgCoordinateTransformer.kt` / `net.osmand.core.jni.CoordinateTransformer` resolution issue.


Chumva:
This should be fixed here - https://github.com/osmandapp/OsmAnd/pull/25293

PR #25328, opened 2026-06-29
Optimize serialization and favorites saving

alex-dev-neo:
## 1. XML/GPX serialization buffering

In OsmAnd-shared/src/androidMain/kotlin/net/osmand/shared/xml/XmlSerializer.kt:12, both output paths now use a `BufferedWriter` with a 32K-character buffer:

```text
KXmlSerializer
    → BufferedWriter (32K characters)
        → UTF-8 writer / Okio Sink
```

Previously, KXml could send many tiny writes—sometimes individual characters—to the UTF-8 encoder. Now they accumulate and are encoded/written in approximately 32K-character chunks.

Memory cost is approximately:

- 32K Java/Kotlin characters × 2 bytes = 64 KB.
- Plus small Writer and Okio buffers.
- No complete GPX document is accumulated.

Ownership remains:

- File output is flushed and closed by `XmlSerializer`.
- Caller-provided `Sink` is flushed but not closed.

`SinkStringWriter` now extends `Writer` directly and uses Okio’s ranged `writeUtf8(str, start, end)`, eliminating `substring()` allocations.

The same 32K buffering was added to legacy JVM/server `GPXUtilities.writeGpxFile()`.

## 2. KML conversion

Previously, KML conversion created several complete representations:

```text
KML → GPX Okio Buffer → String → UTF-8 ByteArray → second Buffer → parser
```

Now OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/helper/ImportGpx.kt:43 does:

```text
KML → one GPX Okio Buffer → parser
```

The buffer size is recorded before parsing because parsing consumes the buffer.

This removes:

- Complete GPX `String`.
- Complete UTF-8 `ByteArray`.
- Second GPX `Buffer`.

One complete generated GPX buffer still exists because it becomes the parser input.

## 3. Settings JSON export

Previously:

```java
json.toString(2).getBytes("UTF-8")
```

This held both:

- The complete JSON `String`.
- A second complete UTF-8 `byte[]`.

Now OsmAnd/src/net/osmand/plus/settings/backend/backup/SettingsHelper.java:125 writes the JSON string through:

```text
JSONObject
    → formatted String
        → BufferedWriter (32K characters)
            → destination stream
```

`JSONObject` and the formatted `String` still exist. Only the additional full UTF-8 byte-array copy was removed.

When progress is needed:

- `Utf8.size()` calculates byte length without allocating a byte array.
- `ProgressOutputStream` counts bytes actually written.
- `flush()` now reaches the underlying stream.
- The caller’s ZIP/network stream is not closed.

## 4. Favorites coalescing

The coordinator state is in OsmAnd/src/net/osmand/plus/myplaces/favorites/FavouritesFileHelper.java:55:

```java
boolean saveRunning;
SaveBatch pendingSave;
```

All save requests enter `enqueueSave()`.

### First request

When nothing is running:

1. `saveRunning` becomes `true`.
2. A `SaveBatch` is created.
3. `SaveFavoritesTask` starts on the existing single-thread executor.

### Requests arriving while saving

Only one `pendingSave` is retained:

- First additional request creates it.
- Further requests merge into it.
- No additional `SaveFavoritesTask` objects enter the executor queue.

For ten rapid operations:

```text
Request 1 → running save
Requests 2–10 → one merged pending SaveBatch
Request 1 finishes → one follow-up save
```

Changes arriving during the follow-up save can create one new pending batch, so the invariant remains one running save plus one pending save.

## 5. What `SaveBatch` contains

`SaveBatch` is defined at OsmAnd/src/net/osmand/plus/myplaces/favorites/FavouritesFileHelper.java:400.

It contains:

- A shallow `List<FavoriteGroup>`.
- `saveAllGroups`.
- Listeners waiting for that save.
- `CountDownLatch` objects for synchronous callers.

It does not copy points or GPX data:

```java
this.groups = new ArrayList<>(groups);
```

This copies only group references.

Merge behavior:

- A newer full-save request replaces the pending group list.
- A selected-group request replaces/adds those groups by name.
- Once a batch becomes a full save, it remains a full save.
- Listeners and synchronous completions are appended in request order.

Its RAM usage is therefore small, although listeners grow with the number of coalesced requests. There are no repeated 10 MB GPX snapshots.

The temporary `GpxFile` created by `asGpxFile()` still exists during actual serialization. That behavior already existed and was not redesigned.

## 6. Synchronous saves

`saveFavoritesIntoFileSync()` creates a `CountDownLatch`, submits through the same coordinator, then waits.

Therefore a synchronous request:

- Does not bypass the bounded queue.
- Can become part of the pending batch.
- Returns after its physical save attempt finishes.

The latch is released after both success and failure to prevent the caller from waiting forever.

## 7. Save task execution

For a full save, OsmAnd/src/net/osmand/plus/myplaces/favorites/SaveFavoritesTask.java:89 performs:

1. Load the previous internal GPX to identify deleted points/groups.
2. Atomically save the internal GPX.
3. Merge/save external group files atomically.
4. Delete orphan group files only after replacement writes succeed.
5. Create the ZIP backup from the committed internal file.

Selected-group saves only write their external group GPX files.

Backup failure remains best-effort: it is logged but does not invalidate successfully committed GPX files.

## 8. AtomicFile operation

`saveFileAtomic()` is at OsmAnd/src/net/osmand/plus/myplaces/favorites/FavouritesFileHelper.java:281.

Its flow is:

```text
Previous GPX remains available
        ↓
AtomicFile.startWrite()
        ↓
Serialize GPX into temporary/new storage
        ↓
Success?
 ├─ yes → finishWrite(): flush/sync and commit replacement
 └─ no  → failWrite(): discard failed write and restore previous GPX
```

It is used for:

- Internal Favorites GPX.
- Every managed external group GPX.

AtomicFile mainly consumes temporary disk space, potentially close to one additional GPX file. It does not hold another complete GPX in RAM.

## 9. Atomic recovery

Android can leave `.new` or `.bak` files if the process dies.

Before loading internal Favorites, `AtomicFile.openRead()` is called. It restores the previous committed base file and removes stale temporary state.

External Favorites discovery also:

1. Finds `.new` and `.bak` remnants.
2. Derives their base GPX filename.
3. Runs AtomicFile recovery.
4. Returns only actual `.gpx` files to the GPX loader.

Temporary files therefore cannot accidentally be parsed as normal Favorites files.

## 10. Completion behavior

After a coordinated task finishes:

- Synchronous latches are released.
- On success, listeners are posted to the UI thread in request order.
- The pending batch, if present, starts.
- Otherwise `saveRunning` becomes `false`.

One current limitation: if a save fails, its listeners are not transferred to a later pending batch. A later batch may successfully contain the same changes, but those earlier listeners will not fire. Strict “listener completes when any covering save commits” semantics would require correcting this.

## Guarantees and limitations

Guaranteed:

- The save-task queue cannot grow without bound.
- Rapid edits normally result in one current rewrite and one follow-up rewrite.
- Interrupted writes preserve the previous committed GPX.
- Temporary AtomicFile remnants are recovered.
- Serialization uses small fixed buffers rather than document-sized buffers.

Not guaranteed:

- The newest in-flight edit survives process death.
- Internal and external GPX files commit as one transaction.
- Automatic retry when a save fails with no pending request.
- Protection if both the committed file and filesystem-level backup are physically corrupted.
- Zero GPX model allocation during serialization.

This remains the deliberately minimal solution: faster serialization, bounded saves, and valid previous files—without a journal, database, or duplicate persistent Favorites model.

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/ProgressOutputStream.java
+++ b/OsmAnd-java/src/main/java/net/osmand/ProgressOutputStream.java
@@ -45,4 +45,9 @@ public class ProgressOutputStream extends OutputStream {
 		this.out.write(b);
 		submitProgress(1);
 	}
+
+	@Override
+	public void flush() throws IOException {
+		out.flush();
+	}
 }
--- a/OsmAnd-java/src/main/java/net/osmand/gpx/GPXUtilities.java
+++ b/OsmAnd-java/src/main/java/net/osmand/gpx/GPXUtilities.java
@@ -27,6 +27,7 @@ import org.xmlpull.v1.XmlPullParserException;
 import org.xmlpull.v1.XmlSerializer;
 
 import java.io.BufferedInputStream;
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
@@ -993,7 +994,8 @@ public class GPXUtilities {
 			if (fout.getParentFile() != null) {
 				fout.getParentFile().mkdirs();
 			}
-			output = new OutputStreamWriter(new FileOutputStream(fout), "UTF-8"); //$NON-NLS-1$
+			output = new BufferedWriter(
+					new OutputStreamWriter(new FileOutputStream(fout), "UTF-8"), 32 * 1024); //$NON-NLS-1$
 			if (Algorithms.isEmpty(file.path)) {
 				file.path = fout.getAbsolutePath();
 			}
--- a/OsmAnd-shared/src/androidMain/kotlin/net/osmand/shared/io/SinkOutputStream.kt
+++ b/OsmAnd-shared/src/androidMain/kotlin/net/osmand/shared/io/SinkOutputStream.kt
@@ -2,12 +2,18 @@ package net.osmand.shared.io
 
 import okio.BufferedSink
 import okio.IOException
-import okio.Okio
 import okio.Sink
 import okio.buffer
 import java.io.OutputStream
 
-class SinkOutputStream(sink: Sink) : OutputStream() {
+/**
+ * [OutputStream] over an okio [Sink].
+ *
+ * With [closeSink] = false, [close] only flushes — for caller-owned sinks that must
+ * stay open (e.g. AtomicFile streams closed by finishWrite()).
+ */
+class SinkOutputStream(sink: Sink, private val closeSink: Boolean = true) : OutputStream() {
+
 	private val bufferedSink: BufferedSink = sink.buffer()
 
 	@Throws(IOException::class)
@@ -25,8 +31,19 @@ class SinkOutputStream(sink: Sink) : OutputStream() {
 		bufferedSink.flush()
 	}
 
+	/**
+	 * Closes the underlying sink, or — when [closeSink] is false — only flushes it.
+	 *
+	 * The flush-only mode deliberately deviates from the [OutputStream.close] contract:
+	 * all buffered bytes are pushed through, but the sink stays open and the caller
+	 * remains responsible for closing it.
+	 */
 	@Throws(IOException::class)
 	override fun close() {
-		bufferedSink.close()
+		if (closeSink) {
+			bufferedSink.close()
+		} else {
+			flush()
+		}
 	}
 }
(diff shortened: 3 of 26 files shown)


nazar-kutz:
I added a small follow-up adjustment related to stale point key handling during save: https://github.com/osmandapp/OsmAnd/pull/25406.

The stricter `cleanupOrphanedGroupFiles()` behavior is kept unchanged, since a failed orphaned external file deletion can potentially restore a deleted group on the next load.

The only change is that the save task now relies on the deletion journal for deleted point keys instead of reading the internal favorites GPX file again. Since deletions are already recorded in the journal before the async save completes, this keeps the same intended recovery flow while avoiding the extra internal file read during save.

PR #25338, opened 2026-06-30
Routing UI 22987

### LINKED ISSUE #22987: Routing - do not require interaction on route start for inconsistent maps
Summarizing leftovers from #22178 as follows:

When calculating a route, depending on their particular offline map constellation and destination, users may today receive a series of dialogs **they are forced to interact with for route calculation to start at all**. The dialogs inform them about "_outdated maps_", requires them to tap "_Details_" twice, offer to "_download maps now_", and finally you can select "_Use downloaded maps_".

All of this is a nuisance, even a no-go, at a time a user in the field needs a route! (The need to suddenly interact with the app in this fashion has even been observed in route-recalculation cases if near map borders - the app suddenly paused navigating and expected user interaction.)

- [ ] Suggestion is that this situation should **not require any user interaction at all**.
- [ ] Instead, if encountered, the app should resolve the situation independently (using the existing maps) without user interaction. As a last resort it could e.g. temporarily for this route fall back to the A-Star 2 Phase routing mechanism, which has no strict requirements on map version consistency.
- [ ] User should be informed by a toast about any fallback, that e.g. a slower and less performant routing mechanism is temporarily used because some maps needed for this route are not recent/consistent enough. But the user should not be forced to interact by tapping any buttons.

0xRe1nk0:


OsmAnd/src/net/osmand/plus/routepreparationmenu/RequiredMapsFragment.java
```diff
@@ -189,21 +200,22 @@ private void updateToolbarMenu() {
...
 	@NonNull
-	private View createItemView(@NonNull DownloadItem downloadItem, boolean showBottomDivider) {
+	private View createItemView(@NonNull RequiredMapItem mapItem, boolean showBottomDivider) {
 		View view = inflate(R.layout.bottom_sheet_item_with_descr_and_checkbox_and_divider_56dp);
+		DownloadItem downloadItem = mapItem.downloadItem();
 		ImageView icon = view.findViewById(R.id.icon);
-		boolean downloaded = downloadItem.isDownloaded();
-		icon.setImageResource(downloaded ? R.drawable.ic_action_map_update : R.drawable.ic_action_map_download);
+		icon.setImageResource(mapItem.type() == RequiredMapType.OUTDATED
+				? R.drawable.ic_action_map_update : R.drawable.ic_action_map_update);
```
Chumva: Same icon in both cases?

PR #25354, opened 2026-07-03
Online public transport routing via MOTIS/Transitous

### LINKED ISSUE #23121: Feature Request: Integrate Transitous MOTIS 2 API for Public Transport Routing
### Describe the idea

Integrate public-transport routing via the open Transitous MOTIS 2 API (https://transitous.org/api/).  
The API supplies worldwide GTFS-based schedules, real-time delay updates, multimodal journey planning and station/stop metadata through a clean REST/graph interface. Adding it as an online source would give OsmAnd users high-quality, privacy-respecting transit directions without leaving the app.

### Expected behaviour

1. The “Public transport (online)” routing profile should queries the Transitous API.  
2. Journey results show:
   • departure/arrival times (with real-time updates if available)  
   • line names, platforms, transfer points and walking legs
3. Stops and lines appear on the map when the profile is active; tapping a stop reveals live departure boards.  
4. The feature remains fully optional:
   • Works only when the user enables it and has data connectivity.  
   • Falls back to existing offline routing if no internet is available.  
5. No tracking: only the minimal API requests required for the chosen trip are sent; nothing is stored server-side.

• Continuing to rely solely on OsmAnd’s offline public transport support → lacks real-time data
• Relying on third-party apps like KDE Itinerary for transit directions and switch to OsmAnd for better map overview.

### Context

• Transitous is fully open-source (Apache 2.0) and already powers KDE Itinerary and other community projects.  
• The API can be self-hosted, which fits OsmAnd’s privacy philosophy and allows third-party map builders to bundle their own endpoints.  
• With growing demand for multimodal navigation—including mixed car/train/bike trips—this integration would close one of OsmAnd’s last big feature gaps and strengthen its position as a complete, open, privacy-respecting navigation suite.

YourSandwich:
Closes #23121.

Native public-transport routing has no schedule data, so journeys render from `00:00` with no real
times. This adds real-time routing from open online timetables: the existing PT route UI now shows
actual departure and arrival times, alternatives, official line colours, and real leg geometry.

The backend is MOTIS through the public [Transitous](https://transitous.org) instance: Apache-2.0,
keyless, self-hostable, no vendored library, no new dependency. It hooks in at **a single point** in
`TransportRoutingHelper.calculateRouteImpl`, maps each itinerary onto the native
`TransportRouteResult`, and falls through to the native engine on an empty result, so everything
downstream is unchanged. One per-profile switch, `USE_ONLINE_PUBLIC_TRANSPORT`, turns it off.

One call for maintainers: the switch defaults on and the endpoint is the hardcoded volunteer
Transitous instance, which asks to be contacted before large-scale traffic. A wide rollout probably
wants the endpoint made configurable (the way `net.osmand.plus.onlinerouting` ships the OSRM demo
server) or the default flipped. One-line change either way.

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/data/TransportRoute.java
+++ b/OsmAnd-java/src/main/java/net/osmand/data/TransportRoute.java
@@ -16,6 +16,7 @@ public class TransportRoute extends MapObject {
 	private List<TransportStop> forwardStops = new ArrayList<TransportStop>();
 	private String ref;
 	private String operator;
+	private String url;
 	private String type;
 	private Integer dist = null;
 	private String color;
@@ -35,6 +36,7 @@ public class TransportRoute extends MapObject {
 		this.names = r.names;
 		this.id = r.id;
 		this.operator = r.operator;
+		this.url = r.url;
 		this.ref = r.ref;
 		this.type = r.type;
 		this.color = r.color;
@@ -282,6 +284,14 @@ public class TransportRoute extends MapObject {
 		return operator;
 	}
 	
+	public String getUrl() {
+		return url;
+	}
+
+	public void setUrl(String url) {
+		this.url = url;
+	}
+
 	public void setOperator(String operator) {
 		this.operator = operator;
 	}
--- a/OsmAnd-java/src/main/java/net/osmand/data/TransportStop.java
+++ b/OsmAnd-java/src/main/java/net/osmand/data/TransportStop.java
@@ -23,6 +23,7 @@ public class TransportStop extends MapObject {
 	private List<TransportStopExit> exits;
 	private List<TransportRoute> routes = null;
 	private TransportStopAggregated transportStopAggregated;
+	private String platform;
 
 	public TransportStop() {}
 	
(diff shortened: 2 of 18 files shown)


YourSandwich:
Pushed an update. Rebuilt the branch into clean, self-contained per-feature commits and field-tested it on a ~1500 km trip through Vienna and Poland, navigating on nothing but this app.

New since the last push: real departure and arrival times with the total duration, walking legs routed on the road network between transit legs, a later-departures action, platform and track where the feed provides them, and an optional ticket link per leg.

One correctness fix worth flagging: a same-station transfer was printing the previous leg's boarding walk (e.g. "19 min, 975 m") because the transfer row read the wrong segment's walk distance. It now reads the transfer's own walk, so changing between two services at one stop shows no phantom walk and the wait sits in the clock times.

The optional commits (later departures, platforms, ticket link) sit last and drop cleanly. Rebased onto current master.

vshcherb:
No plan to review this PR yet as this feature is not scheduled for backlog this year and it will require separate investigation and preparation

elisa-aleman:
Is there an estimated time for this to be reviewed / implemented? 

Lack of public transit navigation is the only thing keeping me from making the full switch from less private map apps, so this change would seal the deal.

Issue #6878, opened 2019-04-27
Possibility of avoiding traffic jam?


wehkah:
I don't know if it's possible for Osmand, but it would be an enormous relief, if it was possible to access traffic information databases, that include traffic jam, construction sites, road closings, etc, and include these information when calculation the routing.

OpenSourceForEver:
https://github.com/osmandapp/Osmand/issues/1432 was closed because there is no possible traffic source. Does anyone know any possible data source?

mahdi1234:
@vshcherb @OpenSourceForEver I guess it might be country-by-country e.g. Czech Republic has national registry which can be used freely after signing agreement (I suggested this source to other app which already works on it)

http://registr.dopravniinfo.cz/en/index.html
http://registr.dopravniinfo.cz/en/sources/cz-ndic_ddr-common/

wehkah:
I've seen apps in the PlayStore that claim to be able to display traffic jam in realtime. I haven't tested them yet, but I guess there have to be some databases that theses apps retrieve the information from. 

Most FM radio stations also offer web pages that provide traffic information. 

So, I would guess, these information are neither top secret nor anyone's private knowledge. After all: these information can be seen by anyone who is looking at the corresponding streets, so how could those information be private property? Ans on the contrary, the spreading of these information is in everybody's interest, isn't it?

exploide:
> So, I would guess, these information are neither top secret nor anyone's private knowledge. After all: these information can be seen by anyone who is looking at the corresponding streets, so how could those information be private property? Ans on the contrary, the spreading of these information is in everybody's interest, isn't it?

Ideologically you are right that this information should not be private property. However, that's not how it is as of today. Google for example collects its own traffic information utilizing the mass of Android phones with GMaps installed out there. This gives them the advantage over competitors like Osmand to have great traffic information available. Hence, they don't share it.

"Public" information from FM station's websites on the other hand is rather inaccurate, not perfectly machine readable (i.e. not necessarily structured) and lacks sufficient coverage.

So what the world (i.e. we :stuck_out_tongue: ) needs is an open data source for traffic information where everyone can participate. But as stated in the other issue, the OpenTraffic initiative did not succeed (yet?!).

[...]

Issue #3565, opened 2017-03-19
[Feature Request] Traffic data


fatuuse:
The only reason I won't use OSMand as traffic routing in my car is that it miss traffic data.
I suggest an extra plugin for that.

One simple plugin were you can report traffic cues, police, accident etc witch pop up like an POI in one overlay. Just let user push one button in a corner to report in there position.  I do not need speed data from all devices just one heads up for temporary traffic happenings.

lordfolken:
There is now an app called "QZ" which decodes TMC messages from a Bluetooth adapter:
https://gitlab.com/mvglasow/qz

This is a gpl implementation. Maybe a plugin can be written.

eregwizard:
Suprising labeling as nice to have.
Since by now several free Android navigation apps already do provide live traffic data and route recalculation, it becomes clear that osmand is lagging and becoming less useful.

vshcherb:
It is nearly impossible to have free/open application and traffic data, cause there is no open global traffic data. Probably they provide that feature for free now or they exchange all user data for that free traffic data.

eregwizard:
There are already several android navigation apps that do have traffic data, including Magic Earth that claims to require and use no personal data at all. Since it works also fully offline (then no traffic data is used) as well as online (using traffic data and live maps) I tend to believe that claim until proven otherwise.

breversa:
FWIW, the french government has provides traffic data in TraficXml format (even for commercial use), along with lots of technical data : https://diffusion-numerique.info-routiere.gouv.fr/

Licence :
https://diffusion-numerique.info-routiere.gouv.fr/conditions-de-reutilisation-des-donnees-a2.html

zimbelstern:
Similarly, there is open traffic data for North Rhine-Westphalia, Germany's most populated state:
https://www.govdata.de/web/guest/suchen/-/details/verkehrslage-los
supplied by the government with a free-to-use licence:
https://www.govdata.de/dl-de/by-2-0
Would be great to have an option to use Datex II data sources for routing.

eregwizard:
For The Netherlands all kinds of open traffic related data can be found at http://opendata.ndw.nu/ , although I must admit I find it hard to interpret the data. Maybe there are people who are better at reading this?

Z3ROGRAV1TYx:
>One simple plugin were you can report traffic cues, police, accident etc witch pop up like an POI in one overlay. 

[...]

Issue #2068, opened 2015-12-18
Traffic matches routing (visual)


luckydevil13:
When using routing + additional layer with traffic - it impossible to adjust color screen to see traffic color (green or red).
https://dl2.pushbulletusercontent.com/NyJVN6CnsxOF6ERUaUuCdU8ieMXR5n0u/QuickMemo%20_2015-12-18-13-14-56.png

It will be nice to have feature like yandex navi:
https://dl2.pushbulletusercontent.com/RvmELQ7BZJigHVJgpHggsIvmCJQoTWkx/traffic.png

Colored stripe(bar) traffic that matches the current route! It will be awesome

Zireael07:
How do you get an additional layer with traffic?

luckydevil13:
online maps - yandex trafic
-- follow your dream...

Korb:
@luckydevil13, does enabling the `Yandex Traffic` layer only provide visualization of traffic density, but does it not contribute to better optimization of route planning taking into account current or expected traffic congestion?

DmitryAlexei:
Raster map data is not considered when calculating and recalculating routes.

Artem777812:
яндекс трафик работает не во всех странах. Также у OsmAnd идеология другая, это оффлайн карты. Трафик - это онлайн, как тот же Waze или Гугл карты. Но у последних значительно хуже детализация карт и функционала вообще нет.

PR #25356, opened 2026-07-03
Live traffic overlay and traffic-aware routing via TomTom

YourSandwich:
Addresses #3565 (traffic data), #2068 (colour roads by congestion) and #6878 (route around jams).

OsmAnd has no live traffic. This adds it as an opt-in plugin: a TomTom congestion overlay and traffic-aware car routing. It is bring-your-own-key, no bundled token and no new dependency, so each user pastes their own free TomTom key, the same way the app already takes an OSRM/GraphHopper/ORS endpoint.

The overlay is the plugin's own `MapTileLayer` on TomTom's `relative0` flow tiles, so only slow roads colour and the base map shows through. It is additive, never touches the overlay/underlay slots, and stays out of the map-source picker. Routing hooks in at **a single point** in `RouteProvider.calculateRouteImpl`: when the plugin toggle is on it calculates with `TomTomEngine` (a normal `OnlineRoutingEngine`, `traffic=true`, guidance parsed to turn-by-turn), otherwise it falls straight through to the normal engine, so no engine is persisted and no profile is rewritten. Everything is one self-contained plugin, default off, with a single key field and toggles for overlay, routing, and periodic re-routing while driving.

One call for maintainers: keeping the key with the user is what sidesteps the reason traffic has stalled since #3565 (2017) - no free global traffic without trading user data. The only core touch is about 25 lines in `RouteProvider`. TomTom is used over raw REST (free tier, no card), and HERE is a drop-in if its explicit third-party-basemap licence is preferred.

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <resources>
+	<string name="traffic_plugin_name">Traffic</string>
+	<string name="traffic_plugin_description">Shows live road traffic from TomTom as a coloured overlay and factors it into car routing. Requires a personal TomTom API key and an internet connection.</string>
+	<string name="traffic_show_overlay">Show traffic overlay</string>
+	<string name="traffic_routing">Traffic-aware routing</string>
+	<string name="traffic_routing_description">Route the car profile with TomTom live traffic. Sets up the routing engine automatically from your API key.</string>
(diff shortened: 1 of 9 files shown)


vshcherb:
No plan to merge as that would be great if can create support for independent plugin and not maintain tomtom - we're not ready to take care for TomTom separate engine and not particularly interested in it as we would like to find open or paid but collaboritive approach that users would be happy to contribute

YourSandwich:
I understand, thank you for checking out <3

PR #25362, opened 2026-07-03
Fix #24802 Gap between My location and Route line

### LINKED ISSUE #24802: Gap between My location and Route line
### Description

iOS issue https://github.com/osmandapp/OsmAnd-iOS/issues/4798

During navigation, a gap appears between "My location" and the route line.

### Steps to reproduce

1. Start navigation with Car profile (default settings + 0.5 sec of Position animation). 

### Actual result

- With 3D relief:

<table style="width:100%">
  <tr>
<th> </td>
  </tr>
</table>

- Without 3D relief:

<table style="width:100%">
  <tr>
<th> </td>
  </tr>
</table>

### Expected result

Without a gap.

```
OsmAnd Version: OsmAnd 5.3.3, released: 2026-03-31
Android version:
Device model: S22
```

nazar-kutz:
## Problem

During long navigation, the route line could gradually move away from the current location marker.

The issue was more noticeable at high zoom levels and increased with the traveled route distance. The marker stayed correctly positioned on the road, while the rendered route line was clipped slightly too far ahead, creating a visible gap.

## Root Cause

In OpenGL rendering, the route line is created as a native `VectorLine`. During navigation, we do not rebuild the whole line on every location update. Instead, the already passed part of the route is clipped by updating `startingDistance`.

This `startingDistance` is calculated on the Java side in `GeometryWay.cutStartOfCachedPath()` and then passed to native `VectorLine.setStartingDistance(...)`.

The problem was that Java-side route clipping distance and native `VectorLine` distance were calculated with slightly different Earth radius values. The difference is very small, but it accumulates as `startingDistance` grows during navigation.

For example, the radius mismatch is about `0.028%`, which gives roughly:

* `~2.8 m` difference after `10 km`;
* `~8.5 m` difference after `30 km`.

At high zoom levels, this accumulated difference becomes visually noticeable as a gap between the current location marker and the rendered route line.

Screen rotation temporarily fixed the issue because the renderer and cached route geometry were recreated, resetting the accumulated clipping distance.

## Solution

This PR keeps the default `GeometryWay` distance calculation unchanged, but overrides the route clipping distance calculation in `RouteGeometryWay`.

For route lines, `startingDistance` is now calculated with the same Earth radius as native `VectorLine`. This keeps Java-side clipping distance consistent with native route line rendering and prevents the route line from drifting forward over long navigation distances.

## Scope

This is a rendering-only fix.

It does not affect:

* route calculation;
* route progress;
* voice guidance;
* off-route detection;
* distance-to-destination calculation.

The change is localized to route line rendering and avoids changing global distance utilities, since different distance constants are used in different parts of the codebase for different purposes.

## Validation

Tested with long route simulation at maximum zoom.

Before the fix, the route line gradually drifted away from the current location marker as navigation progressed.

After the fix, the route line remains aligned with the marker during navigation.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java
@@ -1,5 +1,7 @@
 package net.osmand.plus.views.layers.geometry;
 
+import static net.osmand.util.MapUtils.VECTOR_LINE_EARTH_RADIUS_METERS;
+
 import android.content.Context;
 import android.graphics.Canvas;
 import android.graphics.Paint;
@@ -39,7 +41,6 @@ import gnu.trove.list.array.TByteArrayList;
 public class RouteGeometryWay extends
 		MultiColoringGeometryWay<RouteGeometryWayContext, MultiColoringGeometryWayDrawer<RouteGeometryWayContext>> {
 
-	private static final double VECTOR_LINE_EARTH_RADIUS = 6371000.0;
 	public static final int MIN_COLOR_SQUARE_DISTANCE = 15_000;
 
 	private final RoutingHelper helper;
@@ -286,16 +287,7 @@ public class RouteGeometryWay extends
 
 	private static double calculateNativeVectorLineDistance(double lat1, double lon1, double lat2, double lon2) {
 		// Keep startingDistance in the same metric as OsmAnd-core VectorLine_P::calculateShortestPath().
-		double dLat = Math.toRadians(lat2 - lat1);
-		double dLon = Math.toRadians(lon2 - lon1);
-		double sinHalfLat = Math.sin(dLat / 2.0);
-		double sinHalfLon = Math.sin(dLon / 2.0);
-		double a = sinHalfLat * sinHalfLat
-				+ Math.cos(Math.toRadians(lat1))
-				* Math.cos(Math.toRadians(lat2))
-				* sinHalfLon
-				* sinHalfLon;
-		return 2.0 * VECTOR_LINE_EARTH_RADIUS * Math.asin(Math.sqrt(a));
+		return MapUtils.getDistance(lat1, lon1, lat2, lon2, VECTOR_LINE_EARTH_RADIUS_METERS);
 	}
 
 	public void clearRoute() {
--- a/OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java
@@ -39,6 +39,7 @@ import gnu.trove.list.array.TByteArrayList;
 public class RouteGeometryWay extends
 		MultiColoringGeometryWay<RouteGeometryWayContext, MultiColoringGeometryWayDrawer<RouteGeometryWayContext>> {
 
+	private static final double VECTOR_LINE_EARTH_RADIUS = 6371000.0;
 	public static final int MIN_COLOR_SQUARE_DISTANCE = 15_000;
 
 	private final RoutingHelper helper;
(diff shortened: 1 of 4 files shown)


OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java
```diff
@@ -271,6 +272,32 @@ protected boolean shouldDrawArrows() {
...
+	}
+
+	@Override
+	protected double calculateProjectionDistance(@NonNull Location projection, int x31, int y31) {
+		return calculateNativeVectorLineDistance(
+				projection.getLatitude(), projection.getLongitude(),
+				MapUtils.get31LatitudeY(y31), MapUtils.get31LongitudeX(x31));
+	}
+
+	private static double calculateNativeVectorLineDistance(double lat1, double lon1, double lat2, double lon2) {
```
Chumva, line 288: So this is similar to MapUtils.getDistance but uses another constant for the Earth's radius? Should we add this to MapUtils.getDistance as a flag which constant use? We could make constants for the Earth's radius & use them in other places where 6371 or 6372.8 are used. 

Also update KMapUtils.getDistance

nazar-kutz: Should be fixed in https://github.com/osmandapp/OsmAnd/pull/25362/changes/257d265e6636192d79a175229edd39f78fe82973

sonora:
@nazar-kutz , @Chumva Great research!!

But conceptionally, you could get rid of the extra wrappers and just use overrides:

(Like we do elsewhere in OsmAnd, removes the extra layers of void complexity)

```
diff --git a/OsmAnd/src/net/osmand/plus/views/layers/geometry/GeometryWay.java b/OsmAnd/src/net/osmand/plus/views/layers/geometry/GeometryWay.java
index e84a6dc60ba..abcd1234567 100644
--- a/OsmAnd/src/net/osmand/plus/views/layers/geometry/GeometryWay.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/geometry/GeometryWay.java
@@ -255,7 +255,7 @@ public void drawSegments(@NonNull RotatedTileBox tb, @Nullable Canvas canvas, double
                    double prevLon = locationProvider.getLongitude(previous);
                    double lat = locationProvider.getLatitude(i);
                    double lon = locationProvider.getLongitude(i);
-                   dist = MapUtils.getDistance(prevLat, prevLon, lat, lon);
+                   dist = getSegmentDistance(prevLat, prevLon, lat, lon);
                }
                if (!previousVisible && !ignorePrevious) {
                    if (previous != -1 && !isPreviousPointFarAway(locationProvider, previous, i)) {
@@ -422,9 +422,7 @@ protected List<List<DrawPathData31>> cutStartOfCachedPath(@NonNull MapRendererView view,
            }
        }
        if (lastProjection != null && lastX31 != 0 && lastY31 != 0) {
-           passedDist += (float) MapUtils.measuredDist31(
-                   MapUtils.get31TileNumberX(lastProjection.getLongitude()),
-                   MapUtils.get31TileNumberY(lastProjection.getLatitude()), lastX31, lastY31);
+           passedDist += (float) getProjectionDistance(lastProjection, lastX31, lastY31);
        }
 
        if (passedLineId > 0) {
@@ -450,6 +448,20 @@ protected boolean shouldDrawArrows() {
        return true;
    }
 
+   protected double getSegmentDistance(double lat1, double lon1, double lat2, double lon2) {
+       return MapUtils.getDistance(lat1, lon1, lat2, lon2);
+   }
+
+   protected double getProjectionDistance(@NonNull Location projection, int x31, int y31) {
+       return MapUtils.measuredDist31(
+               MapUtils.get31TileNumberX(projection.getLongitude()),
+               MapUtils.get31TileNumberY(projection.getLatitude()),
+               x31,
+               y31
+       );
+   }
+
    public void drawRouteSegment(@NonNull RotatedTileBox tb, @Nullable Canvas canvas,
                                 List<GeometryWayPoint> points, double distToFinish) {
        if (points.size() < 2) {
diff --git a/OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java b/OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java
index 773a9a0c7d4..def01234567 100644
--- a/OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/geometry/RouteGeometryWay.java
@@ -1,5 +1,7 @@
 package net.osmand.plus.views.layers.geometry;
 
+import static net.osmand.util.MapUtils.VECTOR_LINE_EARTH_RADIUS_METERS;
+
 import android.content.Context;
 import android.graphics.Canvas;
 import android.graphics.Paint;
@@ -271,6 +273,22 @@ protected boolean shouldDrawArrows() {
        return drawDirectionArrows;
    }
 
+   @Override
+   protected double getSegmentDistance(double lat1, double lon1, double lat2, double lon2) {
+       return MapUtils.getDistance(lat1, lon1, lat2, lon2, VECTOR_LINE_EARTH_RADIUS_METERS);
+   }
+
+   @Override
+   protected double getProjectionDistance(@NonNull Location projection, int x31, int y31) {
+       return MapUtils.getDistance(
+               projection.getLatitude(),
+               projection.getLongitude(),
+               MapUtils.get31LatitudeY(y31),
+               MapUtils.get31LongitudeX(x31),
+               VECTOR_LINE_EARTH_RADIUS_METERS
+       );
+   }
+
    public void clearRoute() {
        if (route != null) {
            route = null;

```

Hrojar:
I don't know what I'm doing wrong. I'm just a navigation user. I did the OsmAnd update yesterday. And unfortunately, nothing has changed! After 100 km of driving, the route departs from the arrow again! I turn the phone to portrait and then back to landscape and the arrow and route are connected again. I am attaching a screenshot. The first is after driving 100 km, the second is when I turn the phone back and forth. I don't know what to do about it?
Screenshot-1
Screenshot-2

Hrojar:
Well, then advise me how to use this program to fix the navigation. I only have the OsmAnd navigation from Google play installed and I am not a SW engineer or a programmer. But if you give me advice, maybe I can do it. Thank you very much.

nazar-kutz:
@Sonora Thanks, good point. I agree — after moving the formula to MapUtils, the extra wrapper is unnecessary. I simplified this and removed the extra wrapper method in https://github.com/osmandapp/OsmAnd/pull/25405

nazar-kutz:
Hi @Hrojar,

As @sonora mentioned, the fix is not included in the current Google Play version yet. It should be available in one of the next OsmAnd updates, most likely 5.4.x.

The fix has already been tested in OsmAnd~ 5.4.0#8191m: after one hour of simulation, no gaps were found.

So for the regular Google Play version, you only need to wait for the next update that includes this fix.

Issue #25331, opened 2026-06-29
Spatial Search - Implementation


vshcherb:
# Preamble
As new search introduces it attempts to fix all existing issue at once. The code name is "Spatial search by name", it works on completely different principle so it is required to be reviewed

### Other Tasks
- https://github.com/osmandapp/web/issues/1893
- https://github.com/osmandapp/OsmAnd/issues/25485

### Issues to be fixed
- https://github.com/osmandapp/OsmAnd/issues/23709
- https://github.com/osmandapp/web/issues/1510
- https://github.com/osmandapp/OsmAnd/issues/11467
- https://github.com/osmandapp/OsmAnd/issues/23697
- https://github.com/osmandapp/OsmAnd/issues/21621

# Main task list
- [x] Analyze Duplicate '10 Am Remsufer Remseck am Neckar', +'138 138 Scott Avenue Bellefonte', +'8 av 8'
- [x] https://github.com/osmandapp/OsmAnd/issues/25065

### Follow up tasks

```
// TODO Analyze Performance bootlenecks VisualVM (Pipeline + Intersection)
// TODO Sorting before load objects (use elo and other buildings?) and limit results
// TODO Auto-Corrections / Suggestion based on common suffixes
// TODO Postcode needs to load street and check buildings! Store postcode as bbox not as City! - '1186RZ 324' (NL, UK) 
// TODO Search near key objects (subway station artificial bbox)
// TODO New Geocoding for cases ("NC 42" == "NC-42") - geo index for prefixes
// TODO Add flats: https://www.openstreetmap.org/node/5843642738
// TODO English postcodes
// TODO Precise Boundary 'Chernihiv sport life' mostly Kyiv - check precise boundary for filter
// TODO Short word split "Ro-ki" vs "Roki" 
```

vshcherb:
Auto Tests Us_penn / Germany - Summary
```
NO FIXES
* 2602 NY Elmira - Data bug
* 400 Susquehanna Boulevard Hazel Township	- Data bug (No Hazel Township!)
* 1E Curtin Road University Park - Deleted
* 1526 Grand Central Avenue Elmira - Deleted

Partial support (show more)?
* 330 Innovation Boulevard University Park - incorrect city
 - 330 Innovation Boulevard Innovation Park
 - 330 Innovation Boulevard State College
* 276 East End Centre Wilkes-Barre - Data deleted (https://www.openstreetmap.org/node/13005339094)
[...]

PR #25395, opened 2026-07-07
Deduplicate spatial search

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/25331


OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResult.java
```diff
@@ -322,5 +385,61 @@ public static int compare(SpatialSearchResult o1, SpatialSearchResult o2, LatLon
...
+				lon = loc.getLongitude();
+			}
+		}
+		if (lat == 0 && lon == 0) {
+			return null;
+		}
+		return MapUtils.createShortLinkString(lat, lon, zoom);
+	}
+
+	private String getShortName() {
```
vshcherb: short name means?

PR #25414, opened 2026-07-10
Add duplicate map warning for Standard and Road-only maps

### LINKED ISSUE #24892: Add notification to deactivate/delete Road Map when Full Map is downloaded
See #15039 (especially https://github.com/osmandapp/OsmAnd/issues/15039#issuecomment-3412938702) and #24627. Both tickets got closed without fixing this problem. Maybe this third ticket will finally lead to a solution :)

nazar-kutz:


OsmAnd/src/net/osmand/plus/download/ui/ItemViewHolder.java
```diff
@@ -522,34 +522,28 @@ protected void showContextMenu(View v,
...
+						indexItem.getType() == DownloadActivityType.NORMAL_FILE
+								? DuplicateMapDownloadDialogController.ConflictDirection.ROAD_TO_STANDARD
+								: DuplicateMapDownloadDialogController.ConflictDirection.STANDARD_TO_ROAD;
+
+				DuplicateMapDownloadDialogController.showDialog(
+						context,
+						indexItem.getVisibleName(context, app.getRegions(), false),
+						direction,
+						() -> context.startReplacementDownload(indexItem, conflict.getTargetFile(app)),
+						() -> startDownload(indexItem)
```
Chumva: Will it work correctly after screen rotation? Will we keep the old context(activity) in runnable/controller after screen rotation?

OsmAnd/src/net/osmand/plus/download/ui/ItemViewHolder.java
```diff
@@ -522,34 +522,28 @@ protected void showContextMenu(View v,
...
-				}
+
+			IndexItem conflict = DuplicateMapHelper.findConflictingItem(indexItem, parentOptional);
+			if (conflict != null) {
+				DuplicateMapDownloadDialogController.ConflictDirection direction =
+						indexItem.getType() == DownloadActivityType.NORMAL_FILE
+								? DuplicateMapDownloadDialogController.ConflictDirection.ROAD_TO_STANDARD
+								: DuplicateMapDownloadDialogController.ConflictDirection.STANDARD_TO_ROAD;
+
+				DuplicateMapDownloadDialogController.showDialog(
```
Chumva: Do we need this check & dialogue only in ItemViewHolder in DownloadActivity, or in all places where we download maps via DownloadValidationManager? 

Should we move these checks to DownloadValidationManager?

OsmAnd/src/net/osmand/plus/download/ui/ItemViewHolder.java
```diff
@@ -522,34 +522,28 @@ protected void showContextMenu(View v,
...
+						context,
+						indexItem.getVisibleName(context, app.getRegions(), false),
+						direction,
+						() -> context.startReplacementDownload(indexItem, conflict.getTargetFile(app)),
```
Chumva: Should this also work for deactivated maps, or only for activated ones?

OsmAnd/src/net/osmand/plus/download/DownloadFileHelper.java
```diff
@@ -240,6 +243,26 @@ public boolean downloadFile(IndexItem.DownloadEntry de, IProgress progress,
 		}
 	}
 
+	private void removeFilesAfterSuccessfulInstall(@NonNull DownloadEntry entry) {
```
Chumva, line 246: Can we try reusing or combining it with removePreviousSrtmFile? They have a similar concept: delete the counterpart after a successful install

OsmAnd/src/net/osmand/plus/download/ui/DuplicateMapHelper.kt
```diff
@@ -0,0 +1,25 @@
+package net.osmand.plus.download.ui
+
+import net.osmand.plus.download.DownloadActivityType
+import net.osmand.plus.download.DownloadResourceGroup
+import net.osmand.plus.download.IndexItem
+
+object DuplicateMapHelper {
+
+    @JvmStatic
+    fun findConflictingItem(target: IndexItem, group: DownloadResourceGroup): IndexItem? {
```
Chumva: Do we really need DownloadResourceGroup for this check? Can we check by filenames similar to removePreviousSrtmFile?

OsmAnd/src/net/osmand/plus/download/ui/ItemViewHolder.java
nazar-kutz: fixed in https://github.com/osmandapp/OsmAnd/pull/25414/changes/7d4d08252fdeb4fdb64081c025d04bf35b99aea7

OsmAnd/src/net/osmand/plus/download/ui/ItemViewHolder.java
nazar-kutz: fixed in https://github.com/osmandapp/OsmAnd/pull/25414/changes/b822479891fdd7cd3c6adeadc6185b26e08eaaa0

OsmAnd/src/net/osmand/plus/download/ui/DuplicateMapHelper.kt
nazar-kutz: I checked the filename-based approach, but kept DownloadResourceGroup intentionally. It provides the exact sibling IndexItem, allowing us to reuse getExistedFile(app) for both active and deactivated maps without duplicating suffix, directory, and backup-path rules.
Unlike SRTM, Standard and Road-only maps use different directories and also need backup resolution, so a filename-based implementation would be more fragile. If this check is moved outside DownloadActivity in the future, we can introduce a dedicated model-level sibling resolver.

OsmAnd/src/net/osmand/plus/download/ui/ItemViewHolder.java
nazar-kutz: The current implementation is intentionally scoped to explicit single-item Standard/Road downloads from DownloadActivity. DownloadValidationManager is also used by batch downloads and several non-DownloadActivity flows, so moving the prompt there would require defining multi-conflict, cancellation, and replacement behavior for those flows. I suggest keeping the current scope for this task and handling global/batch duplicate validation separately if required.

OsmAnd/src/net/osmand/plus/download/DownloadFileHelper.java
nazar-kutz: I kept the high-level cleanup flows separate because SRTM cleanup is implicit, while Duplicate Map cleanup is registered only after explicit Replace. I extracted the common file deletion and resource closing into a shared helper without changing when either cleanup runs.

fixed in https://github.com/osmandapp/OsmAnd/pull/25414/changes/804695168c9f7b35834091b23863d979a98accdc

PR #25423, opened 2026-07-11
Keep quick search FABs above the navigation bar in edge-to-edge mode

### LINKED ISSUE #25382: Custom POI search / Save button (blue icon) has no effect
### Description

From https://osmand.net/docs/user/search/search-poi/#save-new-custom-filters

Image

The blue save button displayed above has no effect.

Workaround: Save the custom POI search via the filter icon / save filter.

### Steps to reproduce

See docs at https://osmand.net/docs/user/search/search-poi/#save-new-custom-filters

Search / Categories / Custom search / (select POIs) / Show/Apply(?) / Blue save button

### Actual result

Nothing happens.

### Expected result

A dialog appears, asking for the name of the new custom POI search.

```
OsmAnd Version: 5.3.10
Android version: 15
Device model: Fairphone 4
```

MiMoHo:
Fixes #25382

## Problem

As @scaidermern diagnosed, the blue save-filter FAB in the quick search results overlaps the Android navigation bar in edge-to-edge mode, so its lower half does not react to taps (screenshot in the issue). The show-on-map FAB is affected the same way.

## Cause

`QuickSearchDialogFragment.getInsetTargets()` correctly declares both FABs via `InsetTarget.createFab()`, and the insets framework initially lifts them above the navigation bar. But `updateFabMargins()` overwrites the bottom margin with static dimension values whenever FAB visibility changes — which happens exactly when the save FAB appears (custom filter applied → results shown). Since no new insets dispatch follows, the FAB stays at 12dp from the screen edge, partially behind the navigation bar.

QA could not reproduce it reliably (#25382 comment) because the outcome depends on timing/ordering of insets dispatches vs. visibility changes — e.g. an IME-close dispatch *after* the FAB was shown re-applies the correct margin via `InsetsUtils.applyMargin()`.

## Fix

`updateFabMargins()` now:
- adds the current bottom inset (same type mask as `InsetTarget.createFab()`: system bars + IME + display cutout, taken from `getLastRootInsets()`) to the computed margin, and
- syncs the `initial_margin_bottom` tag that `InsetsUtils.applyMargin()` uses, so the next insets dispatch recomputes exactly the same margin instead of resetting the stacked offsets (two visible FABs / raised margin above the "send empty search" bar).

This keeps the manual stacking logic and the insets framework consistent in both directions. On devices without edge-to-edge, `getLastRootInsets()` is null and behavior is unchanged.

## Testing

Verified by code-path analysis of the insets framework (`InsetsUtils.applyMargin()` stores initial margins in view tags and recomputes `initial + insets.bottom` on every dispatch); all interleavings of insets dispatch and `updateFabMargins()` now produce the same margin. Happy to adjust if device testing by QA shows anything off.

---
Claude Code — Claude Fable 5, xhigh

### DIFF
--- a/OsmAnd/src/net/osmand/plus/search/dialogs/QuickSearchDialogFragment.java
+++ b/OsmAnd/src/net/osmand/plus/search/dialogs/QuickSearchDialogFragment.java
@@ -29,6 +29,7 @@ import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.widget.Toolbar;
 import androidx.core.content.ContextCompat;
+import androidx.core.view.WindowInsetsCompat;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
 import androidx.fragment.app.FragmentManager;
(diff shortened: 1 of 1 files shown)


Chumva:
@MiMoHo It should already be fixed [here](https://github.com/osmandapp/OsmAnd/commit/71a621fdff42bbda5b909024fa2fc2dbcea814f7)

PR #25425, opened 2026-07-11
Use mapper-provided romanized names (ja-Latn, ko-Latn) instead of machine transliteration

### LINKED ISSUE #25267: Map Localization: Use name:ja-Latn as a fallback for English name if name:en is missing
### Describe the idea

When a node has a tag 'name:ja-Latn' AND name:en is empty, copy ja-Latn tag into empty 'name:en' tag. This procedure eliminates incorrectly auto-generated proper name in English.

### Expected behaviour

A node https://www.openstreetmap.org/way/1175336500/ has a tag in name:ja-Latn, "Oyama Afuri Jinja," which you can read like English. While this node does not have name:en tag, it's nice if OSMAnd app shows this ja-Latn name in the name:en field.

Include name:ja-Latn when it's not empty into download map data.

### Context

_No response_

MiMoHo:
Fixes #25267

## Problem

For objects without `name:en`, OsmAnd falls back to machine transliteration (Junidecode), which produces wrong results especially for Japanese — kanji get Chinese readings. Example from the issue, way 1175336500 (大山阿夫利神社, `name:ja-Latn` = "Oyama Afuri Jinja"):

- current English fallback: **"Da Shan A Fu Li Shen She"**
- mapper-provided romanization: **"Oyama Afuri Jinja"**

The mapper-provided names can't even be used today because `name:ja-Latn` (165,044 uses) and `name:ko-Latn` (138,850 uses) are not in `MapRenderingTypes.langs`, so they are dropped during map creation (unlike `sr-latn` / `zh-hans` / `zh-hant`, which are already indexed).

## Changes

1. **`MapRenderingTypes.langs`**: add `ja-latn` and `ko-latn` (placed next to `ja` / `ko`, matching the existing `sr-latn` style). Takes effect once maps are regenerated.
2. **`EntityParser.parseMapObject()`**: lowercase the language suffix before the `langsSet` lookup and store names under the lowercase code. Production parsing already lowercases tag keys (`Entity.putTag`), but code paths using `putTagNoLC` — e.g. the OSM editing download, which sets `setConvertTagsToLC(false)` — preserve OSM's BCP 47 capitalization (`name:ja-Latn`, `name:sr-Latn`) and previously failed the lookup.
3. **`MapObject.getEnName(true)`**: prefer a stored Latin-script name (key ending in `-latn`) over machine transliteration when `name:en` is missing. Explicit `name:en` still wins; behavior with transliteration disabled is unchanged; objects without a romanized name fall back to Junidecode exactly as before.

`int_name` (907k uses) could be a further fallback candidate, but it is not indexed today and not guaranteed to be Latin script, so it is left out of this PR.

## Testing

Compiled the patched classes against the current master snapshot jar and ran a standalone test (class-shadowing), 14 checks — all pass with the patch; the unpatched jar shows the current behavior:

| Check | Unpatched master | Patched |
|---|---|---|
| `name:ja-Latn` indexed (lowercased keys, production path) | dropped | `ja-latn` ✓ |
| `name:ko-Latn` indexed | dropped | `ko-latn` ✓ |
| `name:ja-Latn`/`name:sr-Latn` via case-preserving path (`putTagNoLC`) | dropped | indexed ✓ |
| `getEnName(true)` for 大山阿夫利神社 with `ja-Latn` | "Da Shan A Fu Li Shen She" | "Oyama Afuri Jinja" |
| `name:sr-latn`, `name:zh-Hans`, `name:de` (production path) | indexed | indexed (unchanged) |
| unknown language suffix | dropped | dropped (unchanged) |
| explicit `name:en` precedence | wins | wins (unchanged) |
| Junidecode fallback without romanized name (Москва) | "Moskva" | "Moskva" (unchanged) |
| `getEnName(false)` without `name:en` | empty | empty (unchanged) |

---
Claude Code — Claude Fable 5, xhigh

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/data/MapObject.java
+++ b/OsmAnd-java/src/main/java/net/osmand/data/MapObject.java
@@ -212,9 +212,23 @@ public abstract class MapObject implements Comparable<MapObject> {
 		if (!Algorithms.isEmpty(enName)) {
 			return unzipContent(this.enName);
 		} else if (!Algorithms.isEmpty(getName()) && transliterate) {
-			return TransliterationHelper.transliterate(getName());
+			String latinScriptName = getLatinScriptName();
+			return latinScriptName != null ? latinScriptName : TransliterationHelper.transliterate(getName());
 		}
 		return ""; //$NON-NLS-1$
+	}
+
+	// mapper-provided romanization (name:ja-Latn, name:ko-Latn, name:sr-Latn, ...) reads correctly,
+	// unlike machine transliteration which e.g. renders Japanese kanji with Chinese readings
+	private String getLatinScriptName() {
+		if (names != null) {
+			for (Entry<String, String> entry : names.entrySet()) {
+				if (entry.getKey().toLowerCase().endsWith("-latn") && !Algorithms.isEmpty(entry.getValue())) {
+					return unzipContent(entry.getValue());
+				}
+			}
+		}
+		return null;
 	}
 
 	public void setEnName(String enName) {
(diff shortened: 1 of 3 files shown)


vshcherb:
It loosk that it completely ignored  and alos getName is only used for POI not for map so solution is not generic... I Will try to provide a better apporach

Current code
```
	public static String transliterate(String text) {
		if (japanese) {
			// do not transliterate japanese for now
			//return japanese2Romaji(text);
			return text;
		} else {
			return Junidecode.unidecode(text);
		}
	}```

PR #25445, opened 2026-07-15
Enhance MSL altitude processing and caching for Android 14+

dmpr0:
### Problem

OsmAnd derives MSL altitude from `Location.getAltitude()` (WGS84) plus an 
optional offline geoid correction file ("World altitude correction"). Without 
the file, users see raw WGS84 altitude — off by the local geoid height 
(~26 m in Kyiv, up to ±100 m globally).

Android 14 (API 34) added `Location.hasMslAltitude()` / `getMslAltitudeMeters()`, 
which provide a ready-to-use MSL altitude (populated primarily by the Fused 
Location Provider), making the correction file unnecessary on supported devices.

### Changes

All changes are localized to `OsmAndLocationProvider.convertLocation()`.

**1. System MSL altitude (API 34+).**
When a fix carries MSL altitude, it is used directly and geoid correction is 
skipped — the value is already MSL; applying correction would double-correct.

**2. Dynamic geoid offset cache.**
Providers deliver fixes with and without MSL interleaved (e.g. fused with MSL, 
network without), which caused the displayed altitude to jump between MSL and 
raw WGS84 values. When a fix has both WGS84 and MSL altitude, the difference 
(local geoid offset) is cached together with the calibration coordinates. 
Fixes without MSL apply this cached offset as a fallback.
- Offset is invalidated beyond `DYNAMIC_GEOID_OFFSET_MAX_DISTANCE` (100 km) 
  from the calibration point (geoid height varies with position; within 100 km 
  the variation is typically < 1 m).
- Calibration is skipped when `getMslAltitudeAccuracyMeters()` exceeds 
  `MSL_CALIBRATION_MAX_ACCURACY` (10 m), so a single low-quality fix cannot 
  poison the offset.
- The cache is in-memory only (not persisted) — a stale persisted offset after 
  a long flight would be worse than a few uncorrected fixes on startup.

**3. Correction priority** (documented in code):
system MSL → geoid correction file → cached dynamic offset → raw WGS84.
The file takes precedence over the cached offset because file correction is 
computed for the current coordinates, while the cached offset is tied to the 
calibration point.

**4. Fix: stale geoid correction after file deletion.**
`getGeoidHeight()` silently returned 0 when the correction file had been 
deleted while the app was running, because ResourceManager keeps the 
`GeoidAltitudeCorrection` instance alive. Added an 
`isGeoidInformationAvailable()` check so an unavailable file no longer 
produces a fake zero correction (which also blocked the cached-offset 
fallback). The underlying ResourceManager caching issue is tracked separately: #XXXX.

**5. Consistent vertical accuracy.**
When MSL altitude is used, `getMslAltitudeAccuracyMeters()` (if available) 
overrides the generic WGS84 vertical accuracy, so altitude and its accuracy 
always come from the same source.

**6. Diagnostic logging.**
Each altitude source (MSL / FILE / CACHED / RAW / NONE) is logged once per 
session on first activation, plus offset (re)calibration and distance 
invalidation events. This identifies from user logs which altitude sources a 
device provides, without per-fix spam from provider flapping.

### Compatibility

- **Pre-API-34 devices:** no behavior change — same file-correction / raw path.
- **Builds without GMS (F-Droid):** `hasMslAltitude()` is a framework API 
  (compiles and runs fine) but is typically only populated by the Fused 
  Location Provider, so it stays false → existing behavior. The correction 
  file flow is intentionally kept as the primary mechanism for these builds.

### Testing

Device: [MODEL], Android 14, GMS build:
- Without correction file: altitude stable at MSL value (~135 m vs 161 m raw 
  WGS84 previously); interleaved no-MSL fixes handled by cached offset.
- With correction file downloaded: FILE branch active, value consistent with MSL.
- File deleted at runtime: correct fallback to cached offset without restart.
- [ ] API 33 emulator smoke test (pre-34 path unchanged).

### DIFF
--- a/OsmAnd/src/net/osmand/plus/OsmAndLocationProvider.java
+++ b/OsmAnd/src/net/osmand/plus/OsmAndLocationProvider.java
@@ -104,10 +104,7 @@ public class OsmAndLocationProvider implements SensorEventListener {
 	private static final float MSL_CALIBRATION_MAX_ACCURACY = 10f;
 	private static final int DYNAMIC_GEOID_OFFSET_MAX_DISTANCE = 100_000;
 	
-	private static Double cachedDynamicGeoidOffset = null;
-	private static Double cachedDynamicGeoidOffsetLat = null;
-	private static Double cachedDynamicGeoidOffsetLon = null;
-	private static final java.util.Set<String> loggedAltitudeSources = new java.util.concurrent.CopyOnWriteArraySet<>();
+	private static Location cachedMLSGeoidLocation = null;
 
 
 	private long lastTimeGPSLocationFixed;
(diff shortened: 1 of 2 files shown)


OsmAnd/src/net/osmand/plus/OsmAndLocationProvider.java
```diff
@@ -571,20 +579,77 @@ public static net.osmand.Location convertLocation(Location l, OsmandApplication
 		if (l.hasSpeed()) {
 			r.setSpeed(l.getSpeed());
 		}
-		if (l.hasAltitude()) {
-			r.setAltitude(l.getAltitude());
+		String branchTaken = "NONE";
+
+		if (VERSION.SDK_INT >= VERSION_CODES.UPSIDE_DOWN_CAKE && l.hasMslAltitude()) {
```
vshcherb: Too much code, what's that branchTaken?

OsmAnd/src/net/osmand/plus/OsmAndLocationProvider.java
```diff
@@ -571,20 +579,77 @@ public static net.osmand.Location convertLocation(Location l, OsmandApplication
...
+		if (!loggedAltitudeSources.contains(branchTaken)) {
+			LOG.info("Altitude source active: " + branchTaken + ", alt=" + r.getAltitude() + ", provider=" + l.getProvider());
+			loggedAltitudeSources.add(branchTaken);
 		}
 		if (l.hasBearing()) {
 			r.setBearing(l.getBearing());
 		}
-		if (l.hasAltitude() && app != null) {
-			double alt = l.getAltitude();
-			GeoidAltitudeCorrection geo = app.getResourceManager().getGeoidAltitudeCorrection();
```
vshcherb, line 582: It's already builtin
```
public float getGeoidHeight(double lat, double lon) {
        if (!this.isGeoidInformationAvailable()) {
            return 0.0F;
        } else {
```

OsmAnd/src/net/osmand/plus/resources/ResourceManager.java
```diff
@@ -1038,7 +1038,12 @@ public GeoidAltitudeCorrection getGeoidAltitudeCorrection() {
 	}
 
 	protected void resetGeoidAltitudeCorrection() {
-		geoidAltitudeCorrection = new GeoidAltitudeCorrection(app.getAppPath(null));
+		GeoidAltitudeCorrection geo = new GeoidAltitudeCorrection(app.getAppPath(null));
+		if (geo.isGeoidInformationAvailable()) {
+			geoidAltitudeCorrection = geo;
+		} else {
+			geoidAltitudeCorrection = null;
```
vshcherb: Why null is correct

OsmAnd/src/net/osmand/plus/OsmAndLocationProvider.java
```diff
@@ -571,20 +579,77 @@ public static net.osmand.Location convertLocation(Location l, OsmandApplication
...
+					branchTaken = "FILE";
+				}
+			}
+			if (!corrected) {
+				Double offset = cachedDynamicGeoidOffset;
+				Double offsetLat = cachedDynamicGeoidOffsetLat;
+				Double offsetLon = cachedDynamicGeoidOffsetLon;
+				if (offset != null && offsetLat != null && offsetLon != null) {
+					double dist = MapUtils.getDistance(offsetLat, offsetLon, l.getLatitude(), l.getLongitude());
+					if (dist < DYNAMIC_GEOID_OFFSET_MAX_DISTANCE) {
```
vshcherb: Constant too high 100km?, if we cache only for

OsmAnd/src/net/osmand/plus/OsmAndLocationProvider.java
```diff
@@ -100,6 +100,14 @@ public interface OsmAndCompassListener {
 	private static final int REQUESTS_BEFORE_CHECK_LOCATION = 100;
 	private final AtomicInteger locationRequestsCounter = new AtomicInteger();
 	private final AtomicInteger staleLocationRequestsCounter = new AtomicInteger();
+	
+	private static final float MSL_CALIBRATION_MAX_ACCURACY = 10f;
+	private static final int DYNAMIC_GEOID_OFFSET_MAX_DISTANCE = 100_000;
+	
+	private static Double cachedDynamicGeoidOffset = null;
```
vshcherb: We can store just Location object instead of 3 doubles

PR #25453, opened 2026-07-16
review fix

### LINKED ISSUE #7125: [Feature request] Show AVNotes within favourites/POIs
When media is taken by "Actions" tab within a saved POI or moved to the exact same location as a POI, it should appear **within** the saved favourites point. The current navigation is impractical as it requires the user to leave the POI, tap somewhere else on the map, before tapping on a bunch of items in order to choose between the POI and one or more photo(s) for example.
When the user deletes the POI, Osmand should ask to delete the images of the same location as well or keep them saved.

0xRe1nk0:


OsmAnd/src/net/osmand/plus/gallery/model/GallerySortMode.kt
```diff
@@ -6,20 +6,30 @@ import net.osmand.plus.R
...
+	val group: Group
 ) {
-	NEAREST(R.string.gallery_sort_nearest, R.drawable.ic_action_nearest_map_center),
-	LAST_MODIFIED(R.string.gallery_sort_last_modified, R.drawable.ic_action_sort_by_date),
-	NAME_A_Z(R.string.gallery_sort_name_a_z, R.drawable.ic_action_sort_by_name_ascending),
-	NAME_Z_A(R.string.gallery_sort_name_z_a, R.drawable.ic_action_sort_by_name_descending),
-	NEWEST_FIRST(R.string.gallery_sort_newest_first, R.drawable.ic_action_sort_date_31),
-	OLDEST_FIRST(R.string.gallery_sort_oldest_first, R.drawable.ic_action_sort_date_1),
+	NEAREST(R.string.gallery_sort_nearest, R.drawable.ic_action_nearest_map_center, Group.LOCATION),
+	LAST_MODIFIED(R.string.gallery_sort_last_modified, R.drawable.ic_action_sort_by_date, Group.LOCATION),
```
Chumva, line 13: Group.LOCATION for LAST_MODIFIED? Should we rename it to something more abstract like DEFAULT/GENERAL?

PR #25487, opened 2026-07-22
Make spatial search cache concurrent for multithread web search

alisa911:


alisa911:

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialTextSearch.java
```diff
@@ -208,8 +210,7 @@ public static SpatialTextSearchSettings suggestionSettings() {
 	}
 
 	public static class SpatialSearchFileCache {
-		public int fileInd = -1; // changing each session - not concurrent !!!
-		public int indexInd = -1; // changing each session - not concurrent !!!
+		public final ReentrantLock lock = new ReentrantLock();
```
vshcherb, line 213: Not correct files are not concurrent

vshcherb:
Search uses always non concurrent files each thread should have own cache! That's why we create limited 4-5 threads each thread should maintain own cache of files

PR #25492, opened 2026-07-23
Expand poi filter groups to child types in spatial poi request

alisa911:


alisa911:

PR #25516, opened 2026-07-28
Improve spatial search android

Corwin-Kh:


Corwin-Kh:
Need to be done
- co-integration with old search (favourites, wpt etc.)
- subsequent search on click on found spatial category

RZR-UA (approved):

PR #25539, opened 2026-07-30
Fix excessive memory usage when loading Cloud file list

### LINKED ISSUE #25392: `java.lang.OutOfMemoryError` during OsmAnd Cloud backup sync and GPX/Favorites rendering
My System 5.3.10 has been crashing repeatedly lately. I have attached some relevant files. Is this happening because I have so many points, or could there be another reason? Thanks for your help.

[exception (7).log](https://github.com/user-attachments/files/29739379/exception.7.log)
[exception (6).log](https://github.com/user-attachments/files/29739378/exception.6.log)
[exception (8).log](https://github.com/user-attachments/files/29739380/exception.8.log)

nazar-kutz:
## Summary

Fixes excessive memory usage during OsmAnd Cloud backup checking by parsing the `/userdata/list-files` response directly from the HTTP `InputStream`.

Previously, a successful response was first converted into a complete `String` and then into a full `JSONObject` tree containing the `allFiles` and `uniqueFiles` arrays. For large Cloud accounts, these temporary representations significantly increased peak memory usage and could contribute to `OutOfMemoryError`.

The new implementation processes `allFiles` sequentially with `JsonReader`, keeping only the current file entry in its temporary JSON representation.

## Changes

- Added `StreamingCloudFileListParser` for the Cloud file-list response.
- Added a compact `CloudFileListParseResult` model.
- Integrated streaming parsing into `BackupHelper.downloadFileList()`.
- Added generic `InputStream` response handling to `AndroidNetworkUtils`.
- Refactored the existing String-based `sendRequest()` to reuse the shared HTTP request implementation.
- Skipped the unused `uniqueFiles` array without materializing it.
- Preserved the existing success, error, callback and exception behavior.
- Ensured success streams, error streams and HTTP connections are always closed.
- Kept non-200 response bodies String-based for compatibility with the existing error-handling flow.
- Did not add a legacy parser fallback, since it could reproduce the original memory problem.

## Resulting flow

```text
HTTP InputStream
→ JsonReader
→ one temporary JSONObject per file entry
→ RemoteFile
→ List<RemoteFile>
````

Instead of:

```text
HTTP InputStream
→ complete response String
→ root JSONObject
→ allFiles and uniqueFiles JSON arrays
→ List<RemoteFile>
```

## Validation

The streaming parser was compared with the legacy parser using the same real Cloud response:

```text
Response size:          1,600,574 bytes
allFiles records:       3,318

Legacy parser files:    3,318
Streaming parser files: 3,318

Root differences:       0
RemoteFile differences: 0
Total differences:      0
Equivalent:             true
```

Three independent memory measurements were performed for both implementations.

Median results:

| Metric                               |    Legacy | Streaming |              Difference |
| ------------------------------------ | --------: | --------: | ----------------------: |
| Maximum Java heap                    |  79.7 MiB |  69.0 MiB |  **−10.7 MiB / −13.4%** |
| Maximum total memory                 | 548.1 MiB | 523.6 MiB |   **−24.5 MiB / −4.5%** |
| Java growth from the pre-Cloud level | +27.6 MiB | +27.5 MiB | approximately unchanged |

The implementation produces the same Cloud file list while reducing the typical memory peak.

## Scope

This change specifically removes the full-response memory amplification during
Cloud file-list loading while preserving the existing downstream backup
processing behavior.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/backup/BackupHelper.java
+++ b/OsmAnd/src/net/osmand/plus/backup/BackupHelper.java
@@ -37,13 +37,11 @@ import net.osmand.plus.settings.backend.backup.items.FileSettingsItem.FileSubtyp
 import net.osmand.plus.settings.backend.preferences.CommonPreference;
 import net.osmand.plus.utils.AndroidNetworkUtils;
 import net.osmand.plus.utils.AndroidNetworkUtils.NetworkResult;
-import net.osmand.plus.utils.AndroidUtils;
 import net.osmand.plus.utils.FileUtils;
 import net.osmand.util.Algorithms;
 import net.osmand.util.CollectionUtils;
 
 import org.apache.commons.logging.Log;
-import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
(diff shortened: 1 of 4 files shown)


Chumva:
Too big a refactoring for ~10% memory improvement in parsing backup response

PR #25554, opened 2026-08-02
Implement: Media - Share favorites bottom sheet

nazar-kutz:


OsmAnd/src/net/osmand/plus/myplaces/favorites/dialogs/share/ShareFavoritesDataPreparer.kt
```diff
@@ -0,0 +1,130 @@
...
+				packedMedia.size + 1,
+				totalSize,
+				hasMissingMedia
+			)
+		} catch (e: Exception) {
+			if (isCancelled()) {
+				return PreparationResult.Cancelled
+			}
+			LOG.error(
+				"Failed to prepare attached media for Favorites; falling back to GPX: ${destination.absolutePath}",
```
Chumva, line 101: If we get an error here, should we also notify the user about it? Not silently switch to GpxOnly sharing mode

OsmAnd/src/net/osmand/plus/myplaces/favorites/dialogs/share/ShareFavoritesFileSession.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import net.osmand.plus.OsmandApplication
+import net.osmand.plus.utils.FileUtils
+import net.osmand.util.Algorithms
+import java.io.File
+import java.util.UUID
+
+class ShareFavoritesFileSession(app: OsmandApplication) {
+
+	private val sessionId = UUID.randomUUID().toString()
```
Chumva, line 11: Do we really need a UUID sessionId for this file sharing?

OsmAnd/src/net/osmand/plus/myplaces/favorites/dialogs/share/ShareFavoritesBottomSheet.kt
```diff
@@ -0,0 +1,136 @@
...
+
+	override fun onAskRefreshDialogCompletely(processId: String) {
+		if (processId == ShareFavoritesController.PROCESS_ID && view != null) {
+			updateContent()
+		}
+	}
+
+	override fun onAskDismissDialog(processId: String) {
+		if (processId == ShareFavoritesController.PROCESS_ID) {
+			dismiss()
```
Chumva, line 76: Should we use dismissAllowingStateLoss to avoid an IllegalStateException if this is called after onPostExecute?

OsmAnd/src/net/osmand/plus/myplaces/favorites/dialogs/share/ShareFavoritesFileSession.kt
nazar-kutz: I would keep the UUID. It is not required to prevent concurrent preparation, but it isolates files after the Android share flow has been launched.

The content:// URI points to the actual GPX or OSF file rather than to a snapshot of its contents. A receiving app may open that URI later, while OsmAnd can already start another share session. Without a unique directory, a subsequent export—especially for the same Favorites folder—could overwrite the file referenced by the previous URI before or while it is being read.

Together with the existing cleanup behavior, a per-session UUID prevents this race condition. Its cost is only one additional directory level, so I think it is justified.

OsmAnd/src/net/osmand/plus/myplaces/favorites/dialogs/share/ShareFavoritesDataPreparer.kt
nazar-kutz: Fixed in https://github.com/osmandapp/OsmAnd/pull/25561/changes/41ae23186414cbcdd4714857f996c72491f80073

OsmAnd/src/net/osmand/plus/myplaces/favorites/dialogs/share/ShareFavoritesBottomSheet.kt
nazar-kutz: Fixed in https://github.com/osmandapp/OsmAnd/pull/25561/changes/2b113ebe282d1dc9cc7581fd1cd42347096293d8

PR #25571, opened 2026-08-04
Fix lost poi results in category tile filter

alisa911:


alisa911:

PR #25574, opened 2026-08-04
Fix compass-dependent UI updates

### LINKED ISSUE #25532: Critical No updates on compass related widgets
Looks like when FPS drop to 1-0 (my location not on map screen) - arrows are not rotated (only if you keep scrolling)
- Map Markers 
- Radius Ruler
- Search results

nazar-kutz:


Chumva:
Should we update not only widgets that depend on Compass, but all of them if the map is not moving?

nazar-kutz:
@Chumva I agree that dynamic widgets should not rely on map movement to update.

For this fix, I suggest keeping the scope focused on the affected compass-dependent components identified in the issue. As a follow-up, we can:

1. inventory dynamic map widgets and their update dependencies;
2. separate compass, location, time, and map-state updates;
3. dispatch updates only to visible widgets that depend on each signal, with appropriate throttling.

This would provide consistent updates while avoiding a full map refresh or unnecessary work for static and viewport-only widgets.

OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
```diff
@@ -192,6 +211,57 @@ public void setMapActivity(@Nullable MapActivity mapActivity) {
...
+				compassRefreshScheduled = true;
+				compassRefreshView = mapView;
+				mapView.postDelayed(compassRefreshRunnable, COMPASS_REFRESH_INTERVAL_MS - elapsedTime);
+			}
+		}
+	}
+
+	private void refreshCompassRuler(long refreshTime) {
+		lastCompassRefreshTime = refreshTime;
+		view.refreshMap();
```
Chumva, line 242: Won't it lead to 10 whole map updates per second?

OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
```diff
@@ -46,12 +48,14 @@
...
 import java.util.List;
 
-public class RadiusRulerControlLayer extends OsmandMapLayer {
+public class RadiusRulerControlLayer extends OsmandMapLayer implements OsmAndCompassListener {
 
 	private static final int TEXT_SIZE = 14;
 	private static final float COMPASS_CIRCLE_FITTING_RADIUS_COEF = 1.25f;
 	private static final float CIRCLE_ANGLE_STEP = 5;
 	private static final int SHOW_COMPASS_MIN_ZOOM = 8;
+	private static final float MIN_COMPASS_HEADING_CHANGE = 1f;
```
Chumva: Should we try to reuse existing constants like [MapViewTrackingUtilities.COMPASS_HEADING_THRESHOLD](https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/plus/base/MapViewTrackingUtilities.java#L59)?

nazar-kutz: Fixed in https://github.com/osmandapp/OsmAnd/pull/25574/commits/3f5cc2e002afc280c9b5cbf2a62af8017e87e516

Issue #24809, opened 2026-04-09
Web displays invisible tags for favorites or has broken UI


Rowin63:
### Describe the idea

As the title says - an option to delete all WIKI content (all non-elementary content) of a favorite.
This will reduce the overall favorites database-size immensely.

Main content of a favorite: 
name, icon, appearance, address, comment. 
Everyting else is optional/non-elementary and should be possible to delete easily in the edit menu.
Keeps the favorite database thin.

### Expected behaviour

In the editing area, give an option to delete e.g. all Wiki content.
It´s not possible to edit a favorite like this (inside the app).

### Alternatives you've considered

Editing each single favorite by an external editor. Very cumbersome, very impractical.

Rowin63:
You don´t see all this in the app, but in Osmand web, like e.g "short description" to this example favorite in 30 languages, including chinese, japanese, korean, swaheli, persian, arabic, etc.

That´s not good.

This is the content of Al Hoceima, one favorite picked from the map and saved on the device. 
Only this single favorite is almost 500KB. Perfect waste of recources.

[Al Hoceima favorite content.txt](https://github.com/user-attachments/files/26614481/Al.Hoceima.favorite.content.txt)

vshcherb:
Please give an example of the path @Zirochkabila and check if it's covered by ongoing task of chaing Web Favorite Editor so we can merge it.

Zirochkabila:
@Rowin63 I tried creating a “Favorite” pin for various cities (London, Berlin, Paris, including Al Hoceima ⵍⵃⵓⵙⵉⵎⴰ الحسيمة (35.24512° N, 3.93019° W)) and with popular landmarks (Eiffel Tower, Big Ben, Brandenburg Gate), both with and without the downloaded Wikipedia articles - the file size does not exceed 5 KB.
Please describe the steps to reproduce this - perhaps I’m missing something.

dmpr0:
### Reproduced

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/cb8c86e3-c814-4381-9928-56fea60abce2"  width="500" height="auto"/> </td>
  </tr>
</table>   


File – [favorites-Morocco.gpx.zip](https://github.com/user-attachments/files/27116331/favorites-Morocco.gpx.zip)



**How to reproduce:**
* Download full map for Morocco
* Download Wikipedia for Morocco
[...]

PR #25577, opened 2026-08-05
Refactor tag filtering logic

Dima-1:
https://github.com/osmandapp/OsmAnd/issues/24809


OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,180 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+				}
+				if (type instanceof PoiType pt) {
+					categoryTypes.add(pt);
+				}
+			}
+			if (categoryTypes.isEmpty()) {
+				continue;
+			}
+			String poiAdditionalCategoryName = categoryTypes.get(0).getPoiAdditionalCategory();
+			// NB: intentionally comparing the collapsable_-prefixed key against bare Amenity.CUISINE/DISH,
```
alisa911: Please don't carry known bugs into the new code — file an issue instead, a plain comment will get lost. Also the comment refers to
AmenityUIHelperSnapshotTest, which doesn't exist in this branch. And when this is fixed, note the server side (groupLocalizedTags) doesn't map collapsableExtraRow, so web would silently diverge from Android.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
alisa911: setShowDefault is now dead: buildInternal takes the flag from
isDefaultForCategory() inside the bundle and never reads this field, but AmenityMenuBuilder:248 still calls the setter. Remove both, or wire the external flag into getVisibleTags — right now a caller that sets it gets silently ignored.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,180 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+		Map<String, String> filtered = new LinkedHashMap<>();
+		boolean anySkipped = false;
+		for (Map.Entry<String, String> loc : ((Map<String, String>) localizationsObj).entrySet()) {
+			if (isKeyToSkip(loc.getKey())) {
+				anySkipped = true;
+			} else {
+				filtered.put(loc.getKey(), loc.getValue());
+			}
+		}
+		if (anySkipped) {
```
alisa911: Why does one skipped localization key remove ALL keys without a lang suffix? This needs a comment or a test — the intent is impossible to guess. 
Also note the header is now picked from the filtered map, while the old code picked it from the full one, so the header choice can change.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -100,120 +91,128 @@ public void setPreferredLang(String lang) {
...
 		}
+		return headerBuilder.build();
+	}
+
+	private boolean isNoteKeyHiddenFromEditing(@NonNull String key) {
+		return "note".equals(key) && !osmEditingEnabled;
+	}
+
+	@NonNull
+	private AmenityRowData.Builder getRowDataBuilder(@NonNull Context context, @NonNull String key, @NonNull String value) {
```
alisa911: getRowDataBuilder never returns null, but the old createPoiAdditionalInfoRow returned null for filter-only types — so localized children and the extra row that used to be skipped are now always built. Please confirm this is intended.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -100,120 +91,128 @@ public void setPreferredLang(String lang) {
...
-			}
-			if (infoRow != null) {
-				if (lastBuiltRowIsDescription) {
-					descriptions.add(infoRow);
-				} else if (Amenity.CUISINE.equals(key)) {
-					cuisineRow = infoRow;
-				} else if (poiType == null) {
-					infoRows.add(infoRow);
-				}
+			boolean isDescription = amenityRow.isText && amenityRow.iconId == R.drawable.ic_action_info_dark;
```
alisa911: The "isText + info icon" description check is duplicated here and in getRowDataBuilder:213 — extract one method, or better set an isDescription flag in the shared code so the server can use it too.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -100,120 +91,128 @@ public void setPreferredLang(String lang) {
...
+			String poiAdditionalCategoryName = baseRow.key;
+			String poiAdditionalIconName = poiTypes.getPoiAdditionalCategoryIconName(poiAdditionalCategoryName);
+			String iconName = resolveExistingIconName(context,
+					poiAdditionalIconName, poiAdditionalCategoryName, firstType.getIconKeyName());
+			int iconId = iconName == null ? R.drawable.ic_action_note_dark : 0;
+			return AmenityRowsBuilder.buildPoiTypesGroupRow(poiAdditionalCategoryName,
+					firstType.getKeyName(), firstType.getPoiAdditionalCategoryTranslation(), categoryTypes,
+					firstType.getOrder(), iconId, iconName, extraRow, true, baseRow.collapsableCategory);
+		}
+		PoiCategory groupCategory = baseRow.collapsableCategory;
```
alisa911: This "last category wins" loop also exists in
AdditionalInfoBundle.getVisibleTags:190 — same logic in two places, and taking the category of the last poi type looks accidental; compute it once in the shared code and add a comment why the last one.

OsmAnd-java/src/main/java/net/osmand/data/AmenityRowData.java
```diff
@@ -0,0 +1,150 @@
+package net.osmand.data;
+
+import net.osmand.osm.PoiCategory;
+import net.osmand.osm.PoiType;
+import net.osmand.util.Algorithms;
+
+import java.util.List;
+
+public class AmenityRowData {
```
alisa911: AmenityRowData is UI wording ("row") in the shared data layer — rename to something like AmenityTagEntry before it spreads to all three platforms.
Same applies to the UI-only fields it carries (textColor,
matchWidthDivider, textLinesLimit) — the server sees them but can never use them.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,180 @@ public Map<String, Object> getFilteredLocalizedInfo() {
 		return localizedAdditionalInfo;
 	}
 
-	public boolean shouldDisplayKey(String key) {
+	private static final String CUISINE_INFO_ID = COLLAPSABLE_PREFIX + Amenity.CUISINE;
+	private static final String DISH_INFO_ID = COLLAPSABLE_PREFIX + Amenity.DISH;
+
+	public List<AmenityRowData> getVisibleTags(boolean allowNoteTag) {
```
alisa911: getVisibleTags is ~100 lines handling three distinct cases — split into collectPlainRows / collectCollapsableGroups / collectPoiTypeGroups, same way as suggested for the server mapper.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -100,120 +91,128 @@ public void setPreferredLang(String lang) {
...
-		for (AmenityInfoRow info : infoRows) {
-			buildAmenityRow(view, info);
+	@Nullable
+	private AmenityRowData buildLocalizedRowData(@NonNull Context context, @NonNull AmenityRowData baseRow) {
+		Map<String, String> localizedAdditionalInfo = new LinkedHashMap<>();
+		for (AmenityRowData child : baseRow.collapsableRows) {
+			localizedAdditionalInfo.put(child.key, child.value);
+		}
+		Collection<String> availableLocales = collectAvailableLocalesFromTags(localizedAdditionalInfo.keySet());
+		Locale prefferedLocale = getPreferredLocale(availableLocales);
```
alisa911: Nit: "prefferedLocale" typo carried over from the old code — good moment to fix it.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/rows/behaviour/DefaultPoiAdditionalRowBehaviour.kt
```diff
@@ -41,30 +41,38 @@ open class DefaultPoiAdditionalRowBehaviour : IPoiAdditionalRowBehavior {
...
                                     iconName = parentType.osmTag + "_" + parentType.osmValue
-                                    builder.setIcon(menuBuilder.getRowIcon(context, iconName))
+                                    resolvedIconName = iconName
+                                    deepIconFound = menuBuilder.getRowIcon(context, iconName) != null
                                 }
                             }
                         }
                     }
                     builder.setIconId(iconId)
+                    builder.setIconName(resolvedIconName)
```
alisa911: Nit: iconName is set even when iconId was found, and resolvedIconName may hold a candidate that failed the lookup — confusing to read; set iconName only when it's the actual result.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -245,125 +244,76 @@ private void initVariables() {
...
-		if (poiType == null && pType == null) {
-			poiType = poiTypes.getPoiTypeByKey(key);
-		}
-		if (pType == null) {
-			String altKey = key.replaceAll(":", "_");
-			pType = additionalInfo.getPoiAdditionalType(altKey, vl);
-			poiType = poiCategory.getPoiTypeByKeyName(altKey);
-			if (poiType == null && pType == null) {
-				poiType = poiTypes.getPoiTypeByKey(altKey);
+	private String resolveExistingIconName(@NonNull Context context, String... candidates) {
```
alisa911: Each candidate is resolved via getRowIcon just to check existence, and the winner is resolved again in toAmenityInfoRow — resolve once and reuse the Drawable.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/1cd15a449fe9fe2a49360b36cb0f7d188205de68

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/3341b849b1bb115f7ad697f237af25d9fd35fca1

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/6e12065119a5b0e4d179153e2dac878571603f00

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/3c1e121efc3a66cf7714355a4042a23f5b3dd61f

OsmAnd-java/src/main/java/net/osmand/data/AmenityRowData.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/9e9ad641eb907ef6d4d04313418d82e7d4b92ce6

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/27dc8e8d3d6802242632443d332da2db7233528f

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/bfcc01ac5523394e5161a39d522c239a001dda87

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/f7a1d5a89dd86480f3905cce37bd7758f15f666a

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/f7a1d5a89dd86480f3905cce37bd7758f15f666a

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/rows/behaviour/DefaultPoiAdditionalRowBehaviour.kt
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/627ebbce0f2de889a5dea6748082474b534bdcd3

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/627ebbce0f2de889a5dea6748082474b534bdcd3

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,208 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+					String categoryKey = categoryType.getCategory().getKeyName();
+					if (!MapPoiTypes.OTHER_MAP_CATEGORY.equals(categoryKey)) {
+						collectedPoiTypes.computeIfAbsent(categoryKey, c -> new ArrayList<>()).add(categoryType);
+					}
+				} else {
+					entries.add(new AmenityTagEntry.Builder(key).setValue(strValue).setOrder(PoiType.DEFAULT_ORDER)
+							.setResolvedType(resolvedType).setIsDescription(key.contains(Amenity.DESCRIPTION)).build());
+				}
+			} else if (value instanceof Map) {
+				Object filtered = filterLocalizations((Map<String, Object>) value);
```
RZR-UA: Unchecked cast: 'java.lang.Object' to 'java.util.Map<java.lang.String,java.lang.Object>'. Please never use type casting in this way. Get the Map<> of the value inside if(). Then why did you use Object as the return type for filterLocalizations? It never returns Object!

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -20,10 +20,12 @@
 import net.osmand.util.Algorithms;
 import net.osmand.util.CollectionUtils;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
```
RZR-UA: Please remove HashMap from this class, completely.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,208 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+		if (filtered.isEmpty()) {
+			return null;
+		}
+		Map<String, Object> result = new HashMap<>();
+		result.put("localizations", filtered);
+		return result;
+	}
+
+	private boolean shouldDisplayKey(String key) {
 		AbstractPoiType t = poiTypes.getAnyPoiAdditionalTypeByKey(key);
```
RZR-UA: Move the getAnyPoiAdditionalTypeByKey() check to the end of this method, because this call might be slow.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,208 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+			PoiCategory category, Map<String, List<PoiType>> collectedPoiTypes) {
+		boolean showDefaultTags = isDefaultForCategory();
+		List<AmenityTagEntry> entries = new ArrayList<>();
+		AmenityTagEntry cuisineEntry = null;
+
+		for (Map.Entry<String, Object> entry : getFilteredLocalizedInfo().entrySet()) {
+			String key = entry.getKey();
+			if (!shouldDisplayKey(key) || isKeyToSkip(key)) {
+				continue;
+			}
```
RZR-UA: Swap if's || elements because isKeyToSkip() is much faster than shouldDisplayKey().

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,208 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+					}
+				}
+			}
+		}
+		return children.get(0);
+	}
+
+	private boolean isFilterOnlyOrGrouped(PoiType pType) {
+		return pType != null && (pType.isFilterOnly()
+				|| (!pType.isText() && !Algorithms.isEmpty(pType.getPoiAdditionalCategory())));
```
RZR-UA: Please use Kotlin-style Algorithms.isNotEmpty() instead of `!Algorithms.isEmpty()` to avoid mistakes. Everywhere in the request.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,208 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+				continue;
+			}
+			if (additionalType == null && categoryType == null && !showDefaultTags) {
+				continue;
+			}
+
+			if (strValue != null) {
+				if (additionalType != null) {
+					AmenityTagEntry tagEntry = new AmenityTagEntry.Builder(key).setValue(strValue).setOrder(additionalType.getOrder())
+							.setResolvedType(resolvedType).setIsDescription(key.contains(Amenity.DESCRIPTION)).build();
```
RZR-UA: Follow 120 characters per line, please. Everywhere.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,208 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+					.setCollapsablePoiTypes(poiTypeList)
+					.setCollapsableCategory(category)
+					.setPoiAdditional(false)
+					.setOrder(PoiType.DEFAULT_GROUP_ORDER)
+					.build());
+		}
+		return entries;
+	}
+
+	@SuppressWarnings("unchecked")
```
RZR-UA: Avoid any of SuppressWarnings! Use `if (... instanceof ... that) { ... }` or the like.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,208 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+		return pType != null && (pType.isFilterOnly()
+				|| (!pType.isText() && !Algorithms.isEmpty(pType.getPoiAdditionalCategory())));
+	}
+
+	private boolean isDefaultForCategory() {
+		PoiCategory category = getCategory();
+		String subtype = get(SUBTYPE);
+		if (category == null || Algorithms.isEmpty(subtype)) {
+			return false;
+		}
```
RZR-UA, line 278: Split into two if() statements to increase performance in case the category is null.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -81,7 +83,208 @@ public Map<String, Object> getFilteredLocalizedInfo() {
...
+		PoiType poiType = category.getPoiTypeByKeyName(subtype);
+		return poiType != null && poiType.isDefaultForCategory();
+	}
+
+	@SuppressWarnings("unchecked")
+	private Map<String, Object> filterLocalizations(Map<String, Object> value) {
+		Object localizationsObj = value.get("localizations");
```
RZR-UA: Create const for "localizations". Use it here and there (AmenityMenuBuilder.java, AmenityUIHelper.java). There are 5 "localizations" to replace with the const.

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
```diff
@@ -150,10 +353,37 @@ public PoiType getPoiAdditionalType(String key, String vl) {
...
+
+	public static final class ResolvedPoiType {
+		public final PoiType additionalType;
+		public final PoiType categoryType;
+
+		private ResolvedPoiType(PoiType additionalType, PoiType categoryType) {
+			this.additionalType = additionalType;
+			this.categoryType = categoryType;
+		}
+	}
```
RZR-UA: Could it be a simple record?

OsmAnd-java/src/main/java/net/osmand/data/AmenityTagEntriesBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+		for (AmenityTagEntry desc : descriptions) {
+			if (desc.key.length() > langSuffix.length() && desc.key.endsWith(langSuffix)) {
+				descInPrefLang = desc;
+				break;
+			}
+		}
+		if (descInPrefLang != null) {
+			descriptions.remove(descInPrefLang);
+			descriptions.add(0, descInPrefLang);
+		}
```
RZR-UA: You can avoid this if() by using an iterator instead of a for-loop and placing remove/add/break directly in the loop.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -60,160 +58,146 @@
...
-import java.util.Map.Entry;
 
 
 public class AmenityUIHelper extends MenuBuilder {
 
 	public static final Log LOG = PlatformUtil.getLog(AmenityUIHelper.class);
 
-	private static final String CUISINE_INFO_ID = COLLAPSABLE_PREFIX + "cuisine";
-	private static final String DISH_INFO_ID = COLLAPSABLE_PREFIX + "dish";
 	public static final String US_MAPS_RECREATION_AREA = "us_maps_recreation_area";
```
RZR-UA: Remove US_MAPS_RECREATION_AREA

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -60,160 +58,146 @@
...
 
 	private final AdditionalInfoBundle additionalInfo;
 
-	private String preferredLang;
 	private Amenity wikiAmenity;
```
RZR-UA: wikiAmenity is never assigned, please check it

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -60,160 +58,146 @@
...
 		Context context = view.getContext();
-		List<AmenityInfoRow> infoRows = new LinkedList<>();
-		List<AmenityInfoRow> descriptions = new LinkedList<>();
-		Map<String, Object> filteredInfo = additionalInfo.getFilteredLocalizedInfo();
-		for (Entry<String, Object> entry : filteredInfo.entrySet()) {
-			String key = entry.getKey();
-			Object value = entry.getValue();
-			if (!additionalInfo.shouldDisplayKey(key)) {
+		List<AmenityTagEntry> infoEntries = new LinkedList<>();
+		List<AmenityTagEntry> descriptions = new LinkedList<>();
```
RZR-UA: Switch both to ArrayList, which is faster in this case.

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
```diff
@@ -60,160 +58,146 @@
...
+			return null;
+		}
+		if (headerBuilder.getCollapsableEntryType() == AmenityTagEntry.CollapsableEntryType.NONE) {
+			headerBuilder.setCollapsableEntries(localizedEntries);
 		}
+		return headerBuilder.build();
+	}
+
+	private boolean isNoteKeyHiddenFromEditing(@NonNull String key) {
+		return "note".equals(key) && !osmEditingEnabled;
```
RZR-UA: Please add a const for "note". 6x times in the Android project,

RZR-UA:
@Dima-1, in addition, please open every modified file in Android Studio and check every warning.

Never used, never assigned, could be final, annotation, etc. - everything should be resolved manually.

Finally, add Unit tests to cover typical cases such as POI hidden, sorting order, collapsible order, etc.

- [x] fix IDE warnings
- [x] add Unit tests

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/b06f745d3530d521eab02c6931208d6f00433ade

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/b06f745d3530d521eab02c6931208d6f00433ade

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/4ed8ccb8661220efd53ae383395a8ca8a1b29f8c

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/4ed8ccb8661220efd53ae383395a8ca8a1b29f8c

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/4ed8ccb8661220efd53ae383395a8ca8a1b29f8c

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/4ed8ccb8661220efd53ae383395a8ca8a1b29f8c

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/4ed8ccb8661220efd53ae383395a8ca8a1b29f8c

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/4ed8ccb8661220efd53ae383395a8ca8a1b29f8c

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/4ed8ccb8661220efd53ae383395a8ca8a1b29f8c

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/b06f745d3530d521eab02c6931208d6f00433ade

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/4ed8ccb8661220efd53ae383395a8ca8a1b29f8c

OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/baf8c584e98a01ec1641622f284b5397ea3a193c

OsmAnd-java/src/main/java/net/osmand/data/AmenityTagEntriesBuilder.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/baf8c584e98a01ec1641622f284b5397ea3a193c

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/baf8c584e98a01ec1641622f284b5397ea3a193c

OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java
Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/baf8c584e98a01ec1641622f284b5397ea3a193c

OsmAnd-java/src/main/java/net/osmand/data/AmenityTagEntriesBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+	private AmenityTagEntriesBuilder() {
+	}
+
+	public static AmenityTagEntry buildPoiTypesGroupEntry(String key, String name, String textPrefix,
+                  List<PoiType> types, int order, int iconId, List<String> iconNameCandidates,
+                  int fallbackIconId, boolean poiAdditional, PoiCategory collapsableCategory) {
+		StringBuilder text = new StringBuilder();
+		for (PoiType pt : types) {
+			String translation = pt.getTranslation();
+			if (!text.isEmpty()) {
```
RZR-UA: Get back length() > 0 to avoid (API added in Android 35, while OsmAnd supports API 24)

Dima-1: https://github.com/osmandapp/OsmAnd/pull/25577/commits/dbe8c169d60a8628f0ae7fd0a66ccb64a6e340f9

PR #25611, opened 2026-08-09
Add BLE power meter support

minecraft2048:
I just bought a Magene P515 bike power meter, thought that OSMAnd external sensors plugin will just work with it. It doesn't, even when Magene and OSMAnd both supposedly supports ANT+ protocol. No idea how to debug it, and ANT+ seems to be going to be deprecated anyway.

In comparison BLE cycling power profile have good documentation and good tooling (nRF Connect), so I've implemented support for BLE cycling power profile on this PR. 

This PR currently only implements power and cadence reading as those two were sent from my power meter.
Cadence is calculated/decoded with copy pasted code from `BLEBikeSensor.java` as the BLE specs for both cadence sensor and power sensor sends the cadence data with the same format. 

Tested working with Magene P515 power meter and Xiaomi Android phone

Also found a bug where OSMAnd track analyzer can't display power graph, fixed that bug


vshcherb:
Needs to be tested - gpx recording first, then we coud add  widgets

Issue #20624, opened 2024-08-16
Initially wrong numbers in Elevation Profile widget with intermediate destinations


mikehgentry:
### Description

The numbers reported by the Elevation Profile widget are wrong, tapping on it seems to fix them, sometimes they fix themselves of their own accord.

### Steps to reproduce

Create a new profile (I used Moped as a base for this test).

Long press somewhere and select 'Navigation'.

Tap in the map, long press somewhere again, choose 'Navigation' and add as another destination (doesn't seem to matter whether subsequent or intermediate).

### Actual result

When the route finishes calculating, the elevation profile will report incorrect numbers. Tapping on the widget updates it to seemingly correct numbers. At least once it spontaneously fixed itself for me, not sure what triggered that.

### Expected result

Numbers are always correct.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: +4.6.8
Android/iOS version: 11
Device model: XCover 4s
Crash-Logs: ?
```

yuriiurshuliak:
Please provide the exact coordinates where you noticed the incorrect data in the widget. Also specify the correct data that should be displayed. Additionally, a video or screenshot showing the bug in action would be helpful for further investigation.

mikehgentry:
It happens absolutely anywhere, it seems to be a UI problem not a data problem. Sometimes it seems like it is using the data up to the first intermediate, sometimes the numbers are inexplicable. It might decrease 'total ascent' by 5 meters, or it might double it from 300 to 600.

Unfortunately as I said it happens more on longer routes, and those tend to crash OsmAnd or the screen recorder on my hardware, so I'm having trouble recording it. Here's a trivial example though - total ascent does down by 2 on tapping.


https://github.com/user-attachments/assets/3a8644e8-6c46-4b57-a651-4032cf42513a

scaidermern:
Another strange thing: The numbers for uphill/downhill differ between the elevation profile widget and the navigation widget. I can reproduce this here (but not the bug with the totally wrong numbers in the elevation profile widget)

mikehgentry:
I didn't even notice the discrepancy between 864 and 28-something, was busy trying to get a video to record without crashing. But you're right, that's definitely significantly broken!

mikehgentry:
[...]

Issue #24916, opened 2026-04-23
Split interval analysis corrupts Track analysis


sonora:
Reproduce as follows:
1. Select any track (recorded, imported from a hiking guide etc.) as visible on the map. (Any one having one segment of trk points will do.)
2. Tap it, and take a note of its "Distance" value - let's say it shows 10.2 km. (The value is consistent when checked on a "What's here" selection dialog, or directly on the track's context card, or e.g. top left on the tracks "Split interval" analysis screen.)
4. Now on the track's context card, select "Options > Analyze by intervals", and set the split interval to e.g. 1 km. Then tap the back button.
5. Once back on the map screen, repeat point 2 above again. I find that the overall "distance" value displayed now for the same track has decreased, like e.g. now only showing as "9.94 km" instead of "10.2 km", in all places. The mistake persists until you restart the app (seems fortunately not written to the track itself, and perhaps also not to the track db, but not sure).

Looks like perhaps we miss resetting relevant data in variables, and somehow carry numbers used in the split interval analysis over into the overall track data shown.

Not sure if this has any impact on other track related issues we have, like e.g. #24078.

yuriiurshuliak:
> OsmAnd~ 5.4.0#7677m, released: 2026-05-01

The issue is that after using **Analyze by intervals** for a visible GPX track, OsmAnd recalculates or displays the total track distance incorrectly. The original distance is shown correctly before interval analysis, but after returning to the map, the same track shows a much smaller distance until the app is restarted.

The bug has been reproduced.

<details><summary>Steps to reproduce</summary>

1. Open OsmAnd.
2. Select any GPX track that has one segment of track points.
3. Make the track visible on the map.
4. Tap the track on the map.
5. Check and note the displayed **Distance** value.

   * In my case, the track distance was **10.8 km**.
   * The same distance was displayed in the **Overview**, **Track**, and **Split interval analysis** screens.
6. Open the track context menu.
7. Tap **Options → Analyze by intervals**.
8. Set the split interval to a value such as **1 km**.
9. Tap the back button to return to the map.
10. Tap the same track again.
11. Check the displayed **Distance** value again.

</details>

<details><summary>Actual result</summary>

The total track distance becomes incorrect and decreases.

In my case, the distance changed from **10.8 km** to **738 m**.

The incorrect value is shown until the app is restarted.

https://github.com/user-attachments/assets/bf0eae9b-5b43-4e9f-9599-acceff6db457

</details>

[...]

PR #25624, opened 2026-08-10
use correct index offset in ElevationDiffsCalculator

elf128:
ElevationDiffsCalculator was calling points.get( index ) starting from 0, while pointsCount was correctly set to
lastPointIndex - firstPointIndex + 1.
This caused the calculator to always read the front of the route, losing uphill/downhill points from the end of the route as navigation progressed.
So in case the route was a valley, downhill then uphill, reported gain fell steadily while the user was still descending toward the valley.

Capture firstPointIndex as startIndex after its Math.max adjustment and
use points.get(startIndex + index) in getPointDistance/getPointElevation.

May be related to #20624


sonora:
Great catch!! Perhaps fixes also https://github.com/osmandapp/Osmand/issues/24916.

Issue #25485, opened 2026-07-22
Android - Spatial Search & Format Results


vshcherb:
# To be designed
- [ ] New type of Cells

# Android
- [x] Setting and convert to old results and test
- [x] Integrate (include regions.ocbf) on client
- [x] Progress / cancel
- [x] Use new search by poi category (much faster for supported) categories
- [x] Memory Speed performance comparison (new / old / server)  
- [x] Use "Street, Housenumber" as name and leave city on 2nd row
- [x] Street selection should suggest Buildings
- [x] `Ocean plaza` - ATM instead of Mall
- [ ] To think display "top=true" as chips if applicable

# Bugs
- [x] Check/fix flickering "chips" during Search autocomplete
- [x] Street buildings are not displayed as in Classic search (Street -> Buildings)
- [x] Split joined Wiki+Address (Building|Street|City) as on "Саксаганського" in Kyiv to keep Address object
- [x] Fix strange unwanted "Street" displayed as **address line** for City objects (Web?)
- [x] Crash in UNKNOWN_PHRASE_MATCH_WEIGHT comparator (caused by 6ca837d27f)
- [x] Fix Search -> Address -> (City) -> (Empty nearest cities, Street list does not load)
- [x] Numeric search queries, such as "1", "11", "22", "33" (in **autocomplete mode**) lead to suggestions to download maps, which do not seem connected to the numbers (both classic and spatial search affected)
- [x] Check/fix/port the previous bug fix with OsmandRegions.getSearchIndex on iOS
- [x] Investigate and fix iOS slow getPoiAdditionalByKey() (open Kyiv leads to ~500k for-cycles) [android fix](https://github.com/osmandapp/OsmAnd/pull/25625/changes)
- [x] (_Not reproduced anymore on 2026/09/17_) Check/fix non-reset matched objects (?) `shell mcdonalds` => "Shell+Mcdonalds" and then `mcdonalds` => "Mcdonalds+Shell" (in the second query, "+Shell" should not be displayed)
- [x] [_The difference is not reproduced anymore on 2026/09/18_] Fix difference Web/Android: `24 Allée des Sorbiers Champs-sur-Marne` (in Paris)
- [ ] By default, ‘Boundary’ should be labeled as ‘Area’ (rather than ‘City’), and if a POI is associated with it, the label is taken from the POI (state, national park, province, etc.)
[...]

PR #25658, opened 2026-08-14
Do not unite main Address objects with Amenity (SpatialSearchResult)

RZR-UA:

Better solution: https://github.com/osmandapp/OsmAnd/pull/25661/changes

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResult.java
+++ b/OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResult.java
@@ -264,7 +264,7 @@ public class SpatialSearchResult implements Comparable<SpatialSearchResult> {
 		return result;
 	}
 
-	public void addExtraResult(SpatialSearchResult other, String lang) {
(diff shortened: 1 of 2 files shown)


OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResult.java
```diff
@@ -282,16 +282,18 @@ public void addExtraResult(SpatialSearchResult other, String lang) {
...
-		} 
+			return false; // nothing to merge
+		}
 		if (object instanceof Amenity a && unitedObject == null) {
 			unitedObject = new BaseDetailsObject(a, lang);
 		}
 		if (unitedObject != null) {
 			unitedObject.addObject(otherObj);
 		} else if (otherObj instanceof Amenity a) {
+			if (object instanceof City || object instanceof Street || object instanceof Building) {
```
vshcherb, line 293: if id is the same we need to merge objects, otherwise there is no sense in iddeduplication

RZR-UA:
Will be fixed better.

PR #25661, opened 2026-08-14
Fix BaseDetailsObject

vshcherb:


RZR-UA:
Will be fixed more simply.

PR #25691, opened 2026-08-19
Optimize TOP_INDEX_CATEGORY search and speed up various Strings methods

RZR-UA:

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchToken.java
+++ b/OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchToken.java
@@ -24,6 +24,7 @@ import net.osmand.data.Building;
 import net.osmand.data.LatLon;
 import net.osmand.data.MapObject;
 import net.osmand.data.Street;
+import net.osmand.osm.MapPoiTypes;
 import net.osmand.search.core.HashQuadTree;
 import net.osmand.search.core.HashSkipTileQuadTree;
 import net.osmand.search.core.spatial.SpatialSearchContext.SpatialSearchStats;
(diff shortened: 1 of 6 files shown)


OsmAnd-java/src/main/java/net/osmand/util/Algorithms.java
```diff
@@ -75,8 +75,8 @@ public static boolean containsChar(String s, char[] chars) {
 		}
 		for (int i = 0; i < s.length(); i++) {
 			char ch = s.charAt(i);
-			for (int j = 0; j < chars.length; j++) {
-				if (ch == chars[j]) {
+			for (char aChar : chars) {
```
vshcherb, line 78: is it faster?

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchToken.java
```diff
@@ -362,11 +346,17 @@ public void addPartialOtherAtom(NameIndexAtom atom, List<SpatialSearchToken> oth
...
 		name = SearchAlgorithms.alignChars(name);
+		if (wordAligned.length() >= name.length()) {
+			return null;
+		}
 		String[] res = null;
-		if (wordAligned.length() < name.length() 
-				&& collatorMain.getCollator().equals(name.substring(0, wordAligned.length()), wordAligned)) {
+		String cutName = name.substring(0, wordAligned.length());
+		boolean fastEquals = wordAligned.equals(cutName);
+		boolean isTopIndex = cutName.startsWith(TOP_INDEX_CATEGORY) || wordAligned.startsWith(TOP_INDEX_CATEGORY);
```
vshcherb, line 355: something new like isTopIndex?

OsmAnd-java/src/main/java/net/osmand/util/Algorithms.java
RZR-UA: This type of for() saves ~250 ms of ~7000 ms in total (brand Shell)

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchToken.java
RZR-UA: Ideas here: 1. Skip collator match for brands (Окко = Okko, McDonald's all ok); 2. Skip collator match if a really fast String.equals() is true.

PR #25700, opened 2026-08-20
Bbox poi spatial

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/25331
_4.2 Search in large parks, neighborhood same as in boundaries (index bbox POI), residential_


vshcherb:
https://github.com/osmandapp/OsmAnd/pull/25732

PR #25721, opened 2026-08-23
Auto-merge independent changes in Favorites groups

alex-dev-neo:
### Summary

Automatically merges a Favorites group when local and cloud changes affect different points.

Related to osmandapp/OsmAnd-iOS#5177.

### Implementation

- Stores one local last-synced GPX snapshot per Favorites group.
- Uses the snapshot as the common base for a three-way merge.
- Updates the snapshot only after a successful upload or cloud import.
- During backup preparation, creates only an in-memory merged upload payload and does not modify local Favorites.
- Applies the merged group locally only after a successful upload and only if it has not changed since preparation.
- Recognizes an otherwise unchanged point with a different name as a rename.
- Treats a cloud deletion marker as remote state, preventing a deleted group from being uploaded again as a new local group.

### Merge rules

Automatically merged:

- Independent additions.
- Edits of different points.
- Coordinate moves of different points.
- Deletions of different points.
- The same point deleted on both devices.
- Renames of different points.
- A one-sided rename when the other device did not change that point.

Kept as conflicts:

- Both devices changed the same point.
- The same point was renamed differently.
- One device renamed a point while the other deleted it.
- Same-name additions.
- Group appearance or style changes.
- A cloud-deleted group that was also changed locally.
- Personal groups.
- Groups containing linked points.
- Duplicate or empty point names.
- Missing or mismatched last-synced snapshot.

### Group deletion behavior

- If a group was deleted in the cloud and is unchanged locally, it is deleted locally.
- If the cloud-deleted group was changed locally, it remains a conflict.
- A cloud deletion marker is not interpreted as a missing remote file, so the group is not incorrectly shown as a new local upload.


OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
```diff
@@ -0,0 +1,398 @@
...
+				setShouldReplace(true);
+				processDuplicateItems();
+				apply();
+				current = app.getFavoritesHelper().getGroup(mergedGroup.getName());
+			}
+			if (sameGroup(mergedGroup, current, defaultColor)) {
+				setLocalModifiedTime(sourceModifiedTime);
+				saveSnapshot(app, mergedGroup, fileName, uploadTime);
+			} else {
+				setLocalModifiedTime(Math.max(System.currentTimeMillis(), uploadTime + 1));
```
alex-dev-neo, line 394: Post-merge review note.

When the group changes between merge preparation and upload completion, this branch skips both `apply()` and `saveSnapshot()` but still stamps the local file newer than `uploadTime` — and the merged payload is already in the cloud at that point.

Scenario: devices A and B both edited `Trip`. `prepareMergeUploads()` builds merged `M` (containing B's remote change `R`) and uploads it; `uploadTime = T`. The user edits `Trip` while the upload is in flight, so the live group is `L'`. Here `sameGroup(mergedGroup, current)` and `sameGroup(localGroup, current)` are both false, so we land in the `else`. On the next `PrepareBackup`: `localModifiedTime > T` while the remote `updatetimems == T`, so `fileChangedRemotely` is false and only `fileChangedLocally` is true — `GenerateBackupInfoTask` puts the file in `filesToUpload` and `L'` overwrites `M` in the cloud. B's change `R` is silently gone, and because no snapshot was written the merger cannot recover it on a later sync either.

Leaving the pair as a conflict here, instead of marking the local file dirty, would stop the un-merged state from winning by default.

OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
```diff
@@ -0,0 +1,398 @@
...
+		}
+
+		void finishUpload(@NonNull String fileName, long uploadTime) {
+			int defaultColor = ContextCompat.getColor(app, R.color.color_favorite);
+			FavoriteGroup current = app.getFavoritesHelper().getGroup(localGroup.getName());
+			if (!sameGroup(mergedGroup, current, defaultColor)
+					&& sameGroup(localGroup, current, defaultColor)) {
+				setShouldReplace(true);
+				processDuplicateItems();
+				apply();
```
alex-dev-neo, line 387: Post-merge review note.

`apply()` mutates `FavouritesHelper` (`deleteFavourite` / `addFavourite` / `saveCurrentPointsIntoFile` / `loadFavorites`) from the upload worker thread, and `BackupExporter.writeItems()` submits light tasks to `new ThreadPoolTaskExecutor<>(null)` (`DEFAULT_THREAD_POOL_SIZE = 4`). With two auto-merged groups in one sync, `finishUpload()` -> `apply()` therefore runs concurrently on shared, non-synchronized state.

A losing interleaving: thread B finishes `apply()` for group Y (old points deleted via `deleteFavourite(p, false)`, merged points added to the live `FavoriteGroup`) but has not yet reached `saveCurrentPointsIntoFile()`. Thread A's `apply()` for group X calls `loadFavorites()`, which replaces `flatGroups`/`favoriteGroups` wholesale from disk and — via `SaveFavoritesTask` — clears the deletions journal. B's later `saveCurrentPointsIntoFile()` then snapshots the *new* `favoriteGroups`, which no longer holds B's merged points: Y's old points are gone from disk and the merged ones were never persisted.

Serializing `finishUpload`, or collecting merged items during upload and applying them once after `writeItems()` returns, would avoid this.

OsmAnd/src/net/osmand/plus/backup/GenerateBackupInfoTask.java
```diff
@@ -128,6 +129,7 @@ protected BackupInfo doInBackground(Void... voids) {
 				}
 			}
 		}
+		FavoritesBackupMerger.prepareMergeUploads(app, backupHelper, info);
```
alex-dev-neo, line 132: Post-merge review note. Two separable concerns with running merge preparation here:

1. `prepareMergeUploads()` performs a synchronous `downloadFile()` per conflicting group inside `doInBackground()`. This task runs on every `PrepareBackupTask` (`BackupHelper.generateBackupInfo`), including auto-sync preparation and simply opening the Cloud screen — so N conflicting groups means N full GPX downloads before backup info is available.

2. It never consults `backupHelper.getBackupTypePref(ExportType.FAVORITES, autoSync)`. During auto-sync that resolves to `getAutoBackupTypePref()`, which defaults to `false`, so `BackupInfo.createFilteredFilesToUpload()` / `createFilteredFilesToMerge()` discard the result — the bytes are downloaded on the user's connection and then thrown away on every auto-sync prepare.

(2) looks like a straightforward fix. (1) is more a question of whether network I/O belongs in `PrepareBackupTask`'s contract.

OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
```diff
@@ -0,0 +1,398 @@
...
+				LOG.warn("Failed to delete temporary Favorites file " + tempFile);
+			}
+		}
+	}
+
+	@Nullable
+	private static FavoriteGroup loadSnapshot(@NonNull OsmandApplication app,
+	                                          @NonNull String fileName, long syncTime) {
+		File file = snapshotFile(app, fileName);
+		return file != null && file.exists() && file.lastModified() == syncTime
```
alex-dev-neo, line 138: Post-merge review note — non-blocking, but worth recording.

This requires `file.lastModified() == syncTime` exactly, while `saveSnapshot()` only checks that `setLastModified()` *returned* true; it never reads the value back.

If the underlying filesystem truncates mtime, `setLastModified(1787483581234)` returns true, the file is kept with no warning, and the next read yields `1787483581000`. `loadSnapshot()` then rejects it, `prepareMergeUploads()` bails at the `base == null` guard for every group on every sync, and the snapshot is rewritten and re-rejected indefinitely — the whole feature degrades to "always a conflict" with no diagnostic.

`getNoBackupFilesDir()` is internal storage (ns-precision ext4/f2fs on any current device), so this most likely never fires in practice. Storing the sync time inside the snapshot, or in the backup DB, would remove the dependency.

alex-dev-neo (commented):
Post-merge review of the Favorites auto-merge feature. The merge algorithm itself holds up — rename/deletion/edit combinations in `mergeGroups` resolve as described, and `copyPoint` / `FavouritePoint(FavouritePoint)` is field-complete. The four notes below are in the surrounding orchestration; the first two are data-loss paths now in `master`.

Also checked and found fine: the `hasRemoteFile` change in `GenerateBackupInfoTask` (loop 1 already covers every deletion-marker case; the change only removes the spurious duplicate `filesToUpload` entry), and the purchase filter in `createFilteredFilesToUpload` (`FavoritesExportType.isAvailableInFreeVersion()` is `true`, so converting a conflict into an upload cannot drop it for free users).

PR #25739, opened 2026-08-25
Implemented AA map day-night setting

### LINKED ISSUE #21448: Android Auto (Carpuride) OsmAnd in Night mode (controlled by car)
### Description

When I turn on the Carpuride (carplay) but I don't open the Osmand app
first, the map I see on the carpuride is only in night mode. If I try to
change to day mode from the phone app, nothing happens, it always stays in
night mode.
Instead, when I open the osmand app first and the map mode is in day mode,
then I turn on the Carplay I see the map in day mode.
However, even in this last case, if I want to put the map in day mode, or I
want to activate the mode change based on the light that the sensor
receives, nothing changes in this case it always stays in day mode.

### Steps to reproduce

Open the Carpuride (carplay) wait for start. Chose Osman app. 

### Actual result

The app is open in night mode also is I blocked in light view. 

### Expected result

The may be open in the same way that it is selected on the app. If I try to change in light modo nothing do. 
But If I start first the Osman app on the phone and choose light color of the map  and after I turn on my carplay the map is in light mode

<table style="width:100%">
  <tr>
<th> </th>
  </tr>
</table>

```
OsmAnd Version: 4.9.7
Android/iOS version: Android 15
Device model: Google Pixel 8 pro
Crash-Logs: no crash
```

Corwin-Kh:


OsmAnd/src/net/osmand/plus/auto/screens/LandingScreen.kt
```diff
@@ -128,7 +128,7 @@ class LandingScreen(
 
     private fun updateCompass() {
         val settings = app.settings
-        val nightMode = carContext.isDarkMode
+        val nightMode = app.settings.AA_MAP_NIGHT_MODE.get().isNightMode(carContext.isDarkMode)
```
Chumva, line 131: Should we move this preference check to DayNightHelper? something like app.getDayNightHelper().isNightMode(CarContext)?

PR #25746, opened 2026-08-26
[r5.4] Cherry-pick Spatial fixes from master (UI, etc)

RZR-UA:


RZR-UA:
unlikely useful

PR #25760, opened 2026-08-27
#25300: Unify map status color coding across Map View and Maps & Resources menu

### LINKED ISSUE #25300: Unify map status color coding across Map View and Maps & Resources menu
### Describe the idea

It would be great, if osmand could use the same color for the same map status everywhere in the app. 

Now it is in the map view:
up-to-date "green"
out-of-date "green"
deactivated "orange"

And in maps & resources / local:
up-to-date "green"
out-of-date "orange" 
deactivated "grey"

Make it the same, please. 

### Expected behaviour

One color tells one status.

No alternatives. 

### Context

Thanks.

kraativ:


OsmAnd/src/net/osmand/plus/download/OutdatedIndexesCollection.java
```diff
@@ -4,15 +4,19 @@
...
 import java.util.List;
+import java.util.Set;
 
 public record OutdatedIndexesCollection(@NonNull List<IndexItem> all,
                                         @NonNull List<IndexItem> activated,
                                         @NonNull List<DownloadItem> groupedAll,
                                         @NonNull List<DownloadItem> groupedActivated,
-                                        @NonNull List<IndexItem> deprecated) {
+                                        @NonNull List<IndexItem> deprecated,
+                                        @NonNull Set<String> allFileNames
```
Corwin-Kh, line 14: can't we reuse "all" param?

kraativ: We could, but `all` is a `List<IndexItem>`, so we'd have to loop through the whole list on every check. Since `isItemOutdated()` is called repeatedly for many regions while rendering the map layer, pre-populating a `Set<String>` makes these lookups instant.

sonora:
Can we perhaps factor in https://github.com/osmandapp/OsmAnd/issues/25300#issuecomment-5449338101 to use the occasion to  deduplicate "grey" across various screen?

Issue #25390, opened 2026-07-06
Overestimated ETA for a short route in Berlin (1 hour vs ~27 mins expected)


desputin:
### Description

Hello,

the routing time in OSMAND calculation is wrong for this:
https://osmand.net/map/navigate/?start=52.423260,13.374245&end=52.533142,13.384974&profile=car#12/52.443842/13.353544

-By Car
-On a Sunday

-Time-Prognosis by Osmand+: 1 Hour and 4 Minuts for this exact way

In comparision Graphhopper says 27 Minuts, which is the correct estimate:
https://graphhopper.com/maps/?point=52.42326%2C13.37425_52.42326%2C+13.37425&point=52.53314%2C13.38497_52.53314%2C+13.38497&profile=car&layer=OpenStreetMap

I tested this with my car. On a day with bad traffic, it might be more like 40 Miuts. But 1:04, that is definitevely wrong.

Best regards
Desputin

### Steps to reproduce

Routing by car from: 52.42326, 13.37425
To: 52.53314, 13.38497
14,6 km • 01:04

### Actual result

Is way too much time calculated by osmand.

### Expected result

Should be much less estimated time.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 5.3.10
Android version: 11
Device model: Samsung Galaxy S
Crash-Logs: ?
```

yuriiurshuliak:
> OsmAnd~ 5.4.0#8217m, released: 2026-07-10

**OsmAnd significantly overestimates the ETA for this 14.7 km car route in Berlin**, showing **1 h 4 min** in both the Android app and OsmAnd Web, while other routing engines estimate approximately **25–34 minutes**.

**Reproduced**

[OsmAnd Web link ](https://osmand.net/map/navigate/?start=52.423260,13.374245&end=52.533142,13.384974&profile=car#12/52.4865/13.3652)

[OSM route link ](https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=52.42326%2C13.374245%3B52.533142%2C13.384974#map=12/52.4781/13.3810&layers=P)

<details>
<summary><strong>Details</strong></summary>

### Steps to reproduce

1. Open OsmAnd.
2. Select the **Car** profile.
3. Set the starting point to:
   `52.423260, 13.374245`
4. Set the destination to:
   `52.533142, 13.384974`
[...]

PR #25764, opened 2026-08-28
Overestimated ETA for bunch of traffic light fix

nnngrach:
[issue](https://github.com/osmandapp/OsmAnd/issues/25390)
[related resources pr](https://github.com/osmandapp/OsmAnd-resources/pull/1485)

# Explanation of this bug

We have a route with a lot of traffic lights in one place:

Screenshot 2026-08-25 at 04 47 53

It looks other routers do not take a lot penalty for them. Speed diagram more correlating with route max speed limits.

Screenshot 2026-08-25 at 04 13 43 copy

Our algorithm looks like way more detailed.

Screenshot 2026-08-28 at 11 51 03

Also if we turn off penalty for traffic light, then our diagram will be more like another routers.

Screenshot 2026-08-28 at 12 13 17

But i don't thing this is a proper way to fix.

[here is pr only with only routing. xml fix but I thin this is also not proper way to fix](url)

---

# Idea of this my fix

In logs i see the most slowing tags for this road. I will focus only on "traffic_signals":

```
20 sec   highway=traffic_signals
15 sec   crossing=traffic_signals
01 sec   highway=crossing
```

My guess, now our route gets penalty, like car stops on every traffic signal. Even if there are 5 signals on 50 meters. But I think, usually if we stop on one red signal, then several next signals are usually green. So in this fix I tried to simulate this. 

```
🔴🔴🔴🔴              🔴🔴🔴🔴             🔴🔴🔴🔴             🔴🔴🔴🔴
🔴🟢🟢🟢              🔴🟢🟢🟢             🔴🟢🟢🟢             🔴🟢🟢🟢
```

This is how it looks like in logs:
[skipped_traffic_signals_log.txt](https://github.com/user-attachments/files/31545429/skipped_traffic_signals_log.txt)

<details><summary>Spoiler</summary>
    Screenshot 2026-08-28 at 12 28 30
</details>

---

# Fix and tests

Full route
https://osmand.net/map/navigate/?start=52.423260,13.374245&end=52.533142,13.384974&profile=car#12/52.443842/13.353544

before:
Screenshot 2026-08-28 at 11 55 13    Screenshot 2026-08-28 at 11 55 36

after: 
Screenshot 2026-08-28 at 11 49 41 Screenshot 2026-08-28 at 11 50 06

---

Short route
https://osmand.net/map/navigate/?start=52.438720,13.386304&end=52.483090,13.386500&profile=car#14/52.4628/13.3666

before:
Screenshot 2026-08-28 at 11 54 26      Screenshot 2026-08-28 at 11 54 42

after:
Screenshot 2026-08-28 at 11 50 42    Screenshot 2026-08-28 at 11 51 03

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
@@ -320,14 +320,32 @@ public class RouteResultPreparation {
 	private static final double SLOW_DOWN_SPEED_THRESHOLD = 15;
 	// reference speed 30ms (108kmh) - 2ms (7kmh)
 	private static final double SLOW_DOWN_SPEED = 2;
-	
+
+	private static boolean segmentHasTag(RouteDataObject road, int segmentIndex, String tag) {
+		int[] segmentPointTypes = road.getPointTypes(segmentIndex);
+		if (segmentPointTypes != null) {
+			for (int type : segmentPointTypes) {
+				RouteTypeRule rule = road.region.quickGetEncodingRule(type);
+				if (rule != null && rule.getValue().equals(tag)) {
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+
 	public static void calculateTimeSpeed(RoutingContext ctx, List<RouteSegmentResult> result) {
+		ctx.currentCalculatedDistance = 0.0;
+		ctx.lastTrafficSignalDistance = -1;
+
 		for (int i = 0; i < result.size(); i++) {
 			RouteSegmentResult rr = result.get(i);
+			if (i > 0) {
+				ctx.currentCalculatedDistance += result.get(i-1).getDistance();
+			}
 			calculateTimeSpeed(ctx, rr);
 		}
 	}
-
 	public static void calculateTimeSpeed(RoutingContext ctx, RouteSegmentResult rr) {
 		// Naismith's/Scarf rules add additional travel time when moving uphill
 		boolean useNaismithRule = false;
@@ -372,6 +390,20 @@ public class RouteResultPreparation {
 			double obstacle = ctx.getRouter().defineObstacle(road, j, !plus);
 			if (obstacle < 0) {
 				obstacle = 0;
+			} else if (obstacle > 0) {
+				if (segmentHasTag(road, j,"traffic_signals")) {
+					// For groups with many traffic signals nearby take penalty only for first one. (After red signal next usually are green)
+					// XXXXX XXXXX   ->   Xxxxx Xxxxx
+					boolean isFirstStop = ctx.lastTrafficSignalDistance == -1;
+					double currentDistance = ctx.currentCalculatedDistance + distance;
+					double distanceFromPreviousStop = currentDistance - ctx.lastTrafficSignalDistance;
+					ctx.lastTrafficSignalDistance = currentDistance;
+					if (isFirstStop || distanceFromPreviousStop >= 100) {
+						obstacle = obstacle * 1.0;
+					} else {
+						obstacle = obstacle * 0.25;
+					}
+				}
 			}
 			distOnRoadToPass += d / speed + obstacle;  //this is time in seconds
 
--- a/OsmAnd-java/src/main/java/net/osmand/router/RoutingContext.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/RoutingContext.java
@@ -85,6 +85,9 @@ public class RoutingContext {
 	public boolean leftSideNavigation;
 	public List<RouteSegmentResult> previouslyCalculatedRoute;
 	public PrecalculatedRouteDirection precalculatedRouteDirection;
+
+	public double currentCalculatedDistance;
+	public double lastTrafficSignalDistance;
 	
 	
 	// 2. Routing memory cache (big objects)


OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -372,6 +390,20 @@ public static void calculateTimeSpeed(RoutingContext ctx, RouteSegmentResult rr)
 			double obstacle = ctx.getRouter().defineObstacle(road, j, !plus);
 			if (obstacle < 0) {
 				obstacle = 0;
+			} else if (obstacle > 0) {
+				if (segmentHasTag(road, j,"traffic_signals")) {
+					// For groups with many traffic signals nearby take penalty only for first one. (After red signal next usually are green)
+					// XXXXX XXXXX   ->   Xxxxx Xxxxx
+					boolean isFirstStop = ctx.lastTrafficSignalDistance == -1;
```
vshcherb, line 397: no  configuration  in routing.xml should be generic I think

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -372,6 +390,20 @@ public static void calculateTimeSpeed(RoutingContext ctx, RouteSegmentResult rr)
 			double obstacle = ctx.getRouter().defineObstacle(road, j, !plus);
 			if (obstacle < 0) {
 				obstacle = 0;
+			} else if (obstacle > 0) {
+				if (segmentHasTag(road, j,"traffic_signals")) {
```
vshcherb, line 394: formatting

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -320,14 +320,32 @@ private void justifyUTurns(boolean leftSide, List<RouteSegmentResult> result) {
...
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+
 	public static void calculateTimeSpeed(RoutingContext ctx, List<RouteSegmentResult> result) {
+		ctx.currentCalculatedDistance = 0.0;
+		ctx.lastTrafficSignalDistance = -1;
```
vshcherb, line 339: These two new fields (currentCalculatedDistance / lastTrafficSignalDistance) are only reset here, in the List-overload. But RouteResultPreparation.calculateTimeSpeed(ctx, RouteSegmentResult) is also called directly (bypassing this reset) from RoutePlannerFrontEnd.makeSegmentPointPrecise() — and that call happens *after* prepareResult() already ran the full list through this reset once. So by the time makeSegmentPointPrecise() recomputes the first/last segment, ctx.lastTrafficSignalDistance/currentCalculatedDistance hold stale leftover values from the end of the previous full-route pass, not state relevant to the segment being recomputed. isFirstStop will basically never be true there even when it should be, and distanceFromPreviousStop is measured against an unrelated traffic signal from a different part of the route.

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -320,14 +320,32 @@ private void justifyUTurns(boolean leftSide, List<RouteSegmentResult> result) {
...
 		for (int i = 0; i < result.size(); i++) {
 			RouteSegmentResult rr = result.get(i);
+			if (i > 0) {
+				ctx.currentCalculatedDistance += result.get(i-1).getDistance();
+			}
 			calculateTimeSpeed(ctx, rr);
 		}
 	}
-
 	public static void calculateTimeSpeed(RoutingContext ctx, RouteSegmentResult rr) {
```
vshcherb, line 349: Same issue from the other end: this single-segment overload reads/mutates ctx.lastTrafficSignalDistance and ctx.currentCalculatedDistance as if they were always properly scoped to 'distance along the route up to this segment', but that invariant only holds when called from the List-overload above. RoutePlannerFrontEnd.makeSegmentPointPrecise() calls this overload directly on an already-prepared result (after prepareResult() finished), so it runs with whatever state was left over from the previous full pass — not -1 / 0.0. Worth either resetting explicitly at that call site, or not exposing this overload for standalone/partial recomputation.

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -320,14 +320,32 @@ private void justifyUTurns(boolean leftSide, List<RouteSegmentResult> result) {
...
+				if (rule != null && rule.getValue().equals(tag)) {
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+
 	public static void calculateTimeSpeed(RoutingContext ctx, List<RouteSegmentResult> result) {
+		ctx.currentCalculatedDistance = 0.0;
```
vshcherb, line 338: Why do you need to this to put in context at all? it looks quite bad

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -377,6 +396,18 @@ public static void calculateTimeSpeed(RoutingContext ctx, RouteSegmentResult rr)
 			double obstacle = ctx.getRouter().defineObstacle(road, j, !plus);
 			if (obstacle < 0) {
 				obstacle = 0;
+			} else if (obstacle > 0 && road.hasTrafficLightAt(j)) {
+				// A driver stops once per intersection, not once per signalled node of it, so only the
```
vshcherb: shorter comments - 1 line only if needed and not readable from code

OsmAnd-java/src/main/java/net/osmand/router/RouteResultPreparation.java
```diff
@@ -325,15 +325,34 @@ private void justifyUTurns(boolean leftSide, List<RouteSegmentResult> result) {
 	private static final double SLOW_DOWN_SPEED_THRESHOLD = 15;
 	// reference speed 30ms (108kmh) - 2ms (7kmh)
 	private static final double SLOW_DOWN_SPEED = 2;
-	
+	// A signalized intersection is described by several nodes within a few tens of meters of each other:
```
vshcherb: shorter comments - 1 line only if needed and not readable from code

PR #25535, opened 2026-07-29
Spatial pipeline: fix token mask algebra (PR #25534)

dmvkosm:
## Summary

Fixes and completes the **token-mask layer** for the spatial pipeline ([PR #25534](https://github.com/osmandapp/OsmAnd/pull/25534)).

- **`SpatialTokenMask`** — pure 64-bit mask algebra: `allowed()`, loop-free `combine()`, duplicate-word variants, per-token contested ambiguous fork
- **`SpatialStagePipeline`** — wired into prepare/join; self-join identity skip; semantic filters restored; stats/debug fixes
- **Tests** — `SpatialTokenMaskTest` (JUnit) + `mask_verify.py` (standalone, no Java required)
- **Docs** — `SPATIAL_PIPELINE_MASK.md`

**Performance work** (incremental join, deferred common-word read) is in a **separate stacked PR**: #25536 — merge this PR first.

## Test plan (no maps required)

```bash
# Java (from repo root)
./gradlew :OsmAnd-java:test --tests net.osmand.search.core.spatial.SpatialTokenMaskTest

# Python cross-check (no JDK)
python OsmAnd-java/src/test/resources/spatial/mask_verify.py
```

Windows one-liner: `.\run-spatial-mask-tests.ps1` (JDK 17 + UTF-8).

Integration queries (`SpatialSearchTestAndDocs`, needs `.obf` maps): see `SPATIAL_PIPELINE_MASK.md` and **#25536** for Portugal benchmark.

## Target

Merge into **`spatialPipeline`**, not `master`.

Author: dmvkmusic@osmand.net

PR #25508, opened 2026-07-27
Improve spatial search android

Corwin-Kh:

PR #25773, opened 2026-08-28
Migrate route details backend to OsmAnd-shared

alex-dev-neo:
Related to https://github.com/osmandapp/OsmAnd-Issues/issues/1794

## Summary

Move platform-independent route-details models and calculations from Android into `OsmAnd-shared`, while preserving the existing Android routing engine, native integration, models, and production UI.

## Changes

- Add immutable, serializable shared contracts for:
  - route points and segments
  - maneuvers and summaries
  - alerts/warnings
  - route statistics
- Port Android-compatible geometry, maneuver, event, slope, and statistics calculations to shared Kotlin.
- Add eager Android adapters that copy Java/native routing results into a platform-neutral `RouteDetailsSnapshot`.
- Route Android distance calculation, maneuver timing, cumulative route information, location lookup, and statistics through the shared backend.
- Preserve Android compatibility for renderer fallback, statistics grouping/order, undefined values, NaN maneuver speeds, and slope calculations.
- Add common and Android compatibility coverage using Android-derived fixtures.
- Add a Developer-plugin-only diagnostic dialog showing all route alerts, their location, source rule, affected road segment, and localized hazard meaning.

## Scope

This does not port the route-search engine or introduce a new production UI. Android continues to own native routing integration and rendering-specific classification; only platform-independent route-details data and calculations are moved to `OsmAnd-shared`.

The alert diagnostics are displayed only when the Developer plugin is enabled.


vshcherb (commented):
Reviewed with the stated goal in mind — `OsmAnd-java` moving to Kotlin and then to iOS. The direction is right, so "this shared API has no production consumer yet" is not a complaint from me: that is the future iOS surface. What follows is what would block that move or would have to be redone later.

**1. The debug dialog does not belong in this PR.** ~250 lines of UI formatting (`AlertDialog`, `MapActivity`) landed in `RoutingHelper` — a class that is itself a candidate for the move to shared. It also builds the snapshot and the whole string even when the switch is off, as long as the development plugin is enabled. And it added two fields to the production `AlarmInfo` model. This should be a separate PR, living in `OsmandDevelopmentPlugin`. The same block still contains commented-out code and empty `if` bodies.

**2. `require` in shared constructors means crashes.** `RouteSegment`, `RouteDetailsSnapshot`, `RouteManeuver` and `RouteGeometryCalculation` validate data coming from OBF/native through `require`, i.e. an `IllegalArgumentException` on the route calculation path. On top of that it is several O(N) passes over the whole route per construction. A routing backend cannot do that — sanitize at the adapter boundary and keep the invariants in tests.

**3. Pick one strategy: accessors or snapshot.** Both are in this PR. `ILocationAccessor` and friends exist explicitly to avoid allocating one `KLatLon` per point — and `RouteCalculationResultSnapshotAdapter` right next to it allocates exactly that for every route point. iOS would have to carry both. Please pick one.

**4. The public `OsmAnd-java` API is broken.** `RouteStatistics`, `RouteSegmentAttribute` and `ExitInfo` are gone, and shared types now appear in public signatures while `OsmAnd-shared` is wired as `implementation` (`OsmAnd-java/build.gradle:131`). It needs `api`, plus a pass over the external consumers (MapCreator, server).

**5. The tests do not prove compatibility.** The expected values were recorded from the new code, and the old implementation is deleted in the same PR, so there is nothing to compare against. What is needed is running the old and the new implementation over a real route and comparing element by element — `OsmAnd-java` already has routing tests with resources for that. As it stands, the one real behaviour difference I found by reading is caught by no test: `copyRouteTypes()` drops encoding rules with `value == null`, while the old code put them into `additional` as `tag=null;`.

**6. `RouteManeuverType` duplicates `TurnType`.** Since we are migrating anyway, port `TurnType` itself instead of keeping a reduced copy of 14 of its values next to it.

On the plus side: `getRouteDirectionCumulativeInfoByPosition` is an honest O(n^2) -> O(n) fix in Route Details and in the HTML print. That part can be split out and merged right away.

_Review prepared with Claude Code (Opus 5). Prompt: "analyse this the way vshcherb usually analyses a backend PR", followed by a clarification that the goal is the java -> kotlin -> iOS migration; the review style was derived from @vshcherb's comments in #25781, #25700, #25691, #25658, #25535, #25508. Every finding was verified against the diff and against master by hand._

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
```diff
@@ -313,9 +326,242 @@ void newRouteCalculated(boolean newRoute, RouteCalculationResult res) {
...
 				app.showToastMessage(msg);
 			}
+			OsmandDevelopmentPlugin developmentPlugin = PluginsHelper.getEnabledPlugin(OsmandDevelopmentPlugin.class);
+			if (developmentPlugin != null) {
+				reportRouteAlertsDebug(res, developmentPlugin.SHOW_ROUTE_ALERTS_DEBUG_DIALOG.get());
+			}
 		});
 	}
 
+	private void reportRouteAlertsDebug(@NonNull RouteCalculationResult route, boolean showDialog) {
```
vshcherb: `AlertDialog` / `TextView` / `MapActivity` в `RoutingHelper` — 246 строк UI-форматирования в классе навигации.

Это мешает ровно той цели, ради которой делается PR: `RoutingHelper` — кандидат на переезд в Kotlin/shared, а сюда заезжает Android UI. Всю диагностику надо вынести в `OsmandDevelopmentPlugin` или в отдельный фрагмент, и лучше отдельным PR — к миграции backend она отношения не имеет.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
```diff
@@ -313,9 +326,242 @@ void newRouteCalculated(boolean newRoute, RouteCalculationResult res) {
...
+	                                      @NonNull RouteDetailsSnapshot snapshot) {
+		List<RouteEvent> events = snapshot.getEvents();
+		List<AlarmInfo> sourceAlarms = route.getAlarmInfo();
+		if (events.isEmpty()) {
+			return "No alerts/warnings found along the route.";
+		}
+		StringBuilder message = new StringBuilder();
+		for (int index = 0; index < events.size(); index++) {
+			RouteEvent event = events.get(index);
+			AlarmInfo sourceAlarm = index < sourceAlarms.size() ? sourceAlarms.get(index) : null;
```
vshcherb: `snapshot.getEvents().get(index)` сопоставляется с `route.getAlarmInfo().get(index)` по индексу — молчаливое допущение, что списки параллельны.

Сейчас это правда (`copyEvent` мапит 1:1), но нигде не зафиксировано. Как только в snapshot появится фильтрация/сортировка событий — диагностика начнёт показывать чужой `sourceTag` и никто этого не заметит. Либо тащить tag/value в `RouteEvent`, либо не смешивать два списка.

OsmAnd/src/net/osmand/plus/routing/AlarmInfo.java
```diff
@@ -22,24 +11,29 @@
...
 
-	private final AlarmInfoType type;
+	private final RouteEventType type;
 	protected final int locationIndex;
 	private int lastLocationIndex = -1;
 	private int intValue;
 	private float floatValue;
 	private double latitude;
 	private double longitude;
+	@Nullable private String sourceTag;
```
vshcherb: `sourceTag` / `sourceValue` — две строки на каждый алерт каждого маршрута в прод-модели, нужны только дев-диалогу.

Если диагностика уезжает в плагин (см. коммент в `RoutingHelper`), то и эти поля туда же — либо доставать tag/value по `RouteTypeRule` в момент показа диалога, а не хранить на всём времени жизни маршрута.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteSegment.kt
```diff
@@ -0,0 +1,61 @@
...
+		require(distanceMeters.isFinite() && distanceMeters >= 0f) {
+			"Route segment distance must be finite and non-negative"
+		}
+		require(segmentTimeSeconds.isFinite() && segmentTimeSeconds >= 0f) {
+			"Route segment time must be finite and non-negative"
+		}
+		require(segmentSpeedMetersPerSecond.isFinite()) { "Route segment speed must be finite" }
+		require(maximumSpeedMetersPerSecond.isFinite()) { "Route segment maximum speed must be finite" }
+		require(heightValues.size % 2 == 0) { "Route segment height values must contain distance/height pairs" }
+		require(heightValues.all { it.isFinite() }) { "Route segment height values must be finite" }
```
vshcherb: `require(...)` на данных, которые приходят из OBF/натива, — это краш приложения, а не валидация.

`heightValues` собираются из `RouteSegmentResult.getHeightValues()`, `distanceMeters` — из натива. Одна кривая высота в чужой карте, и вместо деградации статистики пользователь получает `IllegalArgumentException` на пути расчёта маршрута.

Для роутингового бэкенда так нельзя: либо чистить/клампить значения на границе адаптера, либо логировать и пропускать сегмент.

OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
```diff
@@ -711,38 +712,41 @@ protected void calculatePoi(RouteCalculationResult route, List<LocationPointWrap
...
-					prevRailway = alarmInfo;
-				}
-			} else if (settings.SHOW_TRAFFIC_WARNINGS.getModeValue(mode) || settings.SPEAK_TRAFFIC_WARNINGS.getModeValue(mode)) {
-				addPointWrapper(alarmInfo, array, settings.SPEAK_TRAFFIC_WARNINGS.getModeValue(mode));
+		List<AlarmInfo> alarms = route.getAlarmInfo();
+		if (alarms.isEmpty()) {
+			return;
+		}
+		List<RouteEvent> events = new ArrayList<>(alarms.size());
+		Map<RouteEvent, AlarmInfo> alarmsByEvent = new IdentityHashMap<>(alarms.size());
```
vshcherb: Обратный маппинг `RouteEvent -> AlarmInfo` через `IdentityHashMap` держится ровно на том, что `select()` возвращает те же экземпляры, что получил. Появится внутри `.copy()` или пересоздание — `alarmsByEvent.get()` вернёт null, и `if (alarmInfo != null)` молча проглотит алармы: пользователь просто перестанет получать предупреждения.

И по существу: было 43 строки одного читаемого цикла, стало 47 строк с параллельным списком, `IdentityHashMap` и объектом на 9 булевых полей — при том же поведении. Если `select()` нужен ради iOS, то пусть он возвращает индексы исходного списка, а не объекты, — тогда маппинг обратно тривиален и не хрупок.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
```diff
@@ -1179,6 +1119,18 @@ public int getNextIntermediate() {
...
 
+	int getCurrentDirectionInfoForSnapshot() {
+		return currentDirectionInfo;
+	}
+
+	int[] getIntermediateDirectionIndexesForSnapshot() {
+		return intermediatePoints.clone();
+	}
+
+	int getListDistanceForSnapshot(int index) {
```
vshcherb: Обёртка над `getListDistance(index)` без единого отличия. Суффикс `ForSnapshot` в трёх методах подряд — тоже лишний: это обычные package-private геттеры.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResultSnapshotAdapter.kt
```diff
@@ -0,0 +1,179 @@
...
+	)
+
+	private fun copySummary(
+		source: RouteCalculationResult,
+		directions: List<RouteDirectionInfo>,
+	): RouteSummary = RouteSummary(
+		totalDistanceMeters = source.wholeDistance,
+		totalTimeSeconds = directions.firstOrNull()?.afterLeftTime ?: 0,
+		profileId = source.appMode?.stringKey,
+		routeService = source.routeService?.let { RouteServiceType.valueOf(it.name) },
```
vshcherb: `RouteServiceType.valueOf(it.name)` — связь по имени через рефлексию enum-а. Переименуют константу в `RouteService` — упадёт в рантайме, компилятор не поможет.

Лучше явный `when`, тогда новый вариант в `RouteService` не соберётся молча.

OsmAnd/res/values/strings.xml
```diff
@@ -4604,6 +4604,8 @@ Download tile maps directly, or copy them as SQLite database files to OsmAnd\'s
     <string name="osm_editing_plugin_name">OpenStreetMap editing</string>
     <string name="osm_editing_plugin_description">Make contributions to OpenStreetMap, a global community maintaining a comprehensive map of the world and providing up-to date open-source data to every user. \n\nYou can find instructions on how to edit the map on our website %1$s.</string>
     <string name="osmand_development_plugin_description">Provides functionality to test and debug OsmAnd: simulate navigation routes, monitor the screen rendering, check performance, etc. This plugin is not required for normal app use.</string>
+    <string name="show_route_alerts_debug_dialog">Show route alerts and warnings dialog</string>
```
vshcherb: Новые строки — в начало `strings.xml` (AGENTS.md, раздел Coding Standards), иначе ломается процесс перевода.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteStatisticsCalculator.kt
```diff
@@ -0,0 +1,359 @@
...
+
+	private fun makePartition(attributes: List<MutableRouteAttribute>): List<MutableRouteAttribute> {
+		val partitionByUserName = mutableMapOf<String, MutableRouteAttribute>()
+		for (attribute in attributes) {
+			val partition = partitionByUserName.getOrPut(attribute.userPropertyName) {
+				attribute.copy(distanceMeters = 0f)
+			}
+			partition.distanceMeters += attribute.distanceMeters
+		}
+		val naturallyOrderedKeys = partitionByUserName.keys.sorted()
```
vshcherb: Двойная сортировка: сначала `keys.sorted()`, потом `sortedWith(comparator)`.

Работает (обе стабильны, воспроизводит `TreeMap` + `Collections.sort`), но так это не читается — надо комментарием сказать, что первая сортировка задаёт порядок для равных элементов и повторяет `TreeMap` из Java-версии. Иначе первый же, кто будет это оптимизировать, её выкинет и незаметно поменяет порядок легенды на графике.

vshcherb (commented):
Inline comments on the code, grouped by the summary above.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
```diff
@@ -313,9 +326,242 @@ void newRouteCalculated(boolean newRoute, RouteCalculationResult res) {
 						res.getCalculateTime(), res.getVisitedSegments(), res.getLoadedTiles());
 				app.showToastMessage(msg);
 			}
+			OsmandDevelopmentPlugin developmentPlugin = PluginsHelper.getEnabledPlugin(OsmandDevelopmentPlugin.class);
```
vshcherb: The dialog is a dev feature, but the cost is paid whenever the development plugin is enabled, regardless of the switch.

`reportRouteAlertsDebug()` first calls `route.getRouteDetailsSnapshot()` (a full copy of the route) and builds the whole string; `showDialog` is only checked before showing. Testers keep the plugin on permanently, so this is a full route copy plus a large log string on every recalculation.

The `SHOW_ROUTE_ALERTS_DEBUG_DIALOG.get()` check belongs here, before the call.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
vshcherb: `AlertDialog` / `TextView` / `MapActivity` inside `RoutingHelper` — 246 lines of UI formatting in a navigation class.

This works against the very goal of the PR: `RoutingHelper` is a candidate for the move to Kotlin/shared, and Android UI is moving into it. All of the diagnostics should live in `OsmandDevelopmentPlugin` or a separate fragment, and preferably in a separate PR — it has nothing to do with the backend migration.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
vshcherb: `snapshot.getEvents().get(index)` is paired with `route.getAlarmInfo().get(index)` by index — a silent assumption that the two lists are parallel.

It holds today (`copyEvent` maps 1:1) but nothing records it. As soon as the snapshot starts filtering or sorting events, the diagnostics will show someone else's `sourceTag` and nobody will notice. Either carry tag/value inside `RouteEvent`, or do not mix the two lists.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
```diff
@@ -313,9 +326,242 @@ void newRouteCalculated(boolean newRoute, RouteCalculationResult res) {
...
+					//.append(event.getLastLocationIndex() >= 0 ? "Geometry points: " : "Geometry point: ");
+//			if (geometryPointIndex >= 0) {
+//				message.append(geometryPointIndex);
+//			} else {
+//				message.append("unknown");
+//			}
+			if (event.getLastLocationIndex() >= 0) {
+				message.append("..").append(event.getLastLocationIndex());
+			}
+			if (geometryPointIndex != event.getLocationIndex()) {
```
vshcherb: Empty `if` whose only body is commented out. Same again at line 546, plus a commented `//.append('\n')` in the chain at 391.

`resolvedFromSegment` is computed and both of its `if` blocks are empty — the variable has no effect at all.

Dead code in a new file; please clean it up before merging.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
```diff
@@ -313,9 +326,242 @@ void newRouteCalculated(boolean newRoute, RouteCalculationResult res) {
...
+
+	@NonNull
+	private static String formatRouteSegment(@Nullable RouteSegment segment) {
+		if (segment == null) {
+			return "";
+		}
+		StringBuilder road = new StringBuilder();
+		appendDebugValue(road, segment.getRoadName());
+		appendDebugValue(road, segment.getRef());
+		if (!Algorithms.isEmpty(segment.getHighway())) {
```
vshcherb: Another empty `if` — the condition is evaluated, the body is commented out.

OsmAnd/src/net/osmand/plus/routing/AlarmInfo.java
vshcherb: `sourceTag` / `sourceValue` are two strings per alert per route held in a production model, needed only by the debug dialog.

If the diagnostics move into the plugin (see the comment in `RoutingHelper`), these fields should move with it — or resolve tag/value from `RouteTypeRule` at the moment the dialog is shown, instead of keeping them alive for the whole lifetime of the route.

OsmAnd/src/net/osmand/plus/routing/AlarmInfo.java
```diff
@@ -97,68 +101,57 @@ public void setLatLon(double latitude, double longitude) {
...
-			return type.getPriority();
-		}
-		// 2nd level
-		if (time < 7 || distance < 100) {
-			return type.getPriority() + MAXIMUM.getPriority();
+		return RouteEventHelper.INSTANCE.updateDistanceAndGetPriority(type, time, distance);
+	}
+
+	@NonNull
+	public static String getVisualName(@NonNull Context ctx, @NonNull RouteEventType type) {
```
vshcherb, line 106: Replacing the `AlarmInfoType` enum (priority + titleId in one place) with a static `switch` plus `default: throw IllegalArgumentException` is a step back for both readability and safety: add a value to the shared enum and you get a runtime crash instead of a compiler error.

For the java -> kotlin -> ios goal it is better to keep a thin Android mapper `RouteEventType -> @StringRes` (an `EnumMap` or a static table), so a forgotten value is caught by a test rather than at runtime.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteSegment.kt
```diff
@@ -0,0 +1,61 @@
...
+	val destinationName: String? = null,
+	val destinationRef: String? = null,
+	val highway: String? = null,
+	val maximumSpeedMetersPerSecond: Float = 0f,
+	val lanes: Int = -1,
+	val oneWayDirection: Int = 0,
+	val roundabout: Boolean = false,
+	val tunnel: Boolean = false,
+	val routeTypes: List<RouteTypeAttribute> = emptyList(),
+	val heightValues: FloatArray = floatArrayOf(),
```
vshcherb, line 40: A `data class` holding a `FloatArray`: the generated `equals`/`hashCode` compare the array by reference. Two identical segments are not equal, and `hashCode` stops agreeing with `equals` after deserialization.

Same in `RouteGeometryCalculation` (IntArray), `RouteManeuverUpdate` and `RouteIntermediateCalculation`.

For a type declared as an immutable `@Serializable` contract that is going to iOS, this needs fixing: either `List<Float>`, or hand-written `equals`/`hashCode`.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteSegment.kt
vshcherb: `require(...)` over data that arrives from OBF/native is a crash, not validation.

`heightValues` comes from `RouteSegmentResult.getHeightValues()` and `distanceMeters` from native. One bad elevation value in someone's map and, instead of degraded statistics, the user gets an `IllegalArgumentException` on the route calculation path.

A routing backend cannot do this: either sanitize/clamp at the adapter boundary, or log and skip the segment.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteDetailsSnapshot.kt
```diff
@@ -0,0 +1,66 @@
...
+	val summary: RouteSummary,
+	val statistics: List<RouteStatistic> = emptyList(),
+	val currentRoutePointIndex: Int = 0,
+	val currentDirectionIndex: Int = 0,
+	val nextIntermediateIndex: Int = 0,
+	val intermediateRoutePointOffsets: List<Int> = emptyList(),
+) {
+	init {
+		require(schemaVersion > 0) { "Route details schema version must be positive" }
+		require((1 until points.size).all { index ->
```
vshcherb: Five `require` blocks, each walking the whole route (`points`, `segments`, `maneuvers`, `intermediateRoutePointOffsets`) — several O(N) passes on every snapshot construction.

Plus the same concern as in `RouteSegment`: `require` throws `IllegalArgumentException`. `require(points.first().distanceToFinishMeters == summary.totalDistanceMeters)` depends on `getWholeDistance()` being exactly `listDistance[0]`; change either side and it crashes.

If these are debugging invariants, they belong in a test or behind `assert`, not in a production constructor.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteManeuver.kt
```diff
@@ -0,0 +1,110 @@
+package net.osmand.shared.routing.details
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/** Exact semantic mapping of Android `TurnType` integer constants. */
+@Serializable
+enum class RouteManeuverType(val legacyValue: Int) {
```
vshcherb, line 8: `RouteManeuverType` duplicates the `TurnType` constants (C=1, TL=2, ... RNLB=14) — a second source of truth that has to be kept in sync by hand on every `TurnType` change.

Meanwhile `RouteManeuver.type` is not used in production at all (only by the test at 2653), and `turnTypeValue` is stored as an int anyway.

Since the goal is moving `OsmAnd-java` to Kotlin and on to iOS, the thing to port is `TurnType` itself, not a reduced copy of its values next to it. Otherwise iOS gets a 14-value enum while `TurnType` can express more (roundabout exits, lanes).

OsmAnd-java/src/main/java/net/osmand/router/RouteStatisticsHelper.java
```diff
@@ -50,44 +50,7 @@ public class RouteStatisticsHelper {
...
-		float dist;
-		float h;
-		float[] interpolatedHeightByStep;
-		float[] slopeByStep;
-		String[] slopeClassUserString;
-		int[] slopeClass;
-	}
-
-	public static List<RouteStatistics> calculateRouteStatistic(List<RouteSegmentResult> route,
+	public static List<RouteStatistic> calculateRouteStatistic(List<RouteSegmentResult> route,
```
vshcherb, line 52: This breaks the public `OsmAnd-java` API: `RouteStatisticsHelper.RouteStatistics` and `RouteSegmentAttribute` are removed and the signature now returns `net.osmand.shared.routing.details.RouteStatistic`. The same PR also deletes `net.osmand.router.ExitInfo`.

`OsmAnd-java` is a public library (MapCreator, server, the iOS wrapper), and `OsmAnd-shared` is wired in `OsmAnd-java/build.gradle:131` as `implementation`, so shared types in public signatures will not resolve transitively for consumers.

At minimum `api` instead of `implementation`, plus a pass over the external consumers.

OsmAnd-java/src/main/java/net/osmand/router/RouteStatisticsHelper.java
```diff
@@ -105,21 +68,71 @@ public static List<RouteStatistics> calculateRouteStatistic(List<RouteSegmentRes
...
 			attributesNames = getRouteStatisticAttrsNames(currentRenderer, defaultRenderer, false);
 		}
-		for (String attr : attributesNames) {
-			RouteStatisticComputer statisticComputer =
-					new RouteStatisticComputer(currentRenderer, defaultRenderer, currentSearchRequest, defaultSearchRequest);
-			RouteStatistics routeStatistics = statisticComputer.computeStatistic(routeSegmentWithInclines, attr);
-			Map<String, RouteSegmentAttribute> partitions = routeStatistics.partition;
-			if (!partitions.isEmpty() && (partitions.size() != 1 || !routeStatistics.partition.containsKey(UNDEFINED_ATTR))) {
-				result.add(routeStatistics);
+		List<RouteSegment> sharedRoute = new ArrayList<>(route.size());
```
vshcherb: Every call copies the whole route into a `List<RouteSegment>`, with a `List<RouteTypeAttribute>` per segment.

`ColoringStyleAlgorithms.isAttributeAvailableForDrawing()` calls `calculateRouteStatistic` one attribute at a time (`Collections.singletonList(attributeName)`), so building the coloring list means N full route copies. Previously `RouteSegmentResult` was used as is.

This needs a measurement on a long route (say 5000+ segments). If the number is bad, either convert once and expose the list, or keep `RouteSegmentResult` as the input and port it later, together with `OsmAnd-java` itself.

OsmAnd-java/src/main/java/net/osmand/router/RouteSegmentResultSnapshotAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+		int[] types = road.getTypes();
+		List<RouteTypeAttribute> result = new ArrayList<>(types.length);
+		RouteRegion region = road.region;
+		int encodingRulesSize = region.quickGetEncodingRulesSize();
+		for (int type : types) {
+			if (type < 0 || type >= encodingRulesSize) {
+				continue;
+			}
+			RouteTypeRule rule = region.quickGetEncodingRule(type);
+			if (rule != null && rule.getTag() != null && rule.getValue() != null) {
```
vshcherb: A behaviour difference the PR claims to have preserved.

The old `searchRenderingAttribute()` filtered only on `type < 0 || type >= encodingRulesSize` and appended the rule to `additional` unconditionally — including rules with `getValue() == null`, which ended up in the string as `tag=null;`. Here they are dropped, so the `additional` string handed to the renderer is different and attribute matching can change.

The `LOG.warn` about invalid encoding rules is lost as well — it used to catch broken maps.

Either reproduce the old behaviour exactly, or call this a fix explicitly and cover it with a test on a real map.

OsmAnd-java/src/main/java/net/osmand/router/RouteSegmentResultSnapshotAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+				result.add(new RouteTypeAttribute(rule.getTag(), rule.getValue()));
+			}
+		}
+		return result;
+	}
+
+	private static float[] copyHeightValues(float[] heightValues) {
```
vshcherb: The method copies nothing — it returns its argument. The comment explains why, but then it is not needed: call `source.getHeightValues()` directly.

OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
vshcherb: Mapping `RouteEvent -> AlarmInfo` through an `IdentityHashMap` relies entirely on `select()` returning the same instances it was given. Add a `.copy()` or a rebuild inside it and `alarmsByEvent.get()` returns null, and `if (alarmInfo != null)` silently swallows the alarms — the user simply stops getting warnings.

On substance: this was 43 lines of one readable loop and is now 47 lines with a parallel list, an `IdentityHashMap` and a 9-boolean options object, for the same behaviour. If `select()` is needed for iOS, let it return indices into the source list rather than objects — then mapping back is trivial and not fragile.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
```diff
@@ -52,6 +52,10 @@ public class RouteCalculationResult {
 	private final String errorMessage;
 	private final int[] listDistance;
 	private final int[] intermediatePoints;
+	private final int snapshotCurrentRoutePointIndex;
```
vshcherb: These three fields are written in all three constructors and are always equal to the initial values of `currentRoute` / `currentDirectionInfo` / `nextIntermediate`, i.e. 0 / 0 / 0.

Either they are not needed at all, or the snapshot should be taken from the current state rather than from the state at construction time — in which case they should be read at call time, not stored.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
vshcherb: A wrapper around `getListDistance(index)` with no difference whatsoever. The `ForSnapshot` suffix on three methods in a row is noise too — these are ordinary package-private getters.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
```diff
@@ -1516,4 +1460,24 @@ public void updateNextVisiblePoint(int nextPoint, Location mp) {
 	public GpxFile getGpxFile() {
 		return gpxFile;
 	}
+
+	/** Returns a lazily created immutable platform-neutral copy of this calculated route. */
+	@NonNull
+	public RouteDetailsSnapshot getRouteDetailsSnapshot() {
```
vshcherb: A lazy cache of a snapshot taken from a mutable object: `currentRoute`, `currentDirectionInfo` and `nextIntermediate` change while driving, but the snapshot is computed once and lives until the end of the route.

Nothing bites today because the only consumer is the dev dialog right after calculation. But once the snapshot is used for what it is meant for (iOS included), the cache will hand out stale indices. Either the snapshot is genuinely immutable and carries no "current" indices at all, or it is not cached.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResultSnapshotAdapter.kt
```diff
@@ -0,0 +1,179 @@
...
+	fun create(
+		source: RouteCalculationResult,
+		currentRoutePointIndex: Int,
+		currentDirectionIndex: Int,
+		nextIntermediateIndex: Int,
+	): RouteDetailsSnapshot {
+		val locations = source.immutableAllLocations
+		val directions = source.immutableAllDirections
+		return RouteDetailsSnapshot(
+			points = locations.mapIndexed { index, location ->
```
vshcherb: A full copy of the geometry: one `KLatLon` plus one `RoutePoint` per route point, plus copies of every segment and maneuver.

The strategy needs deciding here, because this contradicts `ILocationAccessor` (see the comment in `SharedRouteDetailsProvider`): either we do not copy and go through accessors, or we copy and the accessors are pointless. Right now the PR has both.

Given that the end state is `RouteCalculationResult` itself becoming Kotlin/shared, the snapshot and both adapters are a temporary bridge that gets deleted later. So it should not be grown: the minimal set of fields, and no serialization/schema until there is a second consumer.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResultSnapshotAdapter.kt
vshcherb: `RouteServiceType.valueOf(it.name)` links the two enums by name through reflection. Rename a constant in `RouteService` and it fails at runtime, with no help from the compiler.

An explicit `when` is better — then a new variant in `RouteService` will not compile silently.

OsmAnd/src/net/osmand/plus/routing/SharedRouteDetailsProvider.kt
```diff
@@ -0,0 +1,149 @@
...
+			currentRoutePointIndex,
+			distanceMeters,
+		)
+		if (index < 0) {
+			return null
+		}
+		return locations[index]
+	}
+
+	private class LocationAccessor(
```
vshcherb, line 109: Five implementations of three accessor interfaces (`ILocationAccessor`, `IManeuverAccessor`, `IManeuverMetricsAccessor`) exist specifically to avoid allocating a `KLatLon` per point — their KDoc says so. And right next to them `RouteCalculationResultSnapshotAdapter` allocates exactly one `KLatLon` per route point.

Two opposite strategies in one PR. One of them has to go, otherwise iOS will have to support both.

My vote is accessors for the hot calculations (`listDistance`, `updateDirectionsTime`) and no snapshot at all until there is a real second consumer.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteGeometryCalculator.kt
```diff
@@ -0,0 +1,169 @@
+package net.osmand.shared.routing.details
+
+import net.osmand.shared.data.KLatLon
+import net.osmand.shared.util.KMapUtils
+import kotlin.math.abs
+import kotlin.math.floor
+
+/** Android-compatible `RouteCalculationResult.listDistance` values. */
+data class RouteGeometryCalculation(
```
vshcherb, line 9: A `data class` holding an `IntArray` — `equals`/`hashCode` by reference, same as in `RouteSegment`. And the class itself is a wrapper around a single array that is unused in production: `RouteCalculationResult` goes through `calculateInto()`. If only the tests need it, drop it.

OsmAnd-java/src/test/java/net/osmand/router/RouteStatisticsHelperCompatibilityTest.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/** Android compatibility fixtures frozen from the pre-shared RouteStatisticsHelper behavior. */
+public class RouteStatisticsHelperCompatibilityTest {
+
+	@Test
+	public void preservesRendererFallbackGroupingUndefinedOmissionAndPartitionOrder() throws Exception {
```
vshcherb: This is not a compatibility test.

The renderer is assembled from an XML string inside the test, the route is three synthetic segments, and the expected values (`current_asphalt:30.0`, `-4% .. 0%:50.0`) were recorded from the new code. The old implementation is deleted in this same PR, so there is nothing to compare against — the test pins down what the code does now and cannot say whether it does what it used to.

What is needed: keep the old `RouteStatisticsHelper` around temporarily (under another name, or in the tests), run both implementations over a real route — `OsmAnd-java` already has routing tests with resources — and compare `elements` / `partition` / `totalDistance` element by element. Then the port is proven and the fixtures can be dropped along with the old code.

Separately: there is no test on a real `.obf` or a real `RouteCalculationResult` anywhere in the PR, and `SharedRouteDetailsProviderCompatibilityTest` is built the same way.

OsmAnd/res/values/strings.xml
vshcherb: New strings go at the beginning of `strings.xml` (AGENTS.md, Coding Standards) — otherwise the translation process breaks.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteEventHelper.kt
```diff
@@ -0,0 +1,186 @@
...
+						previousRailway = event
+					}
+				}
+				options.showTrafficWarnings || options.speakTrafficWarnings -> {
+					selected.add(RouteEventSelection(event, options.speakTrafficWarnings))
+				}
+			}
+		}
+		// MutableList.sortWith is stable, so events with the same route index retain input order.
+		selected.sortWith(compareBy { it.event.locationIndex })
```
vshcherb: A tie-break is lost: the old `WaypointHelper.sortList()` sorted by `routeIndex` and, on ties, by `deviationDistance`. Here it is only `compareBy { locationIndex }`.

For alarms `deviationDistance` is always 0, so the behaviour is in fact the same — but that should be pinned by a test rather than left as "it happens to match today". Also `sortList()` is no longer called for ALARMS at all: if someone changes `addPointWrapper` for alarms, the divergence will surface far from where it will be looked for.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteStatisticsCalculator.kt
vshcherb, line 355: Sorted twice: first `keys.sorted()`, then `sortedWith(comparator)`.

It works (both are stable, reproducing `TreeMap` + `Collections.sort`), but it does not read that way — a comment should say that the first sort establishes the order for equal elements and reproduces the `TreeMap` from the Java version. Otherwise the first person to optimize this will drop it and silently change the legend order on the chart.

OsmAnd/src/net/osmand/plus/routepreparationmenu/RouteDetailsFragment.java
```diff
@@ -1532,32 +1534,13 @@ public void onCardButtonPressed(@NonNull BaseCard card, int buttonIndex) {
...
-		if (position >= routeDirections.size()) {
-			return cumulativeInfo;
-		}
-		for (int i = 0; i < position; i++) {
-			RouteDirectionInfo routeDirectionInfo = routeDirections.get(i);
-			cumulativeInfo.time += routeDirectionInfo.getExpectedTime();
-			cumulativeInfo.distance += routeDirectionInfo.distance;
-		}
-		return cumulativeInfo;
+	public static List<RouteCumulativeInfo> getRouteDirectionCumulativeInfoByPosition(
```
vshcherb, line 1537: Good change, and independent of the migration: `getRouteDirectionCumulativeInfo(position, ...)` used to be called in a loop over all maneuvers, i.e. O(n^2) on every Route Details open and on the HTML print. Now it is O(n).

This could be split into a small separate PR and merged straight away.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteSegment.kt
alex-dev-neo: Resolved in a3a8f2dc6e. Constructor validation was removed from the backend DTOs, including RouteSegment, so OBF/native values are preserved instead of turning compatibility checks into production crashes.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteDetailsSnapshot.kt
alex-dev-neo: Resolved in a3a8f2dc6e. The snapshot constructor validation and its route-wide passes were removed.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
alex-dev-neo: Resolved in 51fb76e7f9. All alert dialog and TextView formatting was removed from RoutingHelper; this PR now contains only the backend migration.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
alex-dev-neo: Resolved in 51fb76e7f9. This second empty branch was removed with the debug formatter.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
alex-dev-neo: Resolved in 51fb76e7f9. The debug dialog and its reporting path were removed entirely, so route recalculation no longer builds a snapshot or debug string for this feature.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
alex-dev-neo: Resolved in 51fb76e7f9. The debug formatter, including this dead code, was removed.

OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java
alex-dev-neo: Resolved in 51fb76e7f9 by removing the diagnostics. The positional pairing between snapshot events and AlarmInfo no longer exists.

OsmAnd/src/net/osmand/plus/routing/AlarmInfo.java
alex-dev-neo: Resolved in 51fb76e7f9. sourceTag and sourceValue were removed from AlarmInfo together with the debug dialog, so the production model no longer retains those strings.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteSegment.kt
alex-dev-neo: Primitive arrays are intentional here to avoid boxing across the shared/native boundary, and none of these calculation/result containers is compared or used as a map key by Android or iOS. I agree that generated data-class equality is not structural for arrays; this PR does not rely on or promise equality for these transport/calculation containers.

OsmAnd/src/net/osmand/plus/routing/AlarmInfo.java
alex-dev-neo: RouteEventType is now the single shared source for event priority and classification. Android resource IDs cannot move to common code, so getVisualName remains a thin Android-only mapper. The default branch fails loudly if a new shared value is not mapped; Java switch statements are not compile-time exhaustive. No separate AlarmInfoType enum remains.

OsmAnd-java/src/main/java/net/osmand/router/RouteStatisticsHelper.java
alex-dev-neo: Resolved in fc5e16b313. Statistics now read RouteSegmentResult directly through IRouteStatisticsAccessor; the full RouteSegment and RouteTypeAttribute copy is gone, and route-type filters are prepared once per statistics calculation. On a 5,000-segment route the removed adapter took about 0.996 ms per conversion and allocated 5,000 RouteSegment objects, 5,000 route-type lists, and 66,000 RouteTypeAttribute objects before height arrays; the new path eliminates those copies.

OsmAnd/src/net/osmand/plus/helpers/WaypointHelper.java
alex-dev-neo: Resolved in 9567e93cb9. RouteEventHelper.select now returns RouteEventSelection with sourceIndex, and WaypointHelper maps back with alarms.get(sourceIndex). The IdentityHashMap and instance-identity dependency are gone.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteManeuver.kt
alex-dev-neo: turnTypeValue remains the authoritative value and preserves every TurnType integer verbatim. RouteManeuverType is only a nullable semantic view for the stable basic categories; it does not normalize or reject roundabout exits, lanes, or unknown values, and production logic does not depend on it. A complete TurnType migration is broader than this route-details backend slice.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResultSnapshotAdapter.kt
alex-dev-neo: Resolved for Android in 59a342857b. RouteCalculationResultSnapshotAdapter and the full geometry copy were removed; hot Android calculations now use the accessor path in SharedRouteDetailsProvider. The shared snapshot schema remains for the iOS consumer, but Android no longer builds or caches it.

OsmAnd-java/src/main/java/net/osmand/router/RouteSegmentResultSnapshotAdapter.java
alex-dev-neo: Resolved in 89e44faa09 and retained through the direct accessor introduced in fc5e16b313. Nullable rule values are preserved as tag=null; input, invalid rule IDs are skipped with the warning retained, and the behavior is covered at the adapter/accessor boundary.

OsmAnd-java/src/main/java/net/osmand/router/RouteStatisticsHelper.java
alex-dev-neo: This is an intentional source/API migration. All in-repository consumers, including tools, web, and MapCreator-related code, were checked and migrated to the shared RouteStatistic and RouteExitInfo types; the legacy RouteStatistics, RouteSegmentAttribute, and ExitInfo wrappers are not retained. External callers must add OsmAnd-shared explicitly and migrate to the shared types, so changing the Gradle edge to api would mask rather than remove that required migration.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
alex-dev-neo: Resolved in 59a342857b. The Android snapshot state fields and snapshot adapter were removed, so no initial navigation indices are stored.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
alex-dev-neo: Resolved in 59a342857b. The lazy Android snapshot cache and its mutable-state indices were removed.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResult.java
alex-dev-neo: Resolved in 59a342857b. The Android snapshot-only getters were removed with the snapshot adapter.

OsmAnd-java/src/main/java/net/osmand/router/RouteSegmentResultSnapshotAdapter.java
alex-dev-neo: Resolved in fc5e16b313. RouteSegmentResultSnapshotAdapter was removed; the statistics accessor now returns RouteSegmentResult.getHeightValues() directly.

OsmAnd-java/src/test/java/net/osmand/router/RouteStatisticsHelperCompatibilityTest.java
alex-dev-neo: Resolved in f6731a46af. The test now routes through the real Routing_test_64.obf with the default renderer, runs the preserved legacy implementation and the shared implementation over the same RouteSegmentResult list, and compares every statistic name, total distance, element, partition entry, property, user property, color, distance, and order.

OsmAnd/src/net/osmand/plus/routing/SharedRouteDetailsProvider.kt
alex-dev-neo: Resolved in 59a342857b. Android now uses only the accessor-based hot calculation path in SharedRouteDetailsProvider; the competing Android full-snapshot adapter was removed.

OsmAnd/src/net/osmand/plus/routing/RouteCalculationResultSnapshotAdapter.kt
alex-dev-neo: Resolved in 59a342857b. RouteCalculationResultSnapshotAdapter was removed, so the RouteServiceType.valueOf name-based conversion no longer exists on Android.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteGeometryCalculator.kt
alex-dev-neo: Production Android and iOS paths use calculateInto/accessors, so RouteGeometryCalculation is not on the hot path. It remains as the shared list-oriented API/result used by the pure common calculations. IntArray is intentional to avoid boxing, and neither consumer relies on generated data-class equality.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteStatisticsCalculator.kt
alex-dev-neo: Resolved in d2dce9f102. The code now documents that the first natural-key sort reproduces the legacy TreeMap order and that the following stable comparator preserves this order for ties.

OsmAnd/res/values/strings.xml
alex-dev-neo: Resolved in 51fb76e7f9. These strings were removed together with the debug dialog.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/details/RouteEventHelper.kt
alex-dev-neo: Pinned by RouteEventHelperTest. The selection sort is stable by locationIndex, and ordinaryWarningsAreNotDeduplicatedAndEqualIndexesRemainStable verifies that equal alarm indices retain source order. select now returns source indices, so WaypointHelper maps the exact original alarm without identity assumptions.

OsmAnd/src/net/osmand/plus/routepreparationmenu/RouteDetailsFragment.java
alex-dev-neo: Kept in this PR intentionally. The batch calculation is the shared backend API used by both Android and iOS Route Details, preserves the legacy position==size boundary result, and is covered by the compatibility tests. Splitting it would temporarily restore the O(n^2) caller or duplicate the bridge work, so there is no benefit to separating it at this stage.

PR #25776, opened 2026-08-29
Restore custom GPX extension fields in point menus

### LINKED ISSUE #25226: Regression: Custom GPX XMLNS extension tags are missing or incorrectly filtered in the Context Menu
## Description
In OsmAnd 4.7, support for custom tags was introduced. By adding a custom schema as an `xmlns`, any custom tags were properly displayed in the **Context Menu** for **Favorites** or **Waypoints**. Currently, this functionality appears to be completely broken.

> Previous task https://github.com/osmandapp/OsmAnd/issues/19027#issuecomment-1968760863

### How to reproduce

* Import file – [custom_extenstion.gpx.zip](https://github.com/osmandapp/OsmAnd/files/14433050/custom_extenstion.gpx.zip)
* Import it as a **Favorite** or a **GPX** file.
* Tap on the point to open the **Context Menu**.

```
    <wpt lat="37.5460870" lon="-77.4532843">
        <time>2023-06-07T12:31:35Z</time>
        <name>Test</name>
        <type>SOTM</type>
        <extensions>
            ...
            <test:country>United States</test:country>
            <test:state>Virginia</test:state>
            <test:telephone>(804) 828-0100</test:telephone>
            <test:postcode>23284</test:postcode>
            <test:start_date>Thursday, June 8, 2023</test:start_date>
        </extensions>
    </wpt>

```
### Actual result

<table style="width:100%">
  <tr>
    <th>❌ 5.1.9</th>
    <th>❌ 5.0.5</th>
    <th>❌ 4.9.10</th>
    <th>✅ 4.7.17</th>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
    <td> </td>
    <td> </td>
  </tr>
</table>  

* **5.1.9** - Does not work; no custom tags are visible.
* **5.0.5**, **4.9.10**, **4.8.6** – Partially working: `<test:postcode>` and `<test:state>` are visible.
* * `<test:country>` and `<test:telephone>` act as filters instead of data fields.
* **4.7.17** – All custom tags are correctly visible.

### Expected result

* 4.7.17 – All custom tags should be visible in the Context Menu, matching the behavior verified in version 4.7.17.

**Environment**
OsmAnd Version: 5.1.9, 5.0.5, 4.9.10, 4.8.6, 4.7.17
Android: 17
Device model: Google Pixel 6A

nazar-kutz:
## Summary

Fixes #25226.

Restores custom GPX extension fields in the Context Menu for Waypoints and Favorites.

The regression had two related causes:

1. External namespaced GPX extensions lost their XML prefix during parsing.

   For example:

   `<test:country>United States</test:country>`

   was read as `country`, which lost part of the field identity and could cause it to be incorrectly interpreted by the standard OsmAnd POI logic or filtered out.

2. Unknown extensions stored directly in a Waypoint/Favorite were filtered when building additional information for the Context Menu.

## Changes

- Preserve qualified keys for external prefixed GPX extensions:
  - `<test:country>` → `test:country`
- Keep existing `osmand` and `gpxtpx` extension handling backward-compatible.
- Do not classify custom GPX fields using `:`, because valid OSM keys such as `contact:phone` may also contain it.
- Keep the existing POI resolution as the first priority.
- If a field stored directly in a Waypoint/Favorite is not recognized by the standard POI logic, display it using a generic text row.
- Keep Amenity-derived metadata behavior unchanged.
- Exclude existing `osm_tag_` / `amenity_` persisted Amenity fields from the new generic fallback.

Conceptually:

- stored field + known POI type → existing specialized rendering
- stored field + unknown POI type → generic rendering
- Amenity-only field → existing behavior

No new persistent marker or provenance metadata is introduced.

## Verification

Tested with the GPX attached to #25226.

The following custom fields are displayed correctly again:

- Country
- Postcode
- Start date
- State
- Telephone

Verified:

- imported GPX Waypoint Context Menu;
- direct import as Favorite;
- reopening the Favorite after application restart;
- known OSM keys containing `:` still use the existing POI resolution;
- unknown Amenity-only fields are not exposed by the new fallback.

Regression tests were added for:

- preserving external GPX prefixes;
- existing POI resolution for known stored fields;
- generic fallback for unknown stored extensions;
- Amenity-only field filtering;
- mixed stored/Amenity data.

### POI regression check

Since `AmenityUIHelper` is shared with regular POI rendering, additional characterization checks were performed to make sure the new stored-extension fallback does not change existing Amenity output.

The comparison was performed in two stages:

1. A sample of 10,000 POIs was rendered before and after the change and the resulting rows were compared.
2. The same comparison was then run for all 138,016 POIs from the Kyiv map.

For the final implementation, the generated output was identical to the baseline:

- no existing rows were removed;
- no existing row values changed;
- no existing rows were reordered;
- no additional Amenity-only/internal fields appeared.

For the full Kyiv dataset, both baseline and updated snapshots contained 47,407,561 bytes and had the same SHA-256 hash.

These snapshot checks were used as development/characterization tests and are not included in the committed test suite.

### Favorites imported by affected older versions

Some previously imported Favorites may already have lost the original namespace identity, for example:

`test:country` → `country`

Once the prefix has been discarded, the original namespace cannot be reliably reconstructed from the stored Favorite data.

If the field itself is still stored, the current fallback may still display it. If the original field data was already lost, re-importing the source GPX is required.

### GPX Waypoint → Add → Favorite

During manual testing, a separate existing data-loss flow was found:

`GPX Waypoint → Context Menu → Add → Save Favorite`

This path goes through `FavoritePointEditor.add(...)`, which creates a new `FavouritePoint` from basic point data and does not copy the source `WptPt` extensions.

Other GPX-to-Favorites flows that use `FavouritePoint.fromWpt(...)` do copy extensions.

This is a separate issue from #25226 and is intentionally not changed in this PR, since fixing it changes the semantics of the generic Add-to-Favorites flow for Waypoints in general, not only custom GPX extensions.

### DIFF
--- a/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
+++ b/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
@@ -61,8 +61,10 @@ object GpxUtilities {
 	const val PINNED_EXTENSION = "pinned"
 	const val POINT_TYPE_EXTENSION = "point_type"
 
-	const val GPXTPX_PREFIX = "gpxtpx:"
-	const val OSMAND_EXTENSIONS_PREFIX = "osmand:"
+	private const val GPXTPX_XML_PREFIX = "gpxtpx"
+	private const val OSMAND_XML_PREFIX = "osmand"
+	const val GPXTPX_PREFIX = "$GPXTPX_XML_PREFIX:"
+	const val OSMAND_EXTENSIONS_PREFIX = "$OSMAND_XML_PREFIX:"
 	const val OSM_PREFIX = "osm_tag_"
 	const val AMENITY_PREFIX = "amenity_"
 	const val ORIGIN_EXTENSION = "origin"
@@ -1688,6 +1690,16 @@ object GpxUtilities {
 		return supportedTag ?: tag.replace(XML_COLON, ":")
 	}
 
+	private fun getQualifiedExtensionTagName(parser: XmlPullParser): String? {
+		val name = parser.getName() ?: return null
+		val prefix = parser.getPrefix()
+		return if (prefix.isNullOrEmpty() || prefix == OSMAND_XML_PREFIX || prefix == GPXTPX_XML_PREFIX) {
+			name
+		} else {
+			"$prefix:$name"
+		}
+	}
+
 	@Throws(XmlParserException::class, IOException::class)
 	private fun readExtensionsText(parser: XmlPullParser, key: String, target: GpxExtensions) {
 		var tok: Int
@@ -1698,7 +1710,7 @@ object GpxUtilities {
 				if (tag != null && text != null) {
 					val value = text.toString()
 					if (!value.isBlank()) {
-						applyExtensionValue(target, tag, value)
+						applyExtensionValue(target, getQualifiedExtensionTagName(parser) ?: tag, value)
 					}
 				}
 				if (tag == key) {
--- a/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
+++ b/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
@@ -63,6 +63,7 @@ object GpxUtilities {
 
 	private const val GPXTPX_XML_PREFIX = "gpxtpx"
 	private const val OSMAND_XML_PREFIX = "osmand"
+	private const val GPXX_XML_PREFIX = "gpxx"
 	const val GPXTPX_PREFIX = "$GPXTPX_XML_PREFIX:"
 	const val OSMAND_EXTENSIONS_PREFIX = "$OSMAND_XML_PREFIX:"
 	const val OSM_PREFIX = "osm_tag_"
@@ -118,6 +119,13 @@ object GpxUtilities {
 			}
 		}
 	}
+
+	// Extensions from these namespaces keep their local tag names, so that a file binding the
+	// same namespace to another prefix (e.g. "ns3:hr" for TrackPointExtension) stays recognized.
+	private val KNOWN_EXTENSION_HOSTS = setOf("osmand.net", "garmin.com")
+
+	private val KNOWN_EXTENSION_PREFIXES =
+		setOf(OSMAND_XML_PREFIX, GPXTPX_XML_PREFIX, GPXX_XML_PREFIX)
 
 	private val SUPPORTED_EXTENSION_TAGS = mapOf(
 		"heartrate" to PointAttributes.SENSOR_TAG_HEART_RATE,
@@ -1693,11 +1701,14 @@ object GpxUtilities {
 	private fun getQualifiedExtensionTagName(parser: XmlPullParser): String? {
 		val name = parser.getName() ?: return null
 		val prefix = parser.getPrefix()
-		return if (prefix.isNullOrEmpty() || prefix == OSMAND_XML_PREFIX || prefix == GPXTPX_XML_PREFIX) {
-			name
-		} else {
-			"$prefix:$name"
+		if (prefix.isNullOrEmpty()) {
+			return name
 		}
+		val host = parser.getNamespace()?.lowercase()
+			?.substringAfter("://")?.substringBefore('/') ?: ""
+		val known = KNOWN_EXTENSION_HOSTS.any { host == it || host.endsWith(".$it") }
+				|| KNOWN_EXTENSION_PREFIXES.contains(prefix)
+		return if (known) name else "$prefix:$name"
 	}
 
 	@Throws(XmlParserException::class, IOException::class)
--- a/OsmAnd-shared/src/commonTest/kotlin/net/osmand/shared/GpxUtilitiesLoadTest.kt
+++ b/OsmAnd-shared/src/commonTest/kotlin/net/osmand/shared/GpxUtilitiesLoadTest.kt
@@ -66,6 +66,30 @@ class GpxUtilitiesLoadTest {
 	}
 
 	@Test
+	fun testLoadGpxFilePreservesCustomExtensionPrefix() {
+		val gpxFile = loadGpx(
+			"""
+			<gpx version="1.1" creator="test" xmlns:test="https://example.com/gpx/test">
+			  <wpt lat="10.0" lon="20.0">
+			    <extensions>
+			      <test:country>United States</test:country>
+			      <test:telephone>+1 804 828 0100</test:telephone>
+			    </extensions>
+			  </wpt>
+			</gpx>
+			""".trimIndent(),
+			addGeneralTrack = false
+		)
+
+		assertNull(gpxFile.error)
+		val extensions = gpxFile.getPointsList().single().getExtensionsToRead()
+		assertEquals("United States", extensions["test:country"])
+		assertEquals("+1 804 828 0100", extensions["test:telephone"])
+		assertNull(extensions["country"])
+		assertNull(extensions["telephone"])
+	}
+
+	@Test
 	fun testLoadGpxFilePreservesGeneralTrackBehaviorForMultipleSegments() {
 		val gpxFile = loadGpx(
 			"""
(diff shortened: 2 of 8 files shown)


alex-dev-neo:
(gen by AI)

Reviewed commit `9ffc4c135aa2686ef6e86ea35e2f112b22539715` against #25226.

The five custom fields from the issue's GPX are restored in the Waypoint/Favorite menus, and the custom-field GPX write/reload check passes. However, I recommend addressing the following findings before merging.

1. **[P2] Standard Garmin extensions stop being recognized when the namespace uses another prefix.**  
   [GpxUtilities.kt, lines 1693–1700](https://github.com/osmandapp/OsmAnd/blob/9ffc4c135aa2686ef6e86ea35e2f112b22539715/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt#L1693-L1700)

   The helper recognizes the literal `gpxtpx` prefix instead of the namespace URI. With the valid binding `xmlns:ns3="http://www.garmin.com/xmlschemas/TrackPointExtension/v1"`, `<ns3:hr>145</ns3:hr>` and `<ns3:cad>91</ns3:cad>` are now stored as `ns3:hr` and `ns3:cad`. Sensor consumers still look up the canonical `gpxtpx:hr` / `gpxtpx:cad` keys, so the measurements stop being recognized for sensor analysis and charts.

   A diagnostic test passes on the PR base (`4be04a4649bc459de26c0761b04ffceb33fbb5e3`) and fails on the head:
   ```text
   {ns3:hr=145, ns3:cad=91} expected:<145> but was:<null>
   ```
   Please recognize supported namespaces by URI and retain their canonical internal keys.

2. **[P2] Generic fallback changes the displayed value of custom fields.**  
   [AmenityUIHelper.java, lines 357–362](https://github.com/osmandapp/OsmAnd/blob/9ffc4c135aa2686ef6e86ea35e2f112b22539715/OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java#L357-L362)

   Arbitrary user values go through `poiTypes.getPoiTranslation(vl)`. Its fallback replaces underscores with spaces and capitalizes the first letter; it can also interpret the value as a translatable POI key. On the emulator, `test:reference=abc_def` renders as `Abc def`. This changes the displayed identifier, although the stored GPX value remains intact.

   Please pass the original `vl` to the generic text row and translate the label independently.

3. **[P2] Internal waypoint/favorite metadata becomes visible.**  
   [AmenityExtensionsHelper.java, lines 108–114](https://github.com/osmandapp/OsmAnd/blob/9ffc4c135aa2686ef6e86ea35e2f112b22539715/OsmAnd/src/net/osmand/plus/helpers/AmenityExtensionsHelper.java#L108-L114)

   Excluding only `amenity_*` and `osm_tag_*` admits internal point metadata into the fallback. The Android test confirms that stored `hidden=true` and `visited_date=2024-01-01T00:00:00Z` produce visible rows. These fields are written by `FavouritePoint.toWpt()`, retained by `fromWpt()`, and are not covered by `AdditionalInfoBundle.HIDDEN_EXTENSIONS`.

   Please exclude internal point fields as well as persisted Amenity metadata.

4. **[P2] Some custom XMLNS fields are still filtered before reaching the fallback.**  
   [AmenityUIHelper.java, lines 346–357](https://github.com/osmandapp/OsmAnd/blob/9ffc4c135aa2686ef6e86ea35e2f112b22539715/OsmAnd/src/net/osmand/plus/mapcontextmenu/builders/AmenityUIHelper.java#L346-L357), [AdditionalInfoBundle.java, lines 153–157](https://github.com/osmandapp/OsmAnd/blob/9ffc4c135aa2686ef6e86ea35e2f112b22539715/OsmAnd-java/src/main/java/net/osmand/data/AdditionalInfoBundle.java#L153-L157)

   `test:route_id=1234` remains invisible even when included in `genericFallbackKeys`: the earlier `isKeyToSkip()` check rejects any key containing `route`. An Android test reproduces the missing row. Similar substring checks exist for content/Wikipedia fields. This is a remaining gap against the requirement to display custom XMLNS fields, rather than a newly introduced filtering rule.

   Please account for field origin at the filtering stage too, while keeping explicit internal-field exclusions and existing Amenity behavior.

**Validation**

- macOS arm64, JDK 17.0.17, repository Gradle wrapper 8.9.
- Built `nightlyFreeLegacyArm64Debug` and its instrumentation APK successfully.
- Ran on Pixel_9a AVD, Android 15 / API 35, arm64, portrait, light theme, English UI, font scale 1.0.
- All 4 tests in the PR's `GpxUtilitiesLoadTest` passed.
- All 5 original `AmenityUIHelperStoredExtensionsTest` tests passed.
- Four additional diagnostic cases reproduced the findings above: one JVM test and three Android tests.
- A separate test using the original GPX attachment verified all five values for the parsed waypoint and a favorite created with `FavouritePoint.fromWpt()`, then opened both actual context menus. It passed.
- The diagnostic tests were added only in local review worktrees; they are not currently part of this PR.

Relevant commands:
```sh
./gradlew :OsmAnd-shared:jvmTest --tests net.osmand.shared.GpxUtilitiesLoadTest --console=plain
./gradlew :OsmAnd:assembleNightlyFreeLegacyArm64Debug :OsmAnd:assembleNightlyFreeLegacyArm64DebugAndroidTest --console=plain
./gradlew :OsmAnd:connectedNightlyFreeLegacyArm64DebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=net.osmand.test.ui.poi.AmenityUIHelperStoredExtensionsTest --console=plain
```

The checks cover parsing, conversion, menu rendering, and GPX round-trip. Full file-picker import, persisted-favorite reopening after application restart, and the Kyiv POI snapshot comparison were not independently repeated. The documented limitations concerning prefixes already lost in older versions and the separate Waypoint → Add → Favorite flow appear reasonable.

Chumva:
(gen by AI)

Head is now `29e6545c99`; the review above was against `9ffc4c135a`.

**Fixed — 3 of the 4 findings**

1. Garmin/OsmAnd namespaces bound to another prefix (`ns3:hr`, `ns2:DisplayColor`): namespaces are matched by URI host, so an alternative prefix keeps the canonical keys. Covered by a test using `ns1`/`ns2`/`ns3` that asserts heart rate, cadence, width and `DisplayColor`.
2. Generic fallback rewrote the value (`abc_def` → `Abc def`): the stored value is now passed through untranslated.
3. Internal point fields (`hidden`, `visited_date`) became visible: only externally qualified keys may fall back, plus a canary test that runs a fully populated `FavouritePoint.toWpt()` and fails naming any key that leaks.

**Still open — finding 4.** A custom key containing `route`, `content` or `wikipedia` is dropped before the fallback: `shouldDisplayKey()` (`AmenityUIHelper.java:110`) and `isKeyToSkip()` (`AmenityUIHelper.java:315`) both run before `genericFallbackKeys` is consulted. It is recorded as a `Known gap` in the commit message and pinned by `customFieldMatchingAmenityFilterIsStillNotShown`. Fixing it means reordering those filters on the code path shared with ordinary POI menus, so it is left to the invisible-tags refactoring in 5.5.

Consequence: the reproduction in #25226 is fixed — all five fields of the attached GPX render, both as a GPX track and as an imported Favorite — but the issue's broader "all custom tags" wording is not fully met. Worth deciding whether `Fixes #25226` should stay, or whether the gap should become a follow-up issue.

**Verification** — variant `androidFullLegacyFatDebug`, JDK 21, repository Gradle wrapper:

- `:OsmAnd-shared:jvmTest` 42/42; `:OsmAnd-java:test` 546/546.
- `AmenityUIHelperStoredExtensionsTest` 11/11 on an API 36 emulator and on a Pixel 8 Pro (Android 17, API 37).
- The GPX attached to #25226 imported on the Pixel: all five fields present in both the waypoint and the Favorite context menu.
- Characterization dumps against the `r5.4` merge base: POI context-menu rows (24 tag maps, including 12 real POI categories) are byte-identical, so ordinary POI rendering is unchanged; GPX parsing over 13 documents differs only in the two intended cases.

**Two notes.** `AmenityUIHelperStoredExtensionsTest` now extends `AndroidTest`, which required backporting the androidx.test versions master already uses (runner/rules 1.7.0, espresso 3.7.0, ext:junit 1.3.0): with 3.5.1 every `AndroidTest` subclass fails in `setup()` on API 35+, because `Espresso.onIdle()` reaches `InputManager.getInstance()`, removed in Android 15. `EditPoiTypeSuggestionsTest` is red after that bump, but it could not run at all before it on these API levels, so no green baseline was available to compare against.

The PR description is also slightly ahead of the code now: the fallback is limited to externally qualified keys, so unknown *unqualified* fields are not shown, and Favorites imported before the prefix was preserved are not recovered.

Issue #25587, opened 2026-08-06
Add support for the panoramax=* key


scaidermern:
[`panoramax=*`](https://wiki.openstreetmap.org/wiki/Key:panoramax) is used for [photo linking](https://wiki.openstreetmap.org/wiki/Photo_linking) from an OSM object to a picture of that object in [Panoramax](https://panoramax.fr/), the open source and federated street-level imagery platform (similar to `image=*`, `wikimedia_commons=*`, `mapillary=*` etc.). The value to use is the picture's unique identifier (UUID).

According to [taginfo](https://taginfo.openstreetmap.org/keys/panoramax), OSM has around 125k elements with the panoramax key.

The UUID of the picture is enough to get access to the picture by querying the Panoramax meta-catalog.
Example: `panoramax=cafb0ec8-51dd-43ac-836c-8cd1f7cb8725`
- Direct access to full definition picture : https://api.panoramax.xyz/api/pictures/cafb0ec8-51dd-43ac-836c-8cd1f7cb8725/hd.jpg
- Direct access to low definition picture : https://api.panoramax.xyz/api/pictures/cafb0ec8-51dd-43ac-836c-8cd1f7cb8725/sd.jpg
- Direct access to the picture thumbnail : https://api.panoramax.xyz/api/pictures/cafb0ec8-51dd-43ac-836c-8cd1f7cb8725/thumb.jpg
- Access to a web viewer: https://api.panoramax.xyz/#focus=pic&pic=cafb0ec8-51dd-43ac-836c-8cd1f7cb8725
- Access to the picture metadata (including links to full definition, low definition and thumbnail): [https://api.panoramax.xyz/api/search?ids=cafb0ec8-51dd-43ac-836c-8cd1f7cb8725](https://api.panoramax.xyz/api/search?limit=10&ids=cafb0ec8-51dd-43ac-836c-8cd1f7cb8725)

Note: This is _not_ a duplicate of #20555. #20555 is about a Panoramax _map layer_ while this support request is solely for the `panoramax=*` key on individual OSM elements for linking individual OSM to specific photos.

mnalis:
> while this support request is solely for the panoramax=* key on individual OSM elements for linking individual OSM to specific photos.

Absolutely support this! It is quite needed, especially as those linked pictures help a lot for orientation / locating of the amenity one is looking for when one does reach the destination. 

[...]

Issue #20555, opened 2024-08-10
Panoramax street-imagery plugin


unnervingduck:
### Describe the idea (required)

A plugin that shows street-imagery from https://panoramax.fr as a map layer, much like the existing Mapillary one. At minimum it would only have to show an overlay on the map with the recorded portions.

### Tell us about the expected behaviour (required)

The plugin would ideally be similar to the Mapillary plugin, but even just an overlay that shows the already recorded streets would suffice. This would greatly help recording new imagery, as you could plan your routes directly on OsmAnd to cover all streets.

### Tell us about alternatives you've considered (required)

Mapillary as a service is not preferable due to the acquisition by Facebook, which as a company has a nasty track record. The alternatives such as KartaView aren't viable options in the long-term either, because stalled development can essentially kill the service.

Panoramax is federated and ensures it's not controlled by a single entity, and has openly licensed imagery ready to be used by everyone.

### Context (optional)

Panoramax was recently integrated into the iD editor: https://www.openstreetmap.org/user/Juicio/diary/404174

Al-Hun:
Hello !
At least the 'panoramax' tag should be used to show in the "on line pictures" displayed when clicking on a POI : at this time this tag is used on nearly 2300 objects in OSM, and will increase since the Panoramax project is yet implemented by new countries.
On the other side it would be a way to get good pictures of POI (as WikiMedia's ones) having the option to send photo notes to Panoramax : always better than pictures taken on the way from a car to illustrate important POI, increasing the touristic feature of Osmand. If done il will be important to confirm the good location and direction of the picture before sending.

pietervdvn:
Since MapComplete uses a panoramax instance, the number of tags is quickly increasing: https://taghistory.raifer.tech/?#***/panoramax/

~~Our instance is however not yet federated with panoramax.xyz (but should be soon) - which should be the main access point for OsmAnd.~~

scaidermern:
To support Panoramax, two things are necessary:
- Support the `panoramax=*` tag on OSM elements. This tag works the same way as the `image=*`, `wikimedia_commons=*` and `mapillary=*` tags. Adding support for this should be easy.
- Support a Panoramax overlay similar to the Mapillary overlay. This probably requires more effort.

Maybe these two tasks should be separated from each other as they don't have much in common.

pietervdvn:
[...]

PR #25777, opened 2026-08-29
Add Panoramax street-level imagery plugin (#20555)

dktue:
# Panoramax street-level imagery plugin (#20555)

Adds a Panoramax layer to OsmAnd, closing the gap described in #20555. Draft rather than
ready-for-review: the marker artwork is placeholder, and several surfaces have not been
exercised on a device yet. Both are listed explicitly below.

Opened at the suggestion of @vshcherb, who saw this being built at State of the Map 2026.

## What it does

Sequence lines at zoom 15-16, individual photo points with heading cones from 17, tapping a
point opens the picture, plus date / 360-only / contributor filters, a map widget and a quick
action. Structure mirrors the existing Mapillary plugin, as a separate
`net.osmand.plus.plugins.panoramax` package, so nothing about Mapillary changes.

https://github.com/user-attachments/assets/9a00a9d0-3fb9-4f6a-8683-3b1bcd233134

*Zoom 15 to 18 over Place Charles de Gaulle: sequence lines resolving into individual photos.*

Sequence lines at zoom 16 Photo points at zoom 17

A 360 picture open in the viewer, Notre-Dame

*Tapping a photo point opens the picture. Notre-Dame, Paris.*

Basemap © OpenStreetMap contributors. Coverage data © Panoramax contributors, CC-BY-SA 4.0.
Photograph © cquest, CC-BY-SA 4.0, via Panoramax.

## Things the tiles do differently from Mapillary

These drove most of the implementation and are worth knowing if you review it:

**Sequences do not exist below zoom 15.** Measured against the live service, zooms 6-14 return
only an aggregated `grid` layer, 15-16 return `sequences`, 17 returns `sequences` + `pictures`,
and 18 returns HTTP 400 - the service caps at 17. Mapillary's constants (`MAX_SEQUENCE_LAYER_ZOOM
= 13`) therefore fetch tiles containing nothing this layer can draw.

**Timestamps are not epoch milliseconds.** `ts` is `2024-07-18 07:48:24.248+00`: a space
separator, and between zero and six fractional digits. Across 1,249 pictures in one tile the
fractional length was three digits 1,037 times, two 125, none 44, six 26, one 11, five 5 and
four once, so a fixed `.SSS` pattern mishandles roughly one in six.

**The two layers date their features with different keys.** `pictures` uses `ts`, `sequences`
uses `date` (a plain day). `filtered()` runs over both, so reading only `ts` scores every
sequence as epoch 0 and the whole overview disappears the moment a date filter is set.

**Picture urls redirect with HTTP 308.** `/api/pictures/{id}/hd.jpg` redirects to whichever
federated instance stores the file. `HttpURLConnection` follows 301/302/303 but not 308, so the
url is resolved through the STAC API first, which also returns the licence and contributor.

**The viewer needs DOM storage.** Unlike the osmand.net Mapillary proxy page, the Panoramax
viewer is a single page app that reads `localStorage` on startup.

## Notes on the approach

- The tile cache extension stays `.pbf`. `GeometryTilesCache` and `BitmapTilesCache` route a
  source by comparing against `MAPILLARY_VECTOR_TILE_EXT`, so any other value would send MVT
  tiles to the bitmap decoder.
- The layer shares the existing `GeometryTilesCache` rather than adding one.
  `ResourceManager.getTilesCache()` returns the first cache whose `isTileSourceSupported()`
  matches, and both vector sources use `.pbf`, so a second instance could never be selected.
  Cached tile ids are prefixed with the source name, so they cannot collide.
- No context menu gallery row. That row is fed by the online photos service through
  `OnlinePhotosGroup`, which has no Panoramax member, so it could only ever render empty.
- No Panoramax app hand-off. There is no Android app, so the widget opens the web viewer.

## Verified on a device

Xiaomi 12T Pro, Android 15, and an API 35 emulator, `androidFullOpenglArm64Debug` and
`androidFullLegacyFatDebug`:

- sequence lines render at zoom 15-16 and photo points at 17+, on **both** the legacy and the
  OpenGL renderer, the latter exercising `PanoramaxTilesProvider` through the native core
- tiles are requested at the intended zooms and cached via `GeometryTilesCache`
- tapping a point produces a "Panoramax image" entry in the context menu and opens the dialog
- the viewer opens and renders the photograph, and dragging rotates the 360 view
- `PanoramaxImage.setData()` accepts all 1,249 picture features of a real tile decoded by
  OsmAnd's own `BinaryVectorTileReader`, with no unparsable timestamps

## Not verified

Being explicit rather than implying completeness:

- the date, 360-only and contributor filters
- the map widget and the quick action
- sequence navigation arrows in the viewer
- offline behaviour, landscape, tablets, any locale other than English

## Known gaps

- **Artwork is placeholder.** The markers are neutral shapes generated in Panoramax's published
  purple (`#54278f`, taken from their MapLibre style); the plugin icon is OsmAnd's generic
  street-imagery icon. Real Panoramax assets are needed before this ships. An earlier iteration
  copied the Mapillary bitmaps and those were removed.
- Below zoom 15 the layer draws nothing, where the official viewer renders the `grid` layer as a
  coverage heatmap. Rendering `grid` would be a reasonable follow-up.
- The `panoramax=*` OSM tag is out of scope here; that is #25587.
- The instance is hardcoded to `api.panoramax.xyz`, isolated in `PanoramaxConstants`.

## Separate upstream bug found

`WeatherTileResourceProvider_P::obtainGeoTile` null-dereferences `TileSqliteDatabase::isOpened()`
when the weather plugin is enabled with no weather data. Two reproducible SIGSEGVs with identical
stacks, unrelated to this change. Happy to file it separately.

## How this was written

The implementation was produced by an AI coding agent, reviewed and directed by me. Flagging it
so you can calibrate review effort: the diff is large, and several defects listed above were
found only by running it rather than by reading it.


nazar-kutz:
## Panoramax — Product & UX review

I'm currently reviewing and polishing the Panoramax implementation from #25777 to prepare it for merge.

First of all, thanks to @dktue for the initial implementation — most of the core functionality is already in place. The current work is focused on fixing remaining issues, polishing the integration, and bringing the feature to a merge-ready state.

During the review and device testing, several questions came up that require product and design decisions before the UI and behavior can be finalized.

This comment collects:
- **product and UX questions** where the desired behavior still needs to be defined;
- **visual and design questions** around the current UI;
- **missing or temporary visual resources** that need to be reviewed or replaced before the feature is finalized.

The items are grouped by feature area, with screenshots or videos shown directly where useful. Additional context, possible options, and implementation details are kept inside expandable sections.

> **Implementation context**
>
> The initial Panoramax integration was largely modeled on the existing Mapillary implementation. This was a useful starting point and allowed most of the feature to be implemented consistently with OsmAnd.
>
> At the same time, Panoramax differs from Mapillary in its data, viewer, available controls, and some interaction patterns. Because of this, several copied UI and behavior decisions now need to be reviewed rather than automatically kept the same.


## 1. Panoramax viewer

<table>
  <tr>
    <th>Current Panoramax viewer</th>
    <th>Current interaction</th>
    <th>Mapillary reference</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      <video src="https://github.com/user-attachments/assets/2256a289-01d3-4f39-9616-aead4dfcd55d" width="300" height="auto"/>
    </td>
    <td>
      
    </td>
  </tr>
</table>

<details>
<summary><b>1.1 Viewer size and Panoramax UI</b></summary>

The current Panoramax image dialog follows the general structure of the existing Mapillary integration.

However, Panoramax has significantly more built-in UI inside its own viewer, so the same layout leaves relatively little space for the actual panorama.

At the moment several UI layers are visible at the same time:

- OsmAnd toolbar and actions;
- Panoramax toolbar and controls;
- Panoramax image information / bottom panel.

A possible direction would be to support two states:

- **Compact view** — keep the OsmAnd map visible and show mainly the panorama with a reduced set of Panoramax controls.
- **Expanded view** — allow the viewer to expand to a larger or full-screen state with additional controls and image information.

The Panoramax Web Viewer officially supports a photo-only component, disabling its default widgets, and selectively adding individual widgets, so this interaction model is technically feasible.

Questions to decide:

1. Should the current viewer stay the same size, become larger, or support an expanded/full-screen state?
2. Which controls should remain visible in compact mode?
3. Should the information panel be hidden in compact mode?
4. Should the expanded state expose the full Panoramax interface?

</details>

<details>
<summary><b>1.2 Synchronization with the OsmAnd map</b></summary>

Currently, the Panoramax viewer and the OsmAnd map become partially independent after the viewer is opened.

For example:

- rotating the panorama does not update the viewing direction / heading marker on the OsmAnd map;
- navigating to another Panoramax image does not move the selected image marker on the map.

The Panoramax viewer API provides the events required for this synchronization:

- current picture ID;
- picture latitude / longitude;
- current absolute viewing heading;
- picture selection / navigation changes.

So synchronizing the OsmAnd selected marker and viewing direction with the embedded viewer is technically feasible.

However, this requires a proper integration layer between the Panoramax Web Viewer and Android rather than the current direct hosted-page WebView approach.

It would be useful to decide whether this synchronization is required for the first merged version or should be implemented as a follow-up improvement.

</details>

<details>
<summary><b>1.3 Implementation scope / follow-up consideration</b></summary>

A technical prototype confirmed that the richer integration is possible, but it is not a small UI-only change.

A robust implementation would likely require an OsmAnd-controlled Panoramax viewer wrapper and a defined WebView ↔ Android communication layer.

This also needs a decision on attribution, because reducing or disabling the standard Panoramax UI may also remove the attribution currently shown by the official viewer.

For this reason, one possible approach is:

- merge the current functional viewer implementation;
- define the desired viewer UX now;
- implement the deeper Panoramax ↔ OsmAnd synchronization as a follow-up.

</details>


## 2. Configure Map

<table>
  <tr>
    <th>Mapillary</th>
    <th>Panoramax</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>

<details>
<summary><b>2.1 Mapillary / Panoramax naming in Configure Map</b></summary>

The Panoramax Configure Map entry was initially modeled after Mapillary.

Originally, both entries used **Street-level imagery** as their title. Apart from being difficult to distinguish visually, this also caused a technical collision because the Configure Map UI uses the title resource ID as part of the entry identity.

The collision itself has already been fixed, but the current result is visually inconsistent:

- **Mapillary**
  - title: `Street-level imagery`
  - description: `Mapillary`
- **Panoramax**
  - title: `Panoramax`
  - description: `Street-level imagery`

So we still need to define how these two providers should be presented consistently.

### Possible direction

My preferred option would be to use the provider as the primary title:

**Mapillary**  
`Street-level imagery: On`

**Panoramax**  
`Street-level imagery: On`

This makes the provider immediately identifiable while the secondary text explains what the toggle controls.

Another possible approach would be to introduce a separate **Street-level imagery** section/category containing Mapillary and Panoramax, but this would add another category to Configure Map and may be unnecessary for only a few provider-specific layers.

It would be useful to decide which hierarchy we want to use before finalizing these entries.

</details>


## 3. Street-level imagery filters

<table>
  <tr>
    <th>Mapillary filters</th>
    <th>Panoramax filters</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
  <tr>
    <td><code>fragment_street_level_imagery_filters.xml</code></td>
    <td><code>fragment_street_level_imagery_filters.xml</code></td>
  </tr>
</table>

<details>
<summary><b>3.1 Visual polish of the filter screen</b></summary>

Mapillary and Panoramax now share the same filter layout, which makes it easier to keep the two integrations visually consistent.

However, several parts of the existing UI could still use some polishing.

In particular:

- the **From / To** date fields do not look fully consistent with each other;
- the contributor / username input still looks different from newer OsmAnd input components;
- some interactive rows, such as **Display only 360° images** and the main Street-level imagery toggle, could use clearer touch/selectable feedback;
- spacing and alignment of some controls could be made more consistent with current OsmAnd screens.

The goal here would not be a major redesign, but to make the existing screen look clean and intentional enough to keep as-is if no larger redesign is planned.

A design review of the current screen would be useful before we finalize these smaller UI changes.

</details>

<details>
<summary><b>3.2 Apply / Reset behavior</b></summary>

The screen currently contains explicit **Apply** and **Reset** actions, so it would be useful to define the expected interaction model clearly.

There are two possible models:

1. **Transactional**
   - the user changes several filters;
   - the map remains unchanged while editing;
   - **Apply** commits all changes at once;
   - **Reset** resets the draft values.

2. **Immediate**
   - every filter change is applied to the map immediately;
   - in this case an explicit **Apply** action may be unnecessary or should have a different purpose.

The current implementation follows the existing Mapillary screen closely, but before polishing the behavior it would be useful to confirm which model we actually want for both providers.

The date filter should also have predictable full-day semantics:
- **From** → beginning of the selected day;
- **To** → end of the selected day.

</details>

<details>
<summary><b>3.3 Panoramax contributor filter</b></summary>

Panoramax additionally supports filtering imagery by contributor.

The contributor lookup is implemented and the filter works against the Panoramax account ID, while Mapillary does not currently expose the same control.

This means the shared screen intentionally differs between the two providers:

- **Mapillary** — contributor control hidden;
- **Panoramax** — contributor control available.

It would be useful to confirm that we want to keep this Panoramax-specific filter exposed in the final UI.

</details>


## 4. First-use / Show widget dialog

<table>
  <tr>
    <th>Mapillary</th>
    <th>Panoramax</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
  <tr>
    <td><code>street_imagery_first_dialog.xml</code></td>
    <td><code>street_imagery_first_dialog.xml</code></td>
  </tr>
</table>

<details>
<summary><b>4.1 First-use flow</b></summary>

Mapillary and Panoramax now use the same first-use dialog, with provider-specific icon, color, title, description and widget text.

The dialog is shown when the street-level imagery layer is enabled for the first time and has not yet been acknowledged for that provider.

Its main purpose is to introduce the feature and optionally enable the corresponding map widget.

Current behavior:

- the widget switch reflects the real widget state for the current profile;
- changing the switch only changes the draft state inside the dialog;
- dismissing the dialog without confirmation does not change the widget configuration;
- pressing **OK** applies the selected widget state;
- after the first-use dialog is acknowledged, it is not normally shown again automatically.

It would be useful to confirm whether this is the desired onboarding flow.

In particular:

1. Do we want to keep this first-use dialog for both Mapillary and Panoramax?
2. Is enabling the optional map widget the right primary action to offer here?
3. Should this dialog remain a one-time onboarding step, or should there be another way to access the same choice later?

</details>

<details>
<summary><b>4.2 Dialog presentation</b></summary>

The dialog has already been made shared between Mapillary and Panoramax and its behavior has been cleaned up, but the presentation could still use a design review.

In particular, it would be useful to review:

- whether the current custom bottom-sheet presentation should remain;
- spacing and visual hierarchy between the description and widget row;
- the placement and style of the confirmation action;
- whether the dialog should follow a more standard OsmAnd bottom-sheet pattern with the usual close/action treatment.

The goal here would not necessarily be a redesign, but to confirm whether the current presentation is good enough to keep or whether a small visual cleanup is needed before the Panoramax integration is finalized.

</details>


## 5. Panoramax map widget

<table>
  <tr>
    <th>Mapillary widget</th>
    <th>Panoramax widget</th>
    <th>Panoramax widget behavior</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
    <td>
        <video src="https://github.com/user-attachments/assets/7dac8037-aef9-4014-a317-681d916cf460" width="300" height="auto"/>
    </td>
  </tr>
</table>

<details>
<summary><b>5.1 Widget title and size</b></summary>

The current widget displays the full **Panoramax** name.

During the UI review, the widget looked relatively wide compared with other map widgets, and the current text takes a noticeable amount of horizontal space.

It would be useful to decide whether:

1. **Panoramax** should remain the visible widget label;
2. a shorter label should be used;
3. the widget should rely more on its icon and use less or no persistent text.

The final choice should also be reviewed together with the dedicated Panoramax widget artwork.

</details>

<details>
<summary><b>5.2 Expected widget behavior</b></summary>

The current Panoramax widget follows the existing Mapillary pattern.

At the moment, tapping it simply opens Panoramax. It does not:

- show or hide the Panoramax map layer;
- reflect whether the layer is currently enabled;
- interact with the currently selected Panoramax image;
- expose any Panoramax state directly on the map.

It would be useful to confirm whether this is the intended role of the widget.

Possible directions include:

- keep it as a simple shortcut to Panoramax;
- use it to show/hide the Panoramax layer;
- make it contextual to the currently selected Panoramax image;
- remove the widget if the Quick Action / Configure Map entry already cover the useful interaction sufficiently.

This does not necessarily require changing the current implementation, but it would be good to define the intended purpose before finalizing the widget UI.

</details>


## 6. Panoramax plugin appearance

<table>
  <tr>
    <th>Mapillary plugin</th>
    <th>Panoramax plugin — current state</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>

<details>
<summary><b>6.1 Panoramax visual identity</b></summary>

The Panoramax integration is functionally available in the Plugins screen, but several visual resources are still temporary or generic.

Compared with Mapillary, Panoramax currently does not yet have a complete OsmAnd-specific visual identity across all entry points.

The main missing / temporary resources include:

- dedicated Panoramax main/action icon;
- dedicated day/night widget icons;
- plugin banner / promo image.

There are also Panoramax-specific map resources already included in the implementation that would benefit from design review:

- selected image marker;
- viewing-direction marker;
- photo point marker;
- Panoramax accent color.

The detailed resource list, current fallbacks, resource IDs, and Mapillary references can be collected separately for design review.

At the product level, it would be useful to confirm that Panoramax should have its own consistent visual identity across the plugin screen, Configure Map, widget, Quick Action, and image viewer.

</details>


## 7. Attribution and image information

<table>
  <tr>
    <th>Current Panoramax viewer</th>
    <th>Image information in Panoramax</th>
  </tr>
  <tr>
    <td>
      
    </td>
        <td>
        <video src="https://github.com/user-attachments/assets/076ca07c-5291-4192-bcbf-7d595e89ccd5" width="300" height="auto"/>
  </tr>
</table>

<details>
<summary><b>7.1 Attribution and metadata</b></summary>

Panoramax imagery contains information such as the image author / producer and license.

With the current implementation, OsmAnd opens the official Panoramax viewer and relies on the information presented by Panoramax itself rather than duplicating this metadata in the OsmAnd UI.

It would be useful to confirm whether this is sufficient for the intended integration.

In particular:

1. Is attribution shown inside the official Panoramax viewer sufficient?
2. Do we want OsmAnd itself to display any image metadata such as:
   - contributor / producer;
   - capture date;
   - license;
   - Panoramax instance / source?
3. If we move toward the proposed compact viewer with parts of the standard Panoramax UI hidden, which attribution information must remain permanently visible?

This is especially relevant to the compact-view proposal described above: technically we can reduce the standard Panoramax UI, but we should not hide attribution that is expected or required to remain visible.

</details>

nazar-kutz:
## Panoramax design resources

There are still several Panoramax visual resources that need either dedicated artwork or design review.

For missing resources, Android currently uses temporary generic OsmAnd assets.  
Mapillary resources are listed only as references for where an equivalent Panoramax resource is expected.

### Resources to add

| Purpose | Mapillary resource | Panoramax — current temporary resource | Needed |
| --- | --- | --- | --- |
| Main / action icon | `ic_action_mapillary` | `ic_action_photo_street` | Add a dedicated Panoramax action icon |
| Widget icon — day | `widget_mapillary_day` | `widget_av_photo_day` | Add a dedicated Panoramax widget icon for day mode |
| Widget icon — night | `widget_mapillary_night` | `widget_av_photo_night` | Add a dedicated Panoramax widget icon for night mode |
| Plugin banner | `mapillary.webp` | Standard OsmAnd plugin placeholder | Add a Panoramax plugin banner / promo image |

These resources are used across the Panoramax plugin UI, Configure Map, widgets, Quick Actions and related entry points.

### Existing Panoramax resources to review

These Panoramax-specific resources already exist and are functional, but it would be useful to review whether they are visually ready to keep.

| Purpose | Mapillary reference | Panoramax resource | Action |
| --- | --- | --- | --- |
| Selected image marker | `map_mapillary_location*` | `map_panoramax_location*` | Review / approve or redesign |
| Viewing-direction marker | `map_mapillary_location_view_angle*` | `map_panoramax_location_view_angle*` | Review / approve or redesign |
| Photo point marker | `map_mapillary_photo_dot*` | `map_panoramax_photo_dot*` | Review / approve or redesign |
| Primary color | Mapillary color resources | `panoramax_color` (`#54278f`) | Confirm or adjust |

For now I mainly want to identify which resources need design work. I can add screenshots / visual comparisons separately if needed.

dmpr0:
- [x] Image viewer should show the image, not the website; if it’s not possible, we should think about whether we really need this integration.


### Resources
We need to replace all resource with vector if it's possible. 
- [x] map_panoramax_photo_dot*

### Remove
- [x] Remove Panoramax widget
- [x] Remove 4. First-use / Show widget dialog

OsmAnd/src/net/osmand/plus/download/local/LocalOperationHelper.java
```diff
@@ -188,7 +188,12 @@ private void clearMapillaryTiles(@NonNull LocalItem item) {
 		ITileSource src = (ITileSource) item.getAttachedObject();
 		ITileSource mapilaryCache = TileSourceManager.getMapillaryCacheSource();
 		ITileSource mapilaryVector = TileSourceManager.getMapillaryVectorSource();
-		if (src != null && (mapilaryVector.getName().equals(src.getName()) || mapilaryCache.getName().equals(src.getName()))) {
+		ITileSource panoramaxCache = TileSourceManager.getPanoramaxCacheSource();
+		ITileSource panoramaxVector = TileSourceManager.getPanoramaxVectorSource();
+		if (src != null && (mapilaryVector.getName().equals(src.getName())
+				|| mapilaryCache.getName().equals(src.getName())
+				|| panoramaxVector.getName().equals(src.getName())
+				|| panoramaxCache.getName().equals(src.getName()))) {
```
Chumva: The entry condition now accepts the two Panoramax sources, but the loop below still maps only `mapilaryCache` / `mapilaryVector` (`if (withoutExt.equals(mapilaryCache.getName())) ... else if (... mapilaryVector ...)`). Clearing or deleting `Panoramax (raster tiles).sqlitedb` therefore wipes both Mapillary caches and leaves the Panoramax sibling cache untouched, so on OpenGL the pre-rendered Panoramax tiles keep being served. The loop needs the two Panoramax names as well, ideally clearing only the siblings of the provider that was tapped. (gen by AI)

OsmAnd/src/net/osmand/plus/plugins/panoramax/PanoramaxVectorLayer.java
```diff
@@ -0,0 +1,674 @@
...
+				if (panoramaxTilesProvider != null) {
+					mapRenderer.resetMapLayerProvider(layerIndex);
+					clearMapMarkersCollections();
+					panoramaxTilesProvider = null;
+				}
+				return;
+			}
+			PanoramaxFilterState currentFilterState = PanoramaxFilterState.read(plugin);
+			if (!currentFilterState.equals(appliedFilterState)) {
+				if (panoramaxTilesProvider != null) {
```
Chumva, line 158: When `map == null` the method returns above without touching `appliedFilterState`, and the provider is dropped. The filter screen is reachable while the layer is off (Configure Map row -> options), so: change "360 only" with the layer hidden, then switch the layer on. The states differ, but `panoramaxTilesProvider` is null, so `clearCache()` is skipped and the new provider opens the same `Panoramax (raster tiles).sqlitedb`, whose `isTileExist()` returns tiles rendered with the previous filter. Same when the filter was changed in a session where the layer was never shown. Traced in code, not reproduced on a device. Suggest persisting the filter state the raster DB was rendered with (a pref or the sqlite info table) and clearing the DB in the provider constructor when it differs, or clearing it from the filter screen on change/reset. Mapillary has the same gap. (gen by AI)

OsmAnd/src/net/osmand/plus/plugins/panoramax/PanoramaxFiltersFragment.java
```diff
@@ -0,0 +1,307 @@
...
+
+
+        Button apply = view.findViewById(R.id.button_apply);
+        disableButtonApply(view);
+        apply.setOnClickListener(v -> {
+            String username = textView.getText().toString();
+            String dateFrom = dateFromEt.getText().toString();
+            String dateTo = dateToEt.getText().toString();
+
+            if (!plugin.PANORAMAX_FILTER_USERNAME.get().isEmpty() || !dateFrom.isEmpty() || !dateTo.isEmpty() || plugin.PANORAMAX_FILTER_PANO.get()) {
```
Chumva, line 217: `PANORAMAX_FILTER_USER_KEY` / `PANORAMAX_FILTER_USERNAME` are written as a side effect of autocompletion (`PanoramaxAutoCompleteAdapter.performFiltering`), and this condition looks at the preference rather than the field. Type a name, delete it, tap Apply: `USE_PANORAMAX_FILTER` becomes true and the map is filtered by the last looked-up account while the field is empty. Copied from Mapillary, where it is dormant because the row is hidden; here it is live. When `username.isEmpty()`, clear both prefs and decide `USE_PANORAMAX_FILTER` from the field. (gen by AI)

OsmAnd/src/net/osmand/plus/plugins/panoramax/PanoramaxFiltersFragment.java
```diff
@@ -0,0 +1,307 @@
...
+
+
+        EditText dateFromEt = view.findViewById(R.id.date_from_edit_text);
+        DatePickerDialog.OnDateSetListener dateFromDialog = (v, year, monthOfYear, dayOfMonth) -> {
+            Calendar from = Calendar.getInstance();
+            from.set(Calendar.YEAR, year);
+            from.set(Calendar.MONTH, monthOfYear);
+            from.set(Calendar.DAY_OF_MONTH, dayOfMonth);
+            dateFromEt.setText(dateFormat.format(from.getTime()));
+            plugin.PANORAMAX_FILTER_FROM_DATE.set(from.getTimeInMillis());
```
Chumva, line 152: `Calendar.getInstance()` with only Y/M/D set keeps the current h:m:s, so `from` is the chosen day at the current wall-clock time (same for `to` below). The `sequences` layer carries only `date` (e.g. `2025-09-05`, checked on live tiles), which `PanoramaxImage.parseCaptureTime()` reads as 00:00 UTC, so any `from` set after midnight UTC hides every sequence of that day at z15-16 while its pictures (`ts` later that day) still show at z17; `to` likewise drops pictures taken after the current time of day. Normalise `from` to the start of the day and `to` to the end of the day, and compare `date`-only features at day granularity. (gen by AI)

OsmAnd/src/net/osmand/core/android/PanoramaxTilesProvider.java
```diff
@@ -0,0 +1,619 @@
...
+	@Override
+	public long obtainImageData(IMapTiledDataProvider.Request request, SWIGTYPE_p_QByteArray byteArray) {
+		IQueryController queryController = request.getQueryController();
+		if (queryController != null && queryController.isAborted()) {
+			return 0;
+		}
+		int requestZoom = request.getZoom().swigValue();
+		ZoomLevel swigZoom = request.getZoom();
+		TileId swigTileId = request.getTileId();
+		if (panoramaxBitmapTileCache.isTileExist(swigTileId, requestZoom)) {
```
Chumva, line 152: Inherited from Mapillary, but worth fixing here since Panoramax coverage grows fast: `isTileExist()` only checks presence, so rendered raster tiles never expire although the vector `.pbf` expires daily, and the "Reload tiles" button (`clearCacheAndTiles(getPanoramaxVectorSource())`) does not touch this DB. On OpenGL new coverage never appears until a filter change, and after "Reload tiles" the dots stay visible but stop being tappable, because `lazyLoadClickMap()` needs the deleted `.pbf` (`imgExist` is false and nothing re-downloads it). Reload should also clear `Panoramax (raster tiles)` (e.g. through the layer's provider `clearCache()`), or the provider should re-render when the geometry tile is missing or expired. (gen by AI)

OsmAnd/src/net/osmand/plus/plugins/panoramax/PanoramaxImageDialog.java
```diff
@@ -0,0 +1,500 @@
...
+			@Override
+			public void onMessage(WebMessagePort port, WebMessage message) {
+				onViewerMessage(message.getData());
+			}
+		}, new Handler(Looper.getMainLooper()));
+		webView.postWebMessage(new WebMessage(viewerNonce, new WebMessagePort[] {ports[1]}),
+				Uri.parse(PanoramaxConstants.INSTANCE_URL));
+	}
+
+	private void releaseViewerPort() {
```
Chumva, line 365: This is only called from `onPageFinished`. When the card is dismissed nothing closes the port or stops the WebView: `ContextMenuCardDialogFragment.onDestroyView()` checks `contentView instanceof WebView`, but the content view is the `panoramax_web_view` FrameLayout, so that branch is dead for both providers. For Mapillary that left a static page alive; here it is a WebGL 360 viewer with an open `WebMessagePort` (callback -> dialog -> MapActivity) until GC. A dialog `onDestroyView` hook that releases the port and calls `webView.destroy()` would fix both. (gen by AI)

OsmAnd/src/net/osmand/plus/plugins/panoramax/PanoramaxVectorLayer.java
```diff
@@ -0,0 +1,674 @@
...
+			}
+		}
+		this.visiblePoints = visiblePoints;
+		drawSelectedPoint(canvas, tileBox);
+	}
+
+	private void drawLines(Canvas canvas, RotatedTileBox tileBox, List<Geometry> geometries,
+	                       int tileX, int tileY, int tileZoom, PanoramaxFilterState filterState) {
+		for (Geometry geometry : geometries) {
+			if (geometry.isEmpty() || filterState.filtered(geometry.getUserData())) {
```
Chumva, line 357: Legacy renderer: `filtered()` runs on every geometry before the type check. Live z15 tiles also contain a `pictures` layer (2,442 points in one Paris tile), so with a date filter this is a regex plus `LocalDateTime` parse per point per frame at z15-16, for points that `drawLines` then ignores anyway. Check `instanceof LineString` / `MultiLineString` first, or parse the timestamp once per feature. (gen by AI)

OsmAnd/src/net/osmand/plus/download/local/LocalOperationHelper.java
Chumva: Fixed in ebfe08ed3a: the sibling pair is now resolved from the tapped source, so only that provider's vector and raster caches are cleared. (gen by AI)

OsmAnd/src/net/osmand/plus/plugins/panoramax/PanoramaxVectorLayer.java
nazar-kutz: Should be fixed in https://github.com/osmandapp/OsmAnd/pull/26215/changes/58c6dbb0a3ed579591df3470f036aa011a96eba1

OsmAnd/src/net/osmand/plus/plugins/panoramax/PanoramaxFiltersFragment.java
nazar-kutz: Should be fixed in https://github.com/osmandapp/OsmAnd/pull/26215/changes/5e958dbf934262143b0f453a60c113c409aee343

OsmAnd/src/net/osmand/plus/plugins/panoramax/PanoramaxFiltersFragment.java
nazar-kutz: Should be fixed in 6e06b1c and 3e0fd5df42799e65b93fc5be1ea521a48a17c548.
The first commit normalizes the selected date boundaries, and the follow-up
compares Panoramax date-only sequence features at calendar-day granularity.

PR #25781, opened 2026-08-29
Group Maps & Resources search results into Regions and Maps

### LINKED ISSUE #25727: Improve search results screen in Maps & Resources
## Description
<table style="width:100%">
  <tr>
    <th>Android</th>
    <th>iOS</th>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
  </tr>
</table>   

Currently, search results in Maps & Resources are shown as one list, and it's hard to understand which is a region and what is a map.

* Divide into two groups: Regions / Maps
* Add a Country name on the second line of the map item.

vshcherb:
## Group Maps & Resources search results into Regions and Maps

Fixes #25727

### Problem

Search results in Maps & Resources were shown as one flat list. A region folder and a
downloadable map with the same name looked identical, and the map with the wider coverage
was indistinguishable from the one covering a city on the other side of the world.

On top of that, the same file could be offered twice: a city found in the world basemap
resolves to the map covering it, and that map was often already listed as a region map —
which is why two "Berlin" rows with an identical size showed up.

### Changes

- Results are split into two sections, **Regions** and **Maps**. Each section is sorted on
  its own and headers are not clickable.
- Every row shows the region it belongs to on its second line: the country for regions and
  maps (`Berlin` → `Germany`), the covering map for cities (`Berlin` → `New Hampshire`).
  A country shows nothing rather than repeating its own name.
- A city is dropped when the map covering it is already listed, and several cities covered
  by the same map produce a single row.
- City rows show the map type the same way map rows do, and the map region moved out of the
  two-line title into the secondary line.

The whole search — matching, grouping, the city lookup in the world basemap and the country
lookup — moved out of `SearchDialogFragment`'s inner filter into `DownloadSearchUIModel`, which
is what the tests drive; the fragment is UI only. `ItemViewHolder` and `DownloadGroupViewHolder`
gained opt-in second-line support; it is off by default, so the region and updates screens
render exactly as before.

### Screenshots

| Before | After | After, dark theme |
| --- | --- | --- |
| before | after-light | after-dark |

The "before" shot is the one from #25727: six rows in one list, two of them the same 127.20 MB
Berlin map. After the change the same query returns two labelled sections and every map is
listed once.

### Performance

Every city is resolved to its map before the list is built, so the second line is there on the
first paint and the old asynchronous `IndexItemResolverTask` is gone. That is affordable because
the lookup walks the region tree that is already in memory instead of reading `regions.ocbf` per
city. Worst case query `"san"` — 167 cities, full catalog of 1083 maps, on an emulator:

| resolving 167 cities | |
| --- | --- |
| walk the in-memory region tree | **19 ms** |
| `OsmandRegions.getCountryName(LatLon)` | 800 ms |
| `OsmandRegions.getSmallestBinaryMapDataObjectAt(LatLon)` | 823 ms |

For scale, the basemap POI search that produces those 167 cities costs 2233 ms on the same
query, so the resolution is under 1% of the screen's cost. The region tree has 1722 regions,
1710 of them with boundaries, depth 5; the descent prunes to 154 point-in-polygon tests per city.

`europe_gb` carries no boundary in `regions.ocbf` (bbox `[0,0 - 0,0]`), which is why a point in
England was never resolved; `WorldRegion.hasBoundaries()` marks such grouping regions and the
walk descends through them.

### Tests

`OsmAnd/test/java/net/osmand/plus/download/DownloadSearchUIModelTest.java`, 6 tests, 32 s.

Nothing in them is simulated. The maps are every map the region tree of `regions.ocbf` offers
(1083 of them) and the cities come from a real `World_basemap_mini.obf` search. Both files are
bundled assets that the app unpacks on start, so **no map has to be downloaded and no network is
needed**. The only thing the tests build is the `IndexItem` of each map; its size and date come
from `indexes.xml` at runtime and neither is part of the rows being asserted.

- `searchResultsMatchExamples` — 18 examples from `OsmAnd/test/assets/download/search_examples.json`,
  each a query and the exact rows it must produce: first the rules of the screen, then the corner
  cases that broke it — `washington`, `san juan`, `santo domingo`, `sanhe`, `holland`, `hamburg`,
  `georgia`.
- `searchIsUnambiguousOnTheFullCatalog` — the broad queries no example spells out: `san`, `santa`,
  `port`, `york`, `kyiv`, `utrecht`. `san` alone puts 195 rows on the screen.
- `citiesResolveToTheMapThatCoversThem` — a table of 14 cities, their coordinates and the country
  each one is in, written from the map rather than from a test run, so a wrong map cannot pass.
- `cityResolutionAgreesWithTheRegionFile` — resolves the 250+ cities three broad queries return and
  compares each map against the one the quad tree in `regions.ocbf` names, a separate
  implementation over the same data.
- `cityIsHiddenWhenItsMapIsAlreadyListed`, `conditionsAreParsedAsAndOfOrs`.

The last two exist because the recorded screens can only confirm that the code still does what it
did, not that it does the right thing. They caught the walk returning the first region covering a
city instead of the smallest — a search for Santo Domingo offered the whole Caribbean archipelago
map — which none of the example screens covered.

Every screen, examples included, is checked against the rules that must hold for any query: one
header per section in a fixed order, no file offered twice, no two rows of a section reading the
same. `WorldRegion.getCountryRegion()` is covered separately by `OsmandRegionsTest` as a JVM test.

The tests write every screen to `download_search_report.html` in the app external files
directory, so the results can be reviewed without taking screenshots:

```
adb shell am instrument -w -e class net.osmand.plus.download.DownloadSearchUIModelTest \
    net.osmand.plus.test/androidx.test.runner.AndroidJUnitRunner
adb pull /sdcard/Android/data/net.osmand.plus/files/download_search_report.html
```

Run through `am instrument` rather than `connectedAndroidTest` — the latter uninstalls the app
at the end and the report goes with it. The Jenkins UI-tests job keeps the app
(`-DuninstallAfterTest=false`) and passes no class filter, so this class runs there
automatically and the report survives the run.

### Not applicable

Android Auto has no Maps & Resources screen; the only map-download touchpoint there is
`MissingMapsScreen`, which shows a message and a "View on phone" action. None of the changed
classes are reachable from a head unit.

### How this PR was written

Written with an AI assistant (Claude Code) driving the build, the emulator and the measurements.
Condensed here so the reasoning is readable — each step is what was asked and what it settled.

1. **Implement the issue, show screenshots, check it on Android Auto, and write tests driven by a
   JSON file of example queries.** Gave the two sections, the second line, the JSON examples and
   the instrumentation test. Android Auto was checked and ruled out — no such screen exists there.
2. **"The duplicate Berlin with the same size looks like a bug. Run the tests over the USA and
   Europe catalogs, and have them generate an HTML file so we don't pass screenshots around."**
   Root cause: a city found in the basemap resolves to a map that is already listed. Produced the
   deduplication and `download_search_report.html`.
3. **First review round.** Less code, as few new classes as possible, a 1000–1200 loc budget, one
   result list instead of two, drop the duplicated country lookup, name classes after the screen,
   add `London` / `New york` / `New` / `San Fr` examples. Produced `DownloadSearchUIModel`,
   `WorldRegion.getCountryRegion()` and `DownloadSearchScreenReport`.
4. **"Why do we simulate maps instead of using WorldRegion, OsmandRegions and World_basemap_mini
   as is? How long do the tests take, and will they run automatically in CI?"** The decisive step.
   The regions were already real; the cities were the made-up part, because `searchCities()` was
   private inside the fragment. Moving it into the model and searching the real basemap
   immediately exposed two defects the hand-written cities had hidden: `washington` produced two
   rows both reading just `Washington`, and York in Great Britain never resolved at all.
5. **"Is a quad tree needed here? Assess the code complexity and how many objects there are."**
   Measured rather than assumed — the table under Performance. Conclusion: the in-memory walk is
   40× faster than the ready-made `regions.ocbf` lookups, and a quad tree would save under 1% of
   the screen's cost, so it was not added. If it is ever wanted, `OsmandRegions` is the place.
6. **"Why is the hand-picked `resources` list still there if everything else is real data?"**
   Removed; every example now runs against the whole catalog, and the invariants are asserted on
   the examples too, not only on the extra ambiguous queries.
7. **"The examples only record what the code does. Think of city cases yourself and check whether
   they actually make sense — the data is real."** The one that found the last defect. Two checks
   that do not use the implementation as their own oracle: a table of cities and the country each
   is in, taken from the map, and a comparison against the quad tree in `regions.ocbf`. 16 of 345
   cities disagreed, all of them offered an aggregate map instead of the country map.
8. **"Why are your corner cases not in `search_examples.json`? Add the most interesting ones."**
   They had been left in Java as assertions on a country name. Each is a query like any other, so
   the seven that carry the most information are now written down as screens in the examples file.

Aspects covered along the way: the duplicate-file defect and its second-order form found by
@alex-dev-neo, first-match versus smallest-match when several regions cover a city, city
resolution cost and where the real bottleneck is, regions with no boundary (`europe_gb`) and with
detached parts (Sanhe, Sansha) in `regions.ocbf`, localized city names versus the name the request
is matched against, CI behaviour of the Jenkins UI-tests job, and Android Auto reachability.

Worth keeping in mind for similar work: the golden screens were generated from a test run, so for
three rounds they confirmed the code was consistent rather than correct. Every defect after the
first review was found by comparing against something outside the implementation — the real data,
a second implementation, or geography written down by hand.


OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
```diff
@@ -0,0 +1,420 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Builds the content of the "Maps & Resources" search results screen.
```
vshcherb: It's bad to create new class with helper name. It looks it could stay in the old class if it's absolutely necessary then better to create with class like DownloadSearchUIModel

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
```diff
@@ -0,0 +1,420 @@
...
+	 */
+	@Nullable
+	private IndexItem resolveCityMap(@NonNull DownloadResources indexes, @NonNull CityItem city) {
+		Amenity amenity = city.getAmenity();
+		if (amenity == null) {
+			return null;
+		}
+		WorldRegion region = null;
+		try {
+			Map.Entry<WorldRegion, BinaryMapDataObject> entry =
```
vshcherb: Lines could be wider we have 120 max width

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
```diff
@@ -0,0 +1,420 @@
...
+	/**
+	 * @param indexes       all available resources.
+	 * @param searchRequest text typed by the user.
+	 * @param cities        cities found in the world basemap, already matched by the caller.
+	 * @return rows to be shown on the screen: section headers, regions and maps.
+	 */
+	@NonNull
+	public List<Object> search(@NonNull DownloadResources indexes, @NonNull String searchRequest,
+	                           @NonNull List<CityItem> cities) {
+		List<Object> matched = new ArrayList<>();
```
vshcherb: why create 2 vars matched,  found?  it looks like unnecessary ineffecient step, however I'm not sure to the end

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
```diff
@@ -0,0 +1,420 @@
...
+		WorldRegion country = getCountryRegion(region);
+		return country != null ? country.getLocaleName() : null;
+	}
+
+	/**
+	 * Countries are the regions directly under a continent. Regions that are not placed under any
+	 * continent are represented by their topmost parent below the world.
+	 */
+	@Nullable
+	public static WorldRegion getCountryRegion(@Nullable WorldRegion region) {
```
vshcherb: Duplicate code???     public WorldRegion getCountryRegion() { - check inside World Region..

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
```diff
@@ -0,0 +1,420 @@
...
+		WorldRegion region = DownloadResourceGroup.getRegion(item.getRelatedGroup());
+		return region != null ? region : osmandRegions.getRegionDataByDownloadName(item.getBasename());
+	}
+
+	/**
+	 * @return localized name of the country the region belongs to. Continents and the world itself
+	 * have no country, for them {@code null} is returned.
+	 */
+	@Nullable
+	public static String getCountryName(@Nullable WorldRegion region) {
```
vshcherb: Duplicate code?   check inside World Region..

OsmAnd/src/net/osmand/plus/download/ui/DownloadSearchUIModel.java
```diff
@@ -0,0 +1,420 @@
...
+		}
+	}
+
+	/**
+	 * Splits the search request into a list of OR-ed conditions, each of them being
+	 * a list of AND-ed terms: "germany berlin, brandenburg" matches either an item
+	 * containing both "germany" and "berlin" or an item containing "brandenburg".
+	 */
+	@NonNull
+	public static List<List<String>> parseConditions(@NonNull String searchRequest) {
```
vshcherb, line 465: Nobody asked so we could drop this part as not requested and simplify code

OsmAnd/test/assets/download/search_examples.json
```diff
@@ -0,0 +1,148 @@
...
+        "Berlin | Germany",
+        "Brandenburg, Berlin | Germany",
+        "[Maps]",
+        "Berlin | New Hampshire",
+        "Berlin | Germany",
+        "Brandenburg, Berlin | Germany"
+      ]
+    },
+    {
+      "name": "Regions are hidden when the search is opened for a single download type",
```
vshcherb, line 155: Add tests like "London", "New york", "New", "San Fr" (San francisco)

OsmAnd/test/java/net/osmand/plus/download/DownloadSearchScreenReport.java
```diff
@@ -0,0 +1,109 @@
+package net.osmand.plus.download;
```
vshcherb, line 1: Class name very generic add names to be more specific that it's download UI search screen

vshcherb:
Overall strategy less code if possible but not reduce readability, and as less separte classes as possible but try to limit up to 1000-1200 loc

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
vshcherb: Renamed to `DownloadSearchUIModel` and moved to `net.osmand.plus.download.ui`.

I kept it as a separate class rather than folding it into `SearchDialogFragment`: it has no `View`/`Fragment` dependency (only `Context` for strings and `OsmandRegions`), and the instrumentation test drives it directly. It is down to 361 loc after the cleanups from the other comments — 984 together with the fragment, inside the limit. Happy to inline it if you prefer zero new classes, it is a mechanical move.

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
vshcherb: Reformatted to 120 columns, no line in the file is longer than that now.

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
vshcherb: Down to a single list — `processGroup()` fills `found`, and the deduplicated cities are inserted at its front:

```java
List<Object> found = new ArrayList<>();
processGroup(indexes, found, parseConditions(searchRequest));
found.addAll(0, filterDuplicateCities(indexes, cities, found));
return groupBySections(found);
```

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
vshcherb: Right, removed. `WorldRegion.getCountryRegion()` now resolves both shapes of the tree — a region under a continent, and a country that is placed next to the continents instead of under one:

```java
if (parent != null && (parent.isContinent() || WORLD.equals(parent.getRegionId()))) {
    return region;
}
```

`LocalItemUtils` and `OutdatedIndexesCollector` call the same method and now get a country where they used to get `null`. Covered by `OsmandRegionsTest.testGetCountryRegion` (JVM test, no emulator needed).

OsmAnd/src/net/osmand/plus/download/DownloadSearchHelper.java
vshcherb: Gone as well. What is left is a three line private wrapper that turns the resolved `WorldRegion` into its localized name for the second line of the row.

OsmAnd/src/net/osmand/plus/download/ui/DownloadSearchUIModel.java
vshcherb: This one is `parseConditions()` / `isMatch()`, moved verbatim from `SearchIndexFilter.performFiltering()` — it is what makes `berlin, munich` an OR and `new hampshire` an AND. Dropping it would change the existing search behaviour, so I kept it, and the new `New york` / `San Fr` examples cover it. I did shrink `isMatch()` from 18 lines to 6 plus a small helper.

If the comment was aimed at the country fallback a few lines above instead, that part is gone from this file — it moved into `WorldRegion.getCountryRegion()`.

OsmAnd/test/assets/download/search_examples.json
vshcherb: Added all four. `London` turned out to be the best illustration of the issue — on the real catalog the query returns six rows that are only told apart by the second line:

```
Regions
  Greater London          United Kingdom
  London                  Canada
Maps
  Derry/Londonderry       Northern Ireland
  New London              Connecticut
  Greater London          United Kingdom
  London                  Canada
```

`New york` also documents that a region whose map is built by joining its subregions has no map row of its own.

OsmAnd/test/java/net/osmand/plus/download/DownloadSearchScreenReport.java
vshcherb: Renamed to `DownloadSearchScreenReport`, and the test class to `DownloadSearchUIModelTest`.

alex-dev-neo:
One functional issue remains in city deduplication:

`SearchDialogFragment.searchCities()` creates every `CityItem` with `indexItem == null`. In `DownloadSearchUIModel.filterDuplicateCities()`, a city is resolved only when its point lies in one of the already matched map regions. If a query returns two same-named places covered by the same download map, but that map/region name does not itself match the query, both cities remain unresolved and both pass `item == null || shownMaps.add(item)`. The lazy resolver later points both rows to the same `IndexItem`, but the list is never deduplicated again, so two indistinguishable rows can still offer the same file.

The JSON case for “Two cities covered by the same map produce a single row” does not exercise the production path because `readCities()` pre-populates each city's `IndexItem`. Please add coverage with unresolved cities/real coordinates and either resolve/cache enough candidates to deduplicate by map, or deduplicate after lazy resolution.

OsmAnd/src/net/osmand/plus/download/ui/DownloadSearchUIModel.java
```diff
@@ -0,0 +1,361 @@
...
+		return result;
+	}
+
+	private static boolean isCoveredByAny(@NonNull List<WorldRegion> regions, @NonNull CityItem city) {
+		Amenity amenity = city.getAmenity();
+		LatLon location = amenity != null ? amenity.getLocation() : null;
+		if (location == null) {
+			return false;
+		}
+		for (WorldRegion region : regions) {
```
vshcherb: this is definitely bad code we should not iterate at all! queryBbox.. is doing search by quad tree which is much much faaster, also if you do it for multiple amenity... I'm not sure it's needed than you need to use HashSkipTileQuadTree and HashSkipTileQuadTreeJoiner to find NxM intersections of bboxes fast

OsmAnd/test/assets/download/search_examples.json
```diff
@@ -0,0 +1,136 @@
+{
+  "comment": [
+    "Examples for DownloadSearchUIModelTest: what the 'Maps & Resources' search results screen shows for a query.",
+    "'resources' is the catalog of downloadable files the search runs against, region names are resolved with the real regions.ocbf.",
+    "Every 'screen' entry is one row of the list: '[Header]' for a section header,",
+    "'Title' for a row without a second line and 'Title | Subtitle' for a row with the region shown in the second line.",
+    "Cities are not listed here - they are searched in the real World_basemap_mini.obf against the whole catalog,",
+    "see searchIsUnambiguousOnTheFullCatalog() and cityIsHiddenWhenItsMapIsAlreadyListed()."
+  ],
+  "resources": [
```
vshcherb: why do you need this section? what does it represent if you take all data from regions.ocbf, basemap_mini now?

OsmAnd/src/net/osmand/plus/download/ui/DownloadSearchUIModel.java
vshcherb: Measured it before deciding. That exact loop (`isCoveredByAny`, a linear scan over the matched regions per city) is gone — it was also functionally wrong, see the reply to @alex-dev-neo below.

What replaced it is a walk down the region tree that is already in memory. Numbers on this emulator, worst case query `"san"` — 167 cities, full catalog of 1083 maps:

| resolving 167 cities | |
|---|---|
| walk the in-memory region tree | **19 ms** |
| `OsmandRegions.getCountryName(LatLon)` | 800 ms |
| `OsmandRegions.getSmallestBinaryMapDataObjectAt(LatLon)` | 823 ms |

The quad tree in `regions.ocbf` is fast at the index level, but every call is file I/O plus re-parsing `BinaryMapDataObject`s in a `synchronized` method, so per point it costs ~4.8 ms against ~0.11 ms in memory. `getCountryName()` also only returns a country *name*: the second line of a city row is the covering **download map** (`Newburgh | Albany and surroundings`), and tapping the row downloads that file, so I need the `WorldRegion` → `IndexItem`, which that method does not give.

How much iteration is actually left, on the real data:

- 1722 regions, 1710 of them with boundaries, 1083 downloadable, tree depth 5, 9 continents
- the descent prunes to **154 point-in-polygon tests per city** — 9% of the tree — because it only enters a subregion whose polygon contains the point

A `HashSkipTileQuadTree` over the 1710 boxes would cut those 154 tests to a handful, saving ~15 ms of a search that spends **2233 ms** in the basemap POI search that produced the cities in the first place. That is under 1% of the screen's cost, against building and keeping an index of 1710 entries plus the joiner. My suggestion is not to add it here. If it is worth having, the right home is `OsmandRegions` itself, where every point → region lookup in the app would benefit, and then this screen just calls it.

One thing the walk did need: `europe_gb` has no boundary in `regions.ocbf` (bbox `[0,0 - 0,0]`), so a point in England was never reached. `WorldRegion.hasBoundaries()` now marks such grouping regions and the walk descends through them.

OsmAnd/test/assets/download/search_examples.json
vshcherb: Fair point — it was a leftover. Removed in 04a4113.

The `resources` list was the catalog the examples ran against, hand picked to keep the expected screens short. Now every example runs against the whole catalog, built from `regions.ocbf` the same way the other tests do, plus the real cities from `World_basemap_mini.obf`. The file is down to what it should be: a query and the screen it produces. Nothing in it is set up by hand any more.

It turned out to cost almost nothing in size, because the search is narrow — `germany` is 2 rows, `berlin` 7, `London` 8, `San Fr` 7. Only the deliberately short `New` is large at 38 rows, and that one is worth reading, it is the only example where cities and regions really mix.

The rules that hold for any query — one header per section, in order; no file offered twice; no two rows in a section reading the same — are now asserted on the examples too, not only on the extra ambiguous queries, so the exact screens are a golden file on top of real invariants rather than the only check.

The trade-off to be aware of: the expected screens now move when the shipped `regions.ocbf` or basemap is regenerated. That is intentional — it is how a change in the data becomes visible — and `download_search_report.html` renders all screens side by side so updating the file is a quick read, not a guess.

vshcherb:
@alex-dev-neo you were right on both counts, and the second one was the reason the first one survived. Fixed in 736b5bf and 04a4113.

**The defect.** Reproduced exactly as described, on the full catalog: `"washington"` produced two rows both reading just `Washington` — Washington PA and Washington DC, neither inside a region whose name matches the query, so neither was resolved and neither was deduplicated. `"berlin"` produced a second Berlin with no second line at all.

**The cause of the blind spot.** As you say, `readCities()` pre-populated each `IndexItem` from the JSON, so the test never ran the production path. That is now gone entirely — `readCities()` and the hand written `cities` blocks in the JSON are deleted, and the cities come from a real `World_basemap_mini.obf` search over the whole 1083-map catalog. `searchCities()` moved out of `SearchDialogFragment` into `DownloadSearchUIModel` so the tests can call the same code the app calls.

**The fix.** Rather than deduplicating after lazy resolution, every city is now resolved before the list is built, which removes the two-phase problem instead of patching it. That was only affordable once the lookup stopped reading the region file per city: it walks the region tree that is already in memory, 19 ms for the 167 cities of `"san"` against 823 ms for `getSmallestBinaryMapDataObjectAt`. `IndexItemResolverTask` is deleted with it, so the second line is filled in on first paint instead of being patched in asynchronously.

That also uncovered a second miss: a city inside a region with no map of its own — York in Great Britain — never resolved, because `europe_gb` carries no boundary in `regions.ocbf` and the lookup stopped there. `WorldRegion.hasBoundaries()` marks such grouping regions now.

**Coverage**, all with unresolved cities and real coordinates:

- `cityIsHiddenWhenItsMapIsAlreadyListed()` — searches the basemap for Berlin and asserts no file is offered twice and no two rows read the same
- `searchIsUnambiguousOnTheFullCatalog()` — the same rules over `berlin, hamburg, kyiv, holland, utrecht, york, washington, georgia`
- the examples in `search_examples.json` now run on the full catalog with real cities and are checked against those rules as well as against their exact screens

`"washington"` today reads `Washington | Pennsylvania`, `Washington | District of Columbia`, `Washington | United States`. Four tests, 3.3 s, no downloaded map needed.

vshcherb:
Good call — the examples recorded what the code already did, so they could not tell whether it did the right thing. I added two checks that do not use the implementation as their own oracle, and they found a real defect. Pushed as 7adf725.

**1. Cities written down from the map, not from a test run.** `citiesResolveToTheMapThatCoversThem()` is a table of a city, its coordinates and the country it is in — Berlin DE, Berlin NH, Hamburg, Washington DC, New York, San Francisco, Kyiv, Lviv, Utrecht, Amsterdam, York, Santo Domingo, Havana, Geneva. The country is the part I can state without running anything, so a wrong map cannot pass.

**2. A second implementation as the oracle.** `cityResolutionAgreesWithTheRegionFile()` takes every city three broad queries return (250+) and compares the map against the one `OsmandRegions.getSmallestBinaryMapDataObjectAt()` names — the quad tree in `regions.ocbf`, separate code over the same data.

**What it found.** 16 of 345 cities disagreed:

```
DISAGREE Santo Domingo   tree=centralamerica_carribean-archipelago-all  ocbf=centralamerica_dominican-republic
DISAGREE Santiago de Cuba tree=centralamerica_carribean-archipelago-all ocbf=centralamerica_cuba
DISAGREE Port-au-Prince  tree=centralamerica_carribean-archipelago-all  ocbf=centralamerica_haiti
DISAGREE Port of Spain   tree=centralamerica_carribean-archipelago-all  ocbf=centralamerica_trinidad-and-tobago
DISAGREE Montego Bay     tree=centralamerica_carribean-archipelago-all  ocbf=centralamerica_jamaica
...
```

The walk returned the **first** region whose boundary contains the point instead of the **smallest** one, so a search for Santo Domingo offered the whole Caribbean archipelago map rather than the Dominican Republic. It now applies the same rule as `getSmallestBinaryMapDataObjectAt()` — of all downloadable regions covering the point, the smallest wins.

Three more cities resolved to nothing: Sanhe in Hebei and the islands of Sansha, each in a detached part of a region that the in-memory boundary does not cover, since only the biggest polygon of a region is kept. The region file holds every part, so it is read for the few points the walk cannot place — a rare fallback, not the hot path.

**Result: 345 of 345 agree**, and all 14 hand written cities land in the right country. Row rendering for the table, for the record:

```
Berlin        | Germany                        Utrecht       | Netherlands
Berlin        | New Hampshire                  Amsterdam     | North Holland
Washington    | District of Columbia           York          | Yorkshire and the Humber
New York      | New York City and surroundings Santo Domingo | Dominican Republic
San Francisco | United States                  Havana        | Cuba
Kyiv          | Ukraine                        Geneva        | Switzerland
Lviv          | Lviv Oblast                    Hamburg       | Germany
```

Where the covering map carries the city's own name — Berlin, Kyiv, San Francisco, Geneva — the second line falls back to the country, which is the line the issue asked for.

Six tests now, 9.8 s; the cross check against the region file is ~6 s of that and is the price of having an independent oracle. None of the example screens changed, so the defect was outside what they covered — which is the point.

vshcherb:
Right — they had no business staying in Java as assertions on a country name. Each one is a query like any other, so they are written down as screens now. 2f6434e, 18 examples.

```
"washington"    →  [Regions]  Washington | United States
                   [Maps]     Washington | Pennsylvania
                              Washington | District of Columbia
                              Washington | United States

"san juan"      →  [Regions]  San Juan | Argentina
                   [Maps]     San Juan | Metro Manila
                              San Juan | Puerto Rico
                              San Juan | Venezuela
                              San Juan del Río | Querétaro
                              St. John's | Newfoundland and Labrador
                              San Juan | Argentina

"santo domingo" →  [Maps]     Santo Domingo | Dominican Republic
"sanhe"         →  [Maps]     三河市 | Hebei Province
"hamburg"       →  [Regions]  Hamburg | Germany
                   [Maps]     Novo Hamburgo | Rio Grande do Sul
                              Hamburg | Germany
"holland"       →  [Maps]     Holland | Michigan …
"georgia"       →  [Regions]  Georgia · Georgia | United States · South Georgia and South Sandwich Islands
```

Why these seven:

- **`washington`** and **`san juan`** are the issue itself at its worst — seven rows of one name, readable only because of the second line. `san juan` also shows a city found under another of its names, St. John's.
- **`santo domingo`** is the regression guard for the defect the cross check found: before the fix this row read `Caribbean archipelago`.
- **`sanhe`** is the other one — a city in a detached part of Hebei that the in-memory boundary does not cover, so it only resolves through the region file. If that fallback is ever dropped, this row loses its second line.
- **`holland`**, **`hamburg`**, **`georgia`** are the everyday confusions: a city carrying a country's name in the wrong hemisphere, a name that merely contains the query, and one name worn by a country, a state and an island group at once.

The broad queries stay in `searchIsUnambiguousOnTheFullCatalog()` where they earn their keep — `san`, `santa`, `port` put 150 to 200 rows on the screen, which is where a duplicate file or an unreadable pair would show up, and spelling those out as screens would be noise. They pass.

Two notes:

- A city is shown under the name the request was matched against, so the test now pins `MAP_PREFERRED_LOCALE` to its default. That is what makes `三河市` rather than `Sanhe` the expected row.
- The class is at 32 s, up from 9.8 s, almost all of it the basemap POI search per query. The searches shared between tests are cached; the rest is the cost of running every example against real data.

alexey-pelykh:
Post-merge review of #25781 (gen by AI). Nine independent reviewers, each in a fresh context, each
blind to the others and to this PR's review history; findings below are only those that survived a
redundancy filter against the 13 points already settled in the pre-merge round.

**Graded against**: issue #25727's two acceptance criteria (verbatim), repo `AGENTS.md` §§7/9/10/11,
and ISO/IEC 25010 + general review criteria where those are silent. Both ACs are implemented.
**Authority**: every individual finding landed on convention or defaults, so all of them are
**advisory**. The PR is merged; this is follow-up input, not a merge objection.

---

### 1. `findMapAt()` does not implement the rule its Javadoc claims — and the fallback that would catch it can't fire

`DownloadSearchUIModel.java:358-389`. The Javadoc says it "applies the rule of
`OsmandRegions#getSmallestBinaryMapDataObjectAt`". It ranks by **bounding-box area**
(`bbox.width() * bbox.height()`, `:386-389`); that method ranks by **shoelace polygon area**
(`OsmandRegions.java:368-380`). The two also draw from different candidate sets — `findMapAt` keeps
only regions with a `NORMAL_FILE`, the region file uses `isRegionMapDownload()`
(`OsmandRegions.java:971`).

The region-file fallback fires only on a null walk (`item != null ? item : readMapAt(...)`,
`:337-338`). But `findMapAt` prunes a subtree when the parent's single in-memory polygon misses the
point — so when a **coarser** downloadable region does contain it, the walk returns that coarser map
and the fallback is never consulted. A city in a detached part of a region can silently offer a
larger, wrong download.

This extends the limitation already noted in the PR (the ocbf fallback kept for Sanhe / Sansha): the
fallback does not fire in part of the case it exists to cover. Two reviewers reached it independently
on different criteria. **Neither could verify it against the binary `regions.ocbf`**, so no concrete
city is demonstrated and confidence is medium — the code paths are the evidence, not a repro.

### 2. `getCountryRegion()`'s widening reaches two screens the description says are unaffected

`WorldRegion.java:161-177`. `isContinent()` excludes exactly one direct child of `world`, so the
`null → russia` flip covers precisely the Russia subtree. Three consumers branch on that null:
`OutdatedIndexesCollector.java:153` (Updates tab now merges N outdated Russian maps into one row),
`LocalItemsCollector.java:111` (local files move into a country folder), `LocalItemUtils.java:296`
(grouped name drops the redundant `, Russia`; the sort key moves with it).

The description says "the region and updates screens render exactly as before" — accurate for the
view holders, not for the Updates screen.

**The behaviour looks right.** A reviewer traced all seven input shapes and all three consumers and
found each one makes Russia consistent with every other country; nobody could construct a regression.
Per-region updating also survives — `SelectIndexesHelper` routes the merged row to a
`MultipleSelectionBottomSheet` with each region individually selectable. So this is undeclared and
untested scope, not a defect. Worth a line in the description and a test at one of the three sites;
whether one merged "Russia — N" row is the wanted Updates behaviour is a product call.

### 3. The acceptance criterion's own behaviour has no test that can fail

Deleting `SearchDialogFragment:369`/`:373` — the country line AC 2 asked for — leaves the suite green:
assertions run through a test-only `render()` helper, and `getTitle` has no production caller.
Separately, the differential test is tautological wherever the walk misses, because the fallback calls
the same function used as the oracle; deleting `findMapAt` also leaves every test green.

This is the point already raised pre-merge about examples only recording what the code does — the fix
(14 hand-written city→country pairs) was applied to the **city** path, and the AC's own path never got
the equivalent anchor.

Two adjacent items: `assumeTrue("Test requires an English locale")` sits in `@Before`, so a
non-English device **silently skips the entire class**, including the one test with no locale
dependence — a green run and an empty run are indistinguishable. And `MAP_PREFERRED_LOCALE` is set
in `@Before` and never restored.

Worth noting the surrounding reality: of the three workflows in `.github/workflows/`, none runs a
test task — not even the JVM-only `:OsmAnd-java:test`. The added `OsmandRegionsTest` did not execute
on this PR.

### 4. Smaller, evidenced

| Where | What |
|---|---|
| `ItemViewHolder.java:272-295` | While a city map downloads, `tvDesc` goes `GONE` — "Berlin \| New Hampshire" and "Berlin \| Germany" (the PR's own fixture pair) become identical. The map name used to be in the title and survived this state. |
| `DownloadSearchUIModel.java:244-246` | Voice prompts, fonts and astronomy maps file under a header reading "Maps" on a screen called "Maps & Resources". Introduced by the AC 1 fix; AC 1 names two groups and is silent on rows that are neither. |
| `DownloadSearchUIModel.java:302-356` | City resolution moved from lazy and abandonable to eager and unabandonable — `cancelCitySearch()` interrupts the POI request but not the resolution loop. The 19 ms figure covers the fast path; the 823 ms fallback is per unresolved city, and `search_examples.json`'s own "sanhe" case shows those exist. Structural reading only — nothing was executed. |
| `DownloadSearchUIModel.java:297-322` | `filterDuplicateCities` also resolves maps and mutates the caller's `CityItem` list; the name and Javadoc describe only dropping. The new test works around both (copies each item, uses a sentinel query). |
| `DownloadSearchUIModel.java:248-283` | `getTitle` throws on `SectionHeader` — a row type the same class produces — while `getSubtitle` survives it only via a short-circuit. The row taxonomy is implemented three times, non-equivalently. |
| `SearchDialogFragment.java:228-237` | The new `isEnabled()` reports city rows actionable, but `onItemClick` has no `CityItem` branch, so only the trailing icon reacts. |
| `SearchDialogFragment.java:431` | `e.printStackTrace()` remains where the extracted twin was moved to `PlatformUtil.getLog` in both catch blocks. Pre-existing line. |
| `WorldRegion.java:161-164` | The new `isContinent()` early return is uncommented while the clause that makes it load-bearing is commented; removing it would silently make every continent its own country. |

Also: `AGENTS.md` §11 asks the AI disclaimer to name the tool and model and to state what the agent
decided on its own — the current section is missing the decided-on-its-own part and does not name a
model.

### Coverage, and one caveat about it

Nine axes: correctness, cross-module blast radius, performance/threading, test quality, project
conventions, UI rendering and accessibility, clarity, acceptance conformance, and an adversarial
pass. Reviewers filed 33 findings from roughly 60 candidates and killed the rest with quoted
counter-evidence — including the classic view-recycling bug in the new code, which is genuinely not
there.

**One reviewer's blinding leaked.** The acceptance reviewer was pointed at a file that did not exist
yet, went looking in the directory it was given, and read a context file carrying this PR's merge
status, reviewer names and the pre-merge resolution list. It re-derived its findings from code and
cites only code, but it noted it cannot claim it would have looked there unprompted — so items 2 and
the AC-conformance points should be weighted accordingly. The other eight attested clean blinding.
That was a dispatch error on the reviewing side, not anything about this PR.

Nothing here was executed — no build, no test run, no device, per `AGENTS.md` §9. Every performance
and reachability claim is a reading of the code, not a measurement.

PR #25788, opened 2026-08-31
Feature/25585 - Include navigation settings in Route link

### LINKED ISSUE #25585: Include navigation settings in Route link
## ToDo
OsmAnd allows sharing a route link that includes coordinates for the start and destination and the navigation type but excludes selected route parameters.  Web and iOS versions support this feature.

iOS task https://github.com/osmandapp/OsmAnd-Issues/issues/1860

- [ ] Include the selected navigation settings in the URL

Kreiri:
[#25585] added routing navigation settings to route link generation and parsing (same as in iOS app); apply these settings when opening route from link


OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
```diff
@@ -161,12 +165,12 @@ private boolean parseNavigationIntent() {
 		Intent intent = mapActivity.getIntent();
 		if (intent != null && isUriHierarchical(intent)) {
 			Uri data = intent.getData();
-			boolean hasNavigationDestination = data.getQueryParameterNames().contains(URL_PARAMETER_END);
-			if (isOsmAndMapUrl(data) && hasNavigationDestination) {
+			if (isOsmAndMapNavigationUrl(data, true)) {
```
Corwin-Kh: used only once and second param is always true. Do we need dedicated function?

OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
```diff
@@ -161,12 +165,12 @@ private boolean parseNavigationIntent() {
...
 				String startLatLonParam = data.getQueryParameter(URL_PARAMETER_START);
 				String endLatLonParam = data.getQueryParameter(URL_PARAMETER_END);
 				String appModeKeyParam = data.getQueryParameter(URL_PARAMETER_MODE);
 				String intermediatePointsParam = data.getQueryParameter(URL_PARAMETER_INTERMEDIATE_POINTS);
+				final String routingParam = data.getQueryParameter(URL_PARAMETER_ROUTING_PARAMS);
```
Corwin-Kh: routingParam better make plural, as parameter is plural. would be clearer understand what can be inside

OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
```diff
@@ -271,7 +275,8 @@ private void registerDevice(@NonNull String token) {
 	}
 
 	private void buildRoute(@Nullable LatLon start, @NonNull LatLon end,
-	                        @Nullable ApplicationMode appMode, @Nullable List<LatLon> points) {
+	                        @Nullable ApplicationMode appMode, @Nullable List<LatLon> points,
+							@Nullable String routingSettingsQueryParam) {
```
Corwin-Kh: also plural

OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
```diff
@@ -285,11 +290,35 @@ private void buildRoute(@Nullable LatLon start, @NonNull LatLon end,
 						null, settings.getIntermediatePoints().size());
 			}
 		}
+		if (routingSettingsQueryParam != null) {
```
Corwin-Kh: for strings better use Algorithms.isEmpty

OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
Kreiri: Previous implementation in android app set path of route url to just "/map". Web and iOS apps set this path to "/map/navigate". On one hand, I want to make links same as on the other platforms; on the other hand, I don't want to break handling of old links generated in android app *yet*; on the third hand, I didn't want to leave "check that path is /map/navigate" to be implemented some time in the future. I chose one method with parameter that can be flipped at any time the decision is made to abandon or re-enable support of old "/map" links.

Corwin-Kh: if  we decide to remove legacy support we'll need to remove the code anyway. so this approach is overthinking. if the param is always true, it is not needed

OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
```diff
@@ -511,6 +546,26 @@ private boolean isOsmAndMapUrl(@NonNull Uri uri) {
 		return isOsmAndSite(uri) && isPathPrefix(uri, "/map");
 	}
 
+	/**
```
Corwin-Kh: better to write code that does not need comments. it should be clear to read and understand without it. Exceptions - thin places that are not obvious to be in that form and need the clarification, why was it written so

OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
```diff
@@ -511,6 +546,26 @@ private boolean isOsmAndMapUrl(@NonNull Uri uri) {
...
+	 * @param acceptLegacyPath if true, also accept urls with /map path
+	 * @return true if url can be treated as a "navigation" url, false otherwise
+	 */
+	@SuppressWarnings("SameParameterValue")
+	private boolean isOsmAndMapNavigationUrl(@NonNull final Uri data, final boolean acceptLegacyPath) {
+		if (!isOsmAndSite(data)) return false;
+		if (acceptLegacyPath){
+			if (!hasPathPrefix(data, URL_PATH)) return false;
+		} else {
+			if (!hasExactPath(data, URL_PATH, URL_PATH_COMPONENT_NAVIGATE)) return false;
```
Corwin-Kh: if acceptLegacy is always true, it looks like app never enters this branch. then why we added URL_PATH_COMPONENT_NAVIGATE

OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
```diff
@@ -285,11 +290,35 @@ private void buildRoute(@Nullable LatLon start, @NonNull LatLon end,
...
 				null, hasIntermediatePoints, true, MapRouteInfoMenu.DEFAULT_MENU_STATE);
 	}
 
+	private void buildRoute(@Nullable LatLon start, @NonNull LatLon end,
+	                        @Nullable ApplicationMode appMode, @Nullable List<LatLon> points) {
+		buildRoute(start, end, appMode, points, null);
+	}
+
+    private void applyRoutingParams(@Nullable final ApplicationMode appMode, @NonNull String routingSettingsQueryParams) {
+		if (appMode == null) return;
```
Corwin-Kh: return on the same line as if statement is visible hidden. better to place it is {} and on separate line

OsmAnd/src/net/osmand/plus/helpers/RoutingUriQueryHandler.kt
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Contains methods for composing, parsing and applying "routing/navigation params" query param value
+ */
+object RoutingUriQueryHandler {
+
+    /**
+     * Utility class for holding key-value pairs parsed from query params
+     */
+    data class KeyValue(val key: String, val value: String)
```
Corwin-Kh: it looks like can be replaced with Pair<String, String>

OsmAnd/src/net/osmand/plus/helpers/IntentHelper.java
```diff
@@ -903,46 +958,83 @@ private boolean isPathPrefix(@NonNull Uri uri, @NonNull String pathPrefix) {
...
+		final String paramsValue = getRoutingParamsQueryValue(app, appMode);
+		if (paramsValue != null) {
+			builder.appendQueryParameter(URL_PARAMETER_ROUTING_PARAMS, paramsValue);
+		}
+
+		final OsmandMapTileView mapTileView = app.getOsmandMap().getMapView();
+		final int zoom = mapTileView.getZoom();
+		final double lat = mapTileView.getLatitude();
+		final double lon = mapTileView.getLongitude();
+		final String fragment = zoom + "/" + getFormattedCoordinate(lat) + "/" + getFormattedCoordinate(lon);
```
Chumva: Why use final here & in other places for local variables?

Kreiri: `final` makes the variable non-reassignable, clearly declaring the intention and preventing accidental reassignments. While is not required in Java, I consider it a good practice to explicitly make locals that aren't reassigned immutable.

OsmAnd/src/net/osmand/plus/helpers/RoutingUriQueryHandler.kt
Kreiri: With Pair<String, String>, purpose of stored values is not clear. With a simple utility `data class` like this, names are explicit: this is the key, and this is the value.

PR #25792, opened 2026-08-31
Fix NPE in AndroidComposeViewAccessibilityDelegateCompat.onViewDetachedFromWindow

### LINKED ISSUE #25667: NPE in AndroidComposeViewAccessibilityDelegateCompat.onViewDetachedFromWindow
Search related, not sure how to reproduce:  [exception.log.txt](https://github.com/user-attachments/files/31110015/exception.log.txt)

vshcherb:
Fixes #25667. Follow-up to #25731, which did not stop the crash.

The `AGENTS.md` convention this PR's disclaimer follows is proposed separately in #25793.

## Root cause

Upstream bug in Compose UI 1.11.3 (`compose-bom:2026.06.00`), `AndroidComposeViewAccessibilityDelegateCompat.android.kt`:

```kotlin
/**
 * Handler returns non-null ONLY when [view] is attached.
 * ... Null means that we are not attached and don't need to process.
 */
private val handler: Handler?
    get() = if (AndroidComposeUiFlags.isViewBasedSemanticsHandlerEnabled) view.handler
            else legacyMainHandler

override fun onViewDetachedFromWindow(view: View) {
    handler!!.removeCallbacks(semanticsChangeChecker)   // line 431 - the crash
```

The doc comment says `null` is expected; the code dereferences it with `!!`. The flag is new in 1.11 and defaults to `true` (`// TODO remove me b/486998514`).

The view that hits this is a `ComposeView` inside a **`ListView` item** — `SearchScopeChip` in `search_list_item.xml`, `GallerySortBarView` in `gallery_sort_bar_item.xml`. `AbsListView.RecycleBin` detaches a recycled item from the window (`removeDetachedView`) and later puts it back with `attachViewToParent()`, which does **not** re-dispatch `onAttachedToWindow()`. The next `ListView.resetList() -> ViewGroup.removeAllViewsInLayout()` detaches that item a second time, `View.getHandler()` is already `null`, and every `ComposeView` in the subtree throws.

That is exactly the reported stack: `removeAllViewsInLayout` -> nested `dispatchDetachedFromWindow` -> `onViewDetachedFromWindow`.

## Why #25731 did not help

Two independent reasons, both verified on a device:

1. **Wrong class.** `ChipsLayout` lives in `button_toolbar_layout` of `search_dialog_fragment.xml`, not inside a `ListView`. It is never on the `AbsListView.resetList()` path from the crash report.
2. **Too late, and not enough.** `ViewGroup.dispatchDetachedFromWindow()` detaches all children *before* calling `super.dispatchDetachedFromWindow()`, which is what triggers the parent's `onDetachedFromWindow()`. By the time the override runs, the inner `AndroidComposeView` has already thrown. And `disposeComposition()` removes neither that child view nor the delegate's `OnAttachStateChangeListener`, which `AndroidComposeViewAccessibilityDelegateCompat` registers in its `init` and never unregisters.

`SearchScopeChip` — the view actually on the crashing path — has had the same `onDetachedFromWindow { disposeComposition() }` override since #25615, and the crash kept arriving on 5403.

## The fix

Turn off `AndroidComposeUiFlags.isViewBasedSemanticsHandlerEnabled` in `OsmandApplication.onCreate()`. This is the escape hatch `AndroidComposeUiFlags` documents for exactly this situation, and it makes `handler` fall back to `legacyMainHandler`, which is never null. The flag only exists to support Compose on a non-main thread, which OsmAnd never does, so the fallback is safe.

It also covers every `ComposeView` in the app at once rather than one class at a time.

## Reproduction and test

The crash needs a specific `RecycleBin` state, which is why manual search scenarios never hit it. `ComposeChipDetachFromWindowTest` replays the `RecycleBin` sequence (`detachViewFromParent` -> `removeDetachedView` -> `attachViewToParent` -> `removeAllViewsInLayout`) on the real `search_list_item` layout:

* `searchListItemSurvivesDetachOfRecycledItem` — the regression test.
* `composeStillCrashesWithoutTheWorkaround` — flips the flag back on and asserts the NPE still happens. It keeps the first test from going vacuous and fails loudly once Compose fixes b/486998514, at which point `applyComposeWorkarounds()` and this test can both be deleted.

Both pass on an API 35 emulator. Before this change, a standalone repro app produced a byte-identical stack trace, including with #25731's `disposeComposition()` override in place.

## Known adjacent issue (not fixed here)

If an `AbstractComposeView` is measured while not attached to a window — e.g. a `ListView` with `wrap_content` height going through `measureHeightOfChildren()` — Compose throws `IllegalStateException: Cannot locate windowRecomposer` from `onMeasure()`. A different crash cluster; none of the current chip-hosting lists use `wrap_content`, so it does not fire today. Worth keeping in mind before putting a `ComposeView` into a self-measuring list.

## AI disclaimer

Produced with Claude Code (Opus 5), driven by the prompts below.

Prompts used (summarised):
1. "Try to reproduce and fix" plus a link to the Play Console crash cluster for this NPE (versionCode 5403). No hypothesis, no pointer to a file - the agent read the stack trace from the console, located the cause in the Compose sources and picked the fix.
2. "Create a branch, commit, open a pull request, link it in issue #25667 and comment there on my behalf (gen by AI); create a UI test that reproduces the problem and verifies the fix; analyse the discussion and why the fix in #25731 did not work, the essence of the problem and the next steps."
3. "Add an AI disclaimer to the PR summarising which prompts were used" - and record that as a standing project rule.

- The choice of fix (the `AndroidComposeUiFlags` escape hatch in `OsmandApplication.onCreate()`) over patching individual `ComposeView` subclasses.
- The shape of the test, including the `composeStillCrashesWithoutTheWorkaround` canary.

Verification performed: `:OsmAnd:connectedNightlyFreeOpenglArm64DebugAndroidTest` for the new test class on an API 35 arm64 emulator (2 tests, 0 failures), plus a standalone repro app that produced a byte-identical stack trace before the fix. Note that this required running Gradle, which `AGENTS.md` section 9 forbids agents from doing on their own - it was done deliberately to satisfy the "reproduce, fix and test it" request, and is called out here rather than hidden.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/OsmandApplication.java
+++ b/OsmAnd/src/net/osmand/plus/OsmandApplication.java
@@ -24,6 +24,7 @@ import android.widget.Toast;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
+import androidx.compose.ui.AndroidComposeUiFlags;
 import androidx.core.app.ActivityCompat.OnRequestPermissionsResultCallback;
 import androidx.lifecycle.DefaultLifecycleObserver;
 import androidx.lifecycle.LifecycleObserver;
@@ -246,6 +247,7 @@ public class OsmandApplication extends MultiDexApplication {
 		}
 		long timeToStart = System.currentTimeMillis();
 		enableStrictMode();
+		applyComposeWorkarounds();
 		super.onCreate();
 
 		LifecycleObserver appLifecycleObserver = new DefaultLifecycleObserver() {
@@ -304,6 +306,26 @@ public class OsmandApplication extends MultiDexApplication {
 
 		SearchUICore.setDebugMode(PluginsHelper.isDevelopment());
 		BackupHelper.DEBUG = PluginsHelper.isDevelopment();
+	}
+
+	/**
+	 * Turns off the "view based semantics handler" Compose feature flag.
+	 * <p>
+	 * With the flag on, AndroidComposeViewAccessibilityDelegateCompat.onViewDetachedFromWindow()
+	 * dereferences View.getHandler() with {@code !!}, although that handler is null whenever the
+	 * view is not attached to a window. AbsListView leaves recycled item views exactly in that
+	 * state: RecycleBin detaches them from the window and later puts them back with
+	 * attachViewToParent(), which does not re-dispatch onAttachedToWindow(). The next
+	 * ListView.resetList() -> removeAllViewsInLayout() then detaches such a view a second time and
+	 * every ComposeView inside a list item (search results, gallery sort bar, chips) crashes with
+	 * a NullPointerException.
+	 * <p>
+	 * The flag only exists to support Compose on a non-main thread, which OsmAnd never does, so
+	 * falling back to the main looper handler is safe. To be removed once the upstream bug
+	 * (b/486998514) is fixed.
+	 */
+	private void applyComposeWorkarounds() {
+		AndroidComposeUiFlags.isViewBasedSemanticsHandlerEnabled = false;
 	}
 
 	public boolean isPlusVersionInApp() {
--- a/OsmAnd/test/java/net/osmand/test/ui/search/ComposeChipDetachFromWindowTest.kt
+++ b/OsmAnd/test/java/net/osmand/test/ui/search/ComposeChipDetachFromWindowTest.kt
@@ -1,0 +1,116 @@
+package net.osmand.test.ui.search
+
+import android.content.Context
+import android.view.View
+import android.view.ViewGroup
+import android.widget.FrameLayout
+import androidx.compose.ui.AndroidComposeUiFlags
+import androidx.compose.ui.ExperimentalComposeUiApi
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.filters.LargeTest
+import net.osmand.plus.R
+import net.osmand.plus.activities.MapActivity
+import net.osmand.plus.search.dialogs.SearchScopeChip
+import net.osmand.test.common.AndroidTest
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * Regression test for https://github.com/osmandapp/OsmAnd/issues/25667: a NullPointerException in
+ * `AndroidComposeViewAccessibilityDelegateCompat.onViewDetachedFromWindow()` raised while a
+ * `ListView` resets its items.
+ *
+ * Compose 1.11 dereferences `View.getHandler()` with `!!` there, although that handler is null for
+ * a view that is not attached to a window. `AbsListView.RecycleBin` produces exactly such a view:
+ * it detaches a recycled item from the window and later puts it back into the list with
+ * `attachViewToParent()`, which does not re-dispatch `onAttachedToWindow()`. The following
+ * `ListView.resetList() -> ViewGroup.removeAllViewsInLayout()` detaches that item a second time
+ * and every `ComposeView` inside it crashes.
+ *
+ * [recycleThroughScrapHeap] replays that `RecycleBin` sequence on the real `search_list_item`
+ * layout, which hosts a [SearchScopeChip].
+ */
+@LargeTest
+@RunWith(AndroidJUnit4::class)
+class ComposeChipDetachFromWindowTest : AndroidTest() {
+
+	@get:Rule
+	val scenarioRule = ActivityScenarioRule(MapActivity::class.java)
+
+	/** The workaround applied in [net.osmand.plus.OsmandApplication] must keep the list alive. */
+	@Test
+	fun searchListItemSurvivesDetachOfRecycledItem() {
+		val error = runRecycledItemDetachScenario()
+		assertNull("Detaching a recycled search list item crashed: $error", error)
+	}
+
+	/**
+	 * Guards the test above from silently becoming vacuous, and doubles as a reminder: as soon as
+	 * this one starts failing, the upstream bug is fixed and
+	 * `OsmandApplication.applyComposeWorkarounds()` can be dropped together with this test.
+	 */
+	@OptIn(ExperimentalComposeUiApi::class)
+	@Test
+	fun composeStillCrashesWithoutTheWorkaround() {
+		val enabled = AndroidComposeUiFlags.isViewBasedSemanticsHandlerEnabled
+		AndroidComposeUiFlags.isViewBasedSemanticsHandlerEnabled = true
+		val error = try {
+			runRecycledItemDetachScenario()
+		} finally {
+			AndroidComposeUiFlags.isViewBasedSemanticsHandlerEnabled = enabled
+		}
+		assertTrue(
+			"Expected the upstream Compose NPE, got $error. If Compose has fixed b/486998514, "
+					+ "remove OsmandApplication.applyComposeWorkarounds() and this test.",
+			error is NullPointerException
+		)
+	}
+
+	private fun runRecycledItemDetachScenario(): Throwable? {
+		val caught = AtomicReference<Throwable?>()
+		scenarioRule.scenario.onActivity { activity ->
+			val content = activity.findViewById<ViewGroup>(android.R.id.content)
+			val container = ScrapContainer(activity)
+			content.addView(container, ViewGroup.LayoutParams(
+				ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT))
+			try {
+				val item = activity.layoutInflater.inflate(R.layout.search_list_item, container, false)
+				item.findViewById<SearchScopeChip>(R.id.search_scope_chip)
+					.setScopeName("Test scope", false)
+				// Composes as soon as the item is attached to the window.
+				container.addView(item)
+
+				container.recycleThroughScrapHeap(item)
+				container.resetList()
+			} catch (error: Throwable) {
+				caught.set(error)
+			} finally {
+				content.removeView(container)
+			}
+		}
+		return caught.get()
+	}
+
+	/** Replays the parts of `AbsListView` / `ViewGroup` that lead to the crash. */
+	private class ScrapContainer(context: Context) : FrameLayout(context) {
+
+		/**
+		 * `RecycleBin` detaches an item from the window and later reattaches it to the list with
+		 * `attachViewToParent()`, which leaves the whole subtree without an `AttachInfo`.
+		 */
+		fun recycleThroughScrapHeap(child: View) {
+			val params = child.layoutParams
+			detachViewFromParent(child)
+			removeDetachedView(child, false)
+			attachViewToParent(child, 0, params)
+		}
+
+		/** `AbsListView.resetList()`. */
+		fun resetList() = removeAllViewsInLayout()
+	}
+}


alexey-pelykh:
## Post-merge review *(gen by AI)*

**Graded against**, in precedence order: issue #25667 (the crash this must stop) → `AGENTS.md` §§7/9/10/11 → upstream Compose UI 1.11.3 sources and Android framework contracts → general practice where no project standard exists. **Advisory only** — the PR is merged, and the last two rungs are convention rather than declared acceptance criteria. Nothing here is a blocking claim.

**The fix is correct.** The diagnosis was checked against the crash log on #25667 and the actual 1.11.3 sources, not taken on faith; the flag is upstream's own documented escape hatch, used exactly as its KDoc prescribes; and bumping Compose was genuinely not an option. The findings below are about a removal trigger that cannot fire, one unnamed cost, and three documentation claims that do not hold.

Full working notes available on request — this is the condensed form.

---

### 🔴 1. The removal trigger cannot fire, for three independent reasons

`applyComposeWorkarounds()` and `composeStillCrashesWithoutTheWorkaround` are both documented as self-retiring — *"as soon as this one starts failing, the upstream bug is fixed"*. That mechanism is inoperative:

1. **Nothing runs the test.** `build-debug-apk.yml` is `on: [workflow_dispatch]` with a single `assembleNightlyFreeLegacyFatDebug` step; the other two workflows are wrapper validation and dependency review. No workflow runs unit or instrumented tests. On this PR only *Validate Gradle Wrapper* and *Dependency Review* ran — the new `AndroidComposeUiFlags` import was never compiled by CI.
2. **The cited bug ID is the wrong signal.** `b/486998514` is the flag-*deletion* tracker — `// TODO remove me b/486998514` sits on the flag itself and on `legacyMainHandler` (`AndroidComposeViewAccessibilityDelegateCompat.android.kt:312`). The NPE was fixed under **`b/486235925`**, commit `4fed5aace9a3ff3214660f0fa7a9e27bdcb52b48` (2026-07-06): *"use safe-call (`handler?`) instead of force-unwrap (`handler!!`) in onViewDetachedFromWindow"*. Someone watching `b/486998514` sees nothing while the NPE is already fixed; and when `b/486998514` does land, the flag is deleted and `OsmandApplication.java:328` becomes a **compile error**, not a failing test.
3. **Nothing tracks the removal.** #25667 is closed and no open issue references the workaround.

**Correct removal criterion**, since the current one points elsewhere: the fix first ships in **1.13.0-alpha01**. 1.11.3 and 1.11.4 sources are byte-identical (MD5 `8d2a809c…`) and 1.12.0 still has `handler!!` at line 429 — no bump before 1.13 helps.

**Cheapest fix that actually works:** put the reminder where the person who will act on it is already looking — `TODO(#25667)` on the BOM line at `build-common.gradle:394`, mirrored on `applyComposeWorkarounds()` and the canary, plus one tracked issue. A scheduled emulator job is stronger but a much heavier ask, and an ignored red is worse than no job.

---

### 🟡 2. Turning the flag off has an app-wide cost the change does not name

The javadoc says falling back to `legacyMainHandler` "is safe". Safe *from the NPE* — not free. Both semantics-scheduling sites guard on `val localHandler = handler; if (… localHandler != null)` (`AndroidComposeViewAccessibilityDelegateCompat.android.kt:2347-2363, 2401-2405`; `AndroidContentCaptureManager.android.kt:99-115`). Flag **on** → `handler` is `view.handler`, null while detached, guard trips, nothing posted. Flag **off** → `legacyMainHandler` is a plain `Handler(Looper.getMainLooper())` that is never null, so the guard never trips and `semanticsChangeChecker` (`view.measureAndLayout()` then `checkForSemanticsChanges()`) is scheduled for **detached** views during scroll and recycle churn.

Upstream added exactly that attach-check in commit `10b903388847c67afc091b8c735760de614eda7e` (2026-07-09) — *"prevents redundant execution and temporary memory leaks when the runnables run after the view has been detached"* — first shipping in **1.13.0-alpha01**. On 1.11.3 the app takes the cost unmitigated. Magnitude is unmeasured; the point is that it is real and unstated.

---

### 🟡 3. The `windowRecomposer` clearance tests the declared height, but measure mode comes from the ancestors

The "known adjacent issue" section clears the risk with *"none of the current chip-hosting lists use `wrap_content`"*. A `ListView`'s measure mode is set by its parents. `search_explore_fragment_layout.xml` (returned by `QuickSearchHistoryListFragment.java:78`) is:

```
FrameLayout    android:layout_height="wrap_content"
└─ LinearLayout   android:layout_height="wrap_content"
   └─ ListView    android:layout_height="match_parent"   ← declared match_parent
```

`getChildMeasureSpec(EXACTLY, WRAP_CONTENT)` → `AT_MOST`, then `getChildMeasureSpec(AT_MOST, MATCH_PARENT)` → `AT_MOST`. So `onMeasure` takes the `measureHeightOfChildren()` path and measures adapter views **never attached to a window**; `AbstractComposeView.onMeasure` is `final` and unconditionally reaches `windowRecomposer` → `check(isAttachedToWindow)`. That adapter is `QuickSearchListAdapter`, which inflates `search_list_item`.

What prevents the crash today is not the stated reason: the chip is `android:visibility="gone"` (`search_list_item.xml:133`) and `LinearLayout` skips `GONE` children. It is made visible only by `SearchResultViewHolder.bindSpatialCategoryPart()`, gated on `spatialSearchResult.isPoiCategory()` with a non-null reference object.

**Latent and pre-existing — not introduced here.** Whether a spatial POI-category result can reach the Explore/History list is a runtime question we did not settle. Worth its own issue; the PR-level point is only that the reasoning given does not clear it.

---

### 🟡 4. The canary asserts the NPE's *type*, not its identity

`assertTrue(…, error is NullPointerException)` is satisfied by any NPE in the scenario. Line 83 is a Kotlin platform-type dereference — `findViewById` returns `SearchScopeChip!` — so if `search_scope_chip` is renamed or moved out of `search_list_item.xml`, that NPE fires *before the recycle sequence runs*. The canary goes green, vouching that the workaround is still needed, while exercising none of the machinery it exists to exercise — the exact vacuity it was added to prevent, one level up.

The message compounds it: *"If Compose has fixed b/486998514, remove …applyComposeWorkarounds() and this test."* The failure this canary is uniquely good at catching (finding 6) is the one its message misdiagnoses, and the prescribed action deletes a still-load-bearing workaround.

Debug builds are unminified (`minifyEnabled true` only under `release`), so frames survive:

```kotlin
val npe = error as? NullPointerException
assertNotNull("Expected the upstream Compose NPE, got $error", npe)
assertTrue(
    "NPE did not come from the Compose accessibility delegate — the reproducer may have stopped " +
        "reproducing. Do NOT remove the workaround on this failure alone.\n" + npe!!.stackTraceToString(),
    npe.stackTrace.any {
        it.className.startsWith("androidx.compose.ui.platform.AndroidComposeViewAccessibilityDelegateCompat")
    })
```

In fairness: a fully silent double-green is unreachable — the sibling requires `error == null` with the flag `false`. The cost is misdiagnosis, not blindness.

---

### 🟡 5. Experimental API used unmarked from Java, with the one check that would catch it disarmed

`AndroidComposeUiFlags` is `@ExperimentalComposeUiApi` (`@RequiresOptIn(level = ERROR)`). Kotlin's compiler check does not reach Java, so `OsmandApplication.java:328` is a silent dependency on an experimental API — the asymmetry is visible inside this PR, where the Kotlin test writes `@OptIn(ExperimentalComposeUiApi::class)` and the Java line carries nothing. Lint's `UnsafeOptInUsage` *does* cover Java callers, but the repo neutralises it: `build-common.gradle:20` `abortOnError false`, `build.gradle:63` `warningsAsErrors false`, and `lint.xml` does not raise it.

That this surface churns is demonstrable: in **1.10.6 the class is `androidx.compose.ui.ComposeUiFlags`** — `AndroidComposeUiFlags` does not exist, the Android flags were split out in 1.11 — and 4 of 1.10.6's 21 flags already carry `@Deprecated("This flag is no longer needed.")`.

Suggested: `@SuppressLint("UnsafeOptInUsage")` plus a one-line "experimental, may vanish on BOM bump" note. Moving the method to Kotlin would carry `@OptIn` naturally but cuts against §10's "preserve existing file language" — hence the annotation rather than a migration.

---

### 🟡 6. The load-bearing step of the replay is missing from the helper's own documentation

The class doc narrates *detach → `attachViewToParent` → `resetList`* and never mentions `removeDetachedView` (line 109) — the step that actually nulls `mAttachInfo`:

| line | call | effect |
|---|---|---|
| 108 | `detachViewFromParent` | pointer surgery only; does **not** dispatch detach |
| 109 | `removeDetachedView(child, false)` | **dispatches** `dispatchDetachedFromWindow()` → nulls `mAttachInfo` |
| 110 | `attachViewToParent` | restores `mParent`, does **not** re-dispatch attach → `mAttachInfo` stays null |
| 114 | `removeAllViewsInLayout` | container still attached → detach dispatched on a child with null `mAttachInfo` → `getHandler()` null → `!!` → NPE |

A future reader tidying the helper to match its own doc deletes line 109; the sequence stops reproducing, both tests pass with the flag `false`, and the canary goes red printing "remove the workaround". Assert the precondition, not just the absence of a crash:

```kotlin
container.recycleThroughScrapHeap(item)
assertSame(container, item.parent)
assertNull("recycleThroughScrapHeap no longer produces a detached-but-parented view", chip.handler)
container.resetList()
```

Stated as **inferred, not verified**: that this four-call sequence is the one `AbsListView.RecycleBin` actually runs. It provably reaches the bug's precondition; the evidence that the production path is real is the stack trace on #25667, not this helper.

---

### 🟡 7. The canary leaks a live composition into the shared instrumentation process

Flag restoration is safe — the `catch (Throwable)` is *inside* the `onActivity` lambda, so the lambda returns normally and the `finally` always runs. The leak is elsewhere: `ViewGroup.dispatchDetachedFromWindow()` recurses into children *before* the parent's own `onDetachedFromWindow()`, so the NPE escapes `removeAllViewsInLayout()` before `SearchScopeChip.onDetachedFromWindow()` (`SearchScopeChip.kt:55`) runs — `disposeComposition()` never executes, and the already-zeroed child count means `finally { content.removeView(container) }` drops the item with its recomposer still registered. All instrumented tests share one process and there is no `@FixMethodOrder` anywhere in `OsmAnd/test/java`, so the canary may run first and a later Compose-touching test can fail for reasons unrelated to itself. Forcing `chip?.disposeComposition()` in the `finally`, and moving the `addView` at :79 inside the `try`, closes both.

---

### 🟢 Smaller items

- **The gallery is not on the described path.** The javadoc, PR body and commit message all attribute the crash to `AbsListView.RecycleBin` reaching "search results, gallery sort bar, chips". `GallerySortBarView` is hosted by `GalleryGridAdapter : RecyclerView.Adapter<…>` (`GalleryGridAdapter.kt:46,146`) — no `RecycleBin`, no `resetList()`; `ChipsLayout` sits directly in `search_dialog_fragment.xml` / `quick_search_history_fragment.xml`, not in list items. The only `AbstractComposeView` reachable by the named mechanism is **`SearchScopeChip`, one widget in one layout**. That does not make the global fix wrong — it does weaken "covers every ComposeView at once" as the justification, and it misleads the next blast-radius audit.
- **Second consumer unmentioned.** `AndroidContentCaptureManager` reads the same flag and has the identical `handler!!` in its own `onViewDetachedFromWindow`. Correctly fixed by the workaround; the javadoc names only the accessibility delegate.
- **Superseded residue.** `SearchScopeChip.kt:56` and `ChipsLayout.kt:177` still carry `onDetachedFromWindow { disposeComposition() }` — the latter from #25731, which this PR's analysis supersedes. `GallerySortBarView` has none. Two of three, unannotated: a future reader cannot tell which are load-bearing.
- **No accessibility verification.** The flag gates nothing but Handler selection, so a functional TalkBack regression is unlikely — but OsmAnd ships `AccessibilityAssistant`, wired into `QuickSearchListFragment`, `QuickSearchListAdapter` and `FavoritesSearchFragment`, i.e. exactly these lists. With finding 2, a TalkBack smoke pass over search and the gallery is cheap insurance.
- **Narrower alternatives unweighed.** `SearchScopeChip.Content()` is a static pill — one `Icon` and one `Text` in a `Row`. A plain `TextView` removes the only `AbstractComposeView` from any `AbsListView` item and retires the whole bug class, finding 3 included. Migrating those lists to `RecyclerView` is the other. For the record, the regression entered with the BOM jump `2024.04.01 → 2026.06.00` in `3d9613ada7` (2026-06-26). Neither is obviously better than the flag flip — each is cheaper to reverse, and neither was named.
- **`AGENTS.md` §11 is ambiguous about the one surface that is permanent.** Its sibling bullets enumerate surfaces in parallel: *English only* names *"commit messages, PR titles and bodies, and issue comments"*; *No tool advertising*, one line below, names only *"PR bodies or issue comments"*. On that reading the `Claude-Session:` trailer in `b09af0bf` is **not** forbidden — but the heading says "Rules for the whole PR" and the bullet ends "the only provenance note needed", which pulls the other way. Under merge-commit strategy that trailer is permanent, so the rule is worth disambiguating before the next such commit. **Not filed as a violation.**
- **`AGENTS.md` §9 leaves an agent no conformant way to verify an instrumentation test.** §9 permits only *"isolated tests that have a direct, already-available runtime"* and forbids Gradle outright (*"EVEN for verifying build errors!!!"*), while §6 names `./gradlew :OsmAnd:connectedDebugAndroidTest` as *the* test command — and this PR's deliverable can only be verified that way. Its "unless the user explicitly requests additional verification" carve-out sits in the *second* bullet, attached to which *checks* are permitted, not to the first bullet's prohibition. A defect in the standard rather than in the PR, and only maintainers can close it.

---

### Confirmed — objections tested that did not survive

These are why the findings above are scoped as narrowly as they are.

- **The diagnosis is right**, checked against the crash log on #25667: `onViewDetachedFromWindow(…:431)` ← `ViewGroup.removeAllViewsInLayout(6056)` ← `AbsListView.resetList(2320)` ← `ListView.resetList(660)`. Line 431 matches the 1.11.3 sources exactly.
- **The javadoc's premise is substantially true.** An exhaustive grep of the 1.11.3 `ui`/`ui-android` trees finds **exactly two** references to the flag, both the same `if (…) view.handler else legacyMainHandler` shape. It does not gate semantics delivery, delegate installation, ContentCapture on/off, or any test branch. The "TalkBack regression" hypothesis is **not supported**; residual risk is finding 2 only.
- **Setting the flag before `super.onCreate()` is exactly what upstream prescribes** — the `AndroidComposeUiFlags` KDoc's own example is `onCreate() { …Flag = false; super.onCreate() }`, warning that changing it after Compose loads is undefined. A sanctioned escape hatch, not a hack.
- **The fix is complete for the reported crash.** Line 431 is the only handler `!!` in the delegate; all four scheduling sites already null-check. The one other `handler!!` in the module (`AndroidContentCaptureManager:178`) reads the same flag, so it is covered too.
- **"Just bump Compose" was not available.** 1.11.3 ≡ 1.11.4 byte-identical; 1.12.0 still has `handler!!`; first released artifact with `handler?.` is 1.13.0-alpha01. This *strengthens* the choice made here.
- **Rejecting the per-class `disposeComposition()` route is evidenced, not asserted.** `SearchScopeChip` already carried that override from `6ca837d27f` (2026-08-13), two days *before* the crash was reported, and it did not prevent it.
- **`MapActivity` as test host is house style**, matching `CleanGpxLayerResourcesTest`, `CrosswalkWarningTooEarlyTest`, `GPXTrackAnalysisUpdatesTest`, `RouteRecalculationFromBeginningTest`. Not flagged.
- **The 5.4 line is covered** — `r5.4` already contains `applyComposeWorkarounds` as of 2026-09-01, so the branch that reported the crash on versionCode 5403 has the fix. No backport needed.
- **`AGENTS.md` conformance otherwise holds**: §10 language, §7 logging and API compatibility (no platform API; minSdk 24 safe), §6 test location, §7 house style, §9 file hygiene, §11 English-only, §11 disclaimer template (tool+model, ordered prompt summary, agent-decided section, placed last) and the `(gen by AI)` marker on the #25667 comment. All satisfied.
- **Not a multi-process, init-ordering or multidex hazard** (minSdk 24 is native multidex; `MultiDex.install` runs in `attachBaseContext`). The `:restart` process early-returns before the workaround, but `RestartActivity` only kills the main pid, relaunches and exits — no Compose, no `ListView`.

---

### Scope

2 files / +138 −0 at `b09af0bf`. Three independent fresh-context axes — Compose correctness and blast radius, test quality, `AGENTS.md` conformance — cross-checked against upstream 1.11.3 sources, the crash log on #25667, and the layouts and adapters in this tree. **Not verified:** that the change compiles or that the tests pass — §9 puts Gradle out of bounds and no CI job does it either (finding 1). **Not settled:** whether finding 3 is reachable at runtime.

PR #25808, opened 2026-09-02
fix: disable trip recording start dialog by default

### LINKED ISSUE #25807: App always creates a push notification when I open the app
### Description

Whenever I open the app, I get a push notification about track recording even though I'm not recording anything. This notification is totally useless in my opinion 

### Steps to reproduce

Open the app on android

### Actual result

Push Notification created

### Expected result

No push notification 

```
OsmAnd Version: 5.4.4
Android version:
Device model: Samsung Galaxy A10
```

MeGaurav4:
Closes #25807

I noticed the GpxNotification path for no recording and no points
still built a Logging no data notification. With
SHOW_TRIP_REC_NOTIFICATION on and no prior dismiss, the gate was
only wasNoDataDismissed or not setting, so it showed on every app
open. I added lastBuiltNoData to the gate so the no data case never
shows when not ongoing, while the paused with data case stays gated
by the setting and dismiss. Let me know if you would prefer it
hidden entirely when not ongoing instead.


sonora:
See https://github.com/osmandapp/OsmAnd/issues/25807#issuecomment-5505984598 and https://github.com/osmandapp/OsmAnd/issues/25807#issuecomment-5506101267.

Issue #25590, opened 2026-08-07
"map center" (reference to +/- buttons) disappears from the visible map area


Rowin63:
### Description

Select a track on display and tap the "track" section in its details. The "info" area now uses the bottom half of the screen. The reference to the zoom buttons +/- (map center?) now disappears from the remaining visible map area (it´s somewhere hidden behind the info area or at it´s border). Anyway, now zoomimg in/out gives you a strange and not helpful result. Here the "map center" (+/- zoom reference) should move to the center of the remaining visible map area.

### Steps to reproduce

see above.

### Actual result

see above.

### Expected result

Set the map center to the center of the actually remaining visible map area.

No, it´s no duplicate to the previously opened issue, as offered while creating this. It´s just similar to it.

### Your Environment (required)

Both, Playstores +5.4.0 and nightly.

Zirochkabila:
@Rowin63 Please attach a video showing the steps and the result so that we can better understand the nature of your issue.

Rowin63:
That's weird and not clearly reproduceable to me. 

Here you see the "map center" ("center of zooming") at the edge of the info screen (after selecting a track that was on display):

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/2c440360-5429-4c5b-af9a-b9da2bb7f43a"  width="400" height="auto"/> </th>
  </tr>
</table>


After minimizing Osmand and bringing it back to full screen this "center" moved to it's expected place:


<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/cc8ab814-b289-40ff-8e38-941547292a84"  width="400" height="auto"/> </th>
  </tr>
</table>

And before, when I created the issue, this "center" must have been hidden somewhere behind the tracks info screen (as I could tell by the behaviour of the zooming), but I can't reproduce this now.

yuriiurshuliak:
> OsmAnd~ 5.4.0#8538m, released: 2026-08-12

The issue is reproduced. When the Track details panel occupies part of the screen, the map center / zoom reference point is positioned incorrectly near the bottom edge of the visible map area. After minimizing OsmAnd and returning to the app, the position is recalculated correctly.

**Steps to reproduce**

1. Open OsmAnd.
2. Display a track on the map.
3. Tap the track to open its details.
4. Open the **Track** tab.
5. Keep the bottom panel expanded so it occupies approximately half of the screen.
6. Check the map center / zoom reference point.
[...]

Issue #25756, opened 2026-08-27
Plan a route saves the profile's default activity instead of the selected one


dmpr0:
## Description
Found in https://github.com/osmandapp/OsmAnd/issues/25717

When a track is saved from **Plan a route**, the activity written into the GPX metadata is taken from the profile's *default* activity instead of the value stored in the profile.

<details>
<summary>Probable cause</summary>

`SaveGpxRouteAsyncTask.java:177` reads the preference via `getProfileDefaultValue(appMode)`:

```java
String activityId = app.getSettings().CURRENT_TRACK_ROUTE_ACTIVITY.getProfileDefaultValue(appMode);
```

`getProfileDefaultValue()` ignores the stored value and falls back to the parent `ApplicationMode` default. It should be `getModeValue(appMode)`, which is what `MonitoringSettingsFragment.java:344` uses to display the same preference.

Trip recording is not affected — `SavingTrackHelper.savePreselectedRouteActivity()` reads the stored value correctly.

</details>

### How to reproduce

* Enable the Trip recording plugin.
* Go to **Settings** → **Cycling** → **Trip recording** → **Activity**.
* Change the activity from the default **Road Cycling** to **Gravel Biking**.
* **Menu** → **Plan a route** → **New route** → **Route between points** → select **Cycling**.
* Add a few points to build a route → tap **Done**
* Open the saved track and check its activity.

### Actual result

The track is saved as **Road Cycling** — the profile's default activity, not the selected one. The value selected in the profile settings is stored correctly, it is only ignored on save.

The same happens on custom profiles, where the default is inherited from the base profile: a Motorcycle-based profile set to Road Motorcycling is saved as Adventure Motorcycling, and a Car-based profile set to Enduro Motorcycling is saved as Car.

### Expected result

The track is saved with the activity selected in the profile settings, for both stock and custom profiles.

**Environment**
OsmAnd Version: OsmAnd~ 5.4.0#8629m, released: 2026-08-27
Android: 16
Device model: Google Pixel 6A

dmpr0:
Testing: Please check iOS version too.

Rowin63:
Thanks for checking and confirming.
So I will close the other issue, there's no need for this extra activity switch 👍

Zirochkabila:
# Testing 1

> OsmAnd~ 5.4.0#8714m, released: 2026-09-07
> OsmAnd Maps 5456, released 07.09.26

Android - **Fixed**

<details><summary>Details</summary>
[...]

Issue #25674, opened 2026-08-17
Missing route calculation algorithm selector


garminbreak:
### Description

OSMAND versions previous to 5.4.1 had route calculation selector in Route Parameters menu with development plugin enabled. Now, in 5.4.1 this feature is missing. As result - route calculation is not optimal, the user can't select A* algorithm for most precise route calculation. All user navigation profiles uses 2 phase A* algorithm, what is terrible slow in comparison with HHx Java. This new version 5 4.1. from 2026-08-13 is much worse than previous in route calculations.

### Steps to reproduce

Try Geocoding profile, this mode was able to calculate shortest possible route between any points in A* routing mode. Now, in 2 phase A* mode calculated route is nit optimal.

### Actual result

-

### Expected result

-

### Your Environment (required)

OsmAnd Version: 5.4.1. 2026-08-13
Android version: 14 and above
Device model : many phones

dmpr0:
Hi!

This was reworked in https://github.com/osmandapp/OsmAnd/issues/22987. You can switch the algorithm in **Navigation settings** → **Route parameters** → **Route calculation method**. Switch to **Standard-only** to always use the **A*** algorithm, or to **Fast-only** to always prefer the **HH** algorithm.

garminbreak:
For 5.4.0 I created very special offroad routing profile based on geocoding profile and A* algorithm with heuristic coefficient set on 1.2. Calculation was reasonable fast and allways found shortest possible way. This same profile in 5.4.1 calculates very different routes, of course not shortest possible. IMHO, 5.4.1 uses 2 phase A*, not old A*.
Please return the algorithm selector back to the application.

dmpr0:
### ToDo

- [ ] In Car profiles, **Standard only** should use A* 2-phase

vshcherb:
A* star 2 phase is enabled by default as we tested for car navigation and it starts from 100km routes. simple A* for long routes won't be avaiable as for now

garminbreak:
> A* star 2 phase is enabled by default as we tested for car navigation and it starts from 100km routes. simple A* for long routes won't be avaiable as for now

This feature doesn't work. I have special offroad profile that works best with A* algorithm. Even on 30 km route uses A* 2 phase routing and not A* classic mode. I'm absolutly sure, because this profile in A* can calculate shortest route, but in A* 2 phase not. None of my test routes 30-70 km long were shortest possible. App version 5.4.4.

[...]

Issue #25374, opened 2026-07-05
UI elements and map text are rendered too small on 10" high-res tablets since v5.4.0


pebogufi:
### Description

Since version
OsmAnd~ 5.4.0#7807m, veröffentlicht: 2026-05-23
and
OsmAnd~ 5.4.0#7923m, veröffentlicht: 2026-06-10
(and until version of today)
there is a change of scaling. Buttons and other symbols are now much smaler and worse to read, especially while driving.
See buttons upper right corner and distance symbol "1 km" lower left corner.
The screenshots where made directly after instslling the different version without making any changes in settings.

**The difference in scaling ONLY occurs with 
10" tablet Samsung tab S5e 10.50 Zoll 16:10, 2560 x 1600 Pixel, Android 10**

<table style="width:100%">
  <tr>
    <th>This is old  (good) version</th>
    <th>This is new version with smaller buttons and symbols</th>
  </tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/79c98784-78ab-4d7c-b2c9-d6eca667d7a8"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/887f19fa-71bc-49e6-96b8-9c70ade46b9d"  width="300" height="auto"/> </th>
  </tr>
</table> 

### Steps to reproduce

Change with new version

### Actual result

Buttons, icons too small, bad to read

### Expected result

Older version was better to read

### Your Environment (required)

OsmAnd Version: see text
Android version: see text
Device model: see text

Zirochkabila:
- [ ] Check for regression
- [ ] Compare with iOS

yuriiurshuliak:
QA Reproduced

Tested on **Android Studio 10" tablet emulator**, `2560 × 1600`.

**Regression range:**

* ✅ `OsmAnd Nightly 5.4.0#7851m`, released 2026-06-01 — not affected

* ❌ `OsmAnd Nightly 5.4.0#7890m`, released 2026-06-05 — affected


**Test location:** Denzlingen / Emmendingen, Germany
`48.06833, 7.88250`

The **Zoom level** developer widget was enabled to ensure both screenshots were compared at exactly the same map zoom:

`14 +0.00`

### Result

A scaling difference is clearly visible in the latest build:

* The **top-right Zoom level widget** is noticeably smaller, including its icon, text, height, and width.
* The **map scale text** in the bottom-left (`0.5 mi`) is also noticeably smaller.
[...]

PR #25811, opened 2026-09-02
Move the map center out from under the track menu right away

vshcherb:
Fixes https://github.com/osmandapp/OsmAnd/issues/25590

`TrackMenuFragment` reports the part of the screen it covers to `MapDisplayPositionManager`, but only asked it to recalculate from `onResume` / `onPause` / `onHiddenChanged`. Opening the **Track** tab expands the menu over the bottom half of the map without any recalculation, so the zoom reference point stayed in the middle of the screen — hidden behind the menu — until the app was sent to background and restored, which is exactly what the reporter observed.

Recalculate in `updateMainViewLayout`, which the base fragment calls whenever the menu is laid out at a new position, including at the end of the expand and collapse animations.

### Tests

`TrackMenuMapCenterTest` opens a track, checks the map center sits in the middle of the visible map, opens the Track tab and checks it again. On an API 35 emulator it passes with the fix; without it the assertion fails with

```
Map center is not centered in the visible part of the map expected:<0.34479168> but was:<0.5>
```

i.e. the center stays in the middle of the screen instead of moving above the menu.

Runtime measured over three runs on that emulator: the test method took **8.7 s, 31.1 s and 31.5 s**; the `connectedGplayFreeOpenglArm64DebugAndroidTest` task took ~1 min once the build was up to date. Most of that is app start-up and GPX import rather than the assertion, so if that is too expensive for the suite the test can be dropped and the fix kept.

### Noticed nearby, not changed here

`TrackDetailsMenuFragment.updateBoundsChangeListener(false)` calls `addOnLayoutChangeListener` instead of removing the listener, so it accumulates one per resume.

Related issues triaged in the same pass: #25756, #25374.

## AI disclaimer

Prompts used (summarised):
1. Investigate whether the reported issues are regressions, reproduce them, prepare fixes on branches off 5.4, and consider automated tests where the problem is a regression.
2. Issues to look at: #25374, #25283, #25674, #25756, #25590.
3. Open pull requests against `r5.4` linking the issues.
4. Measure how long this test takes, and follow the repository AI-disclosure rules.

Decided by the agent, not requested explicitly: hooking the recalculation into `updateMainViewLayout` rather than the menu-state callback, the shape of the UI test and its idling resource, and reporting the `TrackDetailsMenuFragment` listener leak without fixing it here.

### DIFF
--- a/OsmAnd/test/java/net/osmand/test/ui/track/TrackMenuMapCenterTest.java
+++ b/OsmAnd/test/java/net/osmand/test/ui/track/TrackMenuMapCenterTest.java
@@ -1,0 +1,200 @@
+package net.osmand.test.ui.track;
+
+import static net.osmand.test.common.OsmAndDialogInteractions.clickViewWithId;
+import static net.osmand.test.common.OsmAndDialogInteractions.skipAppStartDialogs;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import android.graphics.PointF;
+import android.graphics.Rect;
+import android.os.Handler;
+import android.os.Looper;
+import android.view.View;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+import androidx.test.espresso.Espresso;
+import androidx.test.espresso.IdlingResource;
+import androidx.test.ext.junit.rules.ActivityScenarioRule;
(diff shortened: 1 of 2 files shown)


OsmAnd/test/java/net/osmand/test/ui/track/TrackMenuMapCenterTest.java
```diff
@@ -0,0 +1,200 @@
+package net.osmand.test.ui.track;
```
vshcherb, line 1: Test takes too long investigate whether to accept it or not.

Issue #25283, opened 2026-06-19
Legacy regional map file (.obf) is not deleted after updating to new department-based map structure


Phil2607:
### Description

Since beginning of 2026 download of a region as a single file is no more possible. Instead a file for each included department is downloaded. After updating the old region file is not deleted and remains in memory (in my case France_auvergne-rhone-alpes_europe.obf). As this file remains with the new ones two versions of the map are displayed where there are differences which can be confusing.

### Steps to reproduce

- You need a phone with a map downloaded in 2025. In my case France Auvergne-Rhone-Alpes
- Update the map
- One file for each department is downloaded. In my case Ain, Allier, Ardèche, Cantal, ...
- A new obf file is downloaded for each department but the old obf region file remains

### Actual result

view of [https://www.openstreetmap.org/#map=17/45.190275/5.604926&layers=P](https://www.openstreetmap.org/#map=17/45.190275/5.604926&layers=P)

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/d17b0692-23aa-409f-8c55-b93a353670bc"  width="300" height="auto"/> </th>
  </tr>
</table>  

### Expected result

Same view after manually removing old region file

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/068d992d-dc3e-45a1-86f9-af3c46891925"  width="300" height="auto"/> </th>
  </tr>
</table>  

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: OsmAnd~ 5.3.10
Android version: /e/OS 3.7.3 (Android 15)
Device model: FairPhone 5
Crash-Logs: no crash
```

scaidermern:
Related to #23715 and #25150.

yuriiurshuliak:
> OsmAnd~ 5.4.0#8049m, released: 2026-06-23

After updating **France, Auvergne-Rhône-Alpes** from the old regional map structure to the new department-based structure, the legacy regional `.obf` file may remain on the device and still be used together with the new map files, causing outdated/duplicated map rendering.

I checked the reported area before and after updating the map, but I **could not reproduce a visible rendering difference** in my test.

<details>
<summary>Details</summary>

### Tested area

* `45.190275, 5.604926`

### Test result

I compared the map view:

1. Before updating the map
2. After updating the map

[...]

Issue #25018, opened 2026-05-06
Appearance for all widgets


vshcherb:
Configure for each panel (Related - https://github.com/osmandapp/OsmAnd/issues/18357, https://github.com/osmandapp/OsmAnd/issues/7912, https://github.com/osmandapp/OsmAnd/issues/7111, https://github.com/osmandapp/OsmAnd/issues/24126, https://github.com/osmandapp/OsmAnd/issues/16017)
- Row size
- Background color

# Configure Screen

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/d06f5bf0-51c4-40b1-b008-e9635e66962d"  width="200" height="auto"/> </td>
  </tr>
</table>   

We will rename the "Transparent widgets" item to **Appearance** and use it to open a new configuration screen.

- [x] Rename "Transparent widgets" to **Appearance**
- - [x] Remove the toggle switch
- - [x] Use the same icon
- [x] On tap, open the new Appearance screen

# Appearance Screen, [Figma](https://www.figma.com/design/95sVxDjUfZ0afiW7XmTUYx/Configure-screen?node-id=3300-111871&m=dev)

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/71842c64-ba72-4a0b-b7c5-04ad0bbc0ece"  width="200" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/5bfb4b24-5b65-406f-8927-b96ce99a4977"  width="200" height="auto"/> </td>
  </tr>
</table>   

## AppBar
Use an AppBar with tabs, similar to the one we use on the panels screen, where each tab represents a specific panel.

- [x] Title – shows the corresponding panel name, for example: **Left panel**
- [x] Show a tab for each panel (Left, Right, Top, Bottom)
- [x] Tapping a tab should switch between panels and update all content below accordingly

**Actions:**
- [x] Show Overflow menu icon (`⋮`) in the top right corner.
- [x] On tap, open a dropdown menu with the option to copy settings.
  - [x] **Copy from** (Header/Title)
  - [x] `Another profile` (opens profile selection)
  - [x] Divider line
  - [x] `Right panel` / `Top panel` / `Bottom panel` (depending on the currently active tab, show the other 3 panels).

## Preview

<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
[...]

PR #25812, opened 2026-09-02
Keep the map scale text readable on tablets

vshcherb:
Part of https://github.com/osmandapp/OsmAnd/issues/25374

QA bisected the tablet scaling regression to nightly `#7851m` (ok) → `#7890m` (broken). The change in that window is `5ec248bc5d` "Removed tablet sizes for widget texts" (#25228, 5 June), which dropped `map_widget_text_size` / `map_widget_text_size_small` from `values-large`.

For left/right panel widgets that is intentional — their size is now user-controlled through the new widget appearance menu (#25018). But `map_ruler.xml` uses the same dimension and the map scale is **not** a panel widget, so there is no way to compensate: on a 10" tablet the ruler bar is still `180dp` wide (`values-large`) while its label dropped from `23sp` to `15sp`. That is the "1 km" label the reporter pointed at.

Give the ruler its own `map_ruler_text_size` dimension and restore the tablet value.

`item_elevation_stat_block.xml` reads the same dimension and is not a panel widget either — leaving it alone here, worth a separate look.

The remaining part of the issue (dmpr0's ToDo about width restrictions for the left/right panels depending on screen width) is a product decision and is not touched by this PR.

### Before / after

Tablet emulator, 2560 × 1600 (the reporter's Tab S5e resolution), same map position and zoom, same crop of the bottom-left corner.

| Before (`r5.4`) | After (this PR) |
| --- | --- |
| map scale label at 15sp | map scale label at 23sp |

### Tests

Resource-only change, no automated coverage.

Related issues triaged in the same pass: #25756, #25590.

## AI disclaimer

Prompts used (summarised):
1. Investigate whether the reported issues are regressions, reproduce them, prepare fixes on branches off 5.4, and consider automated tests where the problem is a regression.
2. Issues to look at: #25374, #25283, #25674, #25756, #25590.
3. Open pull requests against `r5.4` linking the issues.
4. Add before/after screenshots of the fix to this PR, and follow the repository AI-disclosure rules.

Decided by the agent, not requested explicitly: narrowing the issue down to the map ruler instead of the panel widgets, introducing a separate `map_ruler_text_size` dimension rather than restoring the shared one, and the choice of device, map position and crop for the screenshots.

### DIFF
--- a/OsmAnd/res/layout/map_ruler.xml
+++ b/OsmAnd/res/layout/map_ruler.xml
@@ -19,7 +19,7 @@
 			android:layout_gravity="bottom"
 			android:gravity="center_horizontal"
 			android:lineSpacingMultiplier="0.9"
-			android:textSize="@dimen/map_widget_text_size_small"
+			android:textSize="@dimen/map_ruler_text_size"
 			tools:text="100 m" />
 
 		<TextView
@@ -29,7 +29,7 @@
 			android:layout_gravity="bottom"
 			android:gravity="center_horizontal"
 			android:lineSpacingMultiplier="0.9"
-			android:textSize="@dimen/map_widget_text_size_small"
+			android:textSize="@dimen/map_ruler_text_size"
 			tools:text="100 m" />
 
 	</FrameLayout>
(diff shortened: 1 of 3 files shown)


Kreiri (approved):
Looks good to me.

Issue #25737, opened 2026-08-25
Glide Ratio docs: incorrect feature information and a hidden tap toggle


onetrev:
The [Informational Widgets](https://osmand.net/docs/user/widgets/info-widgets/#glide-ratio) page has incorrect information about the Average Glide Ratio widget and there's a related problem in the app.

**1. "Average vertical speed" doesn't exist**

The docs say tapping the widget switches it between "Average glide ratio" and "Average vertical speed".  However, there's no vertical speed mode. Tapping does nothing, the context menu has no "By tapping" row (Sunset and others do), and the widget only ever shows a ratio. I can't find vertical speed anywhere else in the app either.

Dropped feature or docs error? Either way, should this line be removed for now?

**2. Glide ratio to target has a tap toggle nobody can find**

Tapping this widget **does** switch between "Glide ratio to target" and "Target elevation" as the docs say, but nothing in the app says so — there is no "By tapping..." line in its "Settings" context menu like other widgets that have tap functionality have. 

NOTE: if needed I can quickly make this second one an issue in the app repo as technically it's a code issue not a docs thing.

EugeneZmeuk:
Hi @onetrev 
The Android version 5.5 requires two updates: implementation of the average vertical speed widget, and improvements to the Glide Ratio widget's context menu.

dmpr0:
### Description
Android https://github.com/osmandapp/OsmAnd/issues/16968
iOS https://github.com/osmandapp/OsmAnd-iOS/issues/3395

iOS implementation has modes for both widgets.


## ToDo

### Glide ratio to target

**Target elevation**

Show elevation of target object, for example 320 m
- [ ] Name on widget – Target elevation

### Average glide ratio

**Average vertical speed**
Show average vertical speed over configurable interval

- [ ] Widget icon - widget_vertical_average_speed

DmitryAlexei:
https://osmand.net/docs/user/widgets/info-widgets?_highlight=glide#glide-ratio

Kreiri:
[Android app]

In addition to "switch the mode by tapping" line in its settings, "glide ratio to target" widget needs to have its description updated with information on which modes it can switch between.

Compare "glide ratio to target" widget settings: 

<img width="337" height="360" alt="Image" src="https://github.com/user-attachments/assets/17b3046b-76dd-426a-991c-69bacc715943" />

and "sunrise" widget settings: 

[...]

PR #25817, opened 2026-09-02
Show "By tapping" hint in Glide ratio to target widget settings

vshcherb:
Follow-up to #25737, which was closed after the docs fix (osmandapp/web#1959 + `c4abb65ff`) while the app-side part was left open.

### Problem

The **Glide ratio to target** widget toggles between *Glide ratio to target* and *Target elevation* when tapped — `GlideTargetWidgetState.changeToNextState()` flips the `glide_widget_show_target_altitude` preference. But nothing in the app tells the user this: `WidgetType.getSettingsFragment()` had no branch for `GLIDE_TARGET`, so its context menu fell through to the generic `BaseSimpleWidgetInfoFragment` with no hint.

Sunrise/Sunset and Zoom level, which have the same kind of tap toggle, do show the note. The docs now state the shortcoming explicitly ("Switching option is not available in the widget's context menu") rather than the app fixing it.

### Change

- New `GlideTargetWidgetInfoFragment` extending `BaseSimpleWidgetInfoFragment`, inflating a one-row layout with the existing `switch_mode_by_tapping_on_widget` string.
- Wire `GLIDE_TARGET` to it in `WidgetType.getSettingsFragment()`.

No new strings, no behaviour change — only the missing hint. 40 lines across 3 files.

**Average glide ratio** is deliberately untouched: it has no `WidgetState` on Android, so there is no tap toggle to document. The *Average vertical speed* mode from the #25737 to-do list is still iOS-only and remains out of scope here.

### Testing

Not run through a full local build — the fresh worktree fails to link resources on unrelated generated POI icons (`mm_storage_tank`, `mx_special_star_stroked`), which are absent from the repo and come from `OsmAnd-resources`. Resource merging accepted the new layout. Please confirm on CI.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/settings/GlideTargetWidgetInfoFragment.java
+++ b/OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/settings/GlideTargetWidgetInfoFragment.java
@@ -1,0 +1,22 @@
+package net.osmand.plus.views.mapwidgets.configure.settings;
+
+import android.view.ViewGroup;
+
+import androidx.annotation.NonNull;
+
+import net.osmand.plus.R;
+import net.osmand.plus.views.mapwidgets.WidgetType;
+
+public class GlideTargetWidgetInfoFragment extends BaseSimpleWidgetInfoFragment {
+
+	@NonNull
+	@Override
+	public WidgetType getWidget() {
+		return WidgetType.GLIDE_TARGET;
+	}
+
+	@Override
+	protected void setupMainContent(@NonNull ViewGroup container) {
+		inflate(R.layout.fragment_widget_settings_glide_target, container);
+	}
+}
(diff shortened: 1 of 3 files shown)


OsmAnd/src/net/osmand/plus/views/mapwidgets/configure/settings/GlideTargetWidgetInfoFragment.java
```diff
@@ -0,0 +1,22 @@
+package net.osmand.plus.views.mapwidgets.configure.settings;
+
+import android.view.ViewGroup;
+
+import androidx.annotation.NonNull;
+
+import net.osmand.plus.R;
+import net.osmand.plus.views.mapwidgets.WidgetType;
+
+public class GlideTargetWidgetInfoFragment extends BaseSimpleWidgetInfoFragment {
```
vshcherb, line 10: We shouldn't create such small clasess (as separate files), check how many we have and probably combine them into groups

Kreiri:
This PR doesn't solve the problem of there being no indication of what modes are these, though. E.g. in "Sunrise" widget its description says "Shows the time of the next sunrise or the time left until sunrise",  explicitly stating which modes are there, but for "Glide ratio to target" it only says "Shows the glide ratio needed to reach the target point" - there no indication of "target elevation" mode existing.

PR #25818, opened 2026-09-02
Avoid numbers in OsmandRegions.getSearchIndex() (LLM)

RZR-UA:

## AI disclaimer

Produced with Codex (GPT-5), including the implementation and tests.

Prompts used (summarised):

1. Investigate the first bug in issue #25485, focusing on region-reference matching and `regions.ocbf` data.
2. Initially implement and test a generated `osmand_region_ref` field based on `ref:USPS`, with an `ISO3166-2` fallback.
3. Analyse and remove the one-character matching bypass, preserve `NameStringMatcher` behaviour, and reuse the same matcher in production and tests.
4. Generate and inspect a fresh local OCBF and run the requested tests.
5. Abandon the generated-field approach, close the tools PR, regenerate the local OCBF from tools master, and revert the Android changes except for tests.
6. Restore the shared matcher and test Pennsylvania, Meuse, Bicol Region, and Central Visayas using real region data.
7. Preprocess semicolon-separated `ref` and `alt_name` elements, removing elements containing numeric tokens detected with `splitAndNormalize()` while preserving the existing `ref` deduplication behaviour.
8. Verify the locally used abbreviation tag and analyse remaining numeric-query matches using generated OCBF data.
9. Handle hyphenated numeric tokens before `splitAndNormalize()` and add real-region regression cases for queries `1`, `4`, `22`, `33`, and `44`.

Decided by the agent, not requested explicitly: None.

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
+++ b/OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
@@ -3,6 +3,7 @@ package net.osmand.binary;
 import net.osmand.PlatformUtil;
 import net.osmand.map.OsmandRegions;
 import net.osmand.map.WorldRegion;
+import net.osmand.util.Algorithms;
 import net.osmand.util.SearchAlgorithms;
 
 import java.io.IOException;
@@ -138,6 +139,10 @@ public class CommonWords {
 		for (WorldRegion s : subregions) {
 			String t = s.getRegionSearchText();
 			if (t != null) {
+				String r = s.getRegionSearchRef();
+				if (Algorithms.isNotEmpty(r)) {
+					t += " " + r;
+				}
 				String[] ns = t.split(" ");
 				for (String n : ns) {
 					if (n.contains(";")) {
--- a/OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
+++ b/OsmAnd-java/src/main/java/net/osmand/binary/CommonWords.java
@@ -3,7 +3,6 @@ package net.osmand.binary;
 import net.osmand.PlatformUtil;
 import net.osmand.map.OsmandRegions;
 import net.osmand.map.WorldRegion;
-import net.osmand.util.Algorithms;
 import net.osmand.util.SearchAlgorithms;
 
 import java.io.IOException;
@@ -139,10 +138,6 @@ public class CommonWords {
 		for (WorldRegion s : subregions) {
 			String t = s.getRegionSearchText();
 			if (t != null) {
-				String r = s.getRegionSearchRef();
-				if (Algorithms.isNotEmpty(r)) {
-					t += " " + r;
-				}
 				String[] ns = t.split(" ");
 				for (String n : ns) {
 					if (n.contains(";")) {
--- a/OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
+++ b/OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
@@ -1,5 +1,6 @@
 package net.osmand.map;
 
+import net.osmand.CollatorStringMatcher;
 import net.osmand.OsmAndCollator;
 import net.osmand.PlatformUtil;
 import net.osmand.ResultMatcher;
@@ -661,6 +662,11 @@ public class OsmandRegions {
 		return ind.toString();
 	}
 
+	public static boolean isRegionNameMatched(String query, String regionName) {
+		return Algorithms.isNotEmpty(regionName)
+				&& new CollatorStringMatcher(query, CollatorStringMatcher.StringMatcherMode.CHECK_EQUALS_FROM_SPACE)
+				.matches(regionName);
+	}
 
 	public boolean isDownloadOfType(BinaryMapDataObject object, String type) {
 		int[] addtypes = object.getAdditionalTypes();
--- a/OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
+++ b/OsmAnd-java/src/main/java/net/osmand/map/OsmandRegions.java
@@ -14,6 +14,7 @@ import net.osmand.data.QuadTree;
 import net.osmand.util.Algorithms;
 import net.osmand.util.MapAlgorithms;
 import net.osmand.util.MapUtils;
+import net.osmand.search.core.SearchPhrase.NameStringMatcher;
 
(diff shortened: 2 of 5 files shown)


vshcherb (commented):
Checked this against the data we actually ship (`OsmAnd-java/src/main/resources/net/osmand/map/regions.ocbf`), not only by reading the diff. Two problems block it, and I think the whole direction (a second parallel search field + an exact-match path) is more machinery than the bug needs.

## Blocking

**1. The OCBF field does not exist, so the feature is a no-op and the ref search regresses.**
`osmand_region_ref` is not present in any `regions.ocbf` in the repo, and the generation side is not part of this PR (it lives in another repo). So `getRegionSearchRef()` returns `null` for every region, while `ref` has just been removed from `getSearchIndex()`. Net effect on master: search by ref stops working entirely, and the two halves can't land atomically.

Actual tags in the shipped file:

```
france_great-east_meuse_europe:  key_name=meuse, ref=55        <- no FR-55 anywhere
us_pennsylvania_northamerica:    ref=PA, short_name=Pa.
us_california_northamerica:      ref=CA, short_name=Calif.
```

**2. `testSearchRegionByRef` cannot pass on CI.** `matches("FR-55", meuse)` — that string is not in the data. `matches("PA", pennsylvania)` — "pa" came precisely from `ref=PA`, which this PR removes from the index.

**3. The download screen is not covered and stays broken afterwards.** `DownloadSearchUIModel:404` reads `getRegionSearchText()` and matches with a plain `contains`; it knows nothing about `regionSearchRef`. "pennsylvania" does not contain "pa", so looking up a map by `PA`/`CA` there dies — and keeps dying even once a regenerated OCBF lands, because that code path never consults the new field.

**4. Ref coverage narrows a lot.** In the current file 694 regions carry `ref`, of which only 161 are numeric; 483 non-US non-numeric ones would be dropped: `afghanistan_baghlan=BGL`, `argentina_cordoba=CBA`, `argentina_santa-fe=SF`, `albania=AL`, … `ref:USPS || ISO3166-2` replaces those local codes with `AR-X`-style codes nobody types.

**5. It only fixes half of the reported bug.** Numbers also come from names and from hyphenated refs. Region hits per query against today's index: `1`→6, `2`→1, `11`→4, `22`→7, `33`→4 — e.g. `indonesia_bali` has `ref=BA-22`, `indonesia_bangka-belitung` has `ref=BB-1`, `philippines_davao-region` has `alt_name=Region XI;Region Ⅺ;Region 11`. Dropping the pure-`ref` half leaves all of those matching.

## Proposal: don't index numbers, instead of routing around them

A bare number is never a useful search key for a region, while every code we actually care about (PA, CA, BGL, CBA) is alphabetic. So drop number words at index time, reusing the same number check the spatial index already uses (`SearchAlgorithms.isNumber2Letters()`, see `SpatialSearchContext:913`), and keep one search field:

```java
private static final Pattern WORD_PATTERN = Pattern.compile("[\\p{L}\\p{N}\\p{M}]+");

// getSearchIndex(...): "ref" stays in the list of indexed tags
final String vl = removeNumbers(it.value().toLowerCase(Locale.US));
if (WORD_PATTERN.matcher(vl).find() && (ind.indexOf(vl) == -1 || tp.tag.equals("ref"))) {
    ind.append(" ").append(vl);
}

// a number is not searchable on its own: "55" (Meuse), "BA-22" (Bali), "Region 11" (Davao)
private static String removeNumbers(String value) {
    Matcher m = WORD_PATTERN.matcher(value);
    StringBuilder res = new StringBuilder();
    while (m.find()) {
        String word = m.group();
        m.appendReplacement(res, SearchAlgorithms.isNumber2Letters(word) ? "" : Matcher.quoteReplacement(word));
    }
    m.appendTail(res);
    return res.toString();
}
```

Measured over all 1710 regions of the shipped OCBF (number of regions matched with `CHECK_EQUALS_FROM_SPACE`):

| query | now | with this patch |
|---|---|---|
| `1` | 6 | 0 |
| `2` | 1 | 0 |
| `11` | 4 | 0 |
| `22` | 7 | 0 |
| `33` | 4 | 0 |
| `55` | 2 | 0 |
| `pa` | 4 | 4 |
| `ca` | 10 | 10 |
| `bgl` / `cba` / `meuse` / `pennsylvania` / `region xi` | 1 | 1 |

No region loses its search text (0 regions end up with an empty index). No OCBF format change, no new `WorldRegion` field, no new matching API, and no follow-up patches in `CommonWords` / `DownloadSearchUIModel`.

For the record, the reason a separate field looked necessary: `CollatorStringMatcher.isSpace()` treats every non-alphanumeric character as a word separator (and `cmatches()` additionally retries with `-` stripped), so `FR-55` sitting in the common blob still matches the query `55`. Removing numbers fixes that at the source instead of building a parallel field to escape the matcher.

If we do want ISO3166-2 / `ref:USPS` in the index later, it can go into the same `regionSearchText` once numbers are gone: `us-pa` then matches both `PA` and `US-PA`, and `fr-55` no longer matches `55`. Still no need for `WorldRegion.regionSearchRef` or `OsmandRegions.isRegionNameMatched`.

## Smaller notes (should the current direction stay)

- `isRegionNameMatched` is `static` in `OsmandRegions` but uses no state of it, so it doesn't belong there; and it constructs a `NameStringMatcher` (with a `Collator`) per region per keystroke inside the recursive walk over the whole download tree (~1700 regions). The matcher should be built once per phrase and passed in — the PR description says "restore shared matcher reuse", but the code does the opposite. It should also take a `WorldRegion` rather than `(name, ref)` so callers don't unpack two fields every time.
- `getRegionSearchRef()` makes a second pass over `getObjectNames()` and decodes tags by hand, while every other `region_*` field is read through `MapIndexFields`/`initTypes`. The ref could be picked up in the same `getSearchIndex()` loop.
- Naming: all neighbouring constants are `region_*` (`region_wiki_link`, `region_population`); the new one is `osmand_region_ref`.
- `regionSearchRef` keeps the original case while `regionSearchText` is lowercased — hence the `equalsIgnoreCase`. Two normalizations in one method.
- Dropping the `phrase.getFullSearchPhrase().length() <= 1 -> return true` bypass is a good change, but it is a separate one (it was the reason single-character queries listed every region) and deserves its own line in the description plus a test.
- Unrelated, but inside the touched code: `searchResult.localeName = name` publishes the concatenated lowercase index blob as the user-visible name of the result.

PR #25828, opened 2026-09-03
Fix #25604: apply Display position setting immediately when map is not linked to GPS

### LINKED ISSUE #25604: Display position (map center) does not always correspond to the setting
**Prerequisite to reproduce:** 
Your actual map center ("Display position) is not centred to the gps position.

When I switch from a profile with "Display position" is set to center, to another profile where "Display position" is set to bottom or automatic, the map center does not follow the new setting.
And also when I change this setting in a profile from center to bottom, the map center doesn’t change straight away.

If i force close and restart Osmand, or when I change the screen orientation (flipping smartphone), the setting will be applied.
Also as soon as I synchronise with the current gps position the expected behavior occours and the map center stay in sync with the setting, also if I select an other profile.

**What else:**
When changing the "Display position" in settings, the status text of the setting will not be updated.
This behavior is also mention here https://github.com/osmandapp/OsmAnd/issues/24703#issuecomment-4192130512
> ### **Severity**: 🔵 Trivial / **Priority**: 🔵 Low
> 
>     * [ ]  1. The “Display position” status is not updated

Here's one about the map center update, that sounds similar: https://github.com/osmandapp/OsmAnd/issues/25589#issue-5085681708
> To make the "map center mark" reappear and get the "+/-" buttons working correct again, you have to tap the "get position" button. But of course then you´re away from where you wanted to be on the map. Or close and restart the app - this at least keeps you in your selected map area.

But this bug appears to have been fixed in the nightly version.

Thank you very much for your attention. I hope the error can be reproduced and fixed.

**My Environment**
Nightly 5.4.0#8507 (2026-08-07), Osmand 5.3.10 on Android 17

vshcherb:
## Summary
- Fixes #25604 — switching a profile (or changing "Display position" directly) had no visible effect while the map was manually panned away from the GPS location, until the user tapped "My location", rotated the screen, or restarted the app.
- Root cause: `MapViewTrackingUtilities.updateSettings()` only called `MapDisplayPositionManager.updateMapDisplayPosition()` when `isMapLinkedToLocation` was `true`. Once the map is manually panned, that flag becomes `false`, so the anchor ratio (Center/Bottom/Automatic) was never recalculated on profile switch. Tapping "My location" happens to flip that flag back to `true` first, which is why it "fixes" it — screen rotation and app restart bypass the gate through separate code paths.
- Fix: always recalculate and refresh the display position in `updateSettings()`, passing `refreshMap=true` so a real ratio change is applied immediately. `updateMapDisplayPosition()` only recomputes the anchor ratio used for future centering — it does not move the map itself — so removing the `isMapLinkedToLocation` gate is safe regardless of GPS-link state.

## Verification (measured, not eyeballed)
Built and installed both the pre-fix (`origin/master`, isolated git worktree) and post-fix APKs on an emulator (API 35), and measured the actual on-screen vertical position of the location marker with a Python/PIL blue-pixel centroid script (`x` range restricted to 0.35w–0.65w to exclude the "My location" FAB), rather than relying on a visual read of the screenshot:

| Profile / Display position | Expected ratioY (`MapPosition`) | Measured marker position |
|---|---|---|
| Browse map (Center) | 0.50 | **49.75%** of screen height |
| Driving (Bottom) | 0.85 | **83.80%** of screen height |

Both measurements match the source constants (`MapPosition.CENTER = 0.5f`, `MapPosition.BOTTOM = 0.85f`) closely enough to confirm the anchor is actually being applied, not just that a refresh occurred.

Behavioral repro, both builds:
1. Pan the map away from the current location (`isMapLinkedToLocation = false`).
2. Switch the active profile from one with Display position = Center to one with Display position = Bottom (or vice-versa), **without** tapping "My location".
3. **Before fix**: marker anchor stays at the old ratio until "My location" is tapped, the screen is rotated, or the app is restarted.
4. **After fix**: marker anchor updates immediately to the new profile's ratio on the profile switch itself.

Build verification: `:OsmAnd:compileAndroidFullLegacyFatDebugJavaWithJavac` compiles cleanly for both the change and the surrounding code.

Before/after screen recordings are available and will be attached directly to issue #25604.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/base/MapViewTrackingUtilities.java
+++ b/OsmAnd/src/net/osmand/plus/base/MapViewTrackingUtilities.java
@@ -455,9 +455,7 @@ public class MapViewTrackingUtilities implements OsmAndLocationListener, IMapLoc
 	}
 
 	private void updateSettings(boolean updateRotationByCompass) {
-		if (isMapLinkedToLocation) {
-			mapDisplayPositionManager.updateMapDisplayPosition();
-		}
+		mapDisplayPositionManager.updateMapDisplayPosition(true);
 		registerUnregisterSensor(app.getLocationProvider().getLastKnownLocation(), false);
 		if (mapView != null && updateRotationByCompass) {
 			mapView.initMapRotationByCompassMode();


Kreiri (approved):
Looks good to me.

PR #25829, opened 2026-09-03
Fix radius ruler near globe poles

### LINKED ISSUE #25657: Radius ruler in Globe View mode misbehaves near poles
### Description

in Globe View mode, radius ruler works fine unless you move the position towards either (North/South) Pole. The ruler circles nearing the Pole become "caved". The closer to the Pole you get, the messier the lines look.

### Steps to reproduce

1) turn on globe view in profile settings
2) add Radius Ruler to screen widgets:  Configure Screen -> Widgets -> Right panel -> "+" (add) -> Radius Ruler
3) get back to map view
4) tap on the added Radius Ruler icon to turn on the "distance circles" display
5) zoom out (if needed) and move the map view towards e.g. the North Pole
6) observe the "distance circles" misbehaving when you get close enough to the Pole

### Actual result

Radius ruler does not show correct "distance circles" near poles

### Expected result

Raduis ruler should always display correct "distance circles", also for positions near the Poles

```
OsmAnd Version: 5.3.10
Android version: 16
Device model: Samsung  Galaxy S24
Crash-Logs: no crash logs - the bug does not cause a crash
```

alex-dev-neo:
Fixes #25657

### Summary

- Generate globe ruler rings with great-circle destinations and extended polar coordinates.
- Calibrate ruler spacing against the active globe projection instead of Web Mercator density.
- Split paths at projection discontinuities and clip flat-map paths at the Mercator latitude boundary.
- Reuse drawing paths to avoid per-frame list and path allocations.

### Screenshots

<table>
<tr>
<td align="center"><strong>North Pole</strong></td>
<td align="center"><strong>South Pole</strong></td>
</tr>
<tr>
<td>Fixed radius ruler near the North Pole in 3D globe view</td>
<td>Fixed radius ruler near the South Pole in 3D globe view</td>
</tr>
<tr>
<td align="center"><strong>Mid-latitude (Europe)</strong></td>
<td align="center"><strong>Equator (Gulf of Guinea)</strong></td>
</tr>
<tr>
<td>Radius ruler at a mid-latitude in 3D globe view</td>
<td>Radius ruler at the equator in 3D globe view</td>
</tr>
</table>

<p align="center"><strong>Close-up 3D Relief (Swiss Alps)</strong></p>
<p align="center">Radius ruler over enabled 3D Relief in the Swiss Alps</p>

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
@@ -56,6 +56,8 @@ public class RadiusRulerControlLayer extends OsmandMapLayer implements OsmAndCom
 	private static final float CIRCLE_ANGLE_STEP = 5;
 	private static final int SHOW_COMPASS_MIN_ZOOM = 8;
 	private static final long COMPASS_REFRESH_INTERVAL_MS = 100L;
+	private static final double MAX_GLOBE_MERCATOR_ANGLE = 2 * Math.PI - 1e-7;
+	private static final long POINT31_FULL_RANGE = 1L << 31;
 
 	private OsmandApplication app;
 	private MapWidgetRegistry widgetRegistry;
@@ -74,6 +76,7 @@ public class RadiusRulerControlLayer extends OsmandMapLayer implements OsmAndCom
 	private float cacheMapDensity;
 	private MetricsConstants cacheMetricSystem;
 	private int cacheIntZoom;
+	private boolean cacheSphericalMap;
 	private LatLon cacheCenterLatLon;
 	private ArrayList<String> cacheDistances;
 
@@ -92,6 +95,7 @@ public class RadiusRulerControlLayer extends OsmandMapLayer implements OsmAndCom
 	private final Path arrow = new Path();
 	private final Path arrowArc = new Path();
 	private final Path redCompassLines = new Path();
+	private final Path rulerCircle = new Path();
 
 	private final double[] degrees = new double[72];
 	public static final String[] CARDINAL_DIRECTIONS = {"N", "NE", "E", "SE", "S", "SW", "W", "NW"};
@@ -103,6 +107,7 @@ public class RadiusRulerControlLayer extends OsmandMapLayer implements OsmAndCom
 	};
 
 	private float cachedHeading;
+	private boolean sphericalMap;
 	private long lastCompassRefreshTime;
 	@Nullable
 	private View compassRefreshView;
@@ -275,6 +280,7 @@ public class RadiusRulerControlLayer extends OsmandMapLayer implements OsmAndCom
 		if (isRulerWidgetOn() && !animatedThread.isAnimatingMapZoom()) {
 			OsmandApplication app = view.getApplication();
 			OsmandSettings settings = app.getSettings();
+			sphericalMap = hasMapRenderer() && settings.SPHERICAL_MAP.get();
 			circleAttrs.updatePaints(app, drawSettings, tb);
 			circleAttrs.paint2.setStyle(Style.FILL);
 			circleAttrsAlt.updatePaints(app, drawSettings, tb);
@@ -423,11 +429,12 @@ public class RadiusRulerControlLayer extends OsmandMapLayer implements OsmAndCom
 			float mapDensity = getMapDensity();
 			boolean updateCache = tb.getZoom() != cacheIntZoom
 					|| !tb.getCenterLatLon().equals(cacheCenterLatLon) || mapDensity != cacheMapDensity
-					|| cacheMetricSystem != currentMetricSystem;
+					|| cacheMetricSystem != currentMetricSystem || cacheSphericalMap != sphericalMap;
 
 			if (!tb.isZoomAnimated() && updateCache) {
 				cacheMetricSystem = currentMetricSystem;
 				cacheIntZoom = tb.getZoom();
+				cacheSphericalMap = sphericalMap;
 				LatLon centerLatLon = tb.getCenterLatLon();
 				cacheCenterLatLon = new LatLon(centerLatLon.getLatitude(), centerLatLon.getLongitude());
 				cacheMapDensity = mapDensity;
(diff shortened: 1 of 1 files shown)


vshcherb:
Need screenshots other latitudes and also screenshots close up with 3d relief different location to test if it's not broken

alex-dev-neo:
Added mid-latitude and equator globe screenshots, plus a Swiss Alps close-up with actual 3D Relief enabled and the terrain heightmap installed. All screenshots use the regular radius-ruler style.

OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
```diff
@@ -458,11 +465,50 @@ private void updateCenter(RotatedTileBox tb, QuadPoint center) {
 
 	private void updateDistance(RotatedTileBox tb) {
 		double pixDensity = tb.getPixDensity();
-		roundedDist = OsmAndFormatter.calculateRoundedDist(maxRadiusInDp / pixDensity, app);
-		radius = (int) (pixDensity * roundedDist);
+		double referenceDistance = maxRadiusInDp / pixDensity;
+		if (sphericalMap) {
+			double globeDistance = getGlobeDistanceForPixelRadius(tb, maxRadiusInDp);
+			if (!Double.isNaN(globeDistance) && globeDistance > 0) {
```
alex-dev-neo: **Unbounded `globeDistance` can hang the UI thread (ANR).**

`globeDistance` is accepted with only `NaN` / `> 0` checks and then passed to `OsmAndFormatter.calculateRoundedDist`, which never terminates for large inputs.

`getGlobeDistanceForPixelRadius` runs two uncorrected iterations of `distance *= pixelRadius / projectedRadius`, and `projectedRadius` is only floored at 1 px, so each iteration can multiply `distance` by up to `pixelRadius` (`map_ruler_width` = 120dp ~ 360 px, 180dp ~ 540 px on large screens) - roughly 1.3e5x over the two rounds. `getGlobePixelRadius` samples with `allowOffscreen = true`, which makes `obtainScreenPointFromPosition` skip the frustum test and project even points behind the globe, where the projection divides by a negative `w` and mirrors them back toward the screen centre. That yields a small `projectedRadius` exactly when the globe is zoomed out - and there the starting distance (`maxRadiusInDp / tb.getPixDensity()`) is already ~1e7 m with the +/-90 degree samples sitting at or behind the limb, so the iteration diverges rather than converging.

`calculateRoundedDist`'s `int generator` overflows at 2^31 and then cycles through values that never exceed the input, so `while (distInMeters * point >= generator)` never exits. Simulating it with Java `int` semantics: it hangs for `distInMeters >= 2.0e9` (metric) and `>= ~6.1e8 m` (miles/feet), and terminates for everything below. This runs inside `onDraw`, so the map freezes.

Suggest capping `globeDistance` at a sane maximum before use - half the earth's circumference is the natural bound, since `greatCircleDestinationPoint` wraps past that and the circle is meaningless anyway.

OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
```diff
@@ -757,10 +821,66 @@ private PointF screenPointFromPoint(double x, double y, boolean compensateMapRot
...
+
+	private double getDistanceForPixelRadius(double pixelRadius, @NonNull RotatedTileBox tb) {
+		return sphericalMap && radius > 0
+				? roundedDist * pixelRadius / radius
+				: pixelRadius / tb.getPixDensity();
+	}
+
+	@Nullable
+	private PointF getRulerPixelFromLatLon(@NonNull RotatedTileBox tb, @NonNull LatLon latLon) {
+		return getRulerPixelFromLatLon(tb, latLon, false);
```
alex-dev-neo: **`allowOffscreen = false` clips circles before the screen edge - behaviour change from the code this replaces.**

The previous call, `NativeUtilities.getElevatedPixelFromLatLon` -> `getElevatedPixelFrom31`, used `getElevatedPointFromLocation(point31, screenPoint, true)` and additionally fell back to tile-box math when the renderer call failed, so it was effectively `@NonNull`. This overload passes `false`, and `AtlasMapRenderer_OpenGL::obtainScreenPointFromPosition` / `obtainElevatedPointFromPosition` return `false` whenever `!checkOffScreen && !globalFrustum2D31.test(position31)`.

`updateText()` deliberately creates circles out to `maxRadius`, the distance from the centre to the farthest screen corner, so the outer circles always leave the viewport. Those samples now return `null`, `drawCircle` flushes and resets the path, and the arc stops at the last in-viewport sample instead of running off the edge - a gap of up to one `CIRCLE_ANGLE_STEP` (5 degrees), which is tens of pixels on the outer circles.

The same `null` propagates through `getPointFromCenterByRadius` / `screenPointFromPoint` into `drawTriangleArrowByRadius` (it returns early if any of its three points is `null`, so the north/heading triangle disappears), `drawCompassCents` and `drawCardinalDirections` (ticks and N/E/S/W letters silently dropped).

Passing `true` here restores the previous behaviour.

OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
```diff
@@ -493,40 +539,45 @@ private void drawRulerCircle(Canvas canvas, RotatedTileBox tb, int circleNumber,
...
+				previousPoint = null;
 				continue;
 			}
-
-			PointF screenPoint = NativeUtilities.getElevatedPixelFromLatLon(getMapRenderer(), tb, latLon);
-			points.add(new QuadPoint(screenPoint.x + getCachedAACanvasOffset().x, screenPoint.y + getCachedAACanvasOffset().y));
-		}
-		if (points.size() > 0) {
-			arrays.add(points);
+			if (previousPoint != null
```
alex-dev-neo: **Discontinuity threshold is a full screen dimension - too coarse to catch the jumps it targets.**

A globe discontinuity (a sample crossing behind the camera or to the far side of the limb) projects with a negative `w` and lands mirrored through the projection centre - typically a few hundred pixels from its neighbour, not a full screen width. On a 1080 px viewport this guard only fires above 1080 px, so such a sample is joined to its neighbour and drawn as a spurious chord across the map.

In the other direction, because this drawing path uses `allowOffscreen = false` (see the comment on `getRulerPixelFromLatLon`), both endpoints have already passed the frustum test, which bounds most legitimate jumps below the threshold - so in the common case the guard never fires at all and the path splitting is inert.

A threshold tied to the expected step size (the circle's pixel radius times the 5 degree step, plus slack) would catch the real cases.

alex-dev-neo (commented):
Self-review of the globe ruler changes. Four findings, one of which I'd fix before merge (the `calculateRoundedDist` hang). Details inline.

One thing I checked and cleared: the `(long)` -> `(int)` narrowing in `calculateGlobePoint31` looks like an overflow for southern polar values (they land in `(2^31, 2^32)` and wrap negative), but it is correct - the native decoder `Utilities::getAnglesFrom31` adds `2^32` back when `p.y < INT32_MIN / 2`, so the wrap *is* the extended-signed encoding the renderer expects. `MAX_GLOBE_MERCATOR_ANGLE = 2 * PI - 1e-7` keeps both caps ~35 units clear of that `-2^30` boundary, and the y31 formula matches `getAnglesFrom31` exactly. Latitudes at +/-90 clamp cleanly rather than producing NaN. It is load-bearing enough to deserve a comment saying so, but it is not a defect.

OsmAnd/src/net/osmand/plus/views/layers/RadiusRulerControlLayer.java
```diff
@@ -423,11 +429,12 @@ private void updateData(RotatedTileBox tb, QuadPoint center) {
 			float mapDensity = getMapDensity();
 			boolean updateCache = tb.getZoom() != cacheIntZoom
 					|| !tb.getCenterLatLon().equals(cacheCenterLatLon) || mapDensity != cacheMapDensity
-					|| cacheMetricSystem != currentMetricSystem;
+					|| cacheMetricSystem != currentMetricSystem || cacheSphericalMap != sphericalMap;
```
alex-dev-neo: **Cache key omits camera tilt, which the new globe calibration depends on.**

`getGlobePixelRadius` measures on-screen pixel distances from the live renderer, and those change substantially with the camera elevation angle (a tilted globe compresses the projected radius). Tilting changes none of `tb.getZoom()`, `tb.getCenterLatLon()`, `getMapDensity()`, the metric system, or `cacheSphericalMap` - so `updateCache` stays `false` and `updateDistance` never re-runs.

The circles themselves still draw correctly, since `drawCircle` derives its distance geodesically (`roundedDist * circleNumber`). But `radius` - the pixel radius used by `calculateTextPosition`, `drawCompassCents`, `drawCardinalDirections` and `getCompassCircleIndex` - keeps the value calibrated at the old tilt, so distance labels, compass ticks and cardinal letters sit off the drawn circle until an unrelated change (zoom, pan, metric switch) invalidates the cache.

Separately: this key uses `tb.getCenterLatLon()` while the calibration uses `getCenterLatLon(tb)` (the renderer-derived elevated centre). Those differ on a tilted 3D map.

PR #25833, opened 2026-09-03
Android Auto: start "Free ride" automatically while driving

### LINKED ISSUE #25783: Automatic Free ride after delay when using Android Auto
### Describe the idea

Adding a setting in the Android Auto interface to have the "Free ride" function automatically activate after a delay of no input (eg driving around with no destination but turning the vehicle on and off making AA reconnect).

### Expected behaviour

After OsmAnd displays in AA, auto-selecting the "Free ride" mode after a pre-defined delay (60secs?)

I've been pressing it manually but it does get annoying when stopping all the time.

### Context

_No response_

vshcherb:
Fixes #25783.

### What it does

While the Android Auto landing screen is on top and no route is calculated, OsmAnd opens the "Free ride" screen (the map) by itself once the vehicle has been moving for a while — no tap on the head unit needed after a reconnect (engine off/on). Always on, no setting.

Constants in `LandingScreen.kt` (easy to tune after testing):

- `AUTO_FREE_RIDE_MIN_SPEED` = 5 km/h — minimal speed to consider the vehicle moving
- `AUTO_FREE_RIDE_DELAY` = 15 s — how long that speed must hold

### Details

- Location updates are only observed while the landing screen is resumed (top of the stack), so browsing POI/Favorites/etc. does not trigger it.
- Any location slower than the threshold (or without speed) resets the timer, so GPS noise while parked can't trigger it.
- Skipped when a route is calculated — that row is "Continue navigation", the user should decide.
- Triggers at most once per Android Auto session, so going back to the landing screen on purpose does not push the user back to the map.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/auto/screens/LandingScreen.kt
+++ b/OsmAnd/src/net/osmand/plus/auto/screens/LandingScreen.kt
@@ -13,22 +13,63 @@ import androidx.car.app.model.Template
 import androidx.car.app.navigation.model.PlaceListNavigationTemplate
 import androidx.core.graphics.drawable.IconCompat
 import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
+import net.osmand.Location
+import net.osmand.plus.OsmAndLocationProvider.OsmAndLocationListener
 import net.osmand.plus.R
 import net.osmand.plus.auto.NavigationSession
 
 class LandingScreen(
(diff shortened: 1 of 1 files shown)


Kreiri:
> Triggers at most once per Android Auto session, so going back to the landing screen on purpose does not push the user back to the map.

I'm not convinced this is desired behaviour. The user asked for this functionality because having to tap "free ride" manually each time was annoying. 

Additionally, the user asked for a toggle in settings to switch this behaviour off or on, but there's no such toggle added by this PR.

PR #25839, opened 2026-09-04
Fix race conditions in Favorites auto-merge uploads

alex-dev-neo:
### Summary

Fixes two post-merge correctness issues in the Favorites auto-merge flow introduced by #25721.

- Preserves conflict state when a Favorites group changes locally while its prepared merge is being uploaded.
- Serializes merged-upload completion so parallel uploads cannot concurrently mutate shared Favorites state.

Related to osmandapp/OsmAnd-iOS#5177.

### Implementation

The prepared merge now retains the timestamp of its common base. If the live group still exists but changes before upload completion, the previous per-file upload timestamp is restored and the existing snapshot is retained. The next backup preparation therefore sees both the local group and cloud group as changed and keeps them in conflict instead of allowing the local state to overwrite the uploaded merge.

If the group is deleted during upload, the successful upload timestamp is retained so the next preparation propagates the deletion instead of restoring the group.

Merged upload completion is protected by a shared lock because Favorites uploads run in parallel while FavoritesSettingsItem.apply() modifies shared FavouritesHelper state.

No changes were made to NetworkWriter, BackupExporter, or the Favorites data model.

### Test plan

- Prepare an automatic merge, edit the same local group while its upload is in progress, and verify the next preparation reports a conflict without losing cloud changes.
- Delete the local group while its prepared merge is being uploaded and verify the next preparation schedules cloud deletion instead of restoring the group.
- Auto-merge two different Favorites groups in one sync and verify both merged results remain present locally and in the cloud.
- Recheck the existing independent-addition, edit, move, rename, and deletion merge scenarios.

Automated tests were not run because Gradle execution is prohibited by the repository instructions for this agent. Static validation with git diff --check passed.

## AI disclaimer

Produced with Codex (GPT-5).

Prompts used (summarised):
1. Analyse the new post-merge review comments on #25721 and determine which should be fixed.
2. Implement only the two correctness fixes concerning in-flight local edits and concurrent merged-upload completion.
3. Commit the changes and create a pull request.
4. Analyse the new review comments on this pull request.
5. Fix the deletion-during-upload edge case, commit it, update the manual test plan, and respond to the specified review comments.

Decided by the agent, not requested explicitly: preserved the original common-base timestamp to force safe conflict detection after an in-flight edit, retained the successful timestamp when the group was deleted during upload, used a shared lock around merged-upload completion, and kept the performance and timestamp-hardening suggestions outside this PR.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
+++ b/OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
@@ -11,6 +11,7 @@ import net.osmand.data.FavouritePoint;
 import net.osmand.plus.OsmandApplication;
 import net.osmand.plus.R;
 import net.osmand.plus.myplaces.favorites.FavoriteGroup;
+import net.osmand.plus.settings.backend.backup.exporttype.ExportType;
 import net.osmand.plus.settings.backend.backup.items.FavoritesSettingsItem;
 import net.osmand.plus.settings.backend.backup.items.SettingsItem;
 import net.osmand.plus.shared.SharedUtil;
@@ -46,10 +47,15 @@ final class FavoritesBackupMerger {
 	// Preparation only changes the in-memory upload payload. It never changes local Favorites.
 	static void prepareMergeUploads(@NonNull OsmandApplication app,
 	                                @NonNull BackupHelper backupHelper,
-	                                @NonNull BackupInfo info) {
+	                                @NonNull BackupInfo info, boolean autoSync) {
 		for (Pair<LocalFile, RemoteFile> conflict : new ArrayList<>(info.filesToMerge)) {
 			LocalFile localFile = conflict.first;
 			if (!(localFile.item instanceof FavoritesSettingsItem localItem)) {
+				continue;
+			}
+			// Merging downloads the remote file, so skip types the backup would filter out anyway.
+			ExportType exportType = ExportType.findBy(localItem);
+			if (exportType == null || !backupHelper.getBackupTypePref(exportType, autoSync).get()) {
 				continue;
 			}
 			try {
--- a/OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
+++ b/OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
@@ -37,6 +37,8 @@ final class FavoritesBackupMerger {
 
 	private static final Log LOG = PlatformUtil.getLog(FavoritesBackupMerger.class);
 	private static final String SNAPSHOT_DIR = "favorites_sync";
+	// Upload callbacks run in parallel, while applying a group mutates shared Favorites state.
+	private static final Object MERGE_FINISH_LOCK = new Object();
 
 	private FavoritesBackupMerger() {
 	}
@@ -64,7 +66,7 @@ final class FavoritesBackupMerger {
 				FavoriteGroup merged = mergeGroups(base, local, remote, defaultColor);
 				if (merged != null) {
 					localFile.item = new MergeUploadItem(app, localItem, local, merged,
-							localFile.localModifiedTime);
+							localFile.localModifiedTime, localFile.uploadTime);
 					info.filesToUpload.add(localFile);
 					info.filesToMerge.remove(conflict);
 				}
@@ -84,7 +86,9 @@ final class FavoritesBackupMerger {
 		}
 		try {
 			if (favoritesItem instanceof MergeUploadItem mergeItem) {
-				mergeItem.finishUpload(fileName, uploadTime);
+				synchronized (MERGE_FINISH_LOCK) {
+					mergeItem.finishUpload(fileName, uploadTime);
+				}
 			} else if (favoritesItem.getLocalModifiedTime() == favoritesItem.getLastModifiedTime()) {
 				saveSnapshot(app, favoritesItem.getSingleGroup(), fileName, uploadTime);
 			} else {
@@ -366,14 +370,16 @@ final class FavoritesBackupMerger {
 		private final FavoriteGroup localGroup;
 		private final FavoriteGroup mergedGroup;
 		private final long sourceModifiedTime;
+		private final long baseSyncTime;
 
 		MergeUploadItem(@NonNull OsmandApplication app, @NonNull FavoritesSettingsItem baseItem,
 		                @NonNull FavoriteGroup localGroup, @NonNull FavoriteGroup mergedGroup,
-		                long sourceModifiedTime) {
+		                long sourceModifiedTime, long baseSyncTime) {
 			super(app, baseItem, Collections.singletonList(mergedGroup));
 			this.localGroup = copyGroup(localGroup);
 			this.mergedGroup = mergedGroup;
 			this.sourceModifiedTime = sourceModifiedTime;
+			this.baseSyncTime = baseSyncTime;
 			setLastModifiedTime(sourceModifiedTime);
 		}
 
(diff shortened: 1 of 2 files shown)


OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
```diff
@@ -391,6 +397,9 @@ && sameGroup(localGroup, current, defaultColor)) {
 				setLocalModifiedTime(sourceModifiedTime);
 				saveSnapshot(app, mergedGroup, fileName, uploadTime);
 			} else {
+				// Preserve the old common base so the next preparation keeps both sides in conflict.
+				app.getBackupHelper().updateFileUploadTime(
```
alex-dev-neo, line 407: The rollback mechanism is right, and I traced the resulting state: on the next `GenerateBackupInfoTask`, `localFile.uploadTime == baseSyncTime`, `localModifiedTime == max(now, T+1) > baseSyncTime` gives `fileChangedLocally`, and the remote `updatetimems == T > baseSyncTime` gives `fileChangedRemotely` — both true, so the pair goes to `filesToMerge` and `L'` never overwrites `M`. Three things that could have broken it and don't: `BackupHelper.uploadFileSync` writes `updateFileUploadTime(type, fileName, T)` at BackupHelper.java:518 *before* firing the listener, so this lands after it; `NetworkWriter.write` runs `uploadItemInfo` afterwards but under the `fileName + ".info"` key, a different row; and `baseSyncTime` can never be 0, because `saveSnapshot` refuses `syncTime <= 0` and `loadSnapshot` demands an exact match, so a `MergeUploadItem` only exists when it is positive. The retained snapshot also stays consistent with the rolled-back time, so the next preparation can retry the merge.

One adjacent edge case inverts, though, and it is not in the test plan. If the user **deletes** the group during the upload window, `current` is nil, both `sameGroup` checks fail, and we reach this branch. On the next preparation there is no `LocalFile`, so loop 1 takes the `localFile == null` path at GenerateBackupInfoTask.java:105: `fileInfo.getUploadTime()` is now `baseSyncTime`, which is `< T`, so the remote file goes to `filesToDownload` and the deleted group is **restored**. Before this change it was `T >= T` and it went to `filesToDelete`, honouring the deletion.

Neither outcome is obviously correct — restoring preserves the other device's changes, deleting honours the user's intent — but it is a silent behaviour inversion, so it is worth making it a deliberate choice rather than a side effect. The iOS port guards exactly this branch with an `if (current)` check (osmandapp/OsmAnd-iOS#5688, `OAFavoritesBackupMerger.mm:437`); Android has no equivalent.

OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
```diff
@@ -84,7 +86,9 @@ static void onUploadSuccess(@NonNull OsmandApplication app, @NonNull SettingsIte
 		}
 		try {
 			if (favoritesItem instanceof MergeUploadItem mergeItem) {
-				mergeItem.finishUpload(fileName, uploadTime);
+				synchronized (MERGE_FINISH_LOCK) {
```
alex-dev-neo, line 95: Confirmed this closes the race I reported: serializing `finishUpload` prevents two `apply()` calls interleaving on `flatGroups` / `favoriteGroups` / `loadFavorites`, which was the data-loss path with `DEFAULT_THREAD_POOL_SIZE = 4` light upload tasks.

No deadlock risk either — `apply()` -> `saveFavoritesIntoFileSync` waits on `singleThreadExecutor`, which never reaches back for this lock, and `SaveFavoritesTask.doInBackground` always counts the latch down, including on exception. Holding the lock across blocking file I/O and a cross-executor `await()` is acceptable here given merged uploads are rare and serializing them is the intent. Leaving the non-merge `else if` / `else` branches outside the lock is also fine: they touch per-group files, and `SaveFavoritesTask.saveLocalGroups` only rewrites groups whose content actually changed, so there is no cross-group mtime churn.

Separate note on coverage: the description says Gradle was not run. That is harmless for compilation — this PR does not touch `mergeGroups`, and `FavoritesBackupMergerTest` only exercises that static method — but it does mean both fixes ship with no automated coverage, and `MergeUploadItem.finishUpload` has none at all. The in-flight-edit path is now load-bearing for data safety and is verified only by a manual test-plan step.

alex-dev-neo (commented):
Reviewed against the post-merge findings on #25721. Both fixes are correct in mechanism, and I traced the resulting state rather than reading them at face value — details in the inline comments.

- The `baseSyncTime` rollback does restore the conflict and stops the un-merged local group overwriting the uploaded merge.
- `MERGE_FINISH_LOCK` does close the concurrent-`apply()` race, with no deadlock against the favorites save executor.

Two things worth resolving before this lands: a behaviour inversion in the delete-during-upload case that falls out of the rollback, and one half of the deferred third finding that is a data-usage bug rather than a performance suggestion.

For cross-reference, the equivalent iOS port (osmandapp/OsmAnd-iOS#5688) still carries both of the issues fixed here, plus an iOS-only one where `copyPoint:` drops point visibility and silently disables auto-merge for every hidden group.

OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
```diff
@@ -64,7 +66,7 @@ static void prepareMergeUploads(@NonNull OsmandApplication app,
 				FavoriteGroup merged = mergeGroups(base, local, remote, defaultColor);
 				if (merged != null) {
 					localFile.item = new MergeUploadItem(app, localItem, local, merged,
-							localFile.localModifiedTime);
+							localFile.localModifiedTime, localFile.uploadTime);
```
alex-dev-neo, line 75: Not a problem with this line — but since this PR consciously scopes out the earlier review notes, worth making sure one of them does not get lost in the process.

The third finding on #25721 was folded into "the performance suggestion" in the description, and half of it is not about performance: `prepareMergeUploads` never checks `backupHelper.getBackupTypePref(ExportType.FAVORITES, autoSync)`. During auto-sync that resolves to `getAutoBackupTypePref()`, which **defaults to false**, so every auto-sync preparation downloads the conflicting Favorites GPX files over the user's connection and then `createItemCollections` filters them straight back out.

That is a data-usage bug with a two-line fix, independent of whether the synchronous `downloadFile` call belongs in `GenerateBackupInfoTask` at all (which is the genuinely architectural half, and reasonable to defer).

OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
alex-dev-neo: (gen by AI)

Thanks for tracing the locking path and confirming that it closes the race without introducing a deadlock. I agree that automated coverage for MergeUploadItem.finishUpload would be valuable. That path currently depends on application Favorites state, filesystem timestamps, and the backup database, while the existing unit test only exercises the pure merge algorithm. To keep this follow-up focused and avoid adding a test-only seam or broader refactoring, I am retaining targeted manual coverage. The test plan now explicitly covers an in-flight edit, an in-flight group deletion, and parallel merged uploads for two groups.

OsmAnd/src/net/osmand/plus/backup/FavoritesBackupMerger.java
alex-dev-neo: (gen by AI)

Thank you. Upstream filtering already prevents disabled Favorites from reaching the normal merge path: local collection, the remote request, and final backup-info filtering all use the same backup-type preference.

I have nevertheless added a direct check in `FavoritesBackupMerger` immediately before merge preparation can download the cloud GPX. Android now passes `autoSync` through and reads `getBackupTypePref(exportType, autoSync)`, so this guard respects the correct automatic or manual backup preference. This is a final-line defense if the preference changes during preparation or if a future caller bypasses the upstream filters.

Implemented in commit 2c0d17e432.

The separate architectural question about synchronous per-conflict downloads remains outside this focused correctness PR.

PR #25844, opened 2026-09-04
Fix HH map selection for countries published as sub-regions (#25841)

### LINKED ISSUE #25841: HH routing selects a neighbouring country's map for a route inside a sub-region-only country (Kyiv → Bukovel picks Moldova/Romania)
### Description

HH routing picks the **wrong country's map** for a route that lies entirely inside another country, whenever that country is distributed only as sub-regions (Ukraine, France, Germany, Poland, Italy, Spain, Russia, GB, US, …).

Reported case: routing **Kyiv → Bukovel** (entirely inside Ukraine) with the Ukraine oblast maps installed plus one Moldova map. HH selects **Moldova only** (or **Romania only**), finds no route, and the app silently falls back to A\*/BRP.

Reproduced with `master` (`0c5b89be13`) using OsmAndMapCreator classes and stock maps from `data.osmand.net`.

### Root cause

`HHRoutePlanner.selectBestRoutingFiles()` groups the candidate HH regions by `(edition, profileParams)` and ranks them by `containsStartEnd` first, `edition` second (`HHRoutePlanner.java:660-680`).

`containsStartEnd` requires `containsStartEndRegion()` (`HHRoutePlanner.java:622-637`), which compares `RoutingContext.regionsCoveringStartAndTargets` against the **OBF routing index names**:

```java
for (RouteRegion index : reader.getRoutingIndexes()) {
    for (String region : regionsCoveringStartAndTargets) {
        if (region.equalsIgnoreCase(index.getName())) {
            return true;
        }
    }
}
```

`regionsCoveringStartAndTargets` is built in `RoutePlannerFrontEnd.calculateRegionsWithAllRoutePoints()` (`RoutePlannerFrontEnd.java:507-526`) as the set of download regions that cover **all** route points. For Kyiv → Bukovel this is:

```
Kyiv    -> ukraine_kyiv_europe, ukraine_europe, ukraine_kyiv-city_europe
Bukovel -> ukraine_europe, ukraine_ivano-frankivsk_europe
regionsCoveringStartAndTargets = [ukraine_europe]
```

The only common covering region is the **country**, but the installed files are named `Ukraine_kyiv_europe`, `Ukraine_ivano-frankivsk_europe`, … — so no name ever matches and `containsStartEndRegion()` returns `false` for the correct group.

There is no `Ukraine_europe_2.obf` download at all (Ukraine ships per-oblast only), so for any cross-oblast route in Ukraine this check is **unsatisfiable by construction**. The same holds for every country published only as sub-regions (France, Germany, Poland, Italy, Spain, Russia, GB, US…) whenever start and finish are in different sub-regions — i.e. for most long routes there.

With `containsStartEnd == false` for every group, ranking degrades to "**newest HH edition wins**", and any neighbouring-country map whose bbox merely touches the start/finish bounding box and whose edition is newer takes the whole route.

The C++ side has identical logic (`core-legacy/native/src/hhRouteDataStructure.h:729`, `hhRoutePlanner.cpp:80`), so Android-native and iOS are affected too.

### Steps to reproduce

Maps (stock, from `https://data.osmand.net/backup-maps/`):

| file | map edition | HH `car` edition |
|---|---|---|
| `Ukraine_kyiv_europe_2.obf`, `Ukraine_zhytomyr…`, `Ukraine_vinnytsya…`, `Ukraine_khmelnytskyy…`, `Ukraine_ternopil…`, `Ukraine_ivano-frankivsk…` | 2026-07-01 | 2026-07-02 |
| `Moldova_europe_2.obf` (2026-09) | 2026-09-01 | 2026-09-02 |
| `Romania_europe_2.obf` (2026-08) | 2026-08-01 | 2026-08-02 |

Route: start `50.4501, 30.5234` (Kyiv), finish `48.3585, 24.4045` (Bukovel).

Driver (java, `RoutePlannerFrontEnd` with `setDefaultHHRoutingConfig()` + `setUseOnlyHHRouting(true)`, `HHRoutingConfig.STATS_VERBOSE_LEVEL = 1`), full source in the collapsed block below.

**A. Ukraine oblast maps only — OK**

```
Selected files: Ukraine_kyiv_europe_2.obf car [], Ukraine_zhytomyr_europe_2.obf car [], Ukraine_vinnytsya_europe_2.obf car [],
                Ukraine_khmelnytskyy_europe_2.obf car [], Ukraine_ternopil_europe_2.obf car [], Ukraine_ivano-frankivsk_europe_2.obf car []
=== RESULT: OK, segments 1555      (719.7 km)
```

**B. Same 6 Ukraine maps + Moldova (2026-09) — BROKEN**

```
Selected files: Moldova_europe_2.obf car []
Loading points...  7,588 - 22.09ms
 Finding first / last segments...
 Routing... finalPnt is null (stop)
=== RESULT: null
=== regionsCoveringStartAndTargets = [ukraine_europe]
```

**C. Same 6 Ukraine maps + Romania (2026-08) — BROKEN**

```
Selected files: Romania_europe_2.obf car []
 Routing... finalPnt is null (stop)
=== RESULT: null
```

**D. The literal user report — "old" Moldova beats older Ukraine maps**

Ukraine oblast maps from 2025-09 (HH edition 2025-09-01) + Moldova from **2026-01** (HH edition 2026-01-02):

```
Selected files: Moldova_europe_2.obf car []
 Routing... finalPnt is null (stop)
=== RESULT: null
=== missingMapsError = To calculate the route maps [ukraine_zhytomyr_europe, ukraine_vinnytsya_europe,
    ukraine_khmelnytskyy_europe, ukraine_ternopil_europe] need to be downloaded
```

An *old* Moldova map is enough — only the relative HH edition matters. Correspondingly, in case B/C nothing at all is reported as missing or outdated: all needed Ukraine maps are installed and current, yet HH still routes on Moldova/Romania alone.

### Group ranking dump

Replicating `selectBestRoutingFiles()` and printing the decision inputs (case B):

```
regionsCoveringStartAndTargets = [ukraine_europe]
>>> SELECTED group #0 params='avoid_toll' hhEdition=Wed Sep 02 17:46:32 EEST 2026
     containsStart=false containsEnd=false containsStartEndRegion=false -> containsStartEnd=false
     files=[Moldova_europe_2.obf]
    ...
             group #4 params='avoid_toll' hhEdition=Thu Jul 02 18:00:50 EEST 2026
     containsStart=true  containsEnd=true  containsStartEndRegion=false -> containsStartEnd=false
     files=[Ukraine_kyiv_europe_2.obf, Ukraine_zhytomyr_europe_2.obf, Ukraine_vinnytsya_europe_2.obf,
            Ukraine_khmelnytskyy_europe_2.obf, Ukraine_ternopil_europe_2.obf, Ukraine_ivano-frankivsk_europe_2.obf]
```

The Ukraine group has real route data at both endpoints (`containsStart/End = true`) and is beaten only because `containsStartEndRegion` is `false`. Neutralising **only** that one term in the same probe makes the Ukraine group rank first again:

```
>>> SELECTED group #0 params='avoid_toll' hhEdition=Thu Jul 02 18:00:50 EEST 2026
     containsStart=true containsEnd=true containsStartEndRegion=true -> containsStartEnd=true
     files=[Ukraine_kyiv_europe_2.obf, … , Ukraine_ivano-frankivsk_europe_2.obf]
```

### Impact

- Fast (HH) routing silently fails for long routes in every sub-region-only country as soon as one neighbouring map is newer than the local ones — a very common state, since users update maps at different times.
- The app then falls back to A\*, which for a 700 km route is dramatically slower, and no message explains why.
- With HH-only usage (`useOnlyHHRouting`) the route simply is not calculated.
- Also explains the "why does it use the France map instead of my English maps" class of reports, cf. #22178 (closed by #22920, which introduced this check).

### Possible directions (for discussion)

1. Match `regionsCoveringStartAndTargets` against the map's **region hierarchy** rather than by exact index name — a `ukraine_kyiv_europe` file does satisfy a `ukraine_europe` requirement.
2. Or make `regionsCoveringStartAndTargets` the set of the **smallest** covering regions per point (the group must cover the region of the start *and* the region of each target), instead of requiring a single region covering all points.
3. Or drop the name check when `contains(start) && contains(end)` already holds on actual route data, and add a hard guard: never rank a group above one that does contain both endpoints.

The same change is needed in `core-legacy/native/src/hhRouteDataStructure.h` (`containsStartEndRegion`).

<details>
<summary>Repro driver source</summary>

```java
public class HHRepro {
    public static void main(String[] args) throws Exception {
        LatLon start = new LatLon(50.4501, 30.5234);  // Kyiv
        LatLon end = new LatLon(48.3585, 24.4045);    // Bukovel
        List<BinaryMapIndexReader> readers = new ArrayList<>();
        for (String a : args) {
            File f = new File(a);
            readers.add(new BinaryMapIndexReader(new RandomAccessFile(f, "r"), f));
        }
        HHRoutingConfig.STATS_VERBOSE_LEVEL = 1;
        RoutePlannerFrontEnd fe = new RoutePlannerFrontEnd();
        RoutePlannerFrontEnd.CALCULATE_MISSING_MAPS = true;
        fe.setDefaultHHRoutingConfig();
        fe.setUseOnlyHHRouting(true);
        RoutingConfiguration.Builder builder = RoutingConfiguration.getDefault();
        int mem = RoutingConfiguration.DEFAULT_NATIVE_MEMORY_LIMIT * 8 * 2;
        RoutingConfiguration config = builder.build("car", new RoutingMemoryLimits(mem, mem), new LinkedHashMap<String, String>());
        RoutingContext ctx = fe.buildRoutingContext(config, null,
                readers.toArray(new BinaryMapIndexReader[0]), RoutePlannerFrontEnd.RouteCalculationMode.NORMAL);
        RouteCalcResult res = fe.searchRoute(ctx, start, end, null, null);
        System.out.println("=== RESULT: " + (res == null ? "null" : (res.isCorrect() ? "OK, segments " + res.getList().size() : "ERROR: " + res.getError())));
        System.out.println("=== regionsCoveringStartAndTargets = " + Arrays.toString(ctx.regionsCoveringStartAndTargets));
    }
}
```

Run: `java -Xmx6g -cp OsmAndMapCreator.jar:lib/* HHRepro <Moldova_europe_2.obf> <Ukraine_*_europe_2.obf ...>`

</details>

### OsmAnd version

`master` @ `0c5b89be13` (OsmAnd-java / core-legacy)

## AI disclaimer

Prompts used (summarised):
1. "There seems to be a backend bug — reproduce it in tools MapCreator or Inspector: I was building HH routing in Ukraine and had an old Moldova map (route ~Kyiv → Bukovel), and instead of the Ukrainian maps it wanted to use Romania or Moldova. Reproduce it, and if reproduced, create a GitHub issue."
2. "Maps are in ../maps"
3. "If you need old maps, download them from https://data.osmand.net/backup-maps/"

Decided by the agent, not requested explicitly: the choice of exact start/finish coordinates (Kyiv 50.4501,30.5234 → Bukovel 48.3585,24.4045), the map sets used for each scenario (Ukraine 2026-07 and 2025-09, Moldova 2026-09 and 2026-01, Romania 2026-08), writing the standalone `HHRepro`/`GroupProbe` drivers against the built OsmAndMapCreator classpath instead of extending an existing test, and the three suggested fix directions.

vshcherb:
Draft — fixes #25841.

## Problem

`HHRoutePlanner.selectBestRoutingFiles()` ranks groups of maps by `containsStartEnd` first and by **newest HH edition** second. `containsStartEnd` requires `containsStartEndRegion()`, which compares `RoutingContext.regionsCoveringStartAndTargets` with the **names of the routing sections** of the downloaded maps.

`regionsCoveringStartAndTargets` was built as the regions covering **all** route points. Routing Kyiv → Bukovel that is:

```
Kyiv    -> ukraine_kyiv_europe, ukraine_europe, ukraine_kyiv-city_europe
Bukovel -> ukraine_europe, ukraine_ivano-frankivsk_europe
=> [ukraine_europe]
```

The only shared region is the country, and there is no `Ukraine_europe_2.obf` to download — Ukraine is published as oblasts only. So the check can never match, `containsStartEnd` is `false` for the correct group, ranking degrades to "newest edition wins", and a Moldova or Romania map that merely clips the corner of the start/finish bbox takes the whole route. The same holds for every country published only as sub-regions (France, Germany, Poland, Italy, Spain, Russia, GB, US…) whenever start and finish are in different sub-regions.

## Change

1. `RoutePlannerFrontEnd` — list the regions of **every** route point, at all levels of the hierarchy, instead of the regions common to all of them. A sub-region map now matches, a country-wide map still matches (the country is in the list), a neighbouring country still does not. Completeness is not lost: `contains(start) && contains(end)` already checks the actual routing data of the group.
2. `HHRoutePlanner` — do not rank a group above one that has routing data at start and end (`containsStartEndPoints` before `edition`). A safety net for the case above: a group that cannot serve the endpoints must never win on edition alone.
3. `RegionsOfRoutePointsTest` — unit test on the region list, no maps needed.

Both (1) and (2) fix the reported case **independently** — measured, see below. (2) is defence in depth; drop it if you consider it redundant.

## Verification

Stock maps, Kyiv `50.4501,30.5234` → Bukovel `48.3585,24.4045`, `car`, HH only, real `RoutePlannerFrontEnd`:

| maps | before | after |
|---|---|---|
| 6 Ukraine oblasts (HH 2026-07-02) | 6 Ukraine files, route 719.7 km | unchanged |
| + Moldova (HH 2026-09-02) | `Selected files: Moldova_europe_2.obf car []` → no route | 6 Ukraine files, route 719.7 km |
| + Romania (HH 2026-08-02) | `Selected files: Romania_europe_2.obf car []` → no route | 6 Ukraine files, route 719.7 km |
| Ukraine oblasts 2025-09 + Moldova 2026-01 | Moldova selected, "missing maps" for 4 oblasts | Ukraine files selected, same missing-maps list |

`regionsCoveringStartAndTargets` after the change:
`[ukraine_europe, ukraine_kyiv_europe, ukraine_kyiv-city_europe, ukraine_ivano-frankivsk_europe]`

`./gradlew :OsmAnd-java:test` — 556 tests, 0 failures. The new test fails on `master`'s logic and passes with the fix.

## Companion PRs

- C++ ranking safeguard: osmandapp/OsmAnd-core-legacy#348
- iOS producer of the same list: osmandapp/OsmAnd-iOS#5726

Both are needed — iOS fills `regionsCoveringStartAndTargets` in `OARouteProvider.mm`, Android-native gets it from this Java code over JNI. The field name and type are unchanged, so the repos stay compatible in any combination.

## Open question for review

An alternative, stricter shape: keep the list **per point** (`String[][]`) and require the group to match a region of *each* point. It is more precise, but it changes the field type and therefore the JNI signature and the iOS producer. Happy to switch if preferred.

## AI disclaimer

Prompts used (summarised):
1. "There seems to be a backend bug — reproduce it in tools MapCreator or Inspector: I was building HH routing in Ukraine and had an old Moldova map (route ~Kyiv → Bukovel), and instead of the Ukrainian maps it wanted to use Romania or Moldova. Reproduce it, and if reproduced, create a GitHub issue."
2. "Maps are in ../maps"; "if you need old maps, download them from https://data.osmand.net/backup-maps/"
3. "Create a draft PR as well, proposing how you'd fix it — Java and C++ — and maybe a simple Java unit test."

Decided by the agent, not requested explicitly: choosing the "regions of every point" shape over the per-point alternative, adding the `containsStartEndPoints` ranking safeguard (not asked for, and not required to fix the reported case), the test coordinates and assertions, splitting the work into three repo PRs, and including the iOS producer change (needed for the fix to reach iOS, which was not mentioned in the prompts).


RZR-UA:
Simplified with regression-free https://github.com/osmandapp/OsmAnd/pull/25944 and https://github.com/osmandapp/OsmAnd-core-legacy/pull/359

Issue #23512, opened 2025-09-18
AIS improvements


vshcherb:
> How to test connection:
> For test purposes use test.osmad.net 168.119.77.47 4001 (udp 10110)

## ToDo

- [ ] Test Simulation Server Location (TCP / UDP)
- [x] Review request - https://github.com/osmandapp/OsmAnd/pull/23986 and for UI changes create a separate comment to approve it. 1) Exclude own position 2) work in background (?) 3) own location - black to test 4) drawShape real dimensions
- [ ] Add setting to parse current location (extra setting to ignore phone locations) from NMEA log
```
!AIVDM,1,1,,A,13HOI:0P0000VOHLCnHQKwvL05Ip,0*23
!AIVDM,1,1,,A,133sVfPP00PD>hRMDH@jNOvN20S8,0*7F
!AIVDM,1,1,,B,100h00PP0@PHFV`Mg5gTH?vNPUIp,0*3B
!AIVDM,1,1,,B,13eaJF0P00Qd388Eew6aagvH85Ip,0*45
!AIVDM,1,1,,A,14eGrSPP00ncMJTO5C6aBwvP2D0?,0*7A
$GPRMC,162751.34,A,5222.2678,N,00453.5506,E,10.50,45.00,180925,,,A*55
```
- [ ] Testing see below and comment once completed!
<details>

<summary> Checklist for Testing once it's completed</summary>
- Allow to edit all fields always - UDP port for example
- Test timeout settings like ship disappears -  constants for server - https://github.com/osmandapp/OsmAnd-tools/blob/master/ais/test-server/server.js and file https://github.com/osmandapp/OsmAnd-tools/blob/master/ais/test-server/AIS-nmea-sample.txt.  
- Test settings for ship approaching (alert) 
- Display how many ship updates & my location updates
</details>

-----------

## Done

- [x] Allow to enter hostname https://github.com/osmandapp/OsmAnd/issues/23400
- [x] Create a simple apk to start UDP server on the device
- [x] https://github.com/osmandapp/OsmAnd/issues/23399 - reconnect
- [x] Screen rotate looses the data https://github.com/osmandapp/OsmAnd/pull/23615
- [x] Code is incorrect it never uses plugin parameters and doesn't update from it
```
    /* after this time of missing AIS signal the object is outdated and can be removed: */
    private static int maxObjectAgeInMinutes = AIS_OBJ_LOST_DEFAULT_TIMEOUT;
[...]

PR #25847, opened 2026-09-04
Vessel tracker (AIS): migrate the plugin screens to Material 3

vshcherb:
Migrates the **Vessel tracker (AIS)** plugin from the preference screen to Material 3, following the design in #23512 (comments of 2026-08-31). This is the first OsmAnd screen on Material 3, so most of the work is the reusable pieces the rest of the migration will use.

## Reusable UI kit — `net.osmand.plus.widgets.ui`

| Component | Layout | Rule it carries |
|---|---|---|
| `GroupHeaderView` | `ui_group_header.xml` | title above a card, on the surface, never inside one; optional trailing icon button |
| `GroupFooterView` | `ui_group_footer.xml` | explanation below the card it belongs to; never truncated |
| `ScreenDescriptionView` | `ui_screen_description.xml` | one intro text per screen, above all cards |
| `MainSwitchView` | `ui_main_switch.xml` | master toggle; when off the content below is hidden, not disabled |
| `SegmentedList` | — | picks the row shape from its position in the group, so hiding a row keeps the group correct |
| `SettingRow` | `item_ui_setting_row.xml` | binds a row of a segmented list |

`OsmandMaterialExpressive{Light,Dark}Theme` (parent `Theme.Material3Expressive.*`) is added next to the existing `OsmandMaterial*Theme`. The expressive theme is what the design needs for the segmented button group and the slider. It deliberately **drops** the app wide `materialSwitchStyle` override, because the design asks for the stock Material 3 switch — the OsmAnd one is tinted with the profile colour.

No new colours: the Material 3 palette is already in `colors.xml`, the screens only use theme attributes.

## Screens

`SettingsScreenType.AIS_SETTINGS` now points at a plain fragment instead of a `PreferenceScreen`; `xml/ais_settings.xml` and the old preference fragment are removed. The preference ids are unchanged, so existing settings survive.

* **Vessel tracker (AIS)** — connection card with its state and action, background switch, live vessel count, entries to the other screens, own MMSI.
* **Connection** — TCP / UDP segmented group, validation on blur and on save, Save enabled only for a valid and changed form.
* **Objects visibility** — "Mark as outdated" and "Hide after" sliders, committed immediately.
* **Collision warning (CPA)** — main switch, MMSI banner, TCPA and CPA sliders.
* **Enter MMSI** — MDC alert dialog; empty clears the MMSI, 1–8 digits is rejected.

The plugin is renamed to **Vessel tracker (AIS)**.

## Behaviour behind the screens

The connection state is real rather than guessed from the arrival of messages: `AisMessageListener` reports connecting / connected / failed through the new `AisConnectionListener`, and the plugin turns that into the six states of the design, including "No data received". Connect / Disconnect is explicit and remembered (`AIS_CONNECTION_ENABLED`), and CPA has its own master switch (`AIS_CPA_ENABLED`) instead of encoding "off" as a zero warning time.

The TCP connect gets a 15s timeout — without it an unreachable host stayed in "Connecting…" forever and the "Connection failed" state of the design was unreachable in practice.

The default NMEA host is empty now, so a fresh install starts in the "Not set up" state instead of pointing at a stranger's LAN address.

**The position from the stream is used now.** The shared library already parsed the position sentences (`NmeaLocationParser`, added by @alex-dev-neo in #25558), but the Android side dropped them — `AisDataManager.onNmeaLocationReceived()` was an empty method. It builds a location from the sentence now and, while the new **Use as location source** setting is on, hands it to OsmAnd through `OsmAndLocationProvider.setCustomLocation()`, the same path the AIDL API uses. Every fix keeps the phone GPS ignored for 10s, so a stream that goes silent hands the position back instead of freezing it, and the plugin prefers that position for its own distance and CPA maths. With the setting on, the group footer gains the second half the design asks for: `Vessels on the map: 39 • Receiving position data`.

This is the ToDo of #23512 ("add setting to parse current location … from NMEA log"); on iOS the same scenario is still failing (scenario 02 of the 2026-08-12 test in OsmAnd-iOS#4844).

## Assets

The vessel illustrations had the water colour baked in. It is removed, so the container can take `waterColor` from the selected map style as the design requires, and the picture matches the map the user is looking at. `MapButtonCard.getMapWaterColor()` is extracted for that. `ic_action_clear_field` is added — the only icon named by the design that did not exist yet.

## Testing

Built and driven on an emulator (Pixel, Android 15) against the OsmAnd test server `168.119.77.47:4001` (TCP):

* connection reaches **Connected • TCP**, the footer counts the vessels live, and the vessels render on the map;
* the Connection screen saves a new host and port and the root screen follows the new state;
* an unreachable host (`153.44.253.27:5631` from OsmAnd-iOS#4844, currently down) cycles **Connecting… → Connection failed**;
* both Objects visibility sliders change value, footer and the dependent "Off" wording; the app bar reset restores the defaults;
* the MMSI dialog rejects 5 digits, accepts 9, and the banner disappears once a MMSI is saved;
* **Use as location source**: with the emulator GPS set to Berlin, the position marker sits on the position of the stream (52.3712 / 5.8151) and points along its course; with the setting off, the same build follows the phone and jumps to Berlin.

`AisVisibilitySettingsTest` (`:OsmAnd-shared:jvmTest`, 6 tests) pins what the two timeouts and the CPA thresholds actually do to a vessel, including that the "Off" stop of "Mark as outdated" can never fire before "Hide after" removes the vessel.

## Not in this PR

* The **Location source** screen of the design is marked WIP in the issue. This PR only adds the switch described above, in the Connection group; a full source picker is left for later.

## AI disclaimer

Prompts used (summarised):
1. Take a task from the AIS area that Dima created recently and that is ready to do, make a branch, build the UI properly and according to Figma, and show the result.
2. Clarified that AIS means the nautical plugin, and then that the task is "AIS improvements" — Android, Material 3, make it all look good.
3. Check that the Figma link opens and that the design can be inspected there.
4. Take all the spacing that the issue does not state from the design itself.
5. Test against the OsmAnd test server.
6. Asked whether the new Material 3 colours in `colors.xml` had been noticed.
7. Send screenshots of the result.
8. Reported from the screenshots: the MMSI dialog is still Material 2, the top insets look wrong, no vessels are visible, the cards of the CPA screen are touching, the button shape and the switch height do not match the design. Asked whether the "parse current location from NMEA" setting exists, asked for the CPA and Objects visibility settings to be verified somehow — a test would do — and asked to check what was recently moved from iOS into the shared Kotlin library.
9. Commit and open a draft pull request.
10. Add the "parse current location from NMEA" setting as well and commit it to the pull request.
11. The screenshots in the pull request are too large.

* Replacing the preference screen instead of restyling it, and pointing `SettingsScreenType.AIS_SETTINGS` at a plain fragment.
* Adding separate Material3Expressive themes rather than changing the parent of the existing Material 3 themes, and dropping the switch restyle only for the new ones.
* Removing the baked water colour from the four vessel illustrations.
* The empty default NMEA host, and the `AIS_CONNECTION_ENABLED` / `AIS_CPA_ENABLED` preferences.
* The 15s TCP connect timeout, and widening the `AisObject.lastUpdate` setter to `internal` so a test can age an object.
* Extracting `MapButtonCard.getMapWaterColor()`, and renaming the plugin to "Vessel tracker (AIS)".
* The default CPA safe distance changed from 1.0 nm to 0.02 nm and the default warning time from 0 to 1 min, to match the values the design describes.
* Placing "Use as location source" as a row of the Connection group. The design has no final place for it — its Location source section is WIP — but the group footer it feeds lives there.
* Dropping the two `%1$s • %2$s` strings this branch had added in favour of the existing `ltr_or_rtl_combine_via_bold_point`.


vshcherb:
Screenshots from the emulator (Pixel, Android 15), light theme, connected to the test server `168.119.77.47:4001` over TCP. (gen by AI)

**Vessel tracker (AIS)**

root screen

**Collision warning (CPA)** and **Objects visibility**

collision warning
objects visibility

**Enter MMSI** — Material 3 dialog, Save disabled until the value is valid and changed

enter mmsi

**Vessels from the test feed on the map**

vessels on the map

**Connection screen against the design** (left: Figma, right: build)

connection vs figma

vshcherb:
Follow-up: **Use as location source** is implemented now. (gen by AI)

The row sits in the Connection group, and with it on the footer gains the position half the design asks for.

root screen with the location source row

Verified with the emulator GPS forced to Berlin (`adb emu geo fix 13.4050 52.5200`) while the plugin is connected to `168.119.77.47:4001`. With the setting on, the position marker sits on the position of the stream and points along its course (the sample `$GPRMC` carries 45°); with the setting off, the same build follows the phone.

location source on vs off

pr_8_root_nmea

pr_7_nmea_location

OsmAnd/res/values/strings.xml
```diff
@@ -11,6 +11,62 @@
 	Strings for OsmAnd core
 	Thx - Hardy
 -->
+	<string name="ais_protocol_tcp">TCP</string>
```
vshcherb, line 14: tcp / udp could be no translation

OsmAnd/res/values/strings.xml
```diff
@@ -11,6 +11,62 @@
...
+	<string name="ais_cpa_warning_time_desc">Highlights vessels in red %1$s before closest approach.</string>
+	<string name="ais_cpa_safe_distance">Safe distance (CPA)</string>
+	<string name="ais_cpa_safe_distance_desc">Highlights vessels in red when projected to pass closer than %1$s.</string>
+	<string name="ais_set_your_mmsi">Set your MMSI</string>
+	<string name="ais_set_your_mmsi_desc">Your own vessel may trigger constant warnings. Set your MMSI so OsmAnd can tell it apart from other ships.</string>
+	<string name="ais_set_mmsi">Set MMSI</string>
+	<string name="ais_enter_mmsi">Enter MMSI</string>
+	<string name="ais_enter_mmsi_desc">Enter 9 digits or leave empty to clear.</string>
+	<string name="ais_error_mmsi_length">Must be 9 digits</string>
+	<string name="ais_nautical_miles_short">%1$s nm</string>
```
vshcherb, line 65: looks like non conventional nm / min we already have - other strings need to be checked as well for duplicate

OsmAnd/res/values/osmand_material_light_style.xml
```diff
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8" standalone="no"?>
 <resources xmlns:android="http://schemas.android.com/apk/res/android">
 
-	<style name="OsmandMaterialLightTheme" parent="Theme.Material3.Light">
+	<!-- Material 3 Expressive: the same colour system as Material 3 plus the component styles the
+	     Material 3 screens are designed with (connected button groups, sized sliders, expressive
+	     buttons and dialogs). Screens that need the stock M3 switch set its style explicitly. -->
+	<style name="OsmandMaterialLightTheme" parent="Theme.Material3Expressive.Light">
```
Chumva, line 7: (gen by AI) The description says `OsmandMaterialExpressive{Light,Dark}Theme` were added *next to* the existing themes, but this changes the parent of the existing `OsmandMaterialLightTheme` (and the dark one) to `Theme.Material3Expressive`. These themes are also used by `AstroContextMenuFragment`, `StarMapSearchDialogFragment`, `AddCoordinateFormatFragment`, `BaseCoordinateFormatFragment`, `AstroConfigureViewBottomSheet` and every `BaseMaterialBottomSheetDialogFragment`.

In material 1.14.0 `Base.Theme.Material3Expressive.Light` overrides `appBarLayoutStyle`, `collapsingToolbarLayoutStyle`, `toolbarStyle`, `actionBarSize`, `materialSearchBarStyle`, `materialSearchViewStyle`, the dialog themes, all button styles, sliders and progress indicators - and the astronomy and coordinate-format screens use AppBarLayout, CollapsingToolbar, SearchBar and SearchView, so this line re-themes them. Either add the separate theme the description talks about, or add before/after screenshots of those screens.

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerPlugin.java
```diff
@@ -321,15 +349,197 @@ public int getVesselLostTimeoutInMinutes() {
...
+	public boolean isCpaEnabled() {
+		return AIS_CPA_ENABLED.get();
+	}
+
+	/**
+	 * Turns a position sentence of the NMEA stream into a location and, while the user asked for
+	 * it, feeds it to the app as the current position. Every fix keeps the phone GPS ignored for
+	 * a short while, so a stream that goes silent hands the position back instead of freezing it.
+	 */
+	private void onNmeaLocationReceived(@NonNull AisLocation location) {
```
Chumva, line 505: (gen by AI) Every `PositionSentence` becomes a `Location` here. GGA and GLL carry no speed or course, RMC does (`NmeaLocationParser` sets `hasSpeed` / `hasBearing` accordingly), so with a feed that interleaves them - most GPS receivers and AIS gateways do - the app location alternates `hasBearing()` true/false several times a second and `PointLocationLayer.getBearingToShow()` flips between the course arrow and the plain dot. The test server only sends `$GPRMC`, so it does not show there.

Suggestion: keep the last speed and bearing when a sentence lacks them (within 2-3 s), and throttle to at most 1 Hz - every fix runs the whole `setLocation()` pipeline (track recording, routing, plugins).

Chumva (commented):
Review summary (gen by AI) - details are in the inline comments.

1. **The theme change re-themes existing screens.** The description says separate `OsmandMaterialExpressive*` themes were added next to the existing ones; the diff changes the parent of the existing `OsmandMaterial{Light,Dark}Theme` to `Theme.Material3Expressive`, which also restyles the astronomy, star-map search and coordinate-format screens.
2. **"Use as location source" jitters with a real feed.** GGA/GLL sentences carry no course, RMC does, so the location marker alternates between the course arrow and the dot several times a second.
3. **The profile selection is ignored.** The screens read and write the active profile even when opened from another profile's "Configure profile"; the old preference screen wrote to the selected one.
4. **Existing installs on the default connection values** end up "Not set up" with no socket after the upgrade.
5. **The connection action flips on a timer** while a host is unreachable (Cancel / Connect every 10-15 s), and the toolbar reset wipes every plugin setting without confirmation.

Already on the branch: the CPA migration registered `ais_cpa_warning_time` with a default of 0 that the plugin then inherited for the first start after the upgrade - fixed in e4a6e1a279. The default CPA distance moving from 1 nm to 0.02 nm for users who had the warning on is left as it is and documented in the migration (ed96180723).

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerPlugin.java
```diff
@@ -321,15 +349,197 @@ public int getVesselLostTimeoutInMinutes() {
...
+	public void removeConnectionStateListener(@NonNull AisConnectionStateListener listener) {
+		connectionStateListeners.remove(listener);
+	}
+
+	/**
+	 * The connection is set up once the user saved it on the Connection screen at least once -
+	 * a user of the old preference screen has one of these preferences set as well - and, for
+	 * TCP, named a host. Until then the plugin does not open a socket on its own.
+	 */
+	public boolean isConnectionConfigured() {
```
Chumva, line 378: (gen by AI) Upgrade behaviour: an existing install that never stored one of these four preferences - UDP on the default port 10110, or TCP with the old default host `192.168.200.16` - is "Not set up" after the upgrade and opens no socket until the user saves the Connection screen once. Probably acceptable, but the description only covers fresh installs; worth a line there.

OsmAnd/src/net/osmand/plus/plugins/aistracker/fragments/AisTrackerSettingsFragment.kt
```diff
@@ -0,0 +1,290 @@
...
+		connectionRow.findViewById<View>(R.id.connection_content).setOnClickListener {
+			showFragment(AisConnectionFragment())
+		}
+
+		nmeaLocationRow = SettingRow(view.findViewById(R.id.nmea_location_row))
+		nmeaLocationRow.setIcon(null)
+		nmeaLocationRow.setTitle(R.string.ais_use_nmea_location)
+		nmeaLocationRow.setSubtitle(getString(R.string.ais_use_nmea_location_desc))
+		nmeaLocationRow.setOnClickListener {
+			plugin.AIS_USE_NMEA_LOCATION.set(!plugin.AIS_USE_NMEA_LOCATION.get())
```
Chumva, line 92: (gen by AI) The screens read and write `plugin.AIS_*` through the *active* profile. Opened from Settings > another profile > Plugin settings, `ConfigureProfileFragment` passes `APP_MODE_KEY` and does not switch the active mode (only the `sepAppModeToSelected()` paths do); the old `BaseSettingsFragment` wrote to the selected profile through `applyPreference()` / `getSelectedAppMode()`. `BaseMaterialFragment.currentAppMode` is restored from the arguments but never used here, and "Apply to all profiles" is gone. If that is intended it should be stated in the description; otherwise the rows need `getModeValue(currentAppMode)` / `setModeValue(currentAppMode, ...)`.

OsmAnd/src/net/osmand/plus/plugins/aistracker/fragments/AisTrackerSettingsFragment.kt
```diff
@@ -0,0 +1,290 @@
...
+				} else {
+					vessels
+				})
+		}
+	}
+
+	private fun onConnectionAction(state: AisConnectionState) {
+		when (state) {
+			AisConnectionState.NOT_SET_UP -> showFragment(AisConnectionFragment())
+			AisConnectionState.CONNECTING, AisConnectionState.CONNECTED -> plugin.disconnect()
```
Chumva, line 247: (gen by AI) With an unreachable host the state cycles CONNECTING (15 s, button "Cancel") > FAILED (10 s, button "Connect") > CONNECTING ... forever (`CONNECT_TIMEOUT_MS` plus the 10 s reconnect delay in `AisMessageListener`), so the button under the finger changes meaning on a timer - and "Cancel" sets `AIS_CONNECTION_ENABLED` to false, which stops the retries for good. Consider showing "Connection failed" with a retry hint and keeping one action while the listener retries.

OsmAnd/src/net/osmand/plus/plugins/aistracker/fragments/AisTrackerSettingsFragment.kt
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	private fun protocolName(): String = getString(
+		if (plugin.AIS_NMEA_PROTOCOL.get() == AisTrackerPlugin.AIS_NMEA_PROTOCOL_TCP) {
+			R.string.ais_protocol_tcp
+		} else {
+			R.string.ais_protocol_udp
+		})
+
+	private fun resetPluginSettings() {
```
Chumva, line 281: (gen by AI) One tap on the always-visible toolbar icon resets every plugin setting, host and port included, with no confirmation and no undo. The existing settings screens confirm a reset through `ResetProfilePrefsBottomSheet`; the same is worth doing here - the per-screen resets on the slider screens are cheap to redo, this one is not.

OsmAnd/res/values/osmand_material_light_style.xml
0xRe1nk0: (gen by AI) Right, the description was written before this change and is outdated: the separate `OsmandMaterialExpressive*` themes were dropped on purpose. They were 110-line copies of `OsmandMaterial*Theme` that differed only in the parent and the switch style, so every colour token would have had to be kept in sync in four files. The design asks for Material 3 Expressive as the base for the Material 3 screens, so the existing themes now inherit it, and the AIS layouts pin the stock M3 switch instead of dropping `materialSwitchStyle` from the theme.

Before / after on the screens that use these themes (same build, only the parent changed; the screens were compared pixel by pixel with the status bar cropped out):

| Screen | Difference |
|---|---|
| Coordinate format | none |
| Coordinate format > Add format (SearchBar) | none |
| Add format > search (SearchView) | none |
| Star map > search > Solar system (CollapsingToolbarLayout) | none |
| Star map > search (SearchBar) | the search icon moved ~4 px to the start (expressive `contentInsetEnd`) |
| Star map > Moon context menu | none apart from the live sky and values |
| Star map > Configure view | none apart from the live sky |

Not compared on a device: the `MaterialAlertDialogBuilder` dialog in `StarMapFragment`, the purchase / discount bottom sheets and Edit format. The expected difference there is the expressive button style (label up to two lines in dialogs, shape morph on press).
theme_add_format
theme_add_format_search
theme_astro_configure_view
theme_astro_context_menu
theme_coordinate_format
theme_star_map_search
theme_star_map_solar_system

The description of the PR is updated accordingly.

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerPlugin.java
0xRe1nk0: (gen by AI) Confirmed and fixed in #26219. The last speed and course are now kept for 3 s when a sentence does not carry them, and the location is handed to `setCustomLocation()` at most once a second; the plugin's own position (CPA, the footer) still takes every sentence. Checked on the emulator with a UDP feed that interleaves `$GPRMC` (course 45°) and `$GPGGA` 300 ms apart: the marker stays the course arrow in every frame.
nmea_marker_rmc_gga

OsmAnd/src/net/osmand/plus/plugins/aistracker/AisTrackerPlugin.java
0xRe1nk0: (gen by AI) Agreed that silently losing the connection is too much for users who were on UDP 10110 with the defaults - that is the standard NMEA port, so for many it was a working setup. The follow-up (#26219) adds migration 5.4-03: when the plugin is enabled, the protocol is stored for every profile that has none of the four connection preferences, so such a user stays set up. TCP with the old default host `192.168.200.16` is deliberately not carried over (it was somebody else's LAN address) and ends up "Not set up"; this is stated in the migration and in the description. Checked by removing the connection preferences, setting the installed version back to 5402 and upgrading: the screen shows "Connected • UDP".
upgrade_default_udp_stays_connected

OsmAnd/src/net/osmand/plus/plugins/aistracker/fragments/AisTrackerSettingsFragment.kt
0xRe1nk0: (gen by AI) Fixed in #26219, the screens behave like the old preference screen again: every row, the Connection screen, the sliders, the MMSI dialog and the resets read and write the profile the screen was opened for (`currentAppMode`, restored from `APP_MODE_KEY`), and the units of the distance follow that profile as well. The socket follows the active profile only, so it is restarted only when the edited profile is the active one. For another profile the Connection row shows "Not set up" or "Not connected" without the Connect / Disconnect button, since there is no socket to act on. "Apply to all profiles" is not brought back in this change.

Checked: with Browse map active, the MMSI saved from Driving > Configure profile > Vessel tracker lands in `net.osmand.settings.car.xml`, Browse map is untouched.

OsmAnd/src/net/osmand/plus/plugins/aistracker/fragments/AisTrackerSettingsFragment.kt
0xRe1nk0: (gen by AI) Fixed in #26219 without new strings: after the first failure the state stays "Connection failed" with the Connect action while the listener keeps retrying in the background, and turns into "Connected" as soon as a retry succeeds. Connect restarts the listener, so the user sees "Connecting…" / Cancel again only for an attempt they started. Frames at about 2, 18, 30 and 45 s with an unreachable host (the retry at 25 s is running during the third one):

OsmAnd/src/net/osmand/plus/plugins/aistracker/fragments/AisTrackerSettingsFragment.kt
0xRe1nk0: (gen by AI) Added in #26219: the reset of the whole plugin asks first (Material 3 dialog, existing strings `ais_reset_plugin_settings`, `reset_confirmation_descr`, `shared_string_reset`). `ResetProfilePrefsBottomSheet` was not reused because its texts are about the whole profile. The resets of the slider screens stay immediate, as suggested.

reset_confirmation

0xRe1nk0:
@Chumva (gen by AI) Thanks - all five points are addressed in #26219 (theme: screenshots and an updated description, no code change). Also thanks for e4a6e1a279 and 6853e952a9: the self-registering `IntPreference` in the migration and the Kotlin default arguments that Swift does not see were both missed in our review.

PR #25243, opened 2026-06-09
Add "Dashed line" track appearance option

rustyqt:
### What
Adds a per-track **Dashed line** option to the track *Appearance* editor. When enabled, the track is rendered as a dashed line instead of solid.

### Why
A simple, frequently-wished way to visually distinguish tracks (e.g. planned vs. recorded, variants) without changing colour or width.

### How
Mirrors the existing `SHOW_ARROWS` appearance option end-to-end:
- New `LINE_DASHED` `GpxParameter` (boolean):
  - persisted in the gpx database — `DB_VERSION` 36 → 37; the generic `onUpgrade` column-add migrates both the gpx and gpx-dir tables,
  - written to / read from the GPX file `<extensions>` as `line_dashed` (`GpxFile.isLineDashed()/setLineDashed()`).
- `DashedLineCard` (`BaseSwitchCard`) added to `TrackAppearanceFragment`, next to "Direction arrows". Persisted on save for saved tracks and for the current recording track (`CURRENT_TRACK_LINE_DASHED`).
- `TrackDrawInfo` carries the flag; `GpxAppearanceHelper.isLineDashedForTrack(...)` resolves it (editor → current track → DB → GPX file).
- Rendering reuses the existing dash pipeline: `GPXLayer` supplies a width-proportional dash pattern to `RenderableSegment.drawGeometry(...)`, which already feeds `VectorLineBuilder.setLineDash(...)` for the OpenGL core. A small `Renderable.setDashed()` change-detector (mirroring `setDrawArrows`) makes a toggle rebuild the geometry.

### Notes
- Single colour for now (an alternating second dash colour is a natural follow-up).
- Only the English string is added; translations go through Weblate.
- Built and verified on-device (`androidFull` / `opengl` / `arm64`).
- Provided under the project's licence (MIT, per `PULL_REQUEST.MIT.LICENSE`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #25852, opened 2026-09-04
Add dashed/dotted line style for tracks (incl. per-folder default appearance)

### LINKED ISSUE #9953: Feature request dashed line for tracks
The users ask to add a dashed line appearance for tracks

Corwin-Kh:
## Summary

Adds a dashed/dotted **line style** option for GPX tracks (Solid / Dashed / Dotted), addressing #9953:

- Line style selector in the single-track "Change appearance" screen.
- Line style now available in the multi-track **Change appearance** dialog and the per-folder **Default appearance** dialog, with the same Unchanged/Original semantics as Color and Width.
- Dash/dot pattern is applied consistently in both the OpenGL (v2) and legacy (v1) map renderers.
- A track with no line style of its own adopts its folder's default line style when moved into that folder (existing generic appearance-adoption mechanism — verified for this parameter specifically).

## 🔒 Blocking behavior for unsupported color/line-style combinations

**1. Single-track "Change appearance" — Line style locked to Solid when color is non-solid (e.g. Speed)**
FINAL_track_appearance_linestyle_blocked

**2. Single-track "Change appearance" — Color locked to solid colors when line style is Dashed/Dotted**
FINAL_track_appearance_color_blocked

**3. Per-folder "Default appearance" — Line style locked to Solid when default color is non-solid (e.g. Speed)**
FINAL_default_appearance_linestyle_blocked

**4. Per-folder "Default appearance" — Color locked to solid colors when default line style is Dashed**
FINAL_default_appearance_color_blocked

## 🤖 AI disclaimer

Part of this branch was developed with AI assistance (Claude Code). Sessions were used for:

- Reviewing the initial dashed-line implementation (from PR by @rustyqt) for correctness before building on it.
- Implementing the `LineStyleCardController` and wiring it into the multi-track "Change appearance" and per-folder "Default appearance" dialogs.
- Diagnosing and fixing a bug where line style was not applied in the legacy (v1, non-OpenGL) map renderer — the dash pattern was computed but never reached the `Paint` used for drawing.
- Diagnosing a separate, pre-existing bug in the "move to folder" flow (a stale in-memory tracks-list reference), and independently confirming — through direct device testing — that the DB-level "adopt folder default on move" mechanism itself works correctly.
- Adding an instrumented Espresso UI test (`DefaultAppearanceLineStyleTest`) covering the popup selector, DB persistence, and the value the renderer would use.
- All AI-assisted changes were reviewed, tested, and committed by the developer; nothing was pushed automatically.

## Key requests from the developer (prompts that drove this work)

- "Check my code for mistakes" — review of the dashed-line diff before commit.
- Implement the unchecked checklist items from #9953: line style in Change/Default appearance dialogs, with Unchanged/Original states.
- "When new line style selected different from original, line style card should become visible, like for line width" — refine the card to show the inline Solid/Dashed/Dotted toggle once a concrete value is picked.
- Explain/verify: does moving a track into a folder apply the folder's default appearance? (Answered and verified on-device.)
- "For v1 (legacy) rendering, line style is not applied for the track line. Fix it." — root-caused and fixed.
- "Is it possible to write a UI test for this?" — yes; wrote and ran a real instrumented test end-to-end on an emulator until green.

## Screenshots

**1. Default appearance dialog — new "Line style" row (below Width)**
Default appearance dialog with Line style row

**2. Line style selector popup — Original / Solid / Dashed / Dotted**
Line style selector popup

**3. "Dashed" selected — inline Solid/Dashed/Dotted toggle becomes visible**
Dashed selected with inline toggle visible

**4. A track with no line style of its own, moved into a folder whose default is "Dashed" — rendered dashed on the map**
Track rendered with dashed line style


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxDatabase.kt
```diff
@@ -20,7 +20,7 @@ class GpxDatabase {
 	companion object {
 		val log = LoggerFactory.getLogger("GpxDatabase")
 
-		const val DB_VERSION = 36
+		const val DB_VERSION = 38
```
Chumva: Why add +2 for db update?

Corwin-Kh: there were 2 db update during development, not needed in clean merge. missed on review, fixed

OsmAnd/res/values/strings.xml
```diff
@@ -11,6 +11,15 @@
...
 -->
+	<string name="gpx_line_style_desc_unavailable_for_color">Dashed and dotted line styles are not available for non-solid track color.</string>
+	<string name="gpx_color_unavailable_for_line_style">This coloring is not available for a non-solid line style.</string>
+	<string name="gpx_color_desc_unavailable_for_style">Only solid track color is available for a dashed or dotted line style.</string>
+	<string name="gpx_line_style">Line style</string>
+	<string name="gpx_line_style_solid">Solid</string>
+	<string name="gpx_line_style_dashed">Dashed</string>
+	<string name="gpx_line_style_dotted">Dotted</string>
+<!--	<string name="gpx_dotted_line">Dotted line</string>-->
+<!--	<string name="gpx_dashed_line">Dashed line</string>-->
```
Chumva, line 22: Is it needed?

rustyqt: No, not needed — they can go.

`gpx_dashed_line` came from my original commit 423ea49 (the first dashed-line implementation in #25243) and was superseded by the `gpx_line_style_*` set. Neither ID is referenced anywhere in the branch, and since #25243 never merged, neither ever reached `master` or Weblate — so removing them has no translation impact.

I've opened #26131 against this branch, which drops both lines and merges current `master` in. `strings.xml` turned out to be the only conflicting file across the gap — both sides had added strings at the top — so that one PR clears this comment and the merge conflict together.

Chumva:
Issues found on the current head (a79ecc4), reproduced on a Xiaomi 2405CRPFDG / Android 16 with `gplayFreeOpenglArm64Debug` from this branch installed over a master (DB v36) build (gen by AI):

1. **Regression, multi-track "Change appearance":** `onAppearanceChanged()` now calls `colorCardController.refreshContent()` on every `setParameter()`; the rebind re-runs `updatePalette(category, DEFAULT_NAME)`, which re-selects the default gradient palette and fires `onPaletteItemSelected` → `COLOR_PALETTE` goes back to `default`. Reproduced: Speed + tap a custom palette → selection reverts. Refresh the cards only when `COLORING_TYPE`/`LINE_STYLE` changed.
2. **DB upgrade:** the appearance triggers are not recreated for v37 (last time in `if (oldVersion < 36)`), so on upgraded DBs `line_style` is missing from `AFTER UPDATE OF …` and a line-style-only change does not bump `appearanceLastModifiedTime` → cloud backup never sees it. Reproduced on a 36→37 upgraded DB. Needs an `if (oldVersion < 37)` drop/recreate block.
3. **Legacy renderer, "Dotted":** `Renderable.updateLocalPaint` now takes the cap from the style paint; styles without `cap="ROUND"` in the `gpx` attribute (Touring view, custom styles) get `Cap.BUTT`, and `{0.1f, w}` renders as invisible 0.1 px slivers; with ROUND caps the dots touch (period = w + 0.1). Also changes the plan-route dashed line to butt caps. Force `Cap.ROUND` in `getLineStylePaint` and tune the intervals.
4. **Folder default ignored by the single-track screen:** `TrackDrawInfo.updateParams` calls `getLineStyleTypeForTrack(gpxFile, item, null)` (no dirItem) while width/colour/arrows resolve the folder. A track inheriting "Dashed" opens as "Solid" and Apply writes `solid`. Same in `GpxUiHelper.addAppearanceToGpx`.
5. **Gradient + dashed is reachable and not guarded:** move-into-folder copies the folder's `line_style=dashed` into a track that keeps its gradient colouring; `GPXLayer` doesn't check `coloringType`, and OpenGL drops colorization for dashed lines, so the track renders as a single-colour dashed line (verified: same track with `solid` shows the altitude gradient). Add `if (!coloringType.isTrackSolid()) lineStyleType = SOLID` in `GPXLayer`.

Minor: `LINE_STYLE` default `"solid"` makes `null` unsupported on `GpxDataItem`, so "Original" can't clear a stored value; the tests leave `line_style_test_*` folders behind and use package `ui.tracks` next to `ui.track`.

PR #25854, opened 2026-09-04
Quick action to show / hide individual tracks (#24990)

### LINKED ISSUE #24990: Quickly toggle one or more GPX files (tracks) on/off
### Describe the idea

I increasingly miss the ability to quickly toggle one or more GPX files (tracks) on/off. I find this very important. Please implement this soon. THANK YOU

### Expected behaviour

see text

see text

### Context

_No response_

vshcherb:
Closes #24990.

### Problem

The only track-related quick action, `Show / Hide - Tracks`, toggles **all** tracks at once (hide everything / restore the previous selection). There is no way to switch a single track, or a small set of tracks, on and off from the map.

### Solution

A new editable quick action in the **Configure map** category with two modes, following the pattern of `Simulate - Location by GPX`:

* **Always ask** (default) - tapping the action opens the tracks dialog, where one or more tracks can be toggled and applied.
* **Select** - one or more tracks are picked when the action is created; a single tap then shows them all, and the next tap hides them again. The button label and icon switch between `Show - <track>` and `Hide - <track>`, and the action name is auto-generated from the chosen tracks (`Kyiv_Loop +1`).

Tracks whose files no longer exist are ignored, and the currently recording track is supported as well.

The existing action that toggles everything is renamed from `Tracks` to `All tracks`, so the two entries can be told apart in the action list.

| Action list | Always ask | Select |
|---|---|---|
|  |  |  |

### Changes

* `ShowHideTracksAction` - new quick action (id 82, `tracks.showhide`).
* `quick_action_show_hide_tracks.xml` - editing UI.
* `MapButtonsHelper`, `QuickActionIds` - registration.
* `ShowHideGpxTracksAction` - name changed to `All tracks`.
* 4 new strings.

### Testing

Manually on an Android 15 emulator (Pixel Tablet + phone, `androidFull-legacy-arm64` debug build) with three test GPX files:

* the action appears in `Add action -> Configure map` next to `All tracks`;
* **Select**: adding / removing tracks updates the list and the auto-generated title; the first tap shows both tracks on the map, the second hides them (verified against the map and the `selected_gpx` preference: `[]` -> both files -> `[]`); the label flips between `Show - Kyiv_Loop +1` and `Hide - Kyiv_Loop +1`;
* **Always ask**: the tap opens the tracks dialog with per-track checkboxes;
* saved parameters are restored correctly when the action is reopened for editing.

Both selected tracks on the map after a single tap of the button:

## AI disclaimer

Prompts used (summarised):
1. Try to fix https://github.com/osmandapp/OsmAnd/issues/24990.
2. Start another emulator (the first one was busy with another task).

- The two-mode design (`Always ask` / `Select`) was taken from the screenshot in the issue comment; `Always ask` reuses the existing tracks dialog, `Select` stores a list of track paths in the action parameters.
- A separate action was added instead of making the existing `Show / Hide - Tracks` editable, so that several per-track buttons can coexist (non-editable actions can only be added once).
- Renaming the existing action to `All tracks` to keep the two entries distinguishable.
- Wording of the four new strings, and the choice to store absolute track paths as a JSON array (same approach as `ShowHidePoiAction`).

### DIFF
--- a/OsmAnd/res/layout/quick_action_show_hide_tracks.xml
+++ b/OsmAnd/res/layout/quick_action_show_hide_tracks.xml
@@ -1,0 +1,117 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+	xmlns:app="http://schemas.android.com/apk/res-auto"
+	xmlns:tools="http://schemas.android.com/tools"
+	android:layout_width="match_parent"
+	android:layout_height="match_parent"
+	android:orientation="vertical">
+
+	<include layout="@layout/list_item_divider" />
+
+	<LinearLayout
+		android:layout_width="match_parent"
+		android:layout_height="wrap_content"
+		android:background="?attr/list_background_color"
+		android:orientation="vertical"
+		android:paddingBottom="@dimen/content_padding_small">
+
+		<net.osmand.plus.widgets.TextViewEx
+			android:layout_width="match_parent"
+			android:layout_height="wrap_content"
+			android:layout_marginStart="@dimen/content_padding"
+			android:layout_marginTop="@dimen/content_padding_small"
+			android:layout_marginEnd="@dimen/content_padding"
+			android:letterSpacing="@dimen/text_button_letter_spacing"
+			android:text="@string/shared_string_tracks"
+			android:textColor="?android:textColorPrimary"
+			android:textSize="@dimen/default_list_text_size"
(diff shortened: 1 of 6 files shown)


Kreiri:
From my point of view as a user of the app, I find names "all tracks" and "tracks" for these actions rather confusing. Description in "All tracks" action also says "A button to show or hide selected tracks on the map", which sounds very much like "select" mode of the "Tracks" action.

Screenshot 2026-09-07 at 18 02 03
Screenshot 2026-09-07 at 18 05 01

Issue #23336, opened 2025-08-25
Android Auto widgets panel (OBD II)


hkusulja:
### Describe the idea

As you can show speedometer on Android Auto, i suggest that you can customize this and show other vehicle obd II parameters, like fuel consumption etc.

### Expected behaviour

on android auto to se vehichel obd II metrics, such as fuel consumption

### Alternatives you've considered

disconnect from big android auto screen and use mobile phone screen like in old days :( :(

vshcherb:
For prototype we can display fixed area with all widgets if screen is big enough

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/647c3d8d-aa0f-4c01-81b7-955d98c688f1"  width="300" height="auto"/> </td>
  </tr>
</table>   

- [ ] Show one panel
- [ ] Use Layouts from Android, support all 3 sizes

hkusulja:
This should be in app settings, what to display and what not

PR #25856, opened 2026-09-04
Android Auto: generic widgets panel over the map (prototype)

vshcherb:
Prototype for #23336 — an OBD II (and any other) widgets panel on the Android Auto screen.

## What it does

Instead of a car-specific implementation per widget, the panel reuses the regular OsmAnd map widgets and draws their views on the car surface, so **any** widget works — OBD II speed / RPM from the Vehicle metrics plugin as well as speed, altitude, GPS info, time, battery, max speed.

- `CarWidgetsPanel` measures, lays out and draws the widget views on the car canvas. The views are created and owned by the panel and never attached to a window, so the phone UI is untouched.
- The widget set is stored per profile in `AA_WIDGETS`, independent of the phone panels.
- The panel is flush with the right edge, under the speedometer, so the navigation card on the left stays free. Rounded on the left, square on the right.
- The panel never moves: its top only depends on the speedometer (always present while driving); rows that the transient alarm widget covers are hidden and their slot is kept, instead of pushing the panel down.
- `SpeedometerWidget` compensates the `* 0.77` factor it applies for Android Auto, which left it much smaller than the alarm widget next to it — on the phone the two are about the same size.

Sizing is done in **car dp**, not in a share of the pixel width: the panel is 200 dp wide (~1.5× the 130 dp phone widget, the car screen is looked at from roughly twice the distance), clamped to 22 % of the visible area so it still fits a 800×480 head unit.

## Verified

Tested on a real phone (Pixel 9 Pro XL, `net.osmand.dev`) driving the Desktop Head Unit at 1920×1080, 1280×720 and 800×480. Six widgets render correctly under the speedometer at 1280×720.

## Not done yet

1. **Widget selection must move into the phone app.** This prototype adds a "Widgets" screen to the Android Auto settings (`CarWidgetsScreen`); per review feedback the configuration belongs in the app, next to the other widget settings, and the Android Auto screen should be dropped.
2. **Paging must be removed.** The current prototype pages through the widgets on a tap (`SurfaceCallback.onClick`, car API level 5). The decision is to drop it and simply cut off the widgets that do not fit — this also removes the open question of whether a tappable surface area passes Google review.
3. Requires a live `MapActivity` on the phone: `MapWidget` needs one to inflate its views, so with the phone activity destroyed the panel stays empty. Needs a headless widget factory (or keeping the widgets alive with the car session) before this is shippable.
4. Not measured: the drawing cost per frame (`measure`/`layout`/`draw` of every row on every rendered frame) and the behaviour on a real head unit — everything above is from the DHU.
5. No tests.

## AI disclaimer

Prompts used (summarised):
1. Look at what the Android project backlog contains and take a task from "In progress / Fixing / Ready to do".
2. Make a prototype for #23336 (Android Auto OBD II widgets panel).
3. Make the UI generic so that any widget works; Dima suggested a panel on the left where there is space.
4. No — the panel must be on the right, so it does not cover the navigation card; 3–4 widgets seem to fit on some screens.
5. Paging on tap is possible, but think about whether it passes Google review; a prototype is fine.
6. Let's look at the UI and the simulator; install the dev nightly on the phone, do not touch production.
7. Separate widget set for Android Auto rather than reusing the phone right panel.
8. Fix the sizes: think about the proportions of a tablet-sized screen and of the DHU; the speedometer is suspiciously small; the alert is now huge, on the phone it is not like that.
9. No right margin for the widgets, never shift the panel — hide the widgets that would be shifted; neat rounded corners on top, bottom and left.
10. Run a smaller DHU to check how paging works with 5–6 widgets.
11. Widget configuration must be in the app, not in Android Auto; drop paging entirely and just cut off what does not fit.
12. Open a draft pull request and write down what is not finished.

- Reusing the existing `MapWidget` views and drawing them on the car canvas, instead of writing car-specific widget rendering.
- The `AA_WIDGETS` / `AA_SHOW_WIDGETS_PANEL` preferences and their defaults (`speed;altitude`, panel on).
- The dp-based sizing rule (200 car dp, 22 % clamp, 400 dp minimum surface width, 70 % maximum height) and the 8 dp corner radius.
- Rounding on the left only, and splitting the panel into separate rounded blocks when the alarm hides a row in the middle.
- Scaling the speedometer by 2× to cancel its Android Auto `0.77` factor, while leaving the alarm widget untouched.
- Adding a default `onSurfaceClick` to `SurfaceRendererCallback` and `SurfaceCallback.onClick` in `SurfaceRenderer` (only needed for the paging that is to be removed).

### DIFF
--- a/OsmAnd/res/values/strings.xml
+++ b/OsmAnd/res/values/strings.xml
@@ -4269,6 +4269,8 @@ You need to activate the sensor so OsmAnd can find it.</string>
     <string name="map_marker_2nd">Second map marker</string>
     <string name="shared_string_toolbar">Toolbar</string>
     <string name="shared_string_widgets">Widgets</string>
+    <string name="aa_show_widgets_panel">Show widgets panel</string>
+    <string name="aa_show_widgets_panel_descr">Draw the selected widgets over the map on the car screen</string>
     <string name="shared_string_add_to_map_markers">Add to map markers</string>
     <string name="select_map_markers">Select map markers</string>
     <string name="shared_string_reverse_order">Reverse order</string>
--- a/OsmAnd/src/net/osmand/plus/auto/CarWidgetsPanel.java
+++ b/OsmAnd/src/net/osmand/plus/auto/CarWidgetsPanel.java
@@ -1,0 +1,295 @@
+package net.osmand.plus.auto;
+
+import android.graphics.Canvas;
+import android.graphics.Path;
+import android.graphics.Rect;
+import android.graphics.RectF;
+import android.view.View;
+import android.view.View.MeasureSpec;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import net.osmand.plus.OsmandApplication;
+import net.osmand.plus.activities.MapActivity;
+import net.osmand.plus.settings.backend.ApplicationMode;
+import net.osmand.plus.settings.enums.ScreenLayoutMode;
+import net.osmand.plus.views.layers.base.OsmandMapLayer.DrawSettings;
+import net.osmand.plus.views.mapwidgets.MapWidgetInfo;
+import net.osmand.plus.views.mapwidgets.WidgetsInitializer;
+import net.osmand.plus.views.mapwidgets.WidgetsPanel;
+import net.osmand.plus.views.mapwidgets.appearance.ResolvedPanelAppearance;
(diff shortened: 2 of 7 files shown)


Kreiri:
NB: MapActivity is used in widgets (various subclasses of MapWidget and its subclasses) for vast variety of purposes, not just for layout inflation. Decoupling MapActivity from widgets and components used in widgets will take a lot of time to rework; may not be possible with current architecture.

PR #25859, opened 2026-09-05
Cache number formatters in OsmAndFormatter (GC ANR in QuickSearch list layout)

vshcherb:
Fixes the third most frequent user-perceived stability issue on Play: the GC ANR at
`art::gc::Heap::WaitForGcToCompleteLocked` — **223 affected users / 407 events over 28 days**
(2.0% of ANRs), issue `8b5d4a20`.

## Root cause

The reported main thread is not doing anything unusual — it is laying out a `ListView`:

```
at java.text.DecimalFormat.<init> (DecimalFormat.java:459)
at net.osmand.plus.utils.OsmAndFormatter.formatValue (OsmAndFormatter.java:789)
at net.osmand.plus.utils.OsmAndFormatter.getFormattedDistanceValue (OsmAndFormatter.java:464)
at net.osmand.plus.utils.UpdateLocationUtils.updateLocationView (UpdateLocationUtils.java:55)
at net.osmand.plus.search.dialogs.QuickSearchListAdapter.updateCompass (QuickSearchListAdapter.java:651)
at net.osmand.plus.search.dialogs.QuickSearchListAdapter.getView (QuickSearchListAdapter.java:247)
at android.widget.ListView.layoutChildren (ListView.java:1802)
...
at android.view.ViewRootImpl.performLayout (ViewRootImpl.java:4175)
```

…and the frames above it are the allocation slow path:

```
art::gc::Heap::WaitForGcToCompleteLocked
art::gc::Heap::AllocateInternalWithGc
artAllocArrayFromCodeResolvedRegionTLAB
at java.lang.AbstractStringBuilder.<init>
at android.icu.impl.LocaleIDParser.<init>
at android.icu.util.Currency.getInstance
at java.text.DecimalFormatSymbols.initializeCurrency
at java.text.DecimalFormat.initPattern
```

`formatValue()` built a `DecimalFormat`, a `DecimalFormatSymbols` and a `MessageFormat` on every
call. Constructing `DecimalFormatSymbols` resolves the locale currency through ICU, which allocates
its way through `LocaleIDParser`.

`QuickSearchListAdapter.updateCompass()` formats the distance of every visible row on every compass
update, from inside `ListView.layoutChildren()`. So the app pays the ICU cost dozens of times per
second on the main thread, during layout.

Measured on an Android 15 arm64 emulator: **0.45 ms per formatted value.**

## Fix

Cache the formatter in `formatValue()`, keyed by everything that configures the instance — pattern,
thousands grouping, locale — so a cached instance is never reconfigured after creation. The key
cannot be the unit: the same `unitId` is formatted with different `decimalPlacesNumber` (see
`getFormattedDistanceValue()`, where km goes through both `decimalPrecision` and
`1 + decimalPrecision`), and the `String unit` overload passes `unitId = -1` for every caller. The
locale has to be part of the key because `preferredLocale` changes at runtime.

`DecimalFormat` is mutable and not thread safe, and formatting also happens off the main thread
(`RoutingHelper`, `SplitTrackAsyncTask`, voice announcements), so the cache is a `ThreadLocal`.

The `MessageFormat` wrapper is dropped too: `MessageFormat("{0}")` with the `DecimalFormat` as its
only subformat calls `DecimalFormat.format(Object)` on a boxed `Float`, which ends in the same
`format(double)` as calling the formatter directly.

## Test

New instrumented test `OsmAndFormatterCachingTest` (`OsmAnd/test/java/net/osmand/test/junit/`).

`formattedValuesAreUnchanged` formats a matrix of 19 values × 4 decimal places × trailing-zeroes
on/off and compares every result against a verbatim copy of the pre-fix implementation kept in the
test. This is the guard that the cache did not change a single output — it passes both before and
after the change.

`repeatedFormattingDoesNotRebuildTheFormatter` measures the production path against that same
reference implementation in the same run, so there is no absolute threshold to tune: the cached path
has to be at least 5× cheaper.

Run on `Claude_Phone_API_35` (Android 15, arm64), variant `nightlyFreeOpenglArm64Debug`, 3000 calls:

**Before the fix** — the production path costs exactly as much as building a formatter per call:

```
repeatedFormattingDoesNotRebuildTheFormatter FAILED
java.lang.AssertionError: formatValue() still builds a DecimalFormat on every call -
building a formatter per call: 816.00006 ms, OsmAndFormatter.formatValue(): 774.0576 ms
for 3000 calls
```

**After the fix** — 13.9× cheaper, both tests pass:

```
building a formatter per call: 1151.7996 ms, OsmAndFormatter.formatValue(): 83.15504 ms for 3000 calls

<testsuite name="net.osmand.test.junit.OsmAndFormatterCachingTest"
           tests="2" failures="0" errors="0" skipped="0" time="2.755">
  <testcase name="formattedValuesAreUnchanged" time="0.329" />
  <testcase name="repeatedFormattingDoesNotRebuildTheFormatter" time="1.302" />
</testsuite>
```

0.45 ms → 0.028 ms per formatted value.

## AI disclaimer

Prompts used (summarised):
1. Asked whether the most frequent crashes in Play Console could be fixed, with links to seven
   crash issues.
2. Asked to review the rest of the top crashes and ANRs, group anything similar, and open a separate
   pull request per issue against the 5.4 branch only (not master).
3. Asked to reproduce the first group with a test and open that pull request; then to file the
   internal task for it.
4. Asked to open the pull request for this group next, and to start studying the renderer ANR.
5. Objected that the fix was over-complicated and asked why it is not simply a cache map inside
   `formatValue()`; after the key and thread-safety constraints were explained, asked to keep the
   `ThreadLocal` and key by pattern.

* Removing the `MessageFormat` wrapper.
* The test design: a verbatim pre-fix reference implementation used both as the correctness oracle
  and as the self-calibrating performance control, instead of an absolute time threshold.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/utils/OsmAndFormatter.java
+++ b/OsmAnd/src/net/osmand/plus/utils/OsmAndFormatter.java
@@ -46,14 +46,15 @@ import java.text.DateFormat;
 import java.text.DateFormatSymbols;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
-import java.text.MessageFormat;
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.TimeZone;
 
 public class OsmAndFormatter {
@@ -119,6 +120,18 @@ public class OsmAndFormatter {
 		Locale preferredLocale = localeHelper.getPreferredLocale();
 		return preferredLocale != null ? preferredLocale : localeHelper.getDefaultLocale();
 	}
+
+	/**
+	 * Creating a {@link DecimalFormat} resolves the locale through ICU ({@link DecimalFormatSymbols}
+	 * looks up the currency, which builds a {@code LocaleIDParser}), which is far too expensive for
+	 * code that runs for every visible list row on every compass update: the allocation churn showed
+	 * up as an ANR inside {@code art::gc::Heap::WaitForGcToComplete} during {@code ListView} layout.
+	 *
+	 * <p>The key covers everything that configures the instance, so a cached formatter is never
+	 * reconfigured after creation. {@link DecimalFormat} is mutable and not thread safe, hence one
+	 * cache per thread.
+	 */
+	private static final ThreadLocal<Map<String, DecimalFormat>> DECIMAL_FORMATS = new ThreadLocal<>();
 
 	@NonNull
 	private static DecimalFormat createDecimalFormat(@NonNull String pattern, int minFractionDigits,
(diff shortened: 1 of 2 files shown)


vshcherb:
Internal task: osmandapp/OsmAnd-Issues#3316 (gen by AI)

PR #25884, opened 2026-09-07
Ios kotlin xml parser

### LINKED ISSUE #4095: Add progress indicator for My Places – Tracks
Image

When tracks are indexed, add cell with circular indicator and with Text "Tracks stats are being calculated..."

Image

vitaliy-sova-ios:
## Related issue / task

Fixes osmandapp/OsmAnd-iOS#4095

## Summary

* Replaced the iOS `XmlPullParser` actual — which forwarded every call (`next`, `getName`, `getText`, `getAttributeValue`, …) to an Obj-C `OAXmlStreamReader` (Qt's `QXmlStreamReader`) — with a self-contained pure-Kotlin pull parser.
* Parsing no longer crosses the Kotlin/Native ↔ Obj-C boundary: the input is read once via okio (`KFile.source()`) and tokenized entirely in Kotlin, removing the per-call `objc_msgSend` dispatch, the double string copy (C/Obj-C string → `NSString` → Kotlin `String`) and the bridged-object bookkeeping on every token.
* Implements the subset of the `org.xmlpull.v1.XmlPullParser` `next()` contract the shared callers rely on (`GpxUtilities.loadGpxFile`, `ImportGpx.parseKmlStreaming`): only `START_DOCUMENT` / `START_TAG` / `TEXT` / `END_TAG` / `END_DOCUMENT` are surfaced, CDATA is coalesced into text, comments / processing instructions / DOCTYPE are skipped, predefined and numeric (`&#…;` / `&#x…;`) entities are decoded, self-closing tags are reported as `START_TAG` + a synthetic `END_TAG`, and tag balance is validated with descriptive `XmlParserException`s.
* Restored input-encoding auto-detection that `QXmlStreamReader` performed: byte-order mark + `<?xml … encoding="…"?>` declaration. Supported: UTF-8 (+BOM), UTF-16 LE/BE, windows-1251, windows-1252, ISO-8859-1. Unknown labels fall back to UTF-8, as before. `getInputEncoding()` now reports the detected charset.
* Incidentally fixes KML import on iOS: `setFeature(process-namespaces, …)` and `nextText()` were no-op stubs in the previous Obj-C implementation and now work correctly.
* **`NativeFile` (iosMain) file metadata via `stat(2)`.** `isDirectory()` / `exists()` / `length()` / `lastModified()` now issue a single `platform.posix.stat` syscall instead of `NSFileManager.fileExistsAtPath` / `attributesOfItemAtPath`, the latter of which builds an `NSDictionary` of all file attributes and bridges it to a Kotlin `Map` on every call. `KFile.length()` / `lastModified()` are hit ~2× per file while scanning a folder, so a 5000-track folder previously did ~10k dictionary-bridge round-trips — this is the main cost behind entering a large track subfolder taking ~5 s on iOS vs ~0.8 s on Android, and it also feeds into the bulk-indexing time.
* `SensorPointAnalyser` (iosMain): skip the per-tag sensor-attribute loop for track points that carry none of the six sensor extension keys — the point's own (usually tiny) key set is checked against a `HashSet` first, instead of probing every fixed sensor tag on every point during bulk indexing. Mirrors the iOS `VehicleTrackPointsAnalyser` optimization.
* No changes to `commonMain` or `androidMain`; the dormant `getXmlPullParserApi()` / `XmlPullParserAPI` path is left in place (now unused) rather than removed in this PR. Android behavior is unchanged.

## Testing

Tested on:

* Device: iPhone 11
* iOS: 26.5
* `./gradlew :OsmAnd-shared:compileKotlinIosSimulatorArm64` — passes, no new warnings.

Scenarios

* Bulk-index ~5000 GPX tracks placed manually into the app's gpx folder: no jetsam/OOM kill (previously crashed); end-to-end indexing time reduced from ~2 min to ~1 min 10s (combined with the `stat(2)` metadata change and the iOS-side nearest-city cache).
* Enter a subfolder containing ~5000 tracks: opens fast (previously ~5 s while the list populated), close to the Android timing.
* Import a GPX file that contains more than one track (including files with self-closing `<trkpt/>` points): the track-selection dialog opens as expected.
* Import a KML file: parses and imports correctly (namespace processing / `nextText()` were previously inert on iOS).
* Load GPX/KML files that are not UTF-8 (UTF-16 LE/BE, windows-1251, windows-1252, ISO-8859-1, and UTF-8 with a BOM): parsed correctly, no mojibake.

## AI disclaimer

Implementation:

* Tool / Agent: Claude Code
* Model: Claude Sonnet 5

Prompts used (summarised):

1. Investigate the OOM (jetsam) crash during bulk GPX-track indexing on iOS and why it is much slower than Android on the same shared Kotlin code.
2. Explain why the custom parser is still slower than Android; is moving the whole flow to Swift viable; describe what happens during GPX analysis.
3. Have the shared library parse the GPX file and persist it directly, instead of parsing on iOS and sending each attribute across to Kotlin — is that feasible?
4. Before implementing, verify it does not break other iOS logic (`VehicleTrackPointsAnalyser`, sensor analyser, KML import), then proceed.
5. Entering a 5000-track subfolder takes ~5 s on iOS vs ~0.8 s on Android for the same shared code — analyse and fix, iosMain only, behaviour-preserving.
6. Non-UTF-8 GPX/KML files no longer load with the new parser — fix.
7. Multi-track GPX import no longer opens the track-selection dialog — fix.
8. Remove the diagnostic logging once the new parser is confirmed to be the active path.
9. Code-review the branch and fix the important findings.

* Chose a fully self-contained pure-Kotlin pull parser (read-once via okio, tokenize in Kotlin) over adding a `cinterop` binding or keeping an Obj-C backend — the build config needs no changes and there is no bridge cost during parsing.
* Scoped the tokenizer to the exact contract the two real `commonMain` callers use; left `nextTag()`, `nextToken()`, namespace-URI resolution and line/column reporting as no-op stubs after confirming they have zero `commonMain` callers.
* Restored BOM / XML-declaration encoding detection (regression surfaced during self-review) and hand-rolled the windows-1251 / windows-1252 / ISO-8859-1 decoders, since Kotlin/Native `decodeToString()` is UTF-8 only.
* Replaced `NativeFile`'s `NSFileManager` metadata calls with `platform.posix.stat` rather than caching the `NSDictionary` result, to drop the per-call dictionary bridge entirely.
* Added the `hasAnySensorKey` fast-path to `SensorPointAnalyser` for symmetry with the iOS `VehicleTrackPointsAnalyser` change made alongside this work.

Final review:

* Tool / Agent: Claude Code
* Model: Claude Sonnet 5
* [x] Final diff reviewed

Significant findings:

* Found and fixed an encoding regression: the new parser decoded all input as UTF-8, whereas the previous `QXmlStreamReader` auto-detected from BOM and the `<?xml encoding>` declaration — non-UTF-8 GPX/KML (UTF-16, windows-1251/1252, ISO-8859-1) would have come back as mojibake.
* Found and fixed a multi-track import regression: the synthetic `END_TAG` emitted for a self-closing element was popping the open-tag stack, so a `<trkpt/>` inside `<trkseg>` produced `Mismatched closing tag: expected </trkseg>, found </trkpt>`; `loadGpxFile` then set `gpxFile.error`, `getTracksCount()` returned 0/1, and `OAGPXImportUIHelper` skipped the track-selection dialog for multi-track files.
* Verified `getLineNumber()` / `getColumnNumber()` / `getInputEncoding()` / `isEmptyElementTag()` / `nextTag()` / `nextToken()` have no `commonMain` callers, so reducing them to stubs is not a functional regression.
* Confirmed the KML import path (`setFeature` / `nextText()`), previously inert via Obj-C no-op stubs, is now exercised and correct.
* Confirmed the change is confined to `iosMain` (`XmlPullParser.kt`, `NativeFile.kt`, `SensorPointAnalyser.kt`); no `commonMain` / `androidMain` behaviour change, and the unused `getXmlPullParserApi()` / `XmlPullParserAPI` / `XmlFactoryAPI` types are left in place.
* Reviewed `SensorPointAnalyser` early-return: the per-tag loop (and its `attribute.setAttributeValue` / `analysis.setHasData` calls) is skipped only for points whose extension key sets contain none of the six sensor tags — matches the already-shipped iOS assumption.


OsmAnd-shared/src/iosMain/kotlin/net/osmand/shared/xml/XmlPullParser.kt
```diff
@@ -22,186 +36,574 @@ actual class XmlPullParser actual constructor() {
...
 
-	actual fun getProperty(name: String): Any? {
-		return xmlPullParserAPI.getProperty(name)
-	}
+	actual fun getProperty(name: String): Any? = null
 
 	@Throws(XmlParserException::class)
 	actual fun setInput(file: KFile, inputEncoding: String?) {
-		xmlPullParserAPI.setInput(file.absolutePath(), inputEncoding)
+		val bytes = file.source().buffer().use { it.readByteArray() }
```
alex-dev-neo: `setInput` reads the whole file into a `ByteArray`, and `decodeBytes()` then materialises a full Kotlin `String` that stays in `text` for the entire parse. Peak is roughly the file size in bytes plus 2 bytes per character of the string.

The non-UTF-8 paths cost more: the `StringBuilder` in `decodeUtf16()` / `decodeSingleByte()` is a second full copy live at the same time as the byte array, and `beginParsing`'s `decoded.substring(1)` for a surviving BOM duplicates the whole string again.

`QXmlStreamReader` streamed, so this peak is new. It does not accumulate across files — `GpxUtilities.kt:1682` calls `parser?.close()` and that clears `text` — but the per-file peak lands in a change series whose starting point was a jetsam OOM during bulk indexing. Worth measuring against the largest GPX you can find before merge.

OsmAnd-shared/src/iosMain/kotlin/net/osmand/shared/xml/XmlPullParser.kt
```diff
@@ -1,12 +1,26 @@
...
+// see KFile.source()) and tokenizes the decoded text entirely in Kotlin.
+//
+// Matches the real, exercised contract of the two commonMain callers (GpxUtilities.loadGpxFile
+// and ImportGpx.parseKmlStreaming): only next()/getEventType() ever surface START_DOCUMENT,
+// START_TAG, TEXT, END_TAG, END_DOCUMENT (comments/processing instructions are skipped,
+// CDATA is coalesced into text, entities are expanded) - this is the standard next() contract
+// used by org.xmlpull.v1.XmlPullParser, i.e. exactly what Android's actual (backed by
+// android.util.Xml) already provides. getAttributeValue(namespace, name) is always called
+// with namespace = "" by real callers, i.e. a plain attribute-name lookup.
 actual class XmlPullParser actual constructor() {
```
alex-dev-neo, line 31: No tests are added, and `OsmAnd-shared` already has `commonTest` and `jvmTest` to model an iOS test source set on.

The PR's own "Significant findings" section lists two regressions that self-review caught: encoding auto-detection lost against `QXmlStreamReader`, and the synthetic `END_TAG` for a self-closing element popping the open-tag stack, which broke multi-track GPX import via `Mismatched closing tag`. Both are a handful of assertions in a unit test. That is the argument for a test file, not against one — a tokenizer this size will keep producing findings of exactly that shape, and right now every one of them has to be found by reading or by a user.

A small fixture set would cover most of the risk: self-closing elements at various nesting depths, CDATA next to text, numeric and predefined entities in text and in attribute values, each supported encoding with and without a BOM, and the malformed cases the parser is expected to reject.

OsmAnd-shared/src/iosMain/kotlin/net/osmand/shared/xml/XmlPullParser.kt
```diff
@@ -22,186 +36,574 @@ actual class XmlPullParser actual constructor() {
...
 	}
 
 	@Throws(XmlParserException::class, IOException::class)
-	actual fun nextTag(): Int {
-		return xmlPullParserAPI.nextTag()
+	actual fun nextTag(): Int = -1 // Not implemented - unused by real callers
+
+	// --- tokenizer internals -------------------------------------------------------------
+
+	private fun applyNamespaceMode(rawName: String?): String? {
```
alex-dev-neo, line 675: With `processNamespaces = true` (the default here), `applyNamespaceMode` strips the prefix off namespace declarations too, so `xmlns:gpxx="..."` is reported as an attribute literally named `gpxx`, and `xmlns="..."` as one named `xmlns`.

`org.xmlpull.v1.XmlPullParser` with namespace processing on does not report `xmlns` declarations as attributes at all. Callers that iterate the attribute list rather than asking for a name will pick up the phantoms — `StringBundleXmlReader.kt:13` and `RoutingConfiguration.kt:525` both do `for (i in 0 until parser.getAttributeCount())`.

Filtering names starting with `xmlns` out of `attrs` in `readStartTag` when `processNamespaces` is on would match the reference behaviour.

OsmAnd-shared/src/iosMain/kotlin/net/osmand/shared/xml/XmlPullParser.kt
```diff
@@ -22,186 +36,574 @@ actual class XmlPullParser actual constructor() {
...
 
-	actual fun getNamespace(prefix: String?): String? {
-		return xmlPullParserAPI.getNamespace(prefix)
-	}
+	actual fun getNamespace(prefix: String?): String? = null // Not implemented
 
-	actual fun getDepth(): Int {
-		return xmlPullParserAPI.getDepth()
-	}
+	actual fun getDepth(): Int = openTags.size
```
alex-dev-neo: `getDepth()` is off by one for empty elements: `readStartTag` deliberately does not push a self-closing tag onto `openTags`, so at the `START_TAG` of `<x/>` this reports the depth of the parent instead.

Unlike `getLineNumber()` / `nextTag()` and the other stubs, this one does not announce itself — it returns a plausible number that happens to be wrong. There are no `commonMain` callers today, so nothing is broken; the risk is the next caller. Either count `pendingEndTagName` in, or make it an explicit stub like its neighbours.

OsmAnd-shared/src/iosMain/kotlin/net/osmand/shared/xml/XmlPullParser.kt
```diff
@@ -22,186 +36,574 @@ actual class XmlPullParser actual constructor() {
...
+		} else {
+			val content = text.substring(contentStart)
+			pos = text.length
+			content
+		}
+	}
+
+	// Balances nested '<'/'>' - used for <!DOCTYPE ...> which may contain an internal
+	// subset ("[ ... ]") with its own '<!ENTITY ...>' declarations.
+	private fun skipBalancedMarkup() {
```
alex-dev-neo, line 801: `skipBalancedMarkup()` counts `<` and `>` without tracking quotes, so a `>` inside a quoted value in a DOCTYPE internal subset ends the skip early and leaves the parser in the middle of the subset:

```xml
<!DOCTYPE gpx [ <!ENTITY x "a > b"> ]>
```

Narrow in practice, but the internal subset is the only reason this function exists — a plain `<!DOCTYPE gpx>` would be handled by a simple scan to the next `>`. Skipping quoted spans would make it do what the comment above it says.

OsmAnd-shared/src/iosMain/kotlin/net/osmand/shared/xml/XmlPullParser.kt
```diff
@@ -22,186 +36,574 @@ actual class XmlPullParser actual constructor() {
...
+		}
+		if (bytes.size >= 2 && bytes[0] == 0xFF.toByte() && bytes[1] == 0xFE.toByte()) {
+			detectedEncoding = "UTF-16LE"
+			return decodeUtf16(bytes, 2, bigEndian = false)
+		}
+		val declared = sniffDeclaredEncoding(bytes)
+		detectedEncoding = declared ?: "UTF-8"
+		return when (declared) {
+			null, "utf-8", "utf8", "us-ascii", "ascii" -> bytes.decodeToString()
+			"utf-16", "utf-16le", "utf16", "unicode" -> decodeUtf16(bytes, 0, bigEndian = false)
```
alex-dev-neo: A document that declares `encoding="UTF-16"` with no BOM is decoded as little-endian unconditionally, so a big-endian one comes out as mojibake rather than an error.

The XML spec makes BOM-less UTF-16 a fatal error, so failing to handle it is defensible — but silently guessing an endianness is a third behaviour. Sniffing the first two bytes for `00 3C` versus `3C 00` is a couple of lines, and otherwise a comment saying the LE default is deliberate would do.

OsmAnd-shared/src/iosMain/kotlin/net/osmand/shared/io/NativeFile.kt
```diff
@@ -20,32 +16,27 @@ actual class NativeFile actual constructor(actual val file: KFile) {
...
-	actual fun lastModified(): Long {
-		val attr: Map<Any?, *>? = NSFileManager.defaultManager.attributesOfItemAtPath(filePath, null)
-		if (attr == null || attr.isEmpty()) {
-			return 0
-		}
-		return ((attr[NSFileModificationDate] as NSDate).timeIntervalSince1970 * 1000.0).toLong()
+	actual fun lastModified(): Long = memScoped {
+		val st = alloc<stat>()
+		if (stat(filePath, st.ptr) != 0) return@memScoped 0L
+		st.st_mtimespec.tv_sec * 1000L + st.st_mtimespec.tv_nsec / 1_000_000L
```
alex-dev-neo: Worth verifying before merge: this may not round identically to the code it replaces.

The old path was `(NSDate.timeIntervalSince1970 * 1000.0).toLong()` — the nanosecond mtime goes through a `Double` and back — while this is exact integer arithmetic. For mtimes whose nanosecond part is not exactly representable as a binary fraction, the two can land a millisecond apart. Whole-second mtimes are unaffected.

That matters because `GpxDbHelper.reconcileFilesystem()` compares exactly:

```kotlin
if (storedModifiedTime != actualModifiedTime || GpxDbUtils.isAnalyseNeeded(item, actualModifiedTime))
```

Every value already in the database was written by the old path. If the new one differs by 1 ms for a meaningful share of files, the first launch after this ships re-reads and re-analyses them — for a 5000-track library, precisely the cost this series is trying to remove. A quick check against a real database would settle it; if it does drift, rounding to match the old result keeps existing rows valid.

alex-dev-neo (commented):
Review of the branch, from the iOS side (this is the companion to osmandapp/OsmAnd-iOS#5737).

Two of the three changes I checked and consider safe as written:

* **`NativeFile` on `stat(2)`** — well-scoped and a clear win. Semantics hold: `isDirectory()` returns `false` on a failed `stat` the same way the old code did for a missing file, `exists()` and `length()` are equivalent, and `lastModified()` keeps millisecond precision.
* **`SensorPointAnalyser`** — behaviour-preserving. The skipped call wrote `Float.NaN`, and the getters already return `sensorData?.x ?: Float.NaN` when unset, so the read-back is identical. `PointAttributes` also lives on the `WptPt` itself (`GpxTrackAnalysis.kt:637`) rather than in a shared pool, so there is no leakage between points.

The parser is the part worth discussing. Replacing `QXmlStreamReader` with 609 lines of hand-written tokenizer puts new code on the path of every GPX and KML file on iOS, and the two items I would resolve before merge are the per-file memory peak and the absence of tests. The rest are smaller, and #7 is a question rather than a defect.

vitaliy-sova-ios:
Memory peak on setInput — The parser now streams: bytes are pulled in 8 KB chunks into an 8 KB sliding character window with bounded lookahead, so peak memory no longer scales with file size (a 32 MB GPX parses under a 12 MB heap, where the previous whole-document approach needed over 96 MB).

No tests added — Added an iosTest source set with 30 iOS-specific tests (encoding auto-detection with and without BOM, buffer-boundary sweeps, malformed input) plus 19 cross-platform tests in commonTest that assert the same event stream on iOS as on Android, and 2 tests pinning lastModified() compatibility — 51 tests, all green on the simulator.

xmlns reported as phantom attributes — readStartTag now filters namespace declarations out of the attribute list when namespace processing is on (checked on the raw name, so both xmlns and xmlns:gpxx are caught), matching the reference parser.

getDepth() off by one for empty elements — Fixed for both cases the tag stack cannot express — an empty element and an END_TAG, which must report the depth of the element it closes — with the exact depth table from the xmlpull docs now asserted in a contract test that kxml2 passes unchanged.

skipBalancedMarkup() ignores quotes — Angle brackets inside quoted literals and inside comments no longer count towards the balance, so <!ENTITY x "a > b"> and <!DOCTYPE a SYSTEM "http://x/a>b.dtd"> are skipped correctly; a mutation check confirms the new test fails against the old logic.

BOM-less encoding="UTF-16" guesses little-endian — The endianness is now sniffed from the first two bytes (00 3C vs 3C 00, unambiguous because the document starts with <), with the little-endian fallback documented as deliberate.

lastModified() rounding may not match the old path — Verified: exact integer arithmetic differs from the old NSDate formula for nanosecond values on a millisecond boundary (0.03% of real files here, 10.6% of files whose mtime was assigned from a millisecond-precision Date), so the implementation now reproduces the old rounding by going through the 2001 reference date — a match in all 138,265 sampled timestamps, pinned by a test that compares against NSFileModificationDate directly.

One extra, found by running the suite against master rather than by review:

getNamespace() returned null — master added a caller (GpxUtilities.getQualifiedExtensionTagName) that identifies known extension namespaces by URI, so the parser now tracks and resolves namespace declarations per scope; without it, Garmin tracks using ns2:/ns3: prefixes would silently lose heart rate, cadence and track colour on iOS.

PR #25892, opened 2026-09-07
Move routing into OsmAnd-shared

alex-dev-neo:
Long running branch for moving the routing code out of `OsmAnd-java` and `OsmAnd/plus/routing` into
the KMP module, so Android, iOS and the server share one implementation instead of the four that
exist today (Java, the C++ port in `core-legacy`, the hand written Objective-C port in
`ios/Sources/Router`, and the app layer on top).

Work lands here step by step, bottom up. This description gets updated as each step goes in.

---

## Landed so far: the foundation

The pieces every later step depends on. Nothing existing changes behaviour yet, and no routing class
is ported.

**`net.osmand.shared.util.collections`** — replacements for `gnu.trove`, which has no KMP equivalent:

| Class | Replaces | Notes |
|---|---|---|
| `KTLongObjectMap` | `TLongObjectHashMap` | open addressing, linear probing, MurmurHash3 key mixing, tombstones |
| `KTLongHashSet` | `TLongHashSet` | same scheme |
| `KTIntArrayList` | `TIntArrayList` | flat `IntArray`, `get` / `getQuick` split as in trove |
| `KBitSet` | `java.util.BitSet` | `java.util.BitSet` semantics, including capacity independent `equals` / `hashCode` |

Keys are mixed through the MurmurHash3 finalizer rather than used raw: routing keys look like
`roadId << 6 | segment`, which is structured enough to cluster badly under linear probing.

**`net.osmand.shared.data`** — ports of `net.osmand.Location` and `net.osmand.data.*`:
`KLocation` (Vincenty inverse solution on WGS84), `KQuadPoint`, `KQuadPointDouble`, `KQuadTree`,
`KDataTileManager`.

### Design notes worth a reviewer's eye

- `KLocation` keeps the Java "primitive plus `hasX` flag" shape instead of nullable primitives. A
  route holds thousands of fixes and `Float?` would box every one of them.
- `KLocation.distanceTo` / `bearingTo` share one solution through an immutable snapshot rather than
  the `synchronized` block of the original, since common code has no `synchronized`. A racing reader
  either sees a complete consistent snapshot or recomputes.
- `KDataTileManager.getObjects` for a lat/lon box resets `tileYUp` where the Java original resets
  `tileXUp`. That looks like a typo; the branch is unreachable for the argument order callers
  actually use.
- Constructors carry `@JvmOverloads` so the existing Java routing code can construct these without
  passing a capacity.

### Testing

- **67 tests in `commonTest`**, passing on JVM and on Kotlin/Native (iOS simulator). Each collection
  is driven against a reference model (`HashMap` / `HashSet` / `ArrayList` / `HashSet<Int>`) over
  tens of thousands of random operations, so tombstone reuse, probe chains after removal, growth and
  iteration all get exercised.
- **`TroveComparisonTest` in `OsmAnd-java`** puts the shared collections and the originals
  (`gnu.trove`, `java.util.BitSet`) through the same operation stream and asserts they stay in step.
  It lives there because that module is the only place the two share a classpath.

| | tests | skipped | failures |
|---|---|---|---|
| `:OsmAnd-shared:jvmTest` | 131 | 4 | 0 |
| `:OsmAnd-shared:iosSimulatorArm64Test` | 130 | 4 | 6 pre-existing, unrelated |
| `:OsmAnd-java:test --tests "net.osmand.shared.*"` | 9 | 4 | 0 |

The 6 native failures are in `GpxUtilitiesLoadTest`, `KGeoPointParserUtilTest` and `NetworkAPITest`
and fail on `master` as well.

### Benchmarks

Two benchmark classes are included and **annotated `@Ignore`** — the 4 skipped tests in each row
above. Measuring takes seconds and the timings are noise in a normal run. Each class documents how
to run it.

```
JVM, vs the originals
  long -> object map, 300k    KTLongObjectMap  7.74 ms   TLongObjectHashMap  9.29 (1.20x)
  long set, 300k              KTLongHashSet    4.48 ms   TLongHashSet        7.23 (1.61x)
  int list, 3M append+scan    KTIntArrayList   3.48 ms   TIntArrayList       2.77 (0.79x)
  bit mask, 300k rules        KBitSet          2.93 ms   java.util.BitSet    2.08 (0.71x)
```

`KTIntArrayList` trails trove only on the growth path; presized appends and scans match it. Presize
when the length is known.

The build file gains an optimised iOS test binary. The default `iosSimulatorArm64Test` task links
`debugTest`, i.e. without LLVM optimisations, and reports numbers 5 to 13 times worse than a real
build — enough to draw the wrong conclusion about whether Kotlin/Native can carry routing code. The
release binary is not part of the normal build graph and only links when asked for:

```
./gradlew :OsmAnd-shared:linkReleaseTestIosSimulatorArm64
```

Measured that way, Kotlin/Native release sits within 1.3–2.1x of the JVM on these workloads, and the
primitive collections beat the boxing stdlib ones by a wider margin on iOS than on JVM (6.2x vs
`ArrayList<Int>`, against 1.9x on JVM).

---

## Next

Pure model classes with no I/O and no platform dependencies: `TurnType`, `RouteCalculationProgress`,
`RoundaboutTurn`, `RouteImporter` / `RouteExporter`, and from the app layer `AlarmInfo`,
`RouteDirectionInfo`, `CurrentStreetName`.

`TurnType` goes first deliberately: it is read from JNI (`java_wrap.cpp` looks classes up by name and
reads fields directly), so it exercises both the C++ binding update and consuming a Kotlin class from
Objective-C before anything larger depends on that working.

### DIFF
--- a/OsmAnd-java/src/test/java/net/osmand/shared/util/collections/CollectionsTroveBenchmarkTest.java
+++ b/OsmAnd-java/src/test/java/net/osmand/shared/util/collections/CollectionsTroveBenchmarkTest.java
@@ -1,0 +1,364 @@
+package net.osmand.shared.util.collections;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.ArrayList;
+import java.util.BitSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Random;
+
+import org.junit.Ignore;
+import org.junit.Test;
+
+import gnu.trove.list.array.TIntArrayList;
+import gnu.trove.map.hash.TLongObjectHashMap;
+import gnu.trove.set.hash.TLongHashSet;
+
+/**
+ * Timing comparison of the shared primitive collections against gnu.trove and java.util.BitSet.
+ *
+ * <b>Disabled on purpose.</b> Measuring takes seconds and the timings are noise in a normal test
+ * run, so this class is not part of the suite. It is a tool for when you change a collection, not a
+ * regression gate: correctness against the same originals is covered by {@link TroveComparisonTest},
+ * which does run.
+ *
+ * To measure, remove the {@code @Ignore} below, run it, and put the annotation back:
+ * <pre>
+ * ./gradlew :OsmAnd-java:test --tests "*CollectionsTroveBenchmarkTest" -i
+ * </pre>
+ *
+ * The cross platform half of the same measurement, which is what tells us how Kotlin/Native
+ * compares, is in
+ * OsmAnd-shared/src/commonTest/kotlin/net/osmand/shared/util/collections/CollectionsBenchmarkTest.kt
+ * and is disabled for the same reason.
+ *
+ * Nothing here asserts a timing. The assertions only check that both implementations computed the
+ * same answer, so a slow machine can never turn the build red.
+ */
+@Ignore("benchmark, run manually")
+public class CollectionsTroveBenchmarkTest {
+
+	private static final int ENTRIES = 300_000;
+	private static final int LOOKUPS = 600_000;
+	private static final int LIST_ENTRIES = 3_000_000;
+	private static final int RULE_EVALUATIONS = 300_000;
+	private static final int WARMUP_ROUNDS = 3;
+	private static final int MEASURED_ROUNDS = 5;
+
+	@Test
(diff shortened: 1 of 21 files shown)


alex-dev-neo:
Superseded by #25900 — same commit, branch renamed to `shared-routing-1-collections` so the work can land as a series of stacked PRs instead of one long running branch. This PR was closed automatically by the branch rename.

PR #25897, opened 2026-09-07
Keep accuracy circle and view angle sector on the location marker

vshcherb:
Internal issue: osmandapp/OsmAnd-Issues#3325

### Problem

While driving, the view angle sector was left behind the location icon and the gap kept growing —
on a real drive the sector ended up about 20 km away from the arrow.

Reproduced on an emulator (OpenGL renderer, simulated positions) with three conditions:

1. Profile appearance: **Location radius = Off**, **View angle** shown for the current state.
2. **Animate my location = on** (default).
3. The map is **not** following my location (panned away).

Pixel centroids over a screenshot series (7 s apart, ~180 px per 100 m): the arrow moved
591 → 779 → 958 px while the sector stayed at 526 px, i.e. the gap grew 66 → 253 → 431 px in
15 seconds. With `Location radius = Resting and navigation`, or with the map following my location,
the sector stays on the arrow.

### Cause

The location icon is a `MapMarker`, but the accuracy circle and the view angle sector are drawn by
the renderer at `MapRendererState::myLocation31`, which is separate state:

- `PointLocationLayer.updateMarkerPosition()` pushes it only for non-animated moves;
- during an animated move it was pushed by the core markers animator, but only while the accuracy
  circle was visible (see osmandapp/OsmAnd-core#1109);
- `onUpdateFrame()` re-set the position on every frame only while the map was following my location,
  which is why a panned map was needed to see it.

### Fix

`PointLocationLayer.onUpdateFrame()` now syncs the renderer position with the current marker
position on every frame, so the two cannot drift apart even when the position is animated or an
animation is cancelled. The root cause is fixed in core as well (osmandapp/OsmAnd-core#1109), which
also covers iOS; this change makes the app correct with any core build.

### Test

`MyLocationSectorPositionTest` (instrumented, `OsmAnd/test/java/net/osmand/test/ui/map/`) drives a
simulated path at 120 km/h with the accuracy circle switched off and the map not following my
location, and asserts every 500 ms that the renderer's "my location" position stays on the location
marker; it also checks that my location really moved, so the test cannot pass vacuously.

Verified on `Claude_Phone_API_35` (arm64 emulator, `nightlyFreeOpenglArm64Debug`): the test fails
without this change and passes with it.

## AI disclaimer

Prompts used (summarised):
1. Help me reproduce the bug where the location icon and the view angle get separated (screenshot
   from a real drive).
2. Create a pull request for 5.5 and, if possible, a test that catches this regression or similar
   bugs.
3. Also create a bug on the 5.5-map milestone.

- Reproducing with the development plugin's position simulation instead of `adb emu geo fix`
  (emulator fixes carry no timestamp, so the location animation never starts and the bug stays
  hidden).
- Fixing the root cause in core (separate PR) in addition to the app-side per-frame sync, so that
  iOS is covered and the app does not depend on the core snapshot for the fix.
- Writing the test as an instrumented UI test driven by the location simulation, next to the
  existing tests in `net.osmand.test.ui`, instead of adding a new unit-test seam to the layer.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/views/layers/PointLocationLayer.java
+++ b/OsmAnd/src/net/osmand/plus/views/layers/PointLocationLayer.java
@@ -258,7 +258,21 @@ public class PointLocationLayer extends OsmandMapLayer
 		} else if (isMapLinkedToLocation() && !isMovingToMyLocation()) {
 			updateMarker(getPointLocation(), null, 0);
 		}
+		updateMyLocationCirclePosition(mapRenderer);
 		lastMarkerLocation = getCurrentMarkerLocation();
+	}
+
+	/**
+	 * Accuracy circle and view angle sector are drawn by the renderer itself at its own
+	 * "my location" position, which is a state separate from the marker position. That state is
+	 * not updated while the marker position is animated (unless the accuracy circle is visible),
+	 * so the sector could be left behind the location icon. Keep both in sync on every frame.
+	 */
+	private void updateMyLocationCirclePosition(@NonNull MapRendererView mapRenderer) {
+		CoreMapMarker locMarker = getCurrentMarker();
+		if (locMarker != null && locMarker.marker != null) {
+			mapRenderer.setMyLocationCirclePosition(locMarker.marker.getPosition());
+		}
 	}
 
 	private boolean setMarkerState(MarkerState markerState, boolean showHeading, boolean forceUpdate) {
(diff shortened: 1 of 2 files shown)


vshcherb:
Repro screenshots (emulator, OpenGL renderer, simulated positions, Location radius = Off, map panned away from my location): the view angle sector stays where the location icon was and the gap grows with the distance driven; the control run with the map following my location keeps the sector on the arrow.

(gen by AI)

Sector left behind the location icon

Control run: sector on the location icon

PR #25901, opened 2026-09-08
Routing to OsmAnd-shared, step 2: TurnType and ExitInfo

alex-dev-neo:
Second step of the routing migration, stacked on #25900. **Base is `shared-routing-1-collections`,
so the diff here is only this step.** GitHub retargets it onto `master` when #25900 merges.

> [!IMPORTANT]
> Needs osmandapp/OsmAnd-core-legacy#350 to ship **first**. See [Ordering](#ordering).

`TurnType` moves from `net.osmand.router` to `net.osmand.shared.routing`, together with `ExitInfo`.
It goes first on purpose: it is the class the C++ router constructs over JNI, so it proves that path
works while nothing larger depends on it yet.

## The Java visible API does not change

`@JvmStatic` on the companion keeps `TurnType.valueOf(...)`, `TurnType.getPrimaryTurn(...)` and the
constants resolving as statics, and the Kotlin property names produce exactly the previous accessors:

| Kotlin | Java sees |
|---|---|
| `val value` | `getValue()` |
| `var exitOut` | `getExitOut()` / `setExitOut()` |
| `var isSkipToSpeak` | `isSkipToSpeak()` / `setSkipToSpeak()` |
| `var isPossibleLeftTurn` | `isPossibleLeftTurn()` / `setPossibleLeftTurn()` |
| `var lanes` | `getLanes()` / `setLanes()` |

So of the 33 files touched, 27 changed by exactly one line: the import.

## The JNI contract

`java_wrap.cpp` resolves the class by name and its constructor by descriptor:

```c
jclass_TurnType = findGlobalClass(env, "net/osmand/router/TurnType");
jmethod_TurnType_init = env->GetMethodID(jclass_TurnType, "<init>", "(IIFZ[IZZ)V");
```

Only the **name** changes. Kotlin emits `(IIFZ[IZZ)V` for the primary constructor as long as the
parameters keep their order and none of them gains a default value — a default would make Kotlin add
a synthetic mask parameter and silently break the lookup. `TurnTypeJniContractTest` asserts the class
name, the descriptor and the static shape by reflection, because nothing else catches a mismatch
before the first native route calculation on a device.

## Ordering

The app consumes OsmAndCore as a prebuilt snapshot, so:

1. osmandapp/OsmAnd-core-legacy#350 merges
2. the OsmAndCore snapshot rebuilds
3. this PR merges

In any other order, native routing on Android fails at runtime with `NoSuchMethodError` and nothing
warns at build time. This ordering applies to most of the remaining steps too: JNI binds 30 Java
classes by name, and `RouteSegmentResult`, `RoutingContext`, `RoutingConfiguration`, `GeneralRouter`,
`RouteCalculationProgress`, `PrecalculatedRouteDirection` and `RouteDataObject` are all on that list.

## Testing

- 17 behaviour tests in `commonTest`, passing on JVM and Kotlin/Native: lane bit packing and its
  independence per slot, `lanesToString` / `lanesFromString` round trips including a trailing
  separator, `toXmlString` / `fromString` round trips for every turn type, roundabout exits, the
  left hand traffic mapping, direction counting, and the classification helpers.
- 5 contract tests in `jvmTest` for the JNI shape described above.
- The 453 existing routing tests in `OsmAnd-java` still pass, `RouteResultPreparationTest` among them,
  which exercises `TurnType` heavily.
- iOS simulator suite: 147 tests, 4 skipped benchmarks, 6 failures that are pre-existing on `master`
  (`GpxUtilitiesLoadTest`, `KGeoPointParserUtilTest`, `NetworkAPITest`).

## Next step

`shared-routing-3-*`, based on this branch. The remaining unblocked model classes are thinner than
they looked from a distance: `RoundaboutTurn`, `RoadSplitStructure`, `RouteDataResources` and
`RouteImporter` / `RouteExporter` all reach `RouteSegmentResult` or `RouteDataObject`, and
`MissingMapsCalculationResult` needs `WorldRegion`. So the next step is likely `RouteSegmentResult`
plus `RouteDataObject` with its `RouteRegion` / `RouteTypeRule` extracted from the OBF reader — a
bigger JNI surface, probably worth splitting in two.

<sub>Please merge with a merge commit rather than a squash: the follow up branches are stacked on this
history.</sub>

---

**Update — `TurnTypeJniContractTest` removed again**, by the commit now at the tip of this branch.


OsmAnd-java/src/main/java/net/osmand/router/TurnType.java
```diff
@@ -1,625 +0,0 @@
-package net.osmand.router;
```
vshcherb, line 1: Deleting from java could slow down development for some developers it's better to keep history and source code.

It's much better to code to invent some way of testing so TurnType.kt and TurnTYpe.java are the same all the time after commits

alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26000.

PR #25903, opened 2026-09-08
Routing to OsmAnd-shared, step 3: OpeningHoursParser

alex-dev-neo:
Third step of the routing migration, stacked on #25901. **Base is `shared-routing-2-model`, so the
diff here is only this step.** No `core-legacy` change is needed: `OpeningHoursParser` is one of the
few routing-adjacent classes JNI does not bind.

## Why this is a routing step

`RouteTypeRule` keeps the parsed `OpeningHours` for `:conditional` tags, so the whole model cluster —
`RouteRegion`, `RouteDataObject`, `RouteSegmentResult` — is blocked behind the parser. It cannot be
skipped without dropping conditional routing restrictions.

It is not routing code though: of its 18 consumers, most are POI screens and the OSM editor. Worth
knowing when reading the file list.

## The port

Structure and method order follow the Java original so the two read side by side. Two deliberate
differences:

- **`OpeningHoursTime` instead of `java.util.Calendar`**, a small wrapper over `kotlinx-datetime`
  exposing the same fields with the same numbering — month counting from zero, Sunday as weekday 1 —
  so expressions like `(i + 5) % 7` port across unchanged.
- **The canonical OSM day and month names are constants.** The Java version derived them from
  `DateFormatSymbols.getInstance(Locale.US)` and truncated to two or three letters, which made the
  canonical output of `toRuleString()` depend on the JVM's locale data. They are fixed OSM
  vocabulary, so they are now written out.

Localized names and 12 hour formatting still come from the platform, through the new
`PlatformDateNames`: `java.text` on JVM and Android, `NSDateFormatter` on iOS.

`PlatformSerializable` (an `expect interface`, `actual typealias` to `java.io.Serializable` on the
JVM) keeps the classes serializable. The OSM editor relies on that to carry a partially edited rule
through a `Bundle` across a configuration change. The alternative was to round trip those fragments
through a string, which is a change to the editor rather than to the port.

## Call site changes

Java callers pass epoch milliseconds where they passed a `Calendar`, and a language tag where they
passed a `Locale`. Most of the 16 touched app files changed only their import.

## Testing

The **541 assertion suite** from `OsmAnd-java` is ported to `commonTest` and passes on **JVM and
Kotlin/Native**. That corpus is what makes a port this size trustworthy.

Two assertions accept either of two renderings, because `java.text` and Foundation genuinely
disagree on locale data:

| | java.text | Foundation |
|---|---|---|
| `zh-TW` day period | 上午 / 下午 | 凌晨 / 清晨 / 晚上 (finer CLDR periods) |
| `ar` digits | Arabic-Indic ٤:٣٠ | Western 4:30 |

Both spellings are correct in their language. Comparison also ignores bidi marks, which the two
platforms place differently around a right to left day period.

| | tests | skipped | failures |
|---|---|---|---|
| `:OsmAnd-shared:jvmTest` | 154 | 4 | 0 |
| `:OsmAnd-shared:iosSimulatorArm64Test` | 148 | 4 | 6 pre-existing, unrelated |
| `:OsmAnd-java:test` (routing and shared) | 462 | 11 | 0 |

The routing suite now evaluates conditional tags through the Kotlin parser, so it covers the change
end to end. The 6 native failures are in `GpxUtilitiesLoadTest`, `KGeoPointParserUtilTest` and
`NetworkAPITest` and fail on `master` as well.

## Next step

With the parser in place, `RouteTypeRule`, `RouteRegion` and `RouteSubregion` can be extracted from
`BinaryMapRouteReaderAdapter` and moved, followed by `RouteDataObject` and `RouteSegmentResult`. That
one needs a paired `core-legacy` PR again, and its JNI surface is much larger than `TurnType`'s, so
it is likely worth splitting in two.

<sub>Please merge with a merge commit rather than a squash: the follow up branches are stacked on this
history.</sub>


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26001.

PR #25904, opened 2026-09-08
Routing to OsmAnd-shared, step 4: RouteTypeRule

alex-dev-neo:
Fourth step of moving the routing model into `OsmAnd-shared`. Stacked on #25903, review that one first.

## What moves

`RouteTypeRule` and its `RouteTypeCondition` leave `BinaryMapRouteReaderAdapter` for `net.osmand.shared.routing`. It is the leaf of the routing data model — `RouteDataObject`, `RouteRegion` and `RouteSegmentResult` all read encoding rules through it — and, unlike every other class in that cluster, the C++ core does not bind it by name, so this step needs **no paired `core-legacy` PR**.

## Why this is smaller than the plan in #25903

That PR said step 4 would take `RouteTypeRule`, `RouteRegion` and `RouteSubregion` together. Looking at the code, those three cannot go in one step:

- `RouteRegion.adopt()` builds `RouteDataObject`s and `RouteSubregion` holds a `List<RouteDataObject>`, while `RouteDataObject` holds a `RouteRegion` — the three are mutually dependent and have to move in one commit with `RouteDataObject`.
- `RouteRegion extends BinaryIndexPart`, and `java_wrap.cpp` reads the inherited package-private `length` and `filePointer` fields off it. Its `getFieldNumber()` returns a protobuf-generated constant. Neither survives a move to a module that has no protobuf and no `BinaryIndexPart`.

So the cluster is one large step of its own, and `RouteTypeRule` is the part that can be separated cleanly.

## Details that changed shape

- **`parseSpeed` / `NONE_MAX_SPEED`** live on `RouteDataObject`, which is still Java. They move to a shared `RouteDataUtils` that `RouteDataObject` delegates to, so the maxspeed parsing is not duplicated while that class waits its turn. `NONE_MAX_SPEED` stays a compile-time constant — verified with `javap -v`. `Algorithms.findFirstNumberEndIndex` gets a `KAlgorithms` twin for the same reason.
- **`String.intern`** has no common equivalent. `internString` keeps the pooling on jvm and android and is a no-op on ios, where Kotlin/Native has no string pool. Route encoding rules repeat the same tags across every loaded region, so this is worth keeping.
- **`StringExternalizable` is dropped.** `readFromBundle` was dead code — `RouteImporter` reads the two keys itself and calls `initRouteEncodingRule`. `writeToBundle` had one caller, `RouteExporter`, which now writes the keys directly. Keeping the interface would have pulled `RouteDataBundle` and `RouteDataResources` into the shared module.
- **`RouteRegion.completeRouteEncodingRules`** reached into a rule's private `conditions` list, which only worked while both classes were nested in `BinaryMapRouteReaderAdapter`. It goes through `getConditions()` and `setRuleId()` now.
- **`OpeningHoursTime.ofEpochMillis(millis, zoneId)`** is new. See below.

## The tools repository was broken by steps 2 and 3

`OsmAnd-tools` builds `OsmAnd-java` and `OsmAnd-shared` from source (`java-tools/settings.gradle` points at `../../android/`). Steps 2 and 3 moved `TurnType` and `OpeningHoursParser` without updating it, so two files there stopped compiling; this step breaks another twelve.

osmandapp/OsmAnd-tools#1680 fixes all of them. It needs the zone-aware `OpeningHoursTime` factory added here, because `SearchResultConverter` evaluates opening hours in the *client's* time zone, which the device-zone factories cannot express. **Merge order: this stack first, then the tools PR.**

## Verification

- `:OsmAnd-shared:jvmTest` — green, 19 new tests covering rule parsing, the two new `KAlgorithms` helpers and the new factory
- `:OsmAnd-java:test` — green
- `:OsmAnd-shared:iosSimulatorArm64Test` — 176 tests, the same 6 failures as master (`GpxUtilitiesLoadTest`, `KGeoPointParserUtilTest`, `NetworkAPITest`)
- Android app compiles
- `OsmAnd-tools`: all four java-tools projects compile against this branch

## Next

Step 5 is the `RouteRegion` / `RouteSubregion` / `RouteDataObject` cluster, which does need a paired `core-legacy` PR and a rethink of the `BinaryIndexPart` inheritance.

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapRouteReaderAdapter.java
+++ b/OsmAnd-java/src/main/java/net/osmand/binary/BinaryMapRouteReaderAdapter.java
@@ -14,15 +14,13 @@ import net.osmand.binary.OsmandOdb.OsmAndRoutingIndex.RouteEncodingRule;
 import net.osmand.binary.OsmandOdb.RestrictionData;
 import net.osmand.binary.OsmandOdb.RouteData;
 import net.osmand.binary.RouteDataObject.RestrictionInfo;
-import net.osmand.util.Algorithms;
 import net.osmand.util.MapUtils;
-import net.osmand.shared.util.OpeningHoursParser;
+import net.osmand.shared.routing.RouteTypeRule;
 
 import org.apache.commons.logging.Log;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Calendar;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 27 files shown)


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26002.

PR #25905, opened 2026-09-08
Routing to OsmAnd-shared, step 5: KTIntObjectMap

alex-dev-neo:
Fifth step of moving the routing model into `OsmAnd-shared`. Stacked on #25904, review that one first.

Small PR: one collection plus its tests. It exists because the next step cannot start without it.

## Why now

`RouteDataObject` keys its `names` by encoding rule id through a `gnu.trove.map.hash.TIntObjectHashMap`. That class is the centre of the next step, and it cannot move to a module that has no int keyed map. `TIntObjectHashMap` also appears in 19 other files in this repository, so this is not a single-use helper.

This completes the trove replacement started in #25900 (`KTLongObjectMap`, `KTLongHashSet`, `KTIntArrayList`).

## One deliberate difference from `KTLongObjectMap`

Same open addressing, same tombstones, same iterator — but the keys are mixed by **Fibonacci hashing** rather than the MurmurHash3 finalizer.

The first version reused Murmur, and the benchmark said it was the wrong choice. Keys in this map are dense ids — encoding rules, string table indices — which a full mixer scrambles at real cost and no benefit. Measured on the jvm, 300k entries and 600k lookups, best of 5 rounds:

| hash | keys | KTIntObjectMap | vs `TIntObjectHashMap` | vs `HashMap` |
|---|---|---|---|---|
| murmur | dense ids | 7.79 ms | 0.89x | 0.83x |
| **fibonacci** | dense ids | **3.46 ms** | **2.00x** | **2.18x** |
| murmur | random | 6.98 ms | 1.62x | 1.27x |
| **fibonacci** | random | **6.61 ms** | **1.75x** | **1.43x** |

Murmur was *slower than both references* on the distribution this map actually sees. `KTLongObjectMap` keeps Murmur — road ids and `x31 << 31 | y31` tile keys are structured in the low bits and genuinely need the mixing.

Kotlin/Native, **release** binary on the simulator, 200k entries:

| keys | KTIntObjectMap | vs `HashMap<Int, V>` | vs the same test on the jvm |
|---|---|---|---|
| dense ids | 4.70 ms | 3.04x | 1.54x slower |
| random | 7.10 ms | 2.17x | 1.40x slower |

In line with the 1.3–2.1x native/jvm ratio measured in #25900.

## Tests

Mirrors what `KTLongObjectMap` already has:

- `KTIntObjectMapTest` in `commonTest` — 12 tests, including a 60,000 operation differential run against `HashMap`
- `TroveComparisonTest.testIntObjectMapMatchesTrove` in `OsmAnd-java` — 200,000 operations against the real `TIntObjectHashMap`, runs in the normal suite
- a benchmark case in both `@Ignore`d benchmark classes

Verified: `:OsmAnd-shared:jvmTest` green, `:OsmAnd-java:test` green, `:OsmAnd-shared:iosSimulatorArm64Test` 189 tests with the same 6 failures as master.

## Next

Step 6 is the move this unblocks: `RouteRegion`, `RouteSubregion` and `RouteDataObject` together. They are mutually dependent, so they go in one commit, with a paired `core-legacy` PR and a decision on `RouteRegion extends BinaryIndexPart` — see the analysis in #25904.

### DIFF
--- a/OsmAnd-java/src/test/java/net/osmand/shared/util/collections/CollectionsTroveBenchmarkTest.java
+++ b/OsmAnd-java/src/test/java/net/osmand/shared/util/collections/CollectionsTroveBenchmarkTest.java
@@ -14,6 +14,7 @@ import org.junit.Ignore;
 import org.junit.Test;
 
 import gnu.trove.list.array.TIntArrayList;
+import gnu.trove.map.hash.TIntObjectHashMap;
 import gnu.trove.map.hash.TLongObjectHashMap;
 import gnu.trove.set.hash.TLongHashSet;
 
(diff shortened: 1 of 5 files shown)


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26003.

PR #25906, opened 2026-09-08
Routing to OsmAnd-shared, step 6: RouteDataObject and the route regions

alex-dev-neo:
Sixth step of moving the routing model into `OsmAnd-shared`. Stacked on #25905, review that one first.

The big one. Pairs with osmandapp/OsmAnd-core-legacy#351 and osmandapp/OsmAnd-tools#1681.

## What moves, and why all at once

`RouteDataObject`, `RouteRegion` and `RouteSubregion` go to `net.osmand.shared.routing`. They cannot be split: a road holds the region it was read from, the region builds roads in `adopt()`, and a subregion caches the roads it loaded. The three are mutually dependent, so they move in one commit.

`BinaryIndexPart` moves with them, to `net.osmand.shared.binary`. `RouteRegion` extends it, and a class in `commonMain` cannot extend a Java one — so the base had to follow. Its six other subclasses (`MapIndex`, `PoiRegion`, `AddressRegion`, `CitiesBlock`, `TransportIndex`, `HHRouteRegion`) only change an import.

The one thing that could not come along is the protobuf constant behind `getFieldNumber()`. The shared module has no generated obf classes, so `RouteRegion` states the field number directly, with the proto field it mirrors named beside it. It is frozen by the file format.

## Transliteration

`RouteDataObject.getName()` romanises names, and that was `Junidecode`, a jvm library. It becomes `KTransliterationHelper` with an `expect fun toLatin`: junidecode on jvm and android, Foundation's `NSStringTransformToLatin` on ios. `net.osmand.util.TransliterationHelper` keeps its name — the C++ core calls it by name — and delegates, so the other seven Java callers are untouched.

## Two signatures changed shape

| was | is | why |
|---|---|---|
| `bearingVsRouteDirection(Location)` | `bearingVsRouteDirection(Float?)` | the shared module has no `Location` yet; `Location.bearingOf(loc)` builds the argument at the four call sites |
| `calculateHeightArray(LatLon)` | `calculateHeightArray(KLatLon)` | one call site |

## Behaviour differences, all deliberate

- `RouteDataObject.names` is a `KTIntObjectMap` (from #25905) rather than a `TIntObjectHashMap`. Same api, so only declared types change at call sites. It cannot hold a null value, so `RouteRegion.adopt` skips a name key whose value is null instead of storing the key with null. Reads answer null either way.
- `region` is not nullable any more, so the two `IllegalStateException` guards in `compareRoute` are gone — nothing in either repository constructs a road without a region.
- `removePointType` returns early when there are no point types. The original condition was `pointTypes != null || ind < pointTypes.length`, which throws on null rather than skipping.

Two oddities of the original are **kept on purpose**, with a comment: `setRestrictionVia` copies from `restrictions` rather than `restrictionsVia`, and `parseLength` narrows through a double multiply. Changing either would change which turns get restricted or which speeds the router sees.

## A precision fix folded into #25904

Porting `parseLength` showed that `RouteDataUtils.parseSpeed`, added in #25904, wrote `f /= 3.6f` where the Java original had `f /= 3.6` — a float divide instead of a double divide narrowed once. That differs in the last bit for almost every common `maxspeed` value, and those values feed routing cost. #25904 has been amended to narrow the same way Java did, and this branch and #25905 rebased onto it.

## Verification

- `:OsmAnd-java:test` — 570 tests, 0 failures. This is the real check: 230 of them (`RouteTestingTest`, `RouteResultPreparationTest`, `AlternativeRoutesTest`, `ApproximationTest`) read obf files and calculate routes end to end through the ported classes.
- `:OsmAnd-shared:jvmTest` — 217 tests, 0 failures, including a new 22 test `RouteDataObjectTest` in `commonTest` covering the encoding rules, tag lookup, geometry, restrictions, `adopt` and the parsers. That one also runs on ios, which the Java suite cannot do.
- `:OsmAnd-shared:iosSimulatorArm64Test` — 211 tests, the same 6 failures as master.
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1681.
- JNI contract checked with `javap -s`: every field the core reads keeps its name and descriptor, and the two constructors keep their shape apart from the package in their descriptors.

**Merge order:** osmandapp/OsmAnd-core-legacy#351 → OsmAndCore snapshot rebuild → this stack → osmandapp/OsmAnd-tools#1681. The app consumes OsmAndCore as a prebuilt artifact, so the wrong order breaks native routing at runtime with nothing failing at build time.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26004.

PR #25908, opened 2026-09-08
Routing to OsmAnd-shared, step 7: the route bundles

alex-dev-neo:
Seventh step of moving the routing model into `OsmAnd-shared`. Stacked on #25906, review that one first. Pairs with osmandapp/OsmAnd-tools#1682.

Small PR, 13 files. It exists because `RouteSegmentResult` cannot move until this does.

## Why this is its own step

`RouteSegmentResult` is what step 8 is about, and it serialises itself through `StringExternalizable<RouteDataBundle>`, `RouteDataBundle` and `RouteDataResources` — all three Java only. Moving them together with the class would have put a 798 line port and a rewrite of the route gpx serialisation in one diff, and the serialisation is the risky half: these bundles are what route gpx files are written from and read back into.

So the bundles move first and `RouteSegmentResult` stays in Java, implementing the shared interface. Step 8 is then a plain move.

## The two StringBundles

There were two: `net.osmand.binary.StringBundle` (used by the legacy `net.osmand.gpx.GPXUtilities`) and `net.osmand.shared.util.StringBundle` (used by the shared gpx writer). `RouteExporter` already bridged them with `kStringBundleJustStrings`, flattening every item to a string before handing the route to the shared writer.

`writeToBundle` now writes into the shared bundle directly and that bridge is gone. `RouteImporter` still reads through the legacy reader, so the same flattening lives there for now, and leaves with that reader. The Java `StringBundle` stays for `GPXUtilities`, which also needs the Java `StringBundleXmlWriter` and its `org.xmlpull` serializer — untangling that belongs with `net.osmand.gpx`, not here.

## Three things the shared StringBundle was missing

**`putFloat(key, value, maxDigits)` printed different text.** It used `round(value * factor) / factor` and `toString()`, which gives `"3.0"` where `DecimalFormat("#.##")` gives `"3"`. Route gpx files already in the wild were written by the Java one, so this had to reproduce it exactly: half even rounding of the exact value, trailing zeros and a bare decimal point dropped, and the plain text of the number outside the two-to-six-digit range Java had formatters for.

`StringBundleFormatJvmTest` runs both over a fixed table and 100,000 random values per digit count and asserts the strings are identical. Nothing used the three argument `putFloat` in shared before, so this changes no existing output.

**`asBoolean` ignored its default** and answered null for anything that was not `"true"` or `"false"`. Java read it with `Boolean.parseBoolean`, which answers false.

**`newInstance()` and a constructor over an existing bundle**, which `RouteDataBundle` needs, plus `putMap` for a `KTIntObjectMap` (it was commented out waiting for that type, which #25905 added).

## RouteDataResources holds KLocation

`net.osmand.Location` is not in the shared module. `Location.toShared()` converts, at the three places that build a `RouteDataResources` — `RouteExporter`, `RouteImporter` and the tools' `RoutingService`. `RouteImporter` builds its locations itself, so it just builds `KLocation`s now.

## Verification

`writeToBundle` and `readFromBundle` had no test at all, which is uncomfortable for the code that writes user route files. `RouteBundleRoundTripTest` now drives a segment through both and checks it comes back with the same geometry, tags, names, turn type, times and speeds — and pins the *text* of the numbers in the bundle, which is what lands in the file.

- `:OsmAnd-java:test` — 574 tests, 0 failures, including the 230 routing regression tests
- `:OsmAnd-shared:jvmTest` — 220 tests, 0 failures
- `:OsmAnd-shared:iosSimulatorArm64Test` — 211 tests, the same 6 failures as master
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1682

## Next

Step 8 moves `RouteSegmentResult` itself, with a paired `core-legacy` PR for its class name in the JNI constructor descriptor.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26013.

PR #25911, opened 2026-09-08
Routing to OsmAnd-shared, step 8: RouteSegmentResult

alex-dev-neo:
Eighth step of moving the routing model into `OsmAnd-shared`. Stacked on #25908, review that one first. Pairs with osmandapp/OsmAnd-core-legacy#352 and osmandapp/OsmAnd-tools#1683.

`RouteSegmentResult` — the type a calculated route is a list of — moves to `net.osmand.shared.routing`. With it, the whole routing *result* model is shared: `TurnType`, `ExitInfo`, `RouteTypeRule`, `RouteRegion`, `RouteSubregion`, `RouteDataObject`, the bundles, and now this.

## What callers notice

The point accessors. `getStartPoint()`, `getEndPoint()` and `getPoint(i)` answer with a `KLatLon`, because the shared module has no `LatLon`. Fourteen places wanted a `LatLon` and convert with the new `LatLon.of(KLatLon)`; the ones that only fed the result to `MapUtils.getDistance` call `KMapUtils` directly instead.

Everything else is an import. 52 files.

## A bug from step 7, found and fixed there

Porting `writeToBundle` surfaced one: `int[][]` rows can be null — that is what point types look like for a point carrying none — and the shared `StringBundle.putArray` I switched to in #25908 declared its rows non-null and would have thrown on the first route with untagged points. Reading them back split differently from `String.split` as well, and a negative value that rounds to zero lost its sign.

#25908 has been amended with all three, plus `StringBundleCompatTest`, which runs both `StringBundle` implementations over the same values and asserts they agree — it is what found them. This branch is rebased onto it.

## Tests

`RouteBundleRoundTripTest` moves from `OsmAnd-java` to `commonTest` along with the class it covers, so the route gpx serialisation is now exercised on ios as well as the jvm. It gained a case for point types with empty rows.

- `:OsmAnd-java:test` — 576 tests, 0 failures, including the 230 routing regression tests that read obf files and calculate real routes through the moved class
- `:OsmAnd-shared:jvmTest` — 224 tests, 0 failures
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1683
- JNI contract checked with `javap -s`: the constructor the core calls keeps its shape, only the package in its descriptor changes

**Merge order:** osmandapp/OsmAnd-core-legacy#350, #351, #352 → OsmAndCore snapshot rebuild → this stack → the tools PRs. The app consumes OsmAndCore as a prebuilt artifact, so the wrong order breaks native routing at runtime with nothing failing at build time.

## What is left

`RoutingContext`, `RoutingConfiguration`, `GeneralRouter`, `RouteCalculationProgress` and `BinaryRoutePlanner` — the router itself rather than its result. All of them are JNI bound, and `RoutingContext` reaches into the obf reader, so they are a different kind of problem from these eight steps.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26017.

PR #25912, opened 2026-09-08
Routing to OsmAnd-shared, step 9a: VehicleRouter

alex-dev-neo:
Ninth step of moving the routing model into `OsmAnd-shared`. Stacked on #25911, review that one first.

Small: two new interfaces, four files touched. It exists to unblock `GeneralRouter`, which cannot move while the interface it implements is Java.

## The problem it solves

`VehicleRouter` is the interface a routing profile answers through — may this road be used, how fast is it taken, what do its obstacles and turns cost. Everything on it already speaks in `RouteDataObject` and `GeneralRouterProfile`, both shared since #25906 and #25901.

Except one method:

```java
double calculateTurnTime(RouteSegment segment, RouteSegment prev);
```

`RouteSegment` here is `BinaryRoutePlanner.RouteSegment` — the A* graph node. It carries a parent pointer, a distance from the start, a distance to the end, the planner's linked lists of connected segments. Moving *that* means moving the planner and its `RoutingContext`, which is a much later step.

## What it actually needs

The method body asks a segment four things: `getRoad()`, `getSegmentStart()`, `getSegmentEnd()`, `isPositive()`. Nothing else on `VehicleRouter` mentions a segment at all.

So `RoadTraversal` is those four — one road, where the route enters and leaves it, and which way that runs. `BinaryRoutePlanner.RouteSegment implements RoadTraversal` with the methods it already had, unchanged, and the planner stays in `OsmAnd-java`.

## No paired core-legacy PR

`VehicleRouter` does not appear in `java_wrap.cpp` — checked. The C++ core reaches the profile through `GeneralRouter` and its nested classes, which is step 9b.

## Verification

- `:OsmAnd-java:test` — 576 tests, 0 failures. The 230 routing regression tests calculate real routes, and `calculateTurnTime` is on that path, so a wrong turn penalty would show up as a changed route.
- `:OsmAnd-shared:jvmTest` — 224 tests, 0 failures
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1684

Nothing new to unit test here — both files are interfaces with no behaviour of their own.

## Next

Step 9b is `GeneralRouter` itself: 1355 lines and five nested types, four of them bound by name in `java_wrap.cpp` with about twenty fields and ten methods between them. That one needs the largest paired core-legacy PR of the series.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26005.

PR #25913, opened 2026-09-08
Routing to OsmAnd-shared, step 9b: GeneralRouter

alex-dev-neo:
Stacked on #25912, review that one first. Pairs with osmandapp/OsmAnd-core-legacy#353 and osmandapp/OsmAnd-tools#1685.

`GeneralRouter` — the routing profile itself: the vehicle attributes `routing.xml` declares, and the rules that decide for every road whether it may be used, how fast it is taken, and what its obstacles and turns cost. 1355 lines, five nested types.

## The largest JNI surface of the series

The core reads sixteen fields off the router and walks its rules through four classes:

| class | what the core touches |
|---|---|
| `GeneralRouter` | `restrictionsAware`, `sharpTurn`, `shortWaySharpTurn`, `slightTurn`, `shortWaySlightTurn`, `roundaboutTurn`, `shortWayRoundaboutTurn`, `minSpeed`, `defaultSpeed`, `maxSpeed`, `maxVehicleSpeed`, `heightObstacles`, `shortestRoute`, `objectAttributes`, `hhNativeFilter`, `hhNativeParameterValues`, `getParameters()`, `getImpassableRoadIds()` |
| `RoutingParameter` | `getType()`, `getDefaultBoolean()`, `getDefaultNumeric()` |
| `RouteAttributeContext` | `getRules()`, `getParamKeys()`, `getParamValues()` |
| `RouteAttributeEvalRule` | `selectValueDef`, `selectType`, `selectExpression`, `getTagValueCondDefTag()`, `getTagValueCondDefValue()`, `getTagValueCondDefNot()`, `getParameters()`, `getExpressions()` |
| `RouteAttributeExpression` | `values`, `expressionType`, `valueType` |

Every name and descriptor is unchanged — checked with `javap -s` against the compiled Kotlin, including that the private fields keep their own names and that `getParameters()` still answers a `java.util.Map` for the core to walk with `entrySet`.

Two of the twelve strings the core-legacy PR changes are the `router` field of `RoutingConfiguration` and of `TransportRoutingConfiguration`, which stay in Java: only the type of the field moved.

**Two `synchronized` blocks.** `commonMain` has none, so they use `KLock`, a plain monitor on jvm and android. The comment on the static one still holds: a built router shares its rule tables with the root it was built from, so the lock has to span all instances.

**`printRules(PrintStream)`.** No callers anywhere, in either repository — a debugging aid. It appends to a `StringBuilder` now.

**`RoutingParameter.getDefaultString()`** used `String.format(Locale.US, "%.1f", value)`. The app stores that text as *the default of the preference the parameter lives in* — `settings.getCustomRoutingProperty(p.getId(), p.getDefaultString())` — so it has to keep saying exactly the same thing.

`GeneralRouterFormatJvmTest` runs the common implementation and `String.format` over a fixed table and 100,000 random values. It immediately caught that `kotlin.math.round` rounds **half to even** where `String.format` rounds **half up**, which would have moved every `.x5` default by a digit and silently changed stored preferences.

## Smaller shape changes

- `RouteAttributeExpression` is a plain nested class holding the router rather than an inner one: Kotlin forbids a companion object in an inner class, and its seven expression constants live there. Its JVM name is unchanged.
- `java.util.BitSet` becomes `KBitSet`, `TIntArrayList` and `TLongHashSet` become their `KT` twins — all three added in #25900 for exactly this class. `TransportRoutingConfiguration` builds one of those bit sets and follows.

## Verification

- `:OsmAnd-java:test` — 576 tests, 0 failures. The 230 routing regression tests calculate real routes, and this class decides every cost in them, so any drift shows up as a changed route.
- `:OsmAnd-shared:jvmTest` — 226 tests, 0 failures
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1685

**Merge order:** osmandapp/OsmAnd-core-legacy#350 to #353 → OsmAndCore snapshot rebuild → this stack → the tools PRs.

## What is left

`RoutingConfiguration`, `RouteCalculationProgress`, then `RoutingContext` and `BinaryRoutePlanner`. The last two hold an open obf reader and a tile cache, so they are a different problem again.

---

**Update — `RouteDataObject.region` is nullable again**, by the commit now at the tip of this branch.

Step 6 ported java's `public final RouteRegion region;` as a non-null `val region: RouteRegion`. That is not a port, it is a change of contract: Kotlin puts a `checkNotNullParameter` in the constructors, and the C++ router legitimately builds roads with **no** region — `convertRouteSegmentResultToJava` leaves it NULL when it cannot match the road's region back — which `NativeLibrary.initRouteRegion` was written to repair, with a comment saying so.

The field and both constructors take `RouteRegion?` now, and every read through it asserts with `!!`, throwing where java threw and nowhere else.

It lands on this branch rather than step 6 because the fix spans four files that arrive in three steps — `RouteDataObject` and `RouteRegion` in #25906, `RouteSegmentResult` in #25911, `GeneralRouter` here — and this is the first branch that has all four.

**How it was found, and why nothing caught it:** the root `build.gradle` sets `isAndroidBuild=true` for every test task, and `RouterUtilTest.getNativeLibPath()` returns null when it sees that — so `ApproximationNativeTest`, `RouteTestingNativeTest` and `RouteResultPreparationNativeTest` have all been running the java planner. With the C++ library built for the host and that property lifted, they are 24, 85 and 114 tests through the real JNI: the 24 fail without this change and all 223 pass with it.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26006.

PR #25917, opened 2026-09-08
Routing to OsmAnd-shared, step 10: RouteCalculationProgress

alex-dev-neo:
Stacked on #25913, review that one first. Pairs with osmandapp/OsmAnd-core-legacy#354 and osmandapp/OsmAnd-tools#1686.

`RouteCalculationProgress` — the live state of one route calculation: how far it has got, what it loaded, and how long that took. `FastRoutingState` comes with it: it is the state machine behind the one int field that says whether the fast hierarchical attempt is still viable. 263 + 83 lines, no dependencies beyond each other.

## The JNI surface

Both routers write here — the Java planner directly, and the C++ core through JNI, which resolves **thirty fields** by name and descriptor, five of them private:

| | fields |
|---|---|
| public | `segmentNotFound`, `distanceFromBegin`, `distanceFromEnd`, `isCancelled`, `totalEstimatedDistance`, `totalApproximateDistance`, `approximatedDistance`, `routingCalculatedTime`, `directSegmentQueueSize`, `reverseSegmentQueueSize`, `visitedSegments`, `visitedDirectSegments`, `visitedOppositeSegments`, `directQueueSize`, `oppositeQueueSize`, `iteration`, `timeNanoToCalcDeviation`, `timeToLoad`, `timeToLoadHeaders`, `timeToFindInitialSegments`, `timeToCalculate`, `loadedTiles`, `unloadedTiles`, `loadedPrevUnloadedTiles`, `distinctLoadedTiles` |
| private | `fastRoutingStatusOrdinal`, `hhIterationStep`, `hhTargetsDone`, `hhTargetsTotal`, `hhCurrentStepProgress` |

`GetFieldID` does not care about visibility, and Kotlin gives a private property a private field of the same name, so the five stay reachable. The traffic runs both ways: the core writes fifteen of them and reads back `isCancelled` and `fastRoutingStatusOrdinal`.

`RoutingContext` also holds the progress in a field the core resolves by the descriptor `Lnet/osmand/router/RouteCalculationProgress;`, so the **package name** is part of the contract too. That is the second of the two strings osmandapp/OsmAnd-core-legacy#354 changes.

Every name and descriptor is unchanged — checked with `javap -s` against the compiled Kotlin, and pinned by `RouteCalculationProgressJniContractTest`, which walks the same thirty pairs by reflection. `FastRoutingState` is **not** bound by name: the native side has its own `FastRoutingState` in C++, and JNI never looks the Java one up.

**`implements Serializable`.** Nothing serializes it — the only mention of `Serializable` anywhere in the router package was that declaration and its `serialVersionUID`. Dropped.

**`getInfo()` built its maps with a `TreeMap`,** and the key order is user visible: `RouteResultPreparation` prints the entries in iteration order into the routing log, and `HHRoutePlanner` prints them to stdout. Common code has no sorted map, so the keys are sorted once on the way out, and `RouteCalculationProgressOrderJvmTest` compares the result against a real `java.util.TreeMap` rather than against a hardcoded list. The boxed types matter too — the log formats a `Float` with one decimal and everything else with `toString` — so the common test asserts `Int` for the counts and `Float` for the seconds.

**`missingMapsCalculationResult`** was typed as `MissingMapsCalculationResult`, which holds a whole `RoutingContext` and reaches into `OsmandRegions` and `WorldRegion`. None of that can move yet, so the field is typed as a new `MissingMapsResult` interface carrying the one method the routing code asks it for, `getErrorMessage()`. The Java class implements it, and the three call sites that want the region lists out of it cast:

- `RouteRecalculationHelper.getCurrentMissingMapsCalculationResult()`
- `RouteProvider`, `CalculateMissingMapsOnlineTask` — both hand it to `RouteCalculationResult`, whose UI needs the regions

## Why not `RoutingConfiguration` in the same step

It was the other half of the plan, and it does not fit yet. Three things block it, all rooted in direction points:

- `RoutingConfiguration.DirectionPoint extends net.osmand.osm.edit.Node`, and `commonMain` cannot extend a Java class — the same wall as `BinaryIndexPart` in #25906. `Node` brings the whole OSM entity model with it.
- `getNativeDirectionPoints()` is bound by the descriptor `()[Lnet/osmand/NativeLibrary$NativeDirectionPoint;`. A class in shared cannot name a type that lives in OsmAnd-java.
- `QuadTree<Node>` is in the public `Builder.setDirectionPoints` API, used by the app's `DirectionPointsHelper` and by `MapRouterLayer` in tools.

The XML itself is not the problem — shared already has `net.osmand.shared.xml.XmlPullParser` — but `parseFromInputStream(InputStream)` and `getResourceAsStream("routing.xml")` need somewhere to live. It is a step of its own anyway: 559 lines and 152 references in `java_wrap.cpp`, the biggest JNI surface in the series.

## Verification

- `:OsmAnd-java:test` — 576 tests, 0 failures. The 230 routing regression tests calculate real routes and the planner writes progress on every one of them.
- `:OsmAnd-shared:jvmTest` — 254 tests, 0 failures (22 new common, 6 new jvm)
- `:OsmAnd-shared:iosSimulatorArm64Test` — 237 tests, the same 6 failures as master
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1686

**Merge order:** osmandapp/OsmAnd-core-legacy#350 to #354 → OsmAndCore snapshot rebuild → this stack → the tools PRs.

## What is left

`RoutingConfiguration` once direction points have somewhere to go, then `RoutingContext` and `BinaryRoutePlanner`, which hold an open obf reader and a tile cache.

---

**Update — `RouteCalculationProgressJniContractTest` removed again**, by the commit now at the tip of this branch.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26007.

PR #25919, opened 2026-09-08
Routing to OsmAnd-shared, step 11: direction points

alex-dev-neo:
Stacked on #25917, review that one first. Pairs with osmandapp/OsmAnd-core-legacy#355 and osmandapp/OsmAnd-tools#1687.

Direction points are the markers the app drops to steer a route around one side of a road — the "avoid roads" files — and the GeoJSON points OsmAndMapCreator loads. They are not in any map: while a routing tile loads, the router finds the road each point belongs to, splices an extra point into that road's geometry and gives it the point's tags, so the routing rules see them as if the map had carried them.

This is the step `RoutingConfiguration` has been waiting on. It is three separate walls, and all three are here.

## `DirectionPoint extends net.osmand.osm.edit.Node`

`commonMain` cannot extend a Java class — the same wall as `BinaryIndexPart` in #25906. But `Node` brings the whole OSM entity model with it (`Entity`, `EntityId`, version and modify flags, changed-tag tracking), and nothing in routing wanted any of it: the router reads a place, a tag map, and the bookkeeping the class adds itself. So the shared `DirectionPoint` carries lat/lon and tags directly, and the entity model stays in OsmAnd-java.

## `getNativeDirectionPoints()` returns `NativeLibrary$NativeDirectionPoint[]`

The core resolves that method by the descriptor `()[Lnet/osmand/NativeLibrary$NativeDirectionPoint;`, and a class in shared cannot name a type from OsmAnd-java. So `NativeDirectionPoint` moves out of `NativeLibrary` as well.

Its three fields — `x31 I`, `y31 I`, `tags [[Ljava/lang/String;` — keep their names and descriptors, checked with `javap -s`, so osmandapp/OsmAnd-core-legacy#355 changes two strings: the class lookup and the method descriptor that names the class in its return type. Both sides are pinned by tests now: `NativeDirectionPointJniContractTest` in shared, and `RoutingConfigurationDirectionPointsTest` in OsmAnd-java for the method descriptor, which lives on the class that has not moved yet.

## `QuadTree<Node>` in the public `Builder` API

`setDirectionPoints` takes a `KQuadTree<DirectionPoint>` now. Both producers already built their points from scratch — the app parses GeoJSON in `DirectionPointsTask`, `MapRouterLayer` parses GeoJSON of its own — so they build direction points instead of OSM nodes. OsmAndMapCreator's map viewer draws `Entity`, so it still shows each point through a node carrying its tags.

## Two things worth naming

**The Builder still copies.** Every point it was given is copied into the configuration it builds, and that is not incidental: the router writes onto a point as it attaches it — the road it is connected to, the projection, the distance a nearer road has to beat — and a `Builder` is reused across calculations (`MapRouterLayer` keeps one loaded file for the session). The copy is what stops one route from starting where the last one finished. `RoutingConfigurationDirectionPointsTest` builds twice from one tree and asserts the two calculations cannot see each other's state.

**Tag keys are still lower cased.** `Entity.putTag` did that, and the router looks the tags up in the routing rules, which are lower case. One difference: `Entity` used `String.toLowerCase()` with the *default locale*, so a key containing `I` becomes a dotless `ı` on a Turkish device and stops matching; Kotlin's `lowercase()` is locale invariant. Keys in these files are already lower case, so this only removes the trap.

## Verification

- `:OsmAnd-java:test` — 580 tests, 0 failures (4 new)
- `:OsmAnd-shared:jvmTest` — 268 tests, 0 failures (11 new common, 3 new jvm)
- `:OsmAnd-shared:iosSimulatorArm64Test` — 248 tests, the same 6 failures as master
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1687

**Merge order:** osmandapp/OsmAnd-core-legacy#350 to #355 → OsmAndCore snapshot rebuild → this stack → the tools PRs.

## What is left

`RoutingConfiguration` itself is now unblocked apart from `parseFromInputStream(InputStream)` and the `getResourceAsStream("routing.xml")` lookup — the XML is not a problem, shared already has `net.osmand.shared.xml.XmlPullParser`. Then `RoutingContext` and `BinaryRoutePlanner`, which hold an open obf reader and a tile cache.

---

**Update — `NativeDirectionPointJniContractTest` removed again**, by the commit now at the tip of this branch.

---

**Update — the `getNativeDirectionPoints` descriptor assertion removed too**, by the commit now at the tip of this branch, for the same reason as the JNI contract tests.

The return type it checked is enforced by javac at every call site, including the three tests left in the class. Its own contribution was the package name as a string, and a test in this build cannot see `java_wrap.cpp` — it pins this side's copy of the name, not the agreement between the two sides. What the builder actually does is still covered: no points, place and tags carried through, and a fresh copy per calculation.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26008.

PR #25920, opened 2026-09-08
Routing to OsmAnd-shared, step 12: RoutingConfiguration

alex-dev-neo:
Stacked on #25919, review that one first. Pairs with osmandapp/OsmAnd-core-legacy#356 and osmandapp/OsmAnd-tools#1688.

`RoutingConfiguration` — everything one route calculation is configured with: which profile to route on, how much memory the tile cache may take, and the tweaks `routing.xml` declares. A configuration is built, not edited: the `Builder` holds what the XML said and hands out a fresh configuration per calculation. 525 lines, of which about 230 are the XML parser.

The three walls this was stuck behind came down in #25919. What was left was the file it reads and the stream it reads it from.

## The JNI surface

The core resolves twelve fields and one method:

| | |
|---|---|
| fields | `nativeMemoryLimitation J`, `heuristicCoefficient F`, `minPointApproximation F`, `minStepApproximation F`, `maxStepApproximation F`, `smoothenPointsNoRoute F`, `penaltyForReverseDirection D`, `ZOOM_TO_LOAD_TILES I`, `planRoadDirection I`, `routeCalculationTime J`, `routerName Ljava/lang/String;`, `router Lnet/osmand/shared/routing/GeneralRouter;` |
| method | `getNativeDirectionPoints ()[Lnet/osmand/shared/routing/NativeDirectionPoint;` |

All unchanged — checked with `javap -s` and pinned by `RoutingConfigurationJniContractTest`. `RoutingContext` also holds the configuration in a `config` field the core resolves by descriptor, so the package is part of the contract; `RoutingConfigurationDefaultTest` pins that from OsmAnd-java, where the field is declared. osmandapp/OsmAnd-core-legacy#356 changes exactly those two strings.

## `routing.xml` has not moved

OsmAnd-java's `collectRoutingResources` task copies it out of the `resources` repository into `net/osmand/router/`, and the shared class reads it from that path through an `expect`/`actual` — the same lookup `getResourceAsStream("routing.xml")` did. Moving the file means changing two builds and the Android resource packaging, which is a build change and belongs on its own.

iOS has no classpath, and its `routing.xml` belongs to the C++ core, so there is nothing to open there and `getDefault()` says so rather than pretending.

The check that matters is the 230 routing regression tests: every one of them calls `RoutingConfiguration.getDefault()`, so if the lookup missed, they would all fail rather than pass with a different route.

## Four things could not be ported as they were

**`parseFromInputStream(InputStream)`** takes an okio `Source` now, with a path overload for Java callers. okio is an `implementation` dependency of OsmAnd-shared, so its `Path` is not on OsmAnd-java's classpath and a `KFile`-only signature does not even resolve there — `new KFile(path)` fails with "cannot access Path". Every caller reads a file, so the path overload covers `AppInitializer`, `TestRouting` and `DataExtractionSettings` alike.

**The source is closed in a `finally`.** Java closed the stream on the way out only, so a failed parse leaked it — on the path where the app then deletes the file.

**`java.util.Stack` is not in common code.** The rule stack is a `MutableList` pushed and popped at the end. It is also read bottom-up in one place, which a list indexes the same way.

**`registerNewRule` and the comparison conditions take non-null strings** — that is how they were ported in #25913. So a `<select>` without a `value`, or a `<gt>` without both, now fails the parse instead of registering a rule that could never evaluate. Every one in `routing.xml` has them (the three that grep finds without a `value` are inside a comment). `AppInitializer` already throws away a custom profile it cannot read, and now catches this too.

One smaller thing: `attributes.put("routerName", …)` can no longer be null. It only could when a config declared no `defaultProfile` and matched no router, and the one place that reads that map skips the key.

## Verification

- `:OsmAnd-java:test` — 583 tests, 0 failures (3 new). The 230 routing regression tests calculate real routes off `getDefault()`, so both the resource lookup and every parsed rule are exercised.
- `:OsmAnd-shared:jvmTest` — 289 tests, 0 failures (16 for the parser and the Builder, 5 for the JNI contract)
- `:OsmAnd-shared:iosSimulatorArm64Test` — 248 tests, the same 6 failures as master
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1688

The parser tests live in `jvmTest`, not `commonTest`: the iOS `XmlPullParser` is an API the host app injects at startup, so a Kotlin/Native test binary cannot parse XML at all — that is why `GpxUtilitiesLoadTest` is one of the six failures on master. Everything they cover is common code.

**Merge order:** osmandapp/OsmAnd-core-legacy#350 to #356 → OsmAndCore snapshot rebuild → this stack → the tools PRs.

## What is left

`RoutingContext` and `BinaryRoutePlanner`, which hold an open obf reader and a tile cache — a different problem again, and worth auditing before promising a shape for them.

---

**Update — `RoutingConfigurationJniContractTest` removed again**, by the commit now at the tip of this branch.

---

**Update — the `RoutingContext.config` field type assertion removed**, by the commit now at the tip of this branch, for the same reason as the JNI contract tests.

It read the declared type of a field that `RoutingContext.java` declares two lines away in this same module, so javac enforces it and the assertion could only fail together with the build. The two tests that stay are the ones only this module can run: the default configuration really is parsed out of the `routing.xml` that `collectRoutingResources` copies into `net/osmand/router/`, and it is parsed once.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26014.

PR #25922, opened 2026-09-08
Routing to OsmAnd-shared, step 13: PrecalculatedRouteDirection, RouteConditionalHelper

alex-dev-neo:
Stacked on #25920, review that one first. Pairs with osmandapp/OsmAnd-core-legacy#357 and osmandapp/OsmAnd-tools#1689.

## Why `RoutingContext` is not in this step

The context and `BinaryRoutePlanner` are one unit: `RoutingSubregionTile.routes` is a map of `BinaryRoutePlanner.RouteSegment`, and the context **builds the planner's graph nodes** as it loads each tile. Neither can move without the other, and together they are about 2300 lines with a JNI surface of their own.

They also need two platform seams first, both interface sized. `RoutingContext` asks `BinaryMapIndexReader` for five things — `getRoutingIndexes`, `searchRouteIndexTree`, `loadRouteIndexData`, `getHHRoutingIndexes().isEmpty()`, `getFile().getName()` — plus a `SearchRequest` it uses only as a bounding box; every type in those signatures is already shared. Of `NativeLibrary` it uses `loadRouteRegion`, the static `deleteNativeRoutingContext`, and `NativeRouteSearchResult`. That is the same narrowing `VehicleRouter` got with `RoadTraversal` in #25912, not a reason to move the obf reader.

So this step moves two of the things the context holds, and leaves the context and the planner for a step that takes them together.

## `PrecalculatedRouteDirection`

A route that has already been found, indexed by place, with each point carrying the time still left to the end. The A* heuristic can then ask "how long from here, if I follow the route I already have" instead of guessing from straight-line distance — which is what makes recalculating a long route cheap: the search stays in a corridor around the old one and only the ends are worked out again.

The core reads eight fields off it and runs the same estimate itself: `tms [F`, `pointsX [I`, `pointsY [I`, `minSpeed F`, `maxSpeed F`, `followNext Z`, `startFinishTime F`, `endFinishTime F`. All private on the java side too, which `GetFieldID` does not mind. Every name and descriptor is unchanged — `javap -s`, plus a reflection test — so osmandapp/OsmAnd-core-legacy#357 changes two strings: the class lookup and the descriptor of the `precalculatedRouteDirection` field `RoutingContext` holds.

## `RouteConditionalHelper`

Applies `:conditional` tags to a road before the router sees it. `processConditionalTags` resolves a condition against a moment in time — a `maxspeed:conditional` that only holds at night — and writes the value onto the road as if the map had carried it plainly. `resolveAmbiguousConditionalTags` does the same for conditions the caller decided to treat as always true, which is how `HHRoutingShortcutCreator` builds a graph no clock can invalidate. It only ever touched `RouteTypeRule` and `RouteDataObject`, both shared since #25904 and #25906.

**`adopt()` took the whole `RoutingContext`** and read four ints and the router off it. It takes those instead — the same narrowing `VehicleRouter` got with `RoadTraversal` in #25912 — so the context can stay in Java.

**`MapUtils.getProjectionPoint31` had no `KMapUtils` twin.** It has one now. The java one stays, because `RoutingContext.connectPoint` attaches direction points with it, so the two have to agree: `KMapUtilsProjectionCompatTest` runs both over a fixed table and 100,000 random segments and asserts they are bit-identical. The projection is a routing cost, so a difference here would move routes.

**`BinaryRoutePlanner.squareRootDist`** is `MapUtils.squareRootDist31` unless `DEBUG_PRECISE_DIST_MEASUREMENT` is set, and the shared class calls `KMapUtils.squareRootDist31` directly. Every assignment of that flag across all three repositories is commented out, so nothing in a build sets it; if someone does, this one class no longer follows it.

One oddity kept: `getDeviationDistance` passes `pointsX` twice where it means `pointsY`. That is how java had it, it feeds a penalty rather than a distance the user sees, and changing it would move every route that leans on a precalculated one. There is a comment on it now.

## Verification

- `:OsmAnd-java:test` — 585 tests, 0 failures (2 new). The 230 routing regression tests recalculate real routes, and `RouteTestingTest` exercises the precalculated path.
- `:OsmAnd-shared:jvmTest` — 316 tests, 0 failures (24 new common, 3 new jvm)
- `:OsmAnd-shared:iosSimulatorArm64Test` — 272 tests, the same 6 failures as master
- Android app compiles; `OsmAnd-tools` compiles via osmandapp/OsmAnd-tools#1689

**Merge order:** osmandapp/OsmAnd-core-legacy#350 to #357 → OsmAndCore snapshot rebuild → this stack → the tools PRs.

## What is left

`RoutingContext` and `BinaryRoutePlanner`, as one step. The way in is to extract the two interfaces above in Java first — a pure refactor, no JNI change, checked by the 230 regression tests — and then move the pair behind them. `RouteCalculationMode` would have to come out of `RoutePlannerFrontEnd` on the way, since the context holds one and the core binds it by its nested name.

---

**Update — `PrecalculatedRouteDirectionJniContractTest` removed again**, by the commit now at the tip of this branch.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26009.

PR #25926, opened 2026-09-08
Reuse the map renderer when the map activity is recreated

alex-dev-neo:
App side of the renderer handover. Requires [osmandapp/OsmAnd-core#1113](https://github.com/osmandapp/OsmAnd-core/pull/1113): merge it first, rebuild and publish
the wrapper artifacts, and only then merge this. The new `suspendRenderer()` returns whether a
renderer is kept, and the handover moved to `exportRenderer()`, so both the OpenGL AAR and the
wrapper JAR used by the legacy flavor need those signatures.

Separate proposal from #25866 and #25921. Neither is modified or closed.

## Why a recreation rebuilds rendering

`MapActivity.onDestroy()` calls `MapViewWithLayers.onDestroy()`, which releases the renderer
through `MapRendererContext.releaseMapRendererView()` / `stopRenderer()`. That tears down the EGL
thread and its contexts on the main thread, and the replacement view then runs a full
`INITIALIZE_RENDERING`. On a recreation the new activity comes immediately, so there is nothing to
release - and `setupAtlasMapRendererView()` already knows how to take a suspended renderer over
from a previous view, which is what an Android Auto transition does.

## Retention is not guessed

`onRetainCustomNonConfigurationInstance()` is called by Android only when the activity is going to
be created again. It *is* the retention decision, so nothing has to be inferred from
`isChangingConfigurations()` or from a `getChangingConfigurations()` mask, and an unexpected
configuration flag can't silently switch the whole thing off.

- `MapActivity.onRetainCustomNonConfigurationInstance()` returns a holder.
- `MapViewWithLayers.onDestroy(holder)` suspends the renderer and puts the view into it, instead
  of releasing rendering.
- The recreated `MapActivity` reads the holder right after `findViewById()` and hands it to its
  own `MapViewWithLayers`, before `checkAppInitialization()` sets the rendering view up.
- `setupAtlasMapRendererView()` uses it as the previous view when the context has none. It is
  taken right before `setupRenderer()`, so an early return or a missing renderer context leaves
  it to the release at the end of `setupRenderingView()`.

The renderer is retained only while it is still owned by `MapRendererContext` and the core
suspends it successfully. A missing or uninitialized renderer, an ownership mismatch and an active
Android Auto session all keep the existing release path. The context reference is cleared with
`setMapRendererView(null)` rather than released, so it never points at a view that is gone -
during the swap `getMapRendererView()` returns null instead of a dead view.

A retained renderer that nothing takes over is released rather than leaked. That covers the
recreated activity not using the OpenGL rendering view, Android Auto owning rendering by then, and
the activity being destroyed again before it sets a rendering view up.

| Event | Renderer retained |
|---|---|
| Configuration-driven recreation (rotation, and `recreate()` for theme, language, settings) | Yes, if the core suspends it |
| Home/background, screen lock, resume | No, existing pause behavior |
| Finish or ordinary destruction | No, existing release behavior |
| Recreation while Android Auto owns rendering | No, existing Android Auto behavior |
| Recreated activity that doesn't set an OpenGL rendering view up | Released, not leaked |

`handleOnPause()` still waits for the frame in flight, so this doesn't claim to fix a GPU driver
stall inside one frame. The deadline for that is in the core PR.

`onRetainCustomNonConfigurationInstance()` and `getLastCustomNonConfigurationInstance()` are
deprecated in favour of `ViewModel`. A `ViewModel` would also give an `onCleared()` release hook,
but there is no `ViewModel` anywhere in the app yet and no `androidx.lifecycle` dependency is
declared, so the deprecated pair is the smaller change here; the cases `onCleared()` would cover
are handled explicitly above.

## Offscreen renderer dropped by `recreateAndroidAutoRenderer()`

Second commit, found while checking the Android Auto paths against the handover. 
`setupOffscreenRenderer()` forgets its offscreen view as soon as the renderer context no longer
refers to it. That is right when the phone took the renderer over, but
`recreateAndroidAutoRenderer()` gets there by clearing the context reference on purpose, and then
the view is still alive: its renderer, EGL thread and GPU resources were leaked and a fresh set was
created next to them. The view is now stopped instead of dropped; on a view whose renderer was
handed over `stopRenderer()` is a no-op with the core change, so the phone path is unaffected.

## Validation

Nothing here was compiled or run - it can't be until the core change is published, since the app
compiles against the wrapper artifacts and this uses the new `suspendRenderer()` signature. What
was done is static: `git diff --check`, a brace-balance check of both files, and a check of
`androidx.activity:activity:1.10.1` sources confirming that
`onRetainCustomNonConfigurationInstance()` is still `open` (only `onRetainNonConfigurationInstance()`
is final) and that `lastCustomNonConfigurationInstance` maps to the getter used here.

Do not merge before a build and a device run with matching artifacts. What needs checking: repeated
rotations with a visible map, `recreate()` through a theme/language/settings change, background and
resume, lock and unlock, final destruction, Android Auto connect and disconnect in both directions
across a rotation, and the OpenGL rendering engine switched off. Confirm the same renderer and EGL
thread are reused, frames continue on the new view, ordinary destruction still releases, and no old
activity is retained after a handover.

## AI disclaimer

Prompts used (summarised):
1. Review #25866 and #25921 and say whether either is right or it should be done differently.
2. No manifest flags. Propose a new correct solution against the `r5.4` branches.
3. Three commits, one pull request (that is the core PR).
4. Go ahead with the app side.
5. Target `master` instead of `r5.4`.
6. Does this break Android Auto? Then: fix the leak found while checking that.

Decided by the agent, not requested explicitly: using the non-configuration instance as the
retention decision instead of a configuration heuristic; passing the renderer explicitly instead of
parking it in `MapRendererContext`; clearing the context reference rather than releasing it;
releasing a retained renderer that nothing takes over, in all three cases where that can happen;
choosing the deprecated retain API over introducing the first `ViewModel` in the app; and opening
this as a draft because nothing was built or run.

### DIFF
--- a/OsmAnd/src/net/osmand/plus/activities/MapActivity.java
+++ b/OsmAnd/src/net/osmand/plus/activities/MapActivity.java
@@ -131,6 +131,7 @@ import net.osmand.plus.views.AddGpxPointBottomSheetHelper.NewGpxPoint;
 import net.osmand.plus.views.AnimateDraggingMapThread;
 import net.osmand.plus.views.MapLayers;
 import net.osmand.plus.views.MapViewWithLayers;
+import net.osmand.plus.views.MapViewWithLayers.RetainedRenderer;
 import net.osmand.plus.views.OsmandMapTileView;
 import net.osmand.plus.views.OsmandMapTileView.OnDrawMapListener;
 import net.osmand.plus.views.controls.VerticalWidgetPanel;
@@ -191,6 +192,7 @@ public class MapActivity extends OsmandActionBarActivity implements DownloadEven
 
 	private AppInitializeListener initListener;
 	private MapViewWithLayers mapViewWithLayers;
+	private RetainedRenderer retainedRenderer;
 	private DrawerLayout drawerLayout;
 	private boolean drawerDisabled;
 
@@ -319,6 +321,7 @@ public class MapActivity extends OsmandActionBarActivity implements DownloadEven
 
 		drawerLayout = findViewById(R.id.drawer_layout);
 		mapViewWithLayers = findViewById(R.id.map_view_with_layers);
+		takeRetainedRenderer();
 
 		checkAppInitialization();
 
@@ -1037,6 +1040,28 @@ public class MapActivity extends OsmandActionBarActivity implements DownloadEven
 		super.onStop();
 	}
 
+	/**
+	 * Keeps the map renderer of the map view across an activity recreation, so that the
+	 * replacement view reuses it instead of releasing rendering and initializing it anew.
+	 * Android calls this only when the activity is going to be created again, so no
+	 * configuration change has to be guessed here. The renderer itself is put into the holder
+	 * by {@link MapViewWithLayers#onDestroy(RetainedRenderer)}, which runs after this call
+	 */
+	@SuppressWarnings("deprecation")
+	@Override
+	public Object onRetainCustomNonConfigurationInstance() {
+		retainedRenderer = new RetainedRenderer();
+		return retainedRenderer;
+	}
+
+	@SuppressWarnings("deprecation")
+	private void takeRetainedRenderer() {
+		Object retained = getLastCustomNonConfigurationInstance();
+		if (mapViewWithLayers != null && retained instanceof RetainedRenderer) {
+			mapViewWithLayers.setRetainedRenderer((RetainedRenderer) retained);
+		}
+	}
+
 	@Override
 	protected void onDestroy() {
 		super.onDestroy();
@@ -1057,7 +1082,7 @@ public class MapActivity extends OsmandActionBarActivity implements DownloadEven
 			getMapViewTrackingUtilities().setMapView(null);
 		}
 		if (mapViewWithLayers != null) {
-			mapViewWithLayers.onDestroy();
+			mapViewWithLayers.onDestroy(retainedRenderer);
 		}
 		lockHelper.setLockUIAdapter(null);
 		keyEventHelper.setMapActivity(null);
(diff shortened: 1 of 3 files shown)


vshcherb:
Built this and osmandapp/OsmAnd-core#1113 and ran them, on an emulator and on a Pixel 9 Pro XL. A
reduced version is now open as #25936, with osmandapp/OsmAnd-core#1114 on the core side. Nothing here
is closed.

**The hand-over works.** Every recreation case tried was clean: 26 consecutive rotations, a
`wm density` change (which makes `NativeCoreContext.setMapRendererContext` build a *new*
`MapRendererContext` while the renderer survives from the old one - the map still rendered
correctly), a relaunch while the app was in the background, and Android Auto driven over the Desktop
Head Unit - connect, disconnect, a phone rotation during projection (the
`carNavigationSession.hasStarted()` gate holds: no `suspendRenderer` at all, pause 6-19 ms), a
rotate/disconnect race and three full cycles. No crash, no `Swig::DirectorException`, no NPE, live
`OpenGLThread` count constant at one.

**But it does not fix the ANR, and on real hardware it is worth less than the emulator suggested.**
With this and #1113 applied and a 20 s stall injected into a frame, a rotation still produced a real
`am_anr` with the production stack (`GLSurfaceView$GLThread.onPause` ->
`MapRendererView.handleOnPause` -> `MapActivity.onPause` -> `handleRelaunchActivityInner`): main
blocks on the GLSurfaceView GL thread, which blocks on the EGL thread, and neither PR touches that
join. Details and the `configChanges` counter-test are in osmandapp/OsmAnd-core#1113.

Over 16 activity recreations on the Pixel:

| | performDestroy avg | max | renderer |
|---|---|---|---|
| master | 72 ms | 88 ms | 16 full re-inits |
| this pair | 22 ms | 39 ms | 16 hand-overs |

~50 ms per recreation, against a change that rewires activity lifetime and renderer ownership. On a
software-GPU emulator the same A/B is 612 ms -> 148 ms per rotation, which is where the original
numbers came from.

#25936 therefore keeps only the second commit of this PR - stopping the offscreen renderer that
`setupOffscreenRenderer()` drops - which is an independent leak fix, verified with four MSAA toggles
while a car session was running (four `Stopping active renderer`, `OpenGLThread` count unchanged).

Two things found along the way, unrelated to this PR and pre-existing on master: ~1 `MapActivity`
leaks per recreation (22 -> 42 live activities over 20 relaunches on the device), and
`androidx.car.app` 1.7.0 leaks one `VirtualDisplay` per Android Auto session (filed separately).

(gen by AI)

PR #25935, opened 2026-09-08
Spatial search: rank by a score below the structural comparator tiers

vshcherb:
Part of #25331.

## The problem

`SpatialSearchResult.compare()` is a lexicographic ladder eleven tiers deep. Whichever tier
differs first decides, and the ten below it are never consulted. That is right for the top of
the ladder, where the tiers are structural — did we explain the whole query — and wrong for the
bottom, where they are opinions:

- **the cliff**: the rating tiers sit above distance and compare the *raw* value, so one point of
  travel elo outranks a result 50 km closer;
- **the flat spot**: `typeOrder()` returns `0` for every POI, so a railway station, a platform
  node and a waste basket are the same kind of thing to the comparator;
- **the arbitrary tail**: on a sample of 300 real user queries, **41 of 77** flagged orderings
  were decided below the printed key — in practice by the OSM id.

There is also a disagreement inside the class: `compareKey()` buckets the rating by 64 while
`compare()` does not bucket it at all, so two rows the UI draws as one visible level were in fact
ordered by a rating difference too small to be shown.

## The change

Tiers 1–3 (`isPoiCategory`, `tCount`, `surplusWords`) stay exactly as they are. Tiers 4–11
(`objs.size`, `sumOther`, `mainRating`, `totalRating`, `sumTypeOrder`, distance,
`biggestCityType`, osm id) become one bounded score of four terms, each in 0…1 with a weight, so
a large gap on one signal outranks a small gap on another and nothing wins by default:

```
score = 2.0·name + 1.5·type + 2.0·rating + 1.5·near
```

The term doing most of the work is `name`, which separates *the query named this object* from
*the query named the KIND of thing this object is*. `atom.name` carries `POI_CATEGORY_PREFIX`
when the match came through a category key, so searching **farm** matches *Podere Colombaio*
through the word for its kind — every farm then scores alike and distance decides, which is what
a person asking for a farm wants. Searching **Berlin** matches *Berlin* through its name, so
prominence is allowed to decide. One number, both behaviours, without a rule saying when rating
may or may not beat distance.

`SpatialSearchRanking` is an ordinary object owned by `SpatialSearchContext`, with the weights as
instance state so they can be tuned per search. The score is computed once per result at the
start of every sort rather than inside the comparator: a sort makes ~12 comparisons per element
(measured: 60,276 calls vs 4,866 results), and a value cached across sorts goes stale as soon as
deduplication attaches a `unitedObject` and changes `getTotalRating()`.

`SCORE_RANKING` in the settings restores the old ladder. It is what makes the two rankings
comparable on the same corpus, so please keep it until the change is settled.

## How it was measured

The ranking was judged against **60 recorded human preferences** — pairs of objects from 300 real
queries out of the web search log, replayed on the map for the region from the position each
query was typed from, and judged one by one. They are committed as
`OsmAnd-java/src/test/resources/spatial_search/preferences.jsonl`, with the format and the
reasoning in `preferences.md` next to it.

A preference records a *judgement*, not an outcome:

> for query Q asked from position L, object B should come before object A

That survives any rewrite of the ranker, so a new preference adds a constraint instead of
invalidating the old ones. Objects are addressed by OSM id. Contradictions are kept rather than
resolved: `pref-0037` and `pref-0057` are the same pair judged both ways by the same reviewer,
cross-linked with `conflictsWith`, and asserted by neither side — that is the measured noise
floor of a human judge.

`SpatialSearchPreferencesTest` prints one number, comparable across ranker versions:

| ranking | satisfied | violated | n/a | not asserted |
|---|---|---|---|---|
| ladder (today) | **16** | 30 | 1 | 13 |
| score (this PR) | **25** | 21 | 1 | 13 |

```
./gradlew :OsmAnd-java:test --tests "*SpatialSearchPreferencesTest*"
OSMAND_SPATIAL_SCORE_RANKING=false ./gradlew :OsmAnd-java:test --tests "*SpatialSearchPreferencesTest*"
```

Maps are not in the repository: the test looks in `~/osmand/maps`, is pointed elsewhere with
`OSMAND_MAPS_DIR`, and **skips** rather than fails when they are absent. It gates on a ratchet
(`MIN_SATISFIED`), so an unrelated reordering cannot break the build — only contradicting a
recorded judgement can.

Which terms actually earn their place, by ablation (zero one weight, re-run all 60):

| variant | satisfied |
|---|---|
| all four | 25 |
| without `type` | 17 |
| `type` alone | 24 |
| without `rating` | 24 |
| without `name` | 26 |
| without `near` | 26 |

`type` carries almost the whole effect, which confirms the diagnosis. A fifth term (a penalty on
stitched answers) moved nothing in either direction and was deleted. `name` and `near` score one
point *worse* on this set — but the set contains no category query at all, and those two terms
exist for exactly that case, so they are kept with the numbers written into the code. Preferences
on `farm` / `ristorante` / `campingplatz` would settle them.

## What this does NOT fix, and one thing to decide

- **`SpatialSearchPipelineTest`: 74 tests, 49 fail** (all 74 pass with `SCORE_RANKING=false`).
  Those cases store a whole ordered result list with the comparator key and the distance in every
  line, so any ranking change rewrites them; the failures are 49 changed snapshots, not 49
  regressions, and nobody has judged them. They point both ways — `germany_burger-king` now
  returns **Burger King** for "burger king" instead of *King American Burger & Pizza*, and
  `sl_shell` now returns **Shell**; but `uk_yaroslaviv_val` returns the Russian duplicate
  *Улица Ярославов Вал* instead of *вулиця Ярославів Вал*, because the score has no language term
  where the old ladder happened to keep the right one. **That one is a real regression and should
  be fixed before this is merged.** The remaining 48 need a decision, which is why this is a
  draft.
- **Duplicate floods are untouched.** 13 of the 21 still-violated preferences are one place spread
  over up to twelve rows (`bundesplatz berlin`, `bölschestr.`). That is not a ranking question —
  it has a right answer and wants an exact deduplication rule (same normalised name within ~500 m
  merged into the best-typed row, the rest attached through the existing `addExtraResult`).
  Measured separately: it would give back 416 top-10 slots over 75 of 276 queries. Deliberately a
  separate change.
- **`berlin` stays wrong**, and not because of the ranking: on `Germany_berlin_europe_2.obf` the
  query returns no city or boundary record for Berlin at all — the city is present only as a POI
  carrying Q64, rated 1495, while an artwork named "Berlin" is rated 2534. The elo data is its
  own work item: it exists for ~7% of results and is city-only.

## AI disclaimer

Prompts used (summarised):
1. Find the earlier session about search unit tests and contrastive pairs, summarise it, and
   propose how to label contrastive pairs whose results look wrong — first pass by the agent, for
   human approval.
2. Take the specifics of our spatial search algorithm into account.
3. After the 60 verdicts were given: what conclusions and rules follow, and how do we fix it?
4. "Rating must not beat distance" is not always true — often, but not always.
5. This is a fuzzy problem and needs a fuzzy ranking; deduplication is probably a crisp one.
6. Make an HTML visualisation of the fix with before/after.
7. Show the fuzzy fix as a local diff.
8. How do we store the labelling results so the information is not lost, given that any new
   preference makes outcome-based tests unstable?
9. Create `preferences.jsonl` in the form you propose.
10. Put it in the android project with a test; drop the changes to `SpatialTextSearch.mainTest`.
11. Why is the score cached — does it actually matter?
12. Then let us simplify, if it is not needed.
13. The `build.gradle` change is not wanted.
14. Do not call the ranking class statically — use an instance; and put the score in a field on
    the result.
15. Open a draft PR linked to the spatial search implementation issue.

- The four terms, their weights, and the three POI subtype tables (infrastructure / stop /
  landmark), including which subtypes belong in each.
- Using `NameIndexReader.POI_CATEGORY_PREFIX` as the signal for "matched through the kind word".
- Keeping `name` and `near` despite the ablation, because the preference set has no category
  query that could judge them; the numbers are recorded in the code so the decision is reversible.
- The preference record schema: OSM id as identity, append-only with `supersedes`, contradictions
  kept and cross-linked with `conflictsWith`, `source`/`confidence` for weighting later.
- The `MIN_SATISFIED` ratchet and its value, and skipping rather than failing without maps.
- Reading `OSMAND_MAPS_DIR` / `OSMAND_SPATIAL_SCORE_RANKING` from the environment so no build
  file has to change.
- Leaving `SCORE_RANKING` defaulting to true.
- Not including the deduplication fix, and not fixing `uk_yaroslaviv_val` yet.


OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchContext.java
```diff
@@ -139,6 +139,9 @@ public String toString() {
 
 	}
 
+	/** ranking weights for this search, see {@link SpatialSearchRanking} */
```
vshcherb: unnecessary comment

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchRanking.java
```diff
@@ -0,0 +1,225 @@
...
+ * default.
+ *
+ * <p>The term that does the most work is {@link #nameScore}: it separates "the query named this
+ * object" from "the query named the KIND of thing this object is". Searching <i>farm</i> finds
+ * <i>Podere Colombaio</i> through the word for its kind, so every farm scores alike on name and
+ * distance decides - the nearest farm wins, which is what a person asking for a farm wants.
+ * Searching <i>Berlin</i> finds <i>Berlin</i> through its name, so prominence is allowed to
+ * decide - which is what a person naming a specific place wants.
+ */
+public class SpatialSearchRanking {
```
vshcherb, line 23: Comments are  too big should be condensed

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResult.java
```diff
@@ -518,17 +521,26 @@ private static long addCompareKey(long key, int bits, int value) {
...
 				Math.min(o.sumOther(), 3), elo, o.sumTypeOrder());
 	}
-	
+
 	public static long compareKey(SpatialSearchResult o) {
 		long key = 0;
 		key = addCompareKey(key, 6, -o.parent.tCount); // 6 bit - 64
 		key = addCompareKey(key, 3, -o.surplusWords); // 3 bit - 8
+		if (o.parent.SCORE_RANKING) {
+			// visibleLevel is grouped by this key, so it has to move with the order the score
```
vshcherb: comment not needed

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResultsList.java
```diff
@@ -626,6 +630,14 @@ public List<SpatialSearchResult> sortResults(SpatialSearchContext ctx, boolean d
 	}
 
 	public List<SpatialSearchResult> sortResults(SpatialSearchContext ctx, List<SpatialSearchResult> finalResult, boolean deduplicate) {
+		if (SCORE_RANKING) {
+			// once per result per sort, rather than on every comparison: a sort makes ~12 of them
```
vshcherb: comment not needed

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialTextSearch.java
```diff
@@ -112,6 +112,10 @@ private SpatialTextSearchSettings() {}
 		// only do incomplete search with 2+ chars
 		public int MIN_CHARACTERS_INCOMPLETE = 2;
 		
+		// false restores the old lexicographic ladder below surplusWords - keep it switchable
```
vshcherb: make comment short one line

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResult.java
```diff
@@ -550,9 +558,28 @@ public static int compare(SpatialSearchResult o1, SpatialSearchResult o2, LatLon
 		if (res != 0) {
 			return res;
 		}
-		res = Integer.compare(o1.objs.size(), o2.objs.size());
-		if (res != 0) {
-			return res;
+		// objs.size stays a hard tier: an answer found in ONE object always beats the same words
```
vshcherb: condense comment and use instance ranking not static method from context

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResultsList.java
```diff
@@ -658,12 +667,88 @@ public List<SpatialSearchResult> sortResults(SpatialSearchContext ctx, List<Spat
...
 	}
-	
-	
+
+	/** how far apart two rows of the same name may be and still be one place */
+	private static final double SAME_PLACE_M = 30;      // two shops, two benches: 57 m are two
+	private static final double SAME_FACILITY_M = 400;  // the platforms of one stop are spread
+	private static final double SAME_STREET_M = 2000;   // one street cut into several OSM ways
+
+	/**
```
vshcherb: condense comment

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResultsList.java
```diff
@@ -658,12 +667,88 @@ public List<SpatialSearchResult> sortResults(SpatialSearchContext ctx, List<Spat
...
+			// the coordinate of a line means little, so the city decides and the distance only
+			// guards against two genuinely different streets of the same name
+			City c1 = ((Street) a.getMainObject()).getCity();
+			City c2 = ((Street) b.getMainObject()).getCity();
+			if (c1 == null || c2 == null || !c1.getName().equals(c2.getName())) {
+				return false;
+			}
+			radius = SAME_STREET_M;
+		} else {
+			radius = SpatialSearchRanking.isSpreadNode(a) || SpatialSearchRanking.isSpreadNode(b)
```
vshcherb: use instance method not static

OsmAnd-java/src/main/java/net/osmand/search/core/spatial/SpatialSearchResultsList.java
```diff
@@ -658,12 +667,88 @@ public List<SpatialSearchResult> sortResults(SpatialSearchContext ctx, List<Spat
...
+	 * keys above cannot see these: a stop stored as a dozen OSM nodes shares none of them.
+	 *
+	 * Same name, and then a distance that depends on what the two rows are - measured on reviewed
+	 * cases: 320 m between a camp site and its bus stop is one place, 218 m between two parcel
+	 * lockers is not, and a street is only ever merged with another way of the same street.
+	 */
+	private List<SpatialSearchResult> deduplicateByProximity(List<SpatialSearchResult> sorted,
+			SpatialSearchContext ctx) {
+		Map<String, List<SpatialSearchResult>> byName = new HashMap<>();
+		List<SpatialSearchResult> out = new ArrayList<>(sorted.size());
```
vshcherb, line 688: do we need to create full copy can we avoid it?

vshcherb:
Superseded by #25953: the same work squashed into one commit, with the deduplication and the preference test that came out of the reviews since. Closing this one.

vshcherb:
Reopened. The squashed attempt (#25953) is closed; this branch stays the single place for the review.

Issue #16322, opened 2023-01-25
Unnecessary turn instructions - following main road


frederikb96:
# 🚀 feature request

### Description

If following a main road that has curves, there are often unnecessary turn navigations (which are not incorrect but just confusing - (there are already issues about wrong turn navigations)).

See this example, where a primary highway needs to be followed but there are turn navigations if there is a small track connected  to it. I tested other routing engines (OSRM, ORS and Graphhopper) which do all not show this issue.

![Screenshot_20230125-130409_OsmAnd~.png](https://user-images.githubusercontent.com/83548283/214573431-a7d7248a-e632-4bc3-ade2-258d41a7c0eb.png)

![Screenshot_20230125-130452_Vespucci.png](https://user-images.githubusercontent.com/83548283/214573457-d66e8a97-d7a9-40e3-b215-892e503196ca.png)

28.76996° N, 17.90393° W

### Describe the solution you'd like

It would be nice, if there is simply no turn navigation if you follow a main road and only way smaller roads like tracks connect to it.

### Describe alternatives you've considered

Or simply give a single info to follow the road instead of giving a turn instruction.

pebogufi:
We had a discussion similar to that here
https://github.com/osmandapp/OsmAnd/issues/8329

frasty:
I find that turn instructions should be avoided particularly in such cases. It doesn't matter if they are semantically correct as they unnecessarily steal the attention of the user whereas attention should only be taken on the road you are following already.

vshcherb:
Reproduced 
https://www.openstreetmap.org/way/51206036 
https://www.openstreetmap.org/way/78168989#map=15/28.7780/-17.9074

Question is that there is a turn, primary road doesn't go straight so for people in other situation or in the dark it could be confusing without that turn instruction. So I would say in this example it's better to keep

frederikb96:
Thanks for investigating.

I actually drove the road in the night and from my perspective I can not imagine that anyone would wonder if to follow the road here, since it is only a track and you are on a primary high way. 

Generally in areas like La Palma or Scotland (two areas I have been lately) there are a lot of roads with many curves and a lot of tracks for forest workers or farmers and there are a lot of turn instructions due to this. I personally never thought that it was ever unclear to simply follow the main road. From my experience, other routing engines/software don't give instructions in those cases. I prefer this behavior.

But I understand that this may vary strongly from area to area and that it is very subjective. So feel free to close the issue, if a change is not desired.

[...]

Issue #25930, opened 2026-09-08
Incorrect “Turn sharply right” instruction and lane guidance despite correct route in Graz (Android 5.4.4 / Web)


shadow-identity:
## Description

At the Parkstraße → Glacisstraße junction in Graz, OsmAnd calculates the correct route through the gentler right turn. However, the turn instruction, maneuver icon, and lane guidance indicate the sharper right-hand branch instead.

## Steps to reproduce

Open [this route in OsmAnd Web](https://osmand.net/map/navigate/?start=47.077984,15.442822&end=47.077274,15.443962&profile=car#18/47.07764/15.44202) and inspect the right-turn instruction and lane guidance. The issue occurs with both Car and Motorcycle profiles and is also reproducible on OsmAnd for Android 5.4.4.

## Actual result

The route follows the correct, gentler right-hand branch, but the instruction approximately 54 m from the start says:

> Turn sharply right and go B67a Glacisstraße

The maneuver icon and highlighted lane arrow also indicate the sharp right turn, rather than the two lanes serving the gentler right turn used by the calculated route. Following these cues would direct the driver onto the wrong branch.

## Expected result

The instruction, maneuver icon, and lane highlighting should agree with the calculated route: indicate the gentler right turn and highlight both lanes serving it.

## Environment

- OsmAnd for Android 5.4.4.
- Also reproducible in OsmAnd Web as of 8 September 2026.
- Affected profiles: Car and Motorcycle.

## Possible regression

I did not encounter this problem at this junction before updating to 5.4 in September; it was working correctly a week earlier. I suspect a regression in 5.4, but have not confirmed it by retesting an older version with identical map data.

## Additional observation — not currently reproducible

On Sunday, 6 September 2026, I also encountered roundabouts where the route used the correct exit, but the voice announcement and advance maneuver diagram indicated the first exit instead. This behavior was also new after the 5.4 update; I had not encountered it a week earlier, before updating. I could not reproduce it again on 8 September on Android or Web, and cannot confirm whether it has the same cause.

## Possibly related issues

- https://github.com/osmandapp/OsmAnd/issues/23787
- https://github.com/osmandapp/OsmAnd/issues/21055

This report provides a different reproducible junction; a shared cause has not been established.

<img width="288" height="595" alt="Image" src="https://github.com/user-attachments/assets/deb678b9-3442-47a9-92ea-8197354f5bf0" />

yuriiurshuliak:
[...]

Issue #25360, opened 2026-07-03
US turn lanes - Says "turn left" when it should say "stay straight" or nothing - don't count service


lunkhub:
### Routing engine

- [x] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile

Offline maps, current, OsmAnd~ from f-droid, current 5.3.10

Settings: No toll roads, No motorways, Fuel-efficient way On or Off

### Start and end points

Start example: 35.23151, -80.92163
End example: 35.25187, -81.02783

### Actual route

Actual result: At two places along the route, it says "turn left" near intersections where you need to stay straight.

### Expected route

Expected result: Stay straight on Rt. 29/74/Wilkinson Blvd for 6+ miles.
Note: Reversing Start and End works as expected.

### Maps you used (online or offline)

- [x] Offline maps offered within the OsmAnd app for download. 
Edition Dates: June 1, 2026 + North Carolina live updates June 23, 2026
- [ ] Online (tile / raster) maps 

### Your Environment

```
OsmAnd Version: 5.3.10 (F-droid)
Android version: 16 (LineageOS)
Device model: Fairphone 4

```

### Anything else?

Note: Reversing Start and End works as expected.

I looked at the first (eastern most) place in edit mode on openstreetmaps, and saw nothing to fix.

DmitryAlexei:
route [OsmAnd map](https://osmand.net/map/navigate/?start=35.247010,-81.015123&end=35.247722,-81.017969&profile=car#16/35.2473/-81.0156)
[Osm object](https://www.openstreetmap.org/way/1072690292)

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/59a697dc-edb4-430c-81a9-ec9c82b09176" />

May be related to
- https://github.com/osmandapp/OsmAnd/issues/21422

vshcherb:
Service road probably should be ignored as it can't produce 3 lanes - https://www.openstreetmap.org/way/723334681 or any other heuristics?

vshcherb:
Checked against osmandapp/Osmand#25878 (unmarked lanes in `turn:lanes`): this route is **not** affected — the output is byte-for-byte the same with and without it. Same for #21422, and for the remaining part of #23922. All three have the same root cause, which is the one named in this title.

[...]

Issue #23131, opened 2025-07-28
Make a toggle to turn on/off the voice announcements for lane guidance


usr97629238:
### Describe the idea

Please make a toggle to turn off the voice announcements for lane guidance. 
Currently, lane‑guidance prompts bombard you with constant announcements about non essential exits even when you’re driving straight - interrupting music or podcasts every few seconds. While some drivers find this helpful, others would prefer a quieter experience and only want to be notified when a lane change or exit is imminent.

Leave turn‑by‑turn directions and exit alerts active so drivers still receive essential guidance without constant interruptions.

Benefits:
- Allows drivers to enjoy audio content uninterrupted during long, straight stretches
- Reduces "announcement fatigue" for users who only need alerts at key moments
- Provides greater personalization of navigation experience

### Expected behaviour

Introduce a user‑configurable setting ("Mute lane guidance") that, when enabled, suppresses all lane‑guidance voice prompts except critical notifications (on exits).

### Alternatives you've considered

Turning it off completely.

franco999:
Duplicated https://github.com/osmandapp/OsmAnd/issues/22223 ?

usr97629238:
> Duplicated [#22223](https://github.com/osmandapp/OsmAnd/issues/22223) ?

Nope, the [#22223](https://github.com/osmandapp/OsmAnd/issues/22223) is about turning off **turn alerts** while this one is about **lane alerts** - the constant "Keep left" / "Keep right" about every exit you pass (and don't care about) while just driving straight.

vshcherb:
Please provide some points https://osmand.net/map/#5/50.000/5.000 to double check it's not an issue with unnecessary keep left / right which we were fixing in the past

usr97629238:
> Please provide some points https://osmand.net/map/#5/50.000/5.000 to double check it's not an issue with unnecessary keep left / right which we were fixing in the past

Will do, however the point is to have the ability to turn all audible lane alerts completely off.

sonora:
What prompts exactly do you mean by "lane alerts"? I am not aware we have anything like this in the app...

usr97629238:
> What prompts exactly do you mean by "lane alerts"? I am not aware we have anything like this in the app...

https://osmand.net/docs/user/widgets/nav-widgets/#lanes

sonora:
Confused. That's an onscreen widget, not a voice announcement...??

usr97629238:
[...]

PR #25937, opened 2026-09-08
Suppress redundant turn instructions on same-road bends

shadow-identity:
Following a named or numbered road around a bend can produce a misleading left/right instruction at a minor side road, or where the main road is split into separate OSM ways. For example, [this B72 / Grazer Straße route](https://osmand.net/map/navigate/?start=47.142409,15.533365&end=47.142044,15.533687&profile=motorcycle) says to turn left onto the road already being followed. The route geometry itself is correct.

Related to https://github.com/osmandapp/OsmAnd/issues/16322, including the [LP-114 example](https://osmand.net/map/navigate/?start=28.7818662,-17.9648754&end=28.7817416,-17.9657042&profile=car).

## Change

Suppress the instruction when the incoming and outgoing road have the same highway class and matching non-empty ref, or matching name when neither has a ref. Matching refs allow a street-name change. This works across different OSM way IDs and also at joins with no attached side road.

Keep existing guidance when an attached road has equal/higher speaking priority or shares the identifying ref/name, or when identity/class changes. Links, roundabouts, explicit turn markings, and lane guidance remain excluded. Suppression runs after turn merging so lane guidance inferred by that stage is preserved. Route selection and geometry are unchanged.

## Validation

- 92 regression cases pass: car/motorcycle in both directions, with official-map geometry fixtures for B72 and LP-114, plus identity, class, branch and lane checks. The 114 existing route-preparation cases also pass.
- Exploratory before/after checks of 68 route variants reconstructed from community reports on official September 1 maps: identical geometry in all variants; 38 instructions removed across nine locations, with no other non-initial instructions added or changed. Point-only reports used approximate reconstructed endpoints, so these are current-map checks rather than exact historical reproductions.
- Backported the same patch to the [release-branch source at the publication date of 5.4.4](https://github.com/osmandapp/OsmAnd/commit/566d11350122b4d9fffae39575aa51d067d92361): Android APK built successfully and all 638 Java-module tests passed. The reporter checked route planning on a Pixel 6 Pro. Voice playback and on-road navigation have not yet been validated.

The PR contains only the Java change, regression tests and two fixtures; local APK build settings and diagnostic tooling are excluded.

## AI disclaimer

Produced with Codex (GPT-6).

Prompts used (summarised):
1. Investigate misleading bend instructions and developer findings in related reports, then implement a small fix using road identity and class.
2. Prefer matching refs, allow street-name changes, fall back to matching names when refs are absent, and preserve guidance at ambiguous or equal/higher-class junctions.
3. Test other reported locations and remove the unnecessary requirement for an attached side road so continuity also works across OSM way boundaries.
4. Build and install a release-based Android APK for manual validation, then open this pull request after the reporter checked route planning.

Decided by the agent, not explicitly requested: placing suppression after turn merging to retain inferred lanes, the fixture format and parameterized test structure, and the local tooling used for comparative route checks and APK assembly.


shadow-identity:
Issue-by-issue results for the reports investigated while preparing this PR:

### Confirmed improvements: [#16322 — Unnecessary turn instructions - following main road](https://github.com/osmandapp/OsmAnd/issues/16322)

- The [B72 / Grazer Straße example added to the discussion](https://github.com/osmandapp/OsmAnd/issues/16322#issuecomment-5589237863): removes the misleading turn instruction when staying on B72 past the service road. Covered by a regression fixture.
- Nine La Palma locations from the original report and [additional coordinates](https://github.com/osmandapp/OsmAnd/issues/16322#issuecomment-1407042667) lose their unnecessary bend instructions:

| Road | Reported coordinates |
| --- | --- |
| LP-4 | 28.76996, -17.90393; 28.78392, -17.91766; 28.78411, -17.91935 |
| LP-1 | 28.77589, -17.95542; 28.77810, -17.94677; 28.78495, -17.92938; 28.78825, -17.92832; 28.65874, -17.93397 |
| LP-114 | 28.78265, -17.96555 |

At LP-114, both the side-road-junction instruction and the instruction about 8 m later at a join between separate OSM ways are removed. This also has a regression fixture. The other three La Palma points already had no bend instruction on the unmodified baseline.

These are confirmed improvements to the reproduced cases in #16322, not a claim that every possible case in the issue is resolved.

### Related reports checked, with no confirmed new fix from this PR

| Issue | Result |
| --- | --- |
| [#8329](https://github.com/osmandapp/OsmAnd/issues/8329) | The bend instruction in the supplied German GPX example was already absent on the baseline. The actual B28-to-L94 turn with lane guidance remains. |
| [#12774](https://github.com/osmandapp/OsmAnd/issues/12774) | The two Quebec examples did not change: one has no bend instruction; the other has an already-muted instruction with calculated lanes nearby, which this patch preserves. |
| [#3663](https://github.com/osmandapp/OsmAnd/issues/3663) | The Scottish example retains an already-muted continue instruction with lane guidance. |
| [#19541](https://github.com/osmandapp/OsmAnd/issues/19541) | The supplied Bulgarian route was already corrected upstream; its muted continue/lane guidance remains. |
| [#25360](https://github.com/osmandapp/OsmAnd/issues/25360) | The service-road / turn-lane interpretation problem is outside this patch's scope; lane guidance is deliberately preserved. |
| [#25930](https://github.com/osmandapp/OsmAnd/issues/25930) | The Graz sharp-right instruction / wrong lane selection is a separate case and is not fixed here. |

Other reports encountered during the investigation were not counted as fixes: [#21031](https://github.com/osmandapp/OsmAnd/issues/21031) involved duplicate map data, with insufficient coordinates for the later Gran Canaria example; [#22995](https://github.com/osmandapp/OsmAnd/issues/22995) concerns BRouter; [#23131](https://github.com/osmandapp/OsmAnd/issues/23131) requests a voice setting, and a reported extra-map interaction was not reproduced here.

The community comparison used official September 1, 2026 maps and Java A* routing: 68 variants, identical route geometry in all of them, and 38 instructions removed across the nine La Palma locations. Point-only reports required approximate reconstructed endpoints; this does not recreate their historical map/software versions.

(gen by AI)

shadow-identity:
The matching C++ port is now available in [OsmAnd-core-legacy#358](https://github.com/osmandapp/OsmAnd-core-legacy/pull/358).

A standalone native test executable reused the inputs and results from all 92 Java regression scenarios and compared turn types, skip-to-speak flags and lane arrays for every segment. Unmodified C++ had 24 mismatches; the patched C++ matches all 92 cases. The native test executed the C++ turn-preparation code directly, with no fallback to Java and no separate copies of the fixture data added to the C++ repository.

Full iOS build/on-device validation is still outstanding. The native PR describes the exact validation scope.

(gen by AI)

PR #25947, opened 2026-09-09
Routing to OsmAnd-shared, step 14: romanise names the same way on every platform

alex-dev-neo:
Stacked on #25922, so the diff to read is the last commit.

`toLatin` was an `expect`/`actual`: junidecode on the jvm and android, and on iOS `NSStringTransformToLatin` followed by a diacritic pass. Those are different engines and they disagree — over **22 121 amenity names** taken off three regional maps they produced the same string for **13 694** of them. A place whose only latin form comes from transliteration could therefore be found by that form on android and not on iOS, from the same map file.

The tables junidecode carries are now generated into common code. There is one implementation, and its answers are the ones android and the server have always given: **nothing about android's or the server's behaviour changes, iOS moves onto junidecode's romanisation.**

### Matching junidecode exactly

The walk is junidecode's own, quirks included, since matching it is the whole point:

- the code point is read at each index but the index advances by **one**, so both halves of a surrogate pair are looked up — the pair as a whole, and then its low half again;
- the block index is **masked to a byte**, which folds anything outside the basic plane back onto a low block;
- there is no ascii shortcut inside junidecode: `U+007F` does not romanise to itself, so the table covers the first code points too.

Printable ascii *is* the identity in the table, which lets a name that is already latin be returned as it came. The test pins that range, so the shortcut cannot outlive the assumption behind it.

### Checks

`KTransliterationTablesTest` romanises **all 65 536 code units** and compares each against junidecode, so the generated file cannot drift from it silently. It also checks whole strings — cyrillic, greek, arabic, burmese, CJK, a surrogate pair, a lone surrogate, an emoji — and that japanese is still left alone.

`KTransliterationTablesGenerator` writes the file and sits beside the test, `@Ignore`d like the other tools here, because a generated file with no way to regenerate it is one nobody can touch.

`SearchUICoreTest` (60 cases) and the rest of OsmAnd-java and OsmAnd-shared pass; the six iOS simulator failures are the ones already failing on this branch's base (`GpxUtilitiesLoad`, `KGeoPointParserUtil`, `NetworkAPI`).

### Cost

junidecode stops being a dependency of the module and becomes a test one — it is only needed now to check that the tables still say what it says. The generated file is 262 KB of source against the 200 KB jar it replaces, and it is no longer carried twice, once for the jvm and once for android.

On Kotlin/Native it is also faster: **0.09 µs a name against 2.3 µs** for the Foundation call. `NSString` takes a transform identifier rather than a handle, so it rebuilds the transliterator on every call, and the POI name search asks for a romanised name once per candidate that has no `name:en` — tens of thousands per query.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26015.

PR #25948, opened 2026-09-09
Fix #25835 night mode check for map controls

### LINKED ISSUE #25835: Map screen buttons now governed by app theme instead of day/night map?
Just to make sure this is intended: While fixing https://github.com/osmandapp/Osmand/issues/25634 we had this behavior change:

Now all buttons on the map screen turned dark in dark theme, even when the map is in day mode:

- Have the app in _dark theme_, but a map in _day mode_: In the last nightly this started producing **dark** menu / search / profile / navigation etc. buttons on the map screen, looks a bit weird.

8/26 build vs. 9/3 build:

Image Image

sonora:


Chumva:
Should this fix also be in 5.4?

PR #25949, opened 2026-09-09
Routing to OsmAnd-shared, step 15: RouteCalculationMode

alex-dev-neo:
Stacked on #25947, so the diff to read is the last commit.

`RouteCalculationMode` says how much of the road network a calculation may see: `BASE` is the coarse network a basemap carries, `NORMAL` the detailed one, and `COMPLEX` means a base route first, which then guides the detailed search as a `PrecalculatedRouteDirection`.

It was a nested enum of `RoutePlannerFrontEnd`, and that class cannot move — it is the java planner, which stays in java as the fallback. So the enum comes out into a class of its own.

### Why this one

It is the **last type the C++ router reads off `RoutingContext` that was not shared yet**. `java_wrap.cpp` takes 21 fields off the context; the other 20 are primitives, arrays, or objects whose classes moved in earlier steps — `RoutingConfiguration`, `RouteCalculationProgress`, `PrecalculatedRouteDirection`. With this one moved, the request half of the context is expressible entirely in shared types, which is what the next step needs.

### The contract is the order, not a field

The core does not read a field off this object. It calls `ordinal()` and casts the result straight onto its own enum:

```cpp
jint value = ienv->CallIntMethod(calculationMode, getOrdinalValueMethod);
c->calculationMode = (RouteCalculationMode)(value);
```

against `enum class RouteCalculationMode { BASE, NORMAL, COMPLEX }` in `native/src/routingContext.h`. Reordering the constants would route in the wrong mode — `BASE` arriving as `NORMAL` — and nothing would fail to build. `RouteCalculationModeJniContractTest` pins the order and the class name for that reason.

### Checks

Every reference is an import change; the enum body is unchanged. `OsmAnd-java`'s router tests pass, `java-tools` compiles.

### Merge order

Paired with osmandapp/OsmAnd-core-legacy#360, which changes two strings and **has to merge first**, followed by the OsmAndCore snapshot rebuild. osmandapp/OsmAnd-tools#1691 follows the same move and merges last.

---

**Update — `RouteCalculationModeJniContractTest` removed again**, by the commit now at the tip of this branch.


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26010.

PR #25950, opened 2026-09-09
Routing to OsmAnd-shared, step 16: HHRoutingConfig

alex-dev-neo:
Stacked on #25949, so the diff to read is the last commit.

`nativeRouting` takes two arguments: the routing context and an `HHRoutingConfig`. This is the second one. It says which search runs over the hub graph — Dijkstra, A\*, contraction hierarchies, midpoints — how far it may go, how much of the detailed road network is rebuilt afterwards, and what an alternative route has to look like to be offered.

The C++ router reads **fourteen fields** off it, so it has to be a shared class before anything outside java can ask for a native route.

It was nested in `HHRouteDataStructure`, which cannot move: that file is the java planner's own graph, welded to `BinaryMapIndexReader` and trove.

### The one test here, and what it is for

`java_wrap.cpp` resolves the class and those fourteen fields **by string**, and nothing in either build reads those strings. A rename, a package move, or a `MAX_DEPTH` widened to a long would compile everywhere and surface on the first native route calculation on a device. `HHRoutingConfigJniContractTest` checks the class name and all fourteen names and descriptors by reflection; `javap -s` agrees.

The fields are `@JvmField`, but that is a separate matter from the JNI: Kotlin names the backing field after the property either way, and `GetFieldID` does not mind a private one — `PrecalculatedRouteDirection` is bound that way today with no `@JvmField` at all. Here it is for the java callers that read and assign these as fields, which javac checks.

There is deliberately **no test of the defaults and builders**. Asserting them would restate the literals in the class rather than check them against anything, and the java routing tests already build these configs and route with them: `defaultHHConfig()` is `astar(0).calcDetailed(CALCULATE_ALL_DETAILED)`, `prepareDefaultRoutingConfig` is `astar(0)` plus `calcDetailed(2)`. Pinning the tuning knobs would also mean editing a test every time one is tuned.

### Two things could not be ported as they were

- **`cacheCtx`** held an `HHRoutingContext` — the java planner's loaded graph, which stays behind. It is an `Object` here, cast in one place in `HHRoutePlanner`. The config only carries it from one calculation to the next; nothing reads into it.
- **`toString(LatLon, LatLon)`** takes a `KLatLon` now, so `LatLon` grew a `toKLatLon` to match the `of(KLatLon)` it already had. Both print the same text.

### Checks

`OsmAnd-java`'s routing tests pass (460), `OsmAnd-shared` passes on the jvm, `java-tools` compiles, the app compiles. On iOS the only failures are the six already failing on this branch's base (`GpxUtilitiesLoad`, `KGeoPointParserUtil`, `NetworkAPI`).

### Merge order

Paired with osmandapp/OsmAnd-core-legacy#361, which changes one string and **has to merge first**, followed by the OsmAndCore snapshot rebuild. osmandapp/OsmAnd-tools#1692 follows the same move and merges last.

---

**Update — `HHRoutingConfigJniContractTest` removed again**, by the commit now at the tip of this branch.


alex-dev-neo:

PR #25952, opened 2026-09-09
Routing to OsmAnd-shared, step 17: the request half of RoutingContext

alex-dev-neo:
Stacked on #25950, so the diff to read is the last commit.

`RoutingContext` is two objects in one file:

- **the request** — where the route starts, where it has to reach, on what profile, where to report progress;
- **the java planner's machinery** — the tile cache, the reader map, and the `BinaryRoutePlanner.RouteSegment` graph nodes it builds as it loads each tile.

Only the first crosses the boundary to the C++ router, and only the first is expressible in shared types now that the last of them, `RouteCalculationMode`, moved in #25949.

So the request becomes `RoutingRequest` in `OsmAnd-shared`, and java's `RoutingContext extends RoutingRequest`. The planner is untouched — it still writes `ctx.startX` and reads `ctx.config` exactly as before.

### core-legacy needs no change at all

This is the first step of the series with no paired PR, and it is worth saying why.

`java_wrap.cpp` resolves **twenty two fields** on `net/osmand/router/RoutingContext` by name, and after this they are every one of them declared in the Kotlin superclass. That still works: `GetFieldID` on a subclass resolves fields declared in a superclass, whatever their visibility.

That was not taken on faith. A throwaway JNI library did `FindClass` on a Java subclass of a Kotlin parent and then read and wrote the parent's fields through it:

```
FindClass(probe/Child) ok
  plainProp      I    kotlin parent, private backing field   read 1100
  jvmFieldProp   I    kotlin parent, @JvmField               read 2200
  objectProp     L…   kotlin parent, @JvmField object        read non-null
  reallyPrivate  I    kotlin parent, private property        read 44
  absent         I    nowhere - control                      NOT FOUND
  wrote plainProp = 9911 / reallyPrivate = 9944
```

The control field coming back NOT FOUND is what says the probe was not passing vacuously. All twenty two names and descriptors were then compared against `java_wrap.cpp` on the compiled class — 22 of 22 match.

`@JvmField` is on them for the other reason: the java planner reads and assigns them as fields, which Kotlin accessors would break, and javac catches that.

### What stayed behind

| | why |
|---|---|
| `map`, `reverseMap`, `mapIndexReaderFilter` | hold `BinaryMapIndexReader`s |
| `nativeLib` | `NativeLibrary`, which step 18 puts behind an interface |
| `dijkstraMode`, `hhHasUnsupportedParameters`, `leftSideNavigation`, `calculationProgressFirstPhase`, `previouslyCalculatedRoute` | the java planner's own knobs; they can follow when something outside java needs them |
| `initStartAndTargetPoints` | takes a `BinaryRoutePlanner.RouteSegmentPoint` |
| `throwNotEnoughMemory` | formats with `%.5f`, and a hand written float formatter is not worth it for an exception message |

The seven accessors that only touch `config` or `calculationProgress` came along.

### One deletion you did not ask for

The first commit removes `RoutingContext.config`'s field type assertion, cherry picked from #25920 where it was already removed. This refactor is exactly the change it mis-reports: it reads `getDeclaredField("config")`, the field is declared one level up now, and the contract it claimed to protect is intact — as the probe above shows. Without it this branch would be red for a reason that is not a defect.

### Checks

459 routing tests in `OsmAnd-java` pass, the app compiles, `java-tools` compiles across all five modules, `OsmAnd-shared` passes on the jvm. On iOS the only failures are the six already failing on this branch's base (`GpxUtilitiesLoad`, `KGeoPointParserUtil`, `NetworkAPI`).

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/router/RoutingContext.java
+++ b/OsmAnd-java/src/main/java/net/osmand/router/RoutingContext.java
@@ -44,9 +44,10 @@ import net.osmand.shared.util.collections.KTIntObjectIterator;
 import net.osmand.shared.routing.RouteSegmentResult;
 import net.osmand.shared.routing.VehicleRouter;
 import net.osmand.shared.routing.GeneralRouter;
+import net.osmand.shared.routing.RoutingRequest;
 
 
-public class RoutingContext {
+public class RoutingContext extends RoutingRequest {
 
 	public static boolean SHOW_GC_SIZE = false;
 	public static boolean PRINT_ROUTING_ALERTS = false;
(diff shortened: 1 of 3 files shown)


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26010.

Issue #25941, opened 2026-09-09
Spatial Search - Ranging and Synonyms


vshcherb:
From https://github.com/osmandapp/OsmAnd/issues/25331
# Ranging
- [ ] Fix order - CC https://github.com/osmandapp/OsmAnd/issues/25021, https://github.com/osmandapp/OsmAnd/issues/24803, https://github.com/osmandapp/OsmAnd/issues/24089, https://github.com/osmandapp/OsmAnd/issues/24803
- [ ] Unable to find `Hardware store` category, city is shown instead of category (from https://github.com/osmandapp/OsmAnd/issues/21621) 
- [ ] Translations 'Bank' (bench), 'Billa' (water_tap), 'Paris' (bookmaker)
- [ ] Test With Category 8

# Synonyms
- [ ] Ranging Compare Unit tests with Live maps
- [ ] INDEX: Find POI Categories translations / synonyms via Common words - Стоматол., Dentist, Basilica 
- [ ] REVIEW: Abbrevations (synonyms / direction words) other languages? - https://github.com/osmandapp/OsmAnd/issues/16359
- [ ] REVIEW: Analyze Abbrevations / common skip (abbrevations 1st=first) 

### Follow up tasks

```
// TODO Analyze Performance bootlenecks VisualVM (Pipeline + Intersection)
// TODO Sorting before load objects (use elo and other buildings?) and limit results
// TODO Auto-Corrections / Suggestion based on common suffixes
// TODO Postcode needs to load street and check buildings! Store postcode as bbox not as City! - '1186RZ 324' (NL, UK) 
// TODO Search near key objects (subway station artificial bbox)
// TODO New Geocoding for cases ("NC 42" == "NC-42") - geo index for prefixes
// TODO Add flats: https://www.openstreetmap.org/node/5843642738
// TODO English postcodes
// TODO Precise Boundary 'Chernihiv sport life' mostly Kyiv - check precise boundary for filter
// TODO Short word split "Ro-ki" vs "Roki" 
```

vshcherb:
Checked the examples of https://github.com/osmandapp/OsmAnd/issues/16359 on the current spatial search: android master 8fdd5f03e7, poi_types from resources master, English phrases. Maps: New York City, Berlin, Detmold, Thüringen from 2026-09-13, Orlando from 2026-09-01. Search point: Times Square, Alexanderplatz, downtown Orlando. "N POI" = objects of the suggested category within 3 km.

### Works

[...]

PR #25953, opened 2026-09-09
Spatial search: rank by a score, deduplicate by name and distance

vshcherb:
Part of #25941.

`SpatialSearchResult.compare()` is a lexicographic ladder: the first differing tier decides and
the rest are never read, so one point of travel elo beats a result 50 km closer and every POI has
typeOrder 0. Below the structural tiers it is replaced by one bounded score - name, kind, fame,
distance - and deduplication learns to see a place returned as several rows.

What the score encodes, all of it read off human verdicts rather than designed:

- between two ordinary POIs the nearer wins: an exact name, a matching category and a higher elo
  all lose to distance;
- a place answers its own name from any distance - "new york" from Amsterdam returned the city
  49th, under 45 outlets of a pizza chain;
- a name is matched against every language the object carries, not the default one: Everest is
  indexed as 珠穆朗玛峰 and scored as an unrelated object;
- a node that only describes a place (platform, stop position, junction) loses to the place even
  from 5.5 km closer.

Deduplication merges rows of the same name by distance, with the radius depending on what they
are: the parts one facility is spread over within 400 m, ways of one street when the city matches,
any other pair within 30 m, and a bridge carrying a street name into the street unless it has a
travel rating of its own. A stop stored as a dozen OSM nodes shares no id, wikidata or route id,
which is why the existing keys could not see it.

The one-object tier keeps "Dr Lucas" off the 74th row, with two corrections computed per result so
the comparator stays transitive: "<object> in <city>" counts as one when the city was named by its
own name, and a subordinate node never counts as one.

**How it is checked.** `SpatialSearchPreferencesTest` reads `spatial_search/preferences.jsonl`:
133 records of the form "for query Q from position L, object B belongs above object A", addressed
by OSM id, append-only, contradictions kept and cross-linked. That sentence survives a rewrite of
the ranker, which a golden result list does not - the corpus in `test-resources/spatial_search`
rewrites 49 of 74 files on any ranking change and says nothing about whether the change was good.
Judgements that a later round reversed are superseded, never edited; every constant in the code
names the records it answers to.

    ladder (master)                 28 satisfied
    score, first weights            35
    score + deduplication, today    66 satisfied, 0 violated

Only what a person can see is asserted: a preference counts when at least one of its objects is
inside the top 10.

## AI disclaimer

Prompts used (summarised):
1. Make the contrastive-pair review easier: show the full engine key, say where each object is,
   suggest an answer, 10-20 cases at a time, and turn my choices into readable rules.
2. Save my verdicts, tune the score to match them, and generate the next batch.
3. Reported cases to fix: a stop named after a street outranking the street; "liechtenstein"
   returning the country 22nd; "new york" returning the city 49th; bridges listed separately from
   their street; farms with a high rating must not push out the nearest farms.
4. Simplify the deduplication, shorten the comments and reference preference ids instead, use the
   ranking as an instance, and assert only changes in the top 10.
5. Squash into one commit and open a new pull request.

Decided by the agent, not requested explicitly: the shape of the score and every weight in it
(fitted offline on dumped ranking features, then confirmed by the test); the preference record
format, `supersedes`/`absorbed`/`note` handling and the top-10 rule in the checker; the
deduplication radii, chosen from the judged distances; making the one-object tier a per-result
number after a pairwise version broke comparator transitivity.


vshcherb:
Closing: the work continues in #25935 on the original branch, reviewed step by step.

PR #25958, opened 2026-09-10
Routing to OsmAnd-shared, step 18: the native router behind an interface

alex-dev-neo:
Stacked on #25952, so the diff to read is the last commit.

There is **one** implementation of the routing itself — the C++ in OsmAnd-core-legacy — and **two** ways of reaching it:

| | how |
|---|---|
| android, jvm | JNI, `NativeLibrary` marshals java objects across |
| iOS | links the same C++ and calls `router->searchRoute(...)` directly |

Nothing shared can name either of them, so nothing shared can ask for a route. `NativeRouting` is that ask, in shared types only: run a route and get the segments back raw, ask whether a private road stands in the way, release the C++ session.

`NativeLibrary` implements it, and `RoutingContext.nativeLib` moves up onto `RoutingRequest` typed as the interface — so the dozen `ctx.nativeLib != null` checks in the planner do not change at all.

### What the interface deliberately does not carry

- **`loadRouteRegion`**, which fills the java planner's tile cache. That is the fallback's own machinery, not the route calculation, and only the JNI library serves it — `RoutingContext` casts for it in the one place it is used.
- **`runNativeSearchGpxRoute`**, which takes `GpxRouteApproximation` and the planner's `GpxPoint`. Neither is a shared type yet, so `GpxRouteApproximation` casts too.
- **the java planner**, which stays in `net.osmand.router` as the fallback for when there is no native library at all — the server, the tools, SAFE_MODE on android — and which nothing outside java calls.

### Two names that could not be kept

- `deleteNativeRoutingContext` is a **static native**, and JNI binds it by the name it is declared under. So the interface method is `releaseNativeRoutingContext` and the static keeps its name; renaming it would have broken native routing at runtime with nothing failing to build.
- `RoutingContext.deleteNativeRoutingContext()` now goes through the instance. `RoutePlannerFrontEnd` briefly nulls `nativeLib` around the base route, and this also runs from `finalize()`, so it checks for null first — the static call could not fail that way.

### The one thing left loaded, and where it gets unloaded

The interface says `RoutingRequest`, but `java_wrap.cpp` still resolves the request's fields on `net/osmand/router/RoutingContext`. So this implementation can only hand the core one of those. Binding them on `RoutingRequest` instead is a paired core-legacy change, and it belongs with **step 19**, where shared code first builds a request of its own. Until then `NativeLibrary` says so and throws, rather than letting the core read fields off the wrong class.

**No core-legacy change in this step** — the second in a row.

### Checks

459 routing tests in `OsmAnd-java` pass, the app compiles, `java-tools` compiles across all five modules **unchanged**, `OsmAnd-shared` passes on the jvm. On iOS the only failures are the six already failing on this branch's base (`GpxUtilitiesLoad`, `KGeoPointParserUtil`, `NetworkAPI`).

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/NativeLibrary.java
+++ b/OsmAnd-java/src/main/java/net/osmand/NativeLibrary.java
@@ -36,6 +36,8 @@ import net.osmand.render.RenderingRulesStorage;
 import net.osmand.shared.routing.GeneralRouter;
 import net.osmand.router.GpxRouteApproximation;
 import net.osmand.shared.routing.HHRoutingConfig;
+import net.osmand.shared.routing.NativeRouting;
+import net.osmand.shared.routing.RoutingRequest;
 import net.osmand.router.HHRoutePlanner;
 import net.osmand.router.NativeTransportRoutingResult;
 import net.osmand.shared.routing.RouteCalculationProgress;
@@ -47,7 +49,7 @@ import net.osmand.router.TransportRoutingConfiguration;
 import net.osmand.util.Algorithms;
 import net.osmand.util.MapUtils;
 
-public class NativeLibrary {
+public class NativeLibrary implements NativeRouting {
 
 
 	public NativeLibrary() {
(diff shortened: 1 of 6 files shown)


alex-dev-neo:

PR #25959, opened 2026-09-10
Routing to OsmAnd-shared, step 19: a shared entry point for a native route

alex-dev-neo:
Stacked on #25958, so the diff to read is the last commit.

`NativeRouteCalculation` takes a `RoutingRequest` and a `NativeRouting` and gives back the segments. Nothing in it names a platform, so android, the jvm and iOS take the same path into the same C++ and get the same route.

What comes back is **raw**: the turns are still worked out afterwards, and that is still two implementations, one per platform. Removing that second one is step 20.

### What moved in

Three things that were inline in `RoutePlannerFrontEnd`:

- the progress reset and the straight line estimate the progress bar is drawn from;
- the normalisation of absent intermediates;
- the call itself.

`refreshProgressDistance` is shared with the java fallback, which calls it in the other branch, so there is one copy of that arithmetic now rather than one per planner.

### What could not

The **list of regions to search** is gathered from the reader map, which belongs to the java planner, so the caller passes it in. The core only needs it to match its own regions back to the java objects by file pointer and length — it opens the files itself.

Splicing the already driven part back onto a recalculated route stays behind too: it walks `BinaryRoutePlanner.RouteSegment`.

### The paired change, and what it unlocks

osmandapp/OsmAnd-core-legacy#362 reads the request's fields off `net/osmand/shared/routing/RoutingRequest` instead of the java subclass. That is what lets a request built by shared code be routed at all, and it is why `NativeLibrary` no longer refuses one — the guard added in #25958 is gone and the native declarations take a `RoutingRequest`.

### The native path is not covered by the tests here

Worth saying plainly. The routing tests fall back to the java planner when `core-legacy/binaries` is absent, which it is on this machine, so the **459 passing tests exercise the java side of this change and not the JNI side.**

What was checked instead:

- all twenty two bound field names and descriptors against the compiled `RoutingRequest` — 22 of 22;
- that each of the twenty one native methods in `NativeLibrary` still has a `Java_net_osmand_NativeLibrary_<name>` in `java_wrap.cpp`, with none of them overloaded, so widening a declared parameter type moved no symbol.

The app compiles, `java-tools` compiles unchanged, `OsmAnd-shared` passes on the jvm. On iOS the only failures are the six already failing on this branch's base.

### DIFF
--- a/OsmAnd-java/src/main/java/net/osmand/NativeLibrary.java
+++ b/OsmAnd-java/src/main/java/net/osmand/NativeLibrary.java
@@ -238,30 +238,14 @@ public class NativeLibrary implements NativeRouting {
 
 	@Override
 	public RouteSegmentResult[] runNativeRouting(RoutingRequest c, HHRoutingConfig hhRoutingConfig, RouteRegion[] regions, boolean basemap) {
-		RoutingContext ctx = asRoutingContext(c);
 		// if hhRoutingConfig == null - process old routing
 		if (hhRoutingConfig != null) {
 			setHHNativeFilterAndParameters(c);
 		}
 		final float CPP_NO_DIRECTION = -2 * (float) Math.PI;
-		return nativeRouting(ctx, hhRoutingConfig, c.config.initialDirection == null ?
+		return nativeRouting(c, hhRoutingConfig, c.config.initialDirection == null ?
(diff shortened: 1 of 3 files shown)


alex-dev-neo:

PR #25961, opened 2026-09-10
Routing to OsmAnd-shared, step 20a: the turn lane logic

alex-dev-neo:
Stacked on #25959, so the diff to read is the last commit.

Step 20 is the last one: `RouteResultPreparation`, the part that turns raw segments into the manoeuvres a driver is actually told about. It runs on **both** paths, so it is what makes android and iOS agree — and today they do not, because the C++ has its own copy of all 2374 lines in `routeResultPreparation.cpp`, android uses the java one and iOS uses that.

### Why this is three PRs and not one

2564 lines do not go in one step. I split them by **what each method holds onto**, not by size:

| | methods | lines |
|---|---|---|
| touch neither the context nor a java type | 60 | 1793 |
| need the context, no java types | 8 | 276 |
| held by the printing, the planner's segment types, or the tile machinery | 8 | 394 |

Nearly all the java-only weight — `RenderingRulesStorage`, `XmlPullParser`, `FileWriter`, `ObfConstants` — turned out to be in the debug printing, not in the turn logic, so it stays behind and blocks nothing.

**This PR is the first closed piece of the first group**: `TurnLanes`, the 25 methods that read `turn:lanes` off a road, decide which lanes the driver may be in, and turn that into the arrows for a manoeuvre. Closed means nothing in it calls anything left behind — checked, not assumed. `RoadSplitStructure` and its `AttachedRoadInfo` come along because two of them take it, and step 20b needs it too; it had no imports beyond `java.util`.

20b will be the turn construction that works through `RoadSplitStructure` (10 methods, 488 lines), 20c `prepareResult` itself plus hooks for the four machinery calls.

### The one trap that was real

Java's `split(regex)` **drops** trailing empty parts; Kotlin's **keeps** them. So `"left;"` would have gained a lane and `"left||"` kept two java never saw. And the same line splits twice with opposite rules — the outer split on `|` uses `split(regex, -1)`, which *does* keep them:

```java
String[] splitLaneOptions = turnLanes.split("\\|", -1);   // keeps empties
String[] laneOptions = splitLaneOptions[i].split(";");    // drops trailing empties
```

`splitDroppingTrailingEmpty` reproduces java's rule including its special case, where a string containing no delimiter comes back whole even when it is empty.

`TurnLanesCompatTest` runs the **java originals** of `calculateRawTurnLanes` and `getUniqTurnTypes` against the ported ones over 21 hand-picked strings and 20 000 random ones. It lives in `OsmAnd-java` because that is the one module where both implementations are on the classpath, next to `TroveComparisonTest` and `StringBundleCompatTest`.

Two smaller ones: java divided by a float `2f` when narrowing the active lane range and the port keeps that, and `MAX_SPEAK_PRIORITY` stays in java as well because 20b still uses it there.

### Checks, with the native path actually running

The C++ library is built for the host for this, so the `*NativeTest` classes are not silently taking the java planner:

| | |
|---|---|
| `TurnLanesCompatTest` | 4 |


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26011.

PR #25962, opened 2026-09-10
Routing to OsmAnd-shared, step 20b: the fork turn construction

alex-dev-neo:
Stacked on #25961, so the diff to read is the last commit.

The second of the three pieces of step 20. Where 20a was the lane arrays on their own, this is the part that needs the picture of the **whole intersection**: what else leaves it, on which side, with how many lanes, and whether any of it is worth speaking. That picture is `RoadSplitStructure`, which moved with 20a because two of its methods already took one.

Ten methods, closed the same way — nothing in it calls anything left behind:

| | |
|---|---|
| `calculateRoadSplitStructure` | builds the picture, skipping roads the previous one is forbidden to turn into |
| `findActiveIndex` + two strategies | which run of lanes the driver has to be in — by lane counts, else by the directions on offer |
| `createKeepLeftRightTurnBasedOnTurnTypes` | the manoeuvre when the road declares its lanes, so the lanes decide |
| `createSimpleKeepLeftRightTurn` | the manoeuvre when it does not, inventing one arrow per lane from the geometry |
| `createCombinedTurnTypeForSingleLane` | the single lane case, showing up to two of the roads left behind |
| `getTurnLanesInfo`, `attachKeepLeftInfoAndLanes`, `getAttachedLanesCount` | the entry points and their arithmetic |

`RouteResultPreparation` is **1338 lines** now, down from 2564 before 20a. What is left is 20c: `prepareResult` itself, the eight context methods, the debug printing and the four machinery calls.

### Two things worth naming

**`normalizeDegrees360` takes a float** and java assigned its result to a `double`, so the port narrows and widens in the same places rather than computing the angle in double throughout. Bearings are floats, and these comparisons decide which side of the road a junction falls on — computing them in double would move the boundary.

**The turn restriction types are declared in shared now**, in `RouteDataObject`'s companion, which is what `getRestrictionType` answers with. `MapRenderingTypes` in `OsmAnd-java` keeps its own seven for the rendering and the map creator, where fourteen places name them through that class; the two sets merge when it moves. Wiring java's copies to the shared ones was tried and dropped — it drags a rendering class into a routing change for no gain today.

### Checks, with the native path running

The C++ library is built for the host, so the `*NativeTest` classes are not silently taking the java planner. The one that matters here is `RouteResultPreparationNativeTest`: it compares the route information this code produces against recorded expectations, **turn lanes included**.

| | |
|---|---|
| `TurnLanesCompatTest` | 4 |


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26012.

PR #25963, opened 2026-09-10
Routing to OsmAnd-shared, step 20c: the turn and time preparation

alex-dev-neo:
Stacked on #25962, so the diff to read is the last commit. **The last step of the migration.**

`TurnPreparation` gives every segment the manoeuvre that leads onto it, then reconciles them with each other — lanes merged across junctions close enough to be one decision, straights that belong to the same intersection silenced, a left then a left over a short segment joined into a u turn, keeps that say nothing dropped — and works out how long each segment takes at what speed. `RoundaboutTurn` goes with it: it had exactly one caller, and that caller moved.

`RouteResultPreparation` is **789 lines**, from 2564 before 20a.

### What is left in java, and why

| | |
|---|---|
| `prepareResult` | the orchestrator — it reaches the routing context's tile machinery and the reader a road's region is initialised from |
| area routing | `combineWayPointsForAreaRouting`, `simplifyAreaRouteWay` walk the planner's segments |
| the debug printing | `RenderingRulesStorage`, `XmlPullParser`, `FileWriter` |
| `convertFinalSegmentToResults` | walks the java planner's own `FinalRouteSegment` |

So `prepareResult` stays and **there are no callbacks in either direction** — it calls the shared preparation the way it called the local methods before. The only field that had to move up to `RoutingRequest` is `leftSideNavigation`, which decides what a sharp turn is called.

### Three things stayed behind on purpose

- **`addTurnInfoDescriptions`** builds `"go 1.2 km"` with `String.format("%.1f")`, and reproducing java's float formatting has bitten this migration before. It is descriptive text nothing in the app reads, so it stays in java and `prepareResult` calls it after the shared preparation, in the same order as before.
- **`validateAllPointsConnected`** prints to stdout.
- **`UNMATCHED_HIGHWAY_TYPE`** is declared in shared and java's constant takes its value from there, so the five places that name it through `RouteResultPreparation` are untouched and there is one definition.

### What this finishes

Turn instructions were computed by two implementations — java's 2564 lines and the C++ 2374 in `routeResultPreparation.cpp` — from segments the same C++ router produced, with android using one and iOS the other. The shared half is now the one that decides the manoeuvre and the lanes.

### Checks, with the native path running

`RouteResultPreparationNativeTest` compares the route information **this code** produces against recorded expectations:

| | |
|---|---|
| `TurnLanesCompatTest` | 4 |


alex-dev-neo:
Closing: the approach changed to copying the classes into OsmAnd-shared without deleting them from OsmAnd-java, and android and tools keep calling the java classes. Replaced by #26016.

PR #25967, opened 2026-09-10
fix switching back to system default language

### LINKED ISSUE #25887: Switching app language back to "System default" has no effect (selecting a specific language works)
## Description
Changing the app language through the Android per-app language settings works only when a specific language is selected. Switching from Ukrainian directly to **English** applies correctly, but switching to **System default** does not: the app keeps using the previously selected language instead of falling back to the system one.

> Previous task https://github.com/osmandapp/OsmAnd/issues/20683

### How to reproduce
* Open **Android Settings** → **OsmAnd** → **Language**
* Switch the language to **Ukrainian**
* Force stop the app
* Launch the app — all menus are displayed in Ukrainian
* Return to **Language** and switch back to **System default**
* Force stop the app and relaunch it

### Actual result

<table style="width:100%">
  <tr>
    <th>After install, before changes</th>
    <th>System default is set in Settings</th>
    <th>Video</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
    <td><video src="https://github.com/user-attachments/assets/5c22519f-7931-4682-818e-116972149740" width="300" height="auto"></video></td>
  </tr>
</table>

* The app interface is still in Ukrainian
* The in-app **Display language** setting (reintroduced earlier as a workaround for manufacturers with incomplete per-app language support) shows `uk_UA`
* Selecting a specific language instead of **System default** (for example, **English**) works as expected, so the issue appears to be limited to the **System default** option

### Expected result
* The app follows the system language (**English** in this case) after **System default** is selected

**Environment**
OsmAnd Version: OsmAnd Nightly 5.4.0#8716m, released: 2026-09-07
Android: 17
Device model: Google Pixel 6A

aditya-mitra:
clear the previously saved language

[Screen_recording_20260910_162853 (1).webm](https://github.com/user-attachments/assets/0b1c96e4-d063-49e4-9e5b-cd637f5d9dc4)

closes https://github.com/osmandapp/OsmAnd/issues/25887

### DIFF
--- a/OsmAnd/src/net/osmand/plus/helpers/LocaleHelper.java
+++ b/OsmAnd/src/net/osmand/plus/helpers/LocaleHelper.java
@@ -79,13 +79,9 @@ public class LocaleHelper {
 			currentLocale = SupportedLocale.normalizeToOsmandLegacy(currentLocale);
 
 			if (!Algorithms.stringsEqual(currentLocale, locale)) {
-				if (Algorithms.isEmpty(currentLocale) && !Algorithms.isEmpty(locale)) {
-					// Ignore empty OS response if vendor firmware rejected a rare tag (e.g., "sc").
-				} else {
-					// Sync with OS if user changed the language via Android App Info.
-					locale = currentLocale;
-					settings.PREFERRED_LOCALE.set(locale);
-				}
+				// Sync with OS if user changed the language via Android App Info.
+				locale = currentLocale;
+				settings.PREFERRED_LOCALE.set(locale);
 			}
 		}
 


Kreiri:
Unfortunately, changes in this PR remove Rare Language Protection feature of OsmAnd. Osmand supports a number of languages, like Sardinian (sc) or Asturian (ast) that may not be supported by an Android vendor. When user selects such locale in OsmAnd, it's passed to the system; however, if such locale is not supported by the system, the system will set "chosen locale" for OsmAnd to "system default"  and LocaleManagerCompat.getApplicationLocales will return empty list. This PR remove handling of the case of "OsmAnd has user-set locale but it is not recognized by the system".

aditya-mitra:
makes sense.
implemented a check for checking if the saved language is supported by the device.

OsmAnd/src/net/osmand/plus/helpers/LocaleHelper.java
```diff
@@ -123,6 +123,20 @@ public void checkPreferredLocale() {
 		}
 	}
 
+	private boolean isLocaleSupportedBySystem(@NonNull String localeId) {
+		Locale locale = SupportedLocale.parseLocale(localeId);
+		if (locale == null) {
+			return false;
+		}
+		for (String systemLocaleId : Resources.getSystem().getAssets().getLocales()) {
+			Locale systemLocale = Locale.forLanguageTag(systemLocaleId.replace('_', '-'));
```
Kreiri: as per https://developer.android.com/reference/android/content/res/AssetManager#getLocales() , 

> On SDK 21 (Android 5.0: Lollipop) and above, Locale strings are valid [BCP-47](https://tools.ietf.org/html/bcp47) language tags and can be parsed using [Locale.forLanguageTag(String)](https://developer.android.com/reference/java/util/Locale#forLanguageTag(java.lang.String)).

minSdk for OsmAnd is 24, so there's no need for `.replace('_', '-')` there.

aditya-mitra: i did not know this. thanks!

OsmAnd/src/net/osmand/plus/helpers/LocaleHelper.java
```diff
@@ -79,7 +79,7 @@ public void checkPreferredLocale() {
 			currentLocale = SupportedLocale.normalizeToOsmandLegacy(currentLocale);
 
 			if (!Algorithms.stringsEqual(currentLocale, locale)) {
-				if (Algorithms.isEmpty(currentLocale) && !Algorithms.isEmpty(locale)) {
+				if (Algorithms.isEmpty(currentLocale) && !isLocaleSupportedBySystem(locale)) {
```
Chumva: (gen by AI) This condition now equates "OS reports no per-app locale" with "the user chose System default" for every language that has framework translations. But the OS side is also empty when the preference exists and was never pushed to the OS:

- Android key/value backup restore: `OsmandBackupAgent` restores `net.osmand.settings` (incl. `preferred_locale`) by writing the prefs file directly, so no listener runs and nothing calls `setApplicationLocales`; the OS-side per-app locale is not part of this app's backup set.
- Upgrade from Android ≤12 to 13+ (by code inspection): `onPreferredLocaleChanged` is `TIRAMISU`-gated and `checkPreferredLocale` never pushes, so the OS side stays empty.

Reproduced on the API 36 emulator: `de` set → `bmgr backupnow` → System default → `bmgr restore` leaves `preferred_locale=de` with `getApplicationLocales()` empty. On that identical state a build with the previous guard (5.4.0 nightly) starts in German; this build resets the preference to `""` and starts in English — the user's language is silently lost.

Suggestion: make the decision state-based instead of language-based. Store the locale the OS last reported (whenever `appLocales` is non-empty) and treat an empty answer as an explicit System default only if it equals the stored preference; otherwise keep the preference (vendor rejected it, or it arrived via backup/OTA and was never pushed). That also resolves System-default-after-Sardinian on firmwares that accept the tag, which this heuristic leaves stuck. The marker must not travel with the backup (separate prefs file, or bound to the device), or a restored marker recreates the same problem.

OsmAnd/src/net/osmand/plus/helpers/LocaleHelper.java
```diff
@@ -123,6 +123,20 @@ public void checkPreferredLocale() {
 		}
 	}
 
+	private boolean isLocaleSupportedBySystem(@NonNull String localeId) {
+		Locale locale = SupportedLocale.parseLocale(localeId);
+		if (locale == null) {
+			return false;
+		}
+		for (String systemLocaleId : Resources.getSystem().getAssets().getLocales()) {
```
Chumva: (gen by AI) `Resources.getSystem().getAssets().getLocales()` lists the languages the firmware's own resources are translated into, not the locales `LocaleManager` accepts — and it varies per OEM.

Measured (`aapt2 dump configurations` on the pulled `framework-res.apk`): the Pixel 8 Pro lacks `ast cy eo hsb kab nn oc sat sc`; the Xiaomi 2405CRPFDG additionally lacks `af` and `is` (and has no `b+sr+Latn`, only `sr`). Consequence, reproduced on the Xiaomi: Afrikaans → App info → System default → the app stays in Afrikaans with `preferred_locale=af`, i.e. #25887 persists there for Afrikaans/Icelandic, while the same steps work on the Pixel.

None of the three firmwares here (Pixel Android 17, Xiaomi HyperOS Android 16, emulator API 36) rejects `sc`/`ast`/`hsb`/`sr-Latn` via `cmd locale set-app-locales`, so "translation present" is not a proxy for "tag accepted" in either direction. The stock Settings picker also offers these rare languages (e.g. Asturianu), so the ambiguous "rare language → System default" case is reachable on stock devices too.

Minor: the language-only comparison treats `sr+Latn` as supported wherever bare `sr` exists, and the scan re-runs on every `checkPreferredLocale()` call (`WebViewEx` constructor, `onConfigurationChanged`) while the app is in the protected state — the set is immutable for the process, so it could be computed once.

Chumva (commented):
(gen by AI) Tested on Pixel 8 Pro / Android 17, Xiaomi 2405CRPFDG / HyperOS Android 16 and an API 36 emulator, built from 37945b4 (`gplayFreeOpenglArm64Debug`, `nightlyFreeOpenglX86Debug`).

The reported flow is fixed: App info → Language → *System default* takes effect after a restart (verified through the real Settings screen), and the rare-language guard still holds (`sc` with an empty OS side stays Sardinian).

Two issues, details in the inline comments:

1. **Every "preference set, OS side empty" state is now treated as the user choosing System default.** A key/value backup restore (or an upgrade from Android ≤12) produces that state without any user action — a build with the previous guard keeps the language there, this build silently resets it to System default (reproduced with `bmgr` on the emulator).
2. **`getAssets().getLocales()` is the firmware's translation coverage, not what `LocaleManager` accepts, and it differs per OEM.** On the Xiaomi tablet `af` and `is` are missing from framework-res, so for Afrikaans/Icelandic the original bug remains on that device (reproduced). None of the three firmwares rejects rare tags at all, so the protected branch could only be simulated.

A state-based check would cover both: remember the locale the OS last reported back, and treat an empty OS answer as an explicit System default only when it matches the stored preference.

OsmAnd/src/net/osmand/plus/helpers/LocaleHelper.java
aditya-mitra: resolved by preserving restored app lang when app locale is empty

aditya-mitra:
implemented changes for preserving translation guess and it now saves the last nonempty reported locale.

for the android <=12 to 13+ upgrade seems like a rare edge case. should we handle it? (we would need to send the current lang to android for confirmation - more case handlings)

OsmAnd/src/net/osmand/plus/helpers/LocaleHelper.java
```diff
@@ -74,23 +77,28 @@ public void checkPreferredLocale() {
...
 			LocaleListCompat appLocales = LocaleManagerCompat.getApplicationLocales(app);
 			String currentLocale = appLocales.isEmpty() ? "" : appLocales.get(0).toLanguageTag();
 			currentLocale = SupportedLocale.normalizeToOsmandLegacy(currentLocale);
 
-			if (!Algorithms.stringsEqual(currentLocale, locale)) {
-				if (Algorithms.isEmpty(currentLocale) && !Algorithms.isEmpty(locale)) {
-					// Ignore empty OS response if vendor firmware rejected a rare tag (e.g., "sc").
-				} else {
-					// Sync with OS if user changed the language via Android App Info.
+			if (!Algorithms.isEmpty(currentLocale)) {
```
Kreiri, line 85: Please explain the logic of your new proposed solution.

aditya-mitra: so the android system reporting nothing for the lang was ambiguous. 
previously, when android returned nothing, it kept the old language in the app and not the user's selected system default lang.
so now when android reports nothing, the app's saved preference is cleared and it follows system default.

PR #25978, opened 2026-09-11
Media gallery 3283

0xRe1nk0:


OsmAnd/src/net/osmand/plus/plugins/audionotes/library/MediaLibraryFragment.kt
```diff
@@ -0,0 +1,223 @@
...
+	private fun updateChips() {
+		val view = chips ?: return
+		val mode = controller.sortMode
+		fun chip(id: String, icon: Int, title: Int, selected: Boolean) = ChipsLayout.ChipData(id,
+			if (selected) 0 else icon, getString(title), selected, true, true, false,
+			ChipsLayout.TextColorStyle.PRIMARY, ChipsLayout.IconColorStyle.DEFAULT)
+		view.updateContent(listOf(
+			ChipsLayout.DropDownChipData(SORT_CHIP, mode.iconId, getString(mode.titleId), false, true, true,
+				ChipsLayout.TextColorStyle.PRIMARY, ChipsLayout.IconColorStyle.ACTIVE,
+				selectionStyle = OsmAndDropdownMenuSelectionStyle.CHECKMARK,
```
Chumva: (gen by AI) `OsmAndDropdownMenuSelectionStyle` no longer exists on master: 9d4934b92d replaced the dropdown with the Compose component, which always puts a trailing checkmark on the selected option. After rebasing, this argument, its import and the `selectionStyle` plumbing that 13f2297d adds to `ChipsLayout` can be dropped, and the sort chip keeps the checkmark look.

OsmAnd/src/net/osmand/plus/plugins/audionotes/library/MediaItemMenu.kt
```diff
@@ -0,0 +1,193 @@
...
+				GalleryPagerController.show(activity, GalleryKey.MediaLibrary, entry.id, orderedIds, autoPlay = true)
+			})
+			if (entry.lat != null && entry.lon != null) add(item(app, R.string.shared_string_show_on_map,
+				R.drawable.ic_show_on_map_outlined, nightMode) { ShowOnMapNavigator.show(activity, entry) })
+			add(item(app, R.string.shared_string_details, R.drawable.ic_action_info_outlined, nightMode, divider = true) {
+				GalleryPagerController.getInstance(app, GalleryKey.MediaLibrary).openDetails(activity, entry.mediaItem, orderedIds)
+			})
+			if (entry.mediaItem is MediaItem.Internal) add(item(app, R.string.shared_string_rename,
+				R.drawable.ic_action_edit_outlined, nightMode) { rename(activity, entry, nightMode) })
+			add(item(app, R.string.attach_to_track, R.drawable.ic_action_track_add, nightMode, divider = true, enabled = false) {})
```
Chumva: (gen by AI) After rebasing, this item is not disabled. On master `PopUpMenu` renders through `showComposeDropdownMenu()`, and `PopUpMenuItem.toDropdownOption()` computes `enabled = hasClickListener || isShowCompoundBtn || shouldDismissOnClick()`, so `setEnabled(false)` is ignored and this placeholder shows as an active item that does nothing (the same goes for the empty-selection Share/Delete in `showSelection()`). `toDropdownOption()` should also honour `isEnabled()`; the Compose menu already dims disabled options. The `PopUpMenuArrayAdapter` changes can go, since master deleted that class in 5756d194d5.

OsmAnd/src/net/osmand/plus/plugins/audionotes/library/MediaLibraryFragment.kt
```diff
@@ -0,0 +1,223 @@
...
+		controller.attach(this)
+		return root
+	}
+
+	override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+		super.onViewCreated(view, savedInstanceState)
+		requireActivity().onBackPressedDispatcher.addCallback(viewLifecycleOwner, backCallback)
+	}
+
+	override fun onResume() {
```
Chumva: (gen by AI) Nothing refreshes the library when My Places comes back to the front. Deleting a note on the map (`AudioVideoNoteMenuController`, `AudioNotesLayer`) only calls `plugin.deleteRecording()`, `RecordingsListener` has no removal callback, and `MapActivity.launchPrevActivityIntent()` brings the existing activity back with `FLAG_ACTIVITY_REORDER_TO_FRONT`, so the deleted note stays in the list. The old `NotesFragment` rebuilt its list in `onResume()`; calling `app.galleryHelper.mediaLibraryRepository.refresh()` here, or adding a removal event to the plugin, would cover it.

OsmAnd/src/net/osmand/plus/gallery/library/MediaLibraryScanner.kt
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}
+		for (source in storage.listMedia(MediaStorageLocation.fromSettings(app))) {
+			add(Link(source.href).apply { text = source.fileName })
+		}
+		for (recording in plugin?.allRecordings.orEmpty().toList()) {
+			add(Link(storage.createMediaFileHref(recording.file)).apply { text = recording.file.name }, recording = recording)
+		}
+		fun addTarget(target: Linkable, kind: MediaAttachment.Kind, name: String, latLon: LatLon, path: String? = null) {
+			for (link in target.links.orEmpty().toList()) {
```
Chumva: (gen by AI) Every new note is also linked to a favourite (`convertRecordingsToFavorites()`), and deleting the note on the map leaves that link behind. This loop has no existence check (the old `NotesFragment` skipped local links whose `getPlaybackUri()` was null), so the missing file comes back as an entry without metadata. Deleting it then always fails: `resolveMediaSource(..., true)` returns null for a missing file, and `MediaItemMenu.delete()` removes links only for entries whose file was deleted, so only Detach in the details sheet clears it. Skipping internal links whose file is gone, or letting Delete drop the links when the file is already missing, would fix it.

OsmAnd/src/net/osmand/plus/plugins/audionotes/library/MediaItemMenu.kt
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	private fun delete(activity: FragmentActivity, entries: List<MediaLibraryEntry>, onDone: () -> Unit) {
+		val app = activity.application as OsmandApplication
+		val storage = MediaStorageHelper(app)
+		val location = MediaStorageLocation.fromSettings(app)
+		app.taskManager.runInBackground(object : OsmAndTaskRunnable<Void, Void, List<MediaLibraryEntry>>() {
+			override fun doInBackground(vararg params: Void?): List<MediaLibraryEntry> = entries.filter { entry ->
+				runCatching {
+					val source = storage.resolveMediaSource(location, entry.href, true)
```
Chumva: (gen by AI) With `includeNonManagedMedia = true` this resolves any existing `file://` path or `content://` URI, so Delete erases files OsmAnd does not own, for example media linked from an imported track's waypoint, wherever Android permits it. The old tab only detached such links, and `DeleteMediaFilesTask` deliberately resolves with `false`. Where Android refuses, and for every http(s) link, the delete fails and the entry stays, because the item menu has no Detach. Suggest limiting Delete to managed media and offering Detach, with the existing `DETACH_EXTERNAL` wording, for everything else.

OsmAnd/src/net/osmand/plus/gallery/ui/GalleryGridRecyclerView.java
```diff
@@ -37,27 +37,38 @@ public void setGestureFinishedListener(@Nullable Runnable listener) {
...
 
 	@Override
 	public boolean onInterceptTouchEvent(MotionEvent e) {
-		scaleDetector.onTouchEvent(e);
-		if (e.getAction() == MotionEvent.ACTION_UP) {
-			isScaling = false;
+		if (scaleDetector != null) {
+			scaleDetector.onTouchEvent(e);
 		}
 		if (isScaling) {
```
Chumva: (gen by AI) Likely issue, not checked on a device: in My Places this grid sits inside `LockableViewPager`, which blocks swipes only in selection mode. Nothing here calls `requestDisallowInterceptTouchEvent(true)`, and during a pinch the layout manager is scroll-locked, so RecyclerView does not request it either. `ViewPager.onInterceptTouchEvent()` follows only the first pointer, so a pinch whose first finger moves mostly sideways can be taken over as a tab swipe. Calling `getParent().requestDisallowInterceptTouchEvent(true)` as soon as a second pointer goes down should prevent it.

OsmAnd/src/net/osmand/plus/gallery/ui/GalleryGridBinder.kt
```diff
@@ -0,0 +1,240 @@
...
+					(recyclerView.layoutManager as? GalleryGridLayoutManager)?.let { manager ->
+						if (manager.extraLayoutSpacePx != px) {
+							manager.extraLayoutSpacePx = px
+							manager.requestLayout()
+						}
+					}
+				}
+
+				override fun commitSpan(span: Int, anchorPosition: Int, anchorOffset: Int, onLaidOut: () -> Unit) {
+					val manager = recyclerView.layoutManager as? GalleryGridLayoutManager ?: return
```
Chumva: (gen by AI) If the layout manager is no longer a `GalleryGridLayoutManager` when a pinch settle ends (the Grid chip tapped during the settle animation, or a third finger), this returns before `onLaidOut`. That callback is the only place `GridPinchController.commit()` clears the captured views' transforms and calls `setScrollLocked(false)`, and `applyLayout()` copies the lock into the new list layout manager, so the list stops scrolling and pooled cells keep their pinch scale until the next pinch. Suggest always invoking `onLaidOut` and releasing the pinch in `updateDisplayMode()`.

Minor, same function: when the target equals the current span, `setItems()` still rebinds every visible cell and reloads its image; it can be skipped in that case.

OsmAnd/src/net/osmand/plus/gallery/library/MediaShareHelper.kt
```diff
@@ -0,0 +1,46 @@
...
+import net.osmand.shared.media.domain.MediaType
+
+object MediaShareHelper {
+	@JvmStatic
+	fun share(activity: FragmentActivity, items: List<MediaItem>) {
+		if (items.isEmpty()) return
+		val app = activity.application as OsmandApplication
+		val uris = items.mapNotNull { app.galleryHelper.mediaSourceResolver.getShareableUri(it) }
+		val intent = Intent()
+		if (uris.size == items.size) {
```
Chumva: (gen by AI) If a single selected item has no shareable URI (a remote attachment, or a local link whose file is gone), every item falls into the text branch, and for `MediaItem.Internal` `MediaUriResolver.getShareUri()` returns the internal path, so the chooser receives lines like `media/<file name>` and no files. The old `ShareRecordingsTask` sent the files as `EXTRA_STREAM` and the remaining links as `EXTRA_TEXT` in one `ACTION_SEND_MULTIPLE` intent; doing the same here keeps the files.

OsmAnd/src/net/osmand/plus/gallery/attached/helpers/AttachedMediaDataHelper.java
```diff
@@ -141,6 +182,86 @@ public void onSavingFavoritesFinished(boolean success) {
...
+			for (Link link : links) {
+				originals.putIfAbsent(link, new Link(link));
+				link.setHref(newHref);
+				link.setText(newName);
+			}
+		}
+		List<Linkable> targets = new ArrayList<>(linksByTarget.keySet());
+		saveTargets(targets, 0, true, saved -> {
+			boolean renamed = Boolean.TRUE.equals(saved) && !destination.exists()
+					&& (recording != null ? recording.setName(name) : source.renameTo(destination));
```
Chumva: (gen by AI) `Recording.setName()` swaps the file, but `RecordingsFileHelper.recordingByFileName` stays keyed by the old name, and `deleteRecording()` later removes by the new name. A recording renamed here and then deleted therefore stays in `getAllRecordings()`, so it reappears in this tab with a missing file and stays on the dashboard. The old rename had the same flaw, but it is now one menu tap away; re-keying the map after a successful rename would close it.

OsmAnd/src/net/osmand/plus/gallery/library/MediaLibraryRepository.kt
```diff
@@ -0,0 +1,136 @@
...
+	private val attachmentsListener: (Set<GalleryKey>) -> Unit = { refresh() }
+	private var metadataUpdatePending = false
+	private val metadataUpdate = Runnable { metadataUpdatePending = false; publish() }
+
+	fun subscribe(listener: (List<MediaLibraryEntry>) -> Unit) {
+		val first = listeners.isEmpty()
+		listeners.add(listener)
+		if (first) {
+			app.favoritesHelper.addListener(favoritesListener)
+			plugin?.addRecordingsListener(recordingsListener)
```
Chumva: (gen by AI) `recordingsListeners` was built for one-shot capture listeners: `cancelPendingRecordingListeners()` replaces the whole list on every cancelled, failed or empty capture (about 20 call sites), so this persistent listener is silently dropped and only re-added after every subscriber has left and a new one arrives. It is hidden today because each new note also becomes a favourite link and triggers the favourites refresh. A separate list for persistent listeners in the plugin would make this robust.

OsmAnd/src/net/osmand/plus/gallery/library/MediaLibraryScanner.kt
```diff
@@ -0,0 +1,71 @@
...
+				val titled = if (link.text.isNullOrBlank()) Link(link).apply { text = link.href?.substringBefore('?')?.substringAfterLast('/') } else link
+				add(titled, MediaAttachment(target, link, kind, name, latLon, path))
+			}
+		}
+		for (point in app.favoritesHelper.favouritePoints.toList()) {
+			addTarget(point, MediaAttachment.Kind.FAVORITE, point.name, LatLon(point.latitude, point.longitude))
+		}
+		for (selected in app.selectedGpxHelper.selectedGPXFiles.toList()) {
```
Chumva: (gen by AI) Only tracks currently shown on the map are scanned, so media attached to waypoints of hidden tracks never appears in the library or under "Attached to", although the tab's own empty state says everything you record or attach appears here. If reading every track here is too expensive, it's worth calling out as a known gap in the PR description.

OsmAnd/res/values/strings.xml
```diff
@@ -11,6 +11,35 @@
...
+	<string name="shared_string_attached_to">Attached to</string>
+	<string name="media_not_attached_title">Not attached to anything yet</string>
+	<string name="media_not_attached_descr">Attach this media to a favorite to see it on the map.</string>
+	<string name="detach_external_media_descr">The file will no longer be shown in this favorite. It stays on your device.</string>
+	<string name="detach_external_media_descr_plural">The files will no longer be shown in this favorite. They stay on your device.</string>
+	<string name="attach_to_track">Attach to track</string>
+	<string name="attach_to_favorite">Attach to favorite</string>
+	<string name="delete_media_title">Delete media?</string>
+	<string name="delete_media_descr">This file will be permanently deleted from your device. It will disappear from all favorites it\'s attached to.</string>
+	<string name="delete_media_title_plural">Delete %1$d media?</string>
```
Chumva: (gen by AI) The counted titles `delete_media_title_plural` and `detach_media_title_plural` are picked with `count == 1`, which cannot be translated correctly for languages with several plural forms (ru, uk, pl, ar...). `<plurals>` with `getQuantityString()` would fix it; this file already has a few `<plurals>` entries.

OsmAnd/src/net/osmand/plus/gallery/library/MediaLibraryRepository.kt
```diff
@@ -0,0 +1,136 @@
...
+	private val listeners = linkedSetOf<(List<MediaLibraryEntry>) -> Unit>()
+	private var scanning = false
+	private var scanAgain = false
+	private val refreshCallbacks = mutableListOf<() -> Unit>()
+	private var metadataRequest: Cancellable? = null
+	private var entries: List<MediaLibraryEntry> = emptyList()
+	var hasSnapshot = false
+		private set
+
+	private val favoritesListener = object : FavoritesListener {
```
Chumva: (gen by AI) Every favourites load, point update and save, every new note and every attachment change runs a full `scan()`: all media folders (with `DocumentFile.listFiles()` for a manually chosen folder, which is slow), every recording, every favourite and every visible track. `ViewerSheetController` also subscribes for any local item, so opening details from a favourite's context menu starts a full scan when the Media tab is not open. Debouncing refreshes, and building the sheet's attachment list for the single item instead of a full scan, would keep this cheap.

OsmAnd/src/net/osmand/plus/plugins/audionotes/AudioVideoNotesPlugin.java
```diff
@@ -205,7 +213,13 @@ public AudioVideoNotesPlugin(@NonNull OsmandApplication app) {
 		registerPreference(recordingsFileHelper.AV_RS_CLIP_LENGTH);
 		registerPreference(recordingsFileHelper.AV_RS_STORAGE_SIZE);
 
-		NOTES_SORT_BY_MODE = registerEnumStringPreference("notes_sort_by_mode", NotesSortByMode.BY_DATE, NotesSortByMode.values(), NotesSortByMode.class);
+		MEDIA_LIBRARY_SORT_MODE = registerEnumStringPreference("media_library_sort_mode", GallerySortMode.NAME_A_Z,
```
Chumva: (gen by AI) The default sort changes from the old tab's date order (`NotesSortByMode.BY_DATE`) to name A-Z, so existing users see a different order after the update. Newest first looks like the closer default; worth confirming with design.

OsmAnd/src/net/osmand/plus/plugins/audionotes/library/MediaItemMenu.kt
```diff
@@ -0,0 +1,193 @@
...
+			})
+		}
+		popup(anchor, nightMode, items)
+	}
+
+	fun showSelection(activity: FragmentActivity, entries: List<MediaLibraryEntry>, anchor: View,
+		nightMode: Boolean, onDone: () -> Unit) {
+		val app = activity.application as OsmandApplication
+		popup(anchor, nightMode, listOf(
+			item(app, R.string.shared_string_share, R.drawable.ic_action_gshare_dark, nightMode, enabled = entries.isNotEmpty()) {
```
Chumva: (gen by AI) The old multi-share also offered a "Locations" GPX with one waypoint per note (`ShareRecordingsTask`), and this PR deletes it together with the `av_locations*` strings. If dropping it is intentional, it's worth stating in the PR description; otherwise it could stay as an extra action here.

PR #26009, opened 2026-09-13
Copy PrecalculatedRouteDirection and RouteConditionalHelper to OsmAnd-shared

alex-dev-neo:

`PrecalculatedRouteDirection` and `RouteConditionalHelper`, with `KMapUtils.getProjectionPoint31` checked bit for bit against `MapUtils` in `KMapUtilsProjectionCompatTest`. `PrecalculatedRouteCompatTest` builds directions from chains of real road segments on both sides, adopts them for a start and a target the way a calculation does, and compares estimates, deviations and indices; and runs the conditional helper over every road at four moments and compares the road's types afterwards.

One thing the converter had to learn from that: the java reader calls `completeRouteEncodingRules()` once a region's rules are in, which gives every conditional rule the id of its value's rule; a region copied without that answers 0 to every conditional.

Supersedes #25922.


alex-dev-neo:

PR #26010, opened 2026-09-13
RoutingRequest and RouteCalculationMode for OsmAnd-shared

alex-dev-neo:

`RouteCalculationMode` and `RoutingRequest`, the request half of java's `RoutingContext`: the profile, the mode, the side of the road, the points and the progress. This is the context the shared turn preparation reads, and what iOS fills in from its C++ context. Java's `RoutingContext` is not changed - the earlier plan to make it extend this class is dropped with the JNI lockstep.

Supersedes #25949 and #25952.


alex-dev-neo:

PR #26011, opened 2026-09-13
Copy the turn lane logic to OsmAnd-shared

alex-dev-neo:

`TurnLanes` and `RoadSplitStructure`: reading `turn:lanes` off a road, which lanes a manoeuvre uses, and the picture of the intersection a keep-left or keep-right is decided from. `TurnLanesCompatTest` runs the copy against `RouteResultPreparation.calculateRawTurnLanes` and `getUniqTurnTypes` over 20 000 random lane strings - the place where java's `split` and Kotlin's disagree about trailing empty parts.

Supersedes #25961.


alex-dev-neo:

PR #26012, opened 2026-09-13
Copy the fork turn construction to OsmAnd-shared

alex-dev-neo:

The ten methods that turn a `RoadSplitStructure` into a keep-left, keep-right or the lanes of a straight-on: `calculateRoadSplitStructure`, `attachedTurnLanes`, `createKeepLeftRightTurn` and what they call. Covered by the route-level comparison in the next step.

Supersedes #25962.


alex-dev-neo:

PR #26013, opened 2026-09-13
Copy the route bundles to OsmAnd-shared

alex-dev-neo:

`RouteDataBundle`, `RouteDataResources` and `StringExternalizable`, the serialisation a route goes through on its way into a gpx. `StringBundleCompatTest` checks the bundle format against java's.

Supersedes #25908.

### DIFF
--- a/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/RouteDataResources.kt
+++ b/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/RouteDataResources.kt
@@ -1,0 +1,55 @@
+package net.osmand.shared.routing
+
+import net.osmand.shared.data.KLocation
+import kotlin.jvm.JvmOverloads
+
+/**
+ * The state shared by every segment of one route while it is written to or read from a gpx file.
+ *
+ * Segments are visited in order and each one consumes a run of [locations], so the reader has to
+ * remember where the previous segment stopped. The encoding rules are collected here too, because
+ * a gpx file carries one table of them for the whole route rather than one per segment.
+ */
+class RouteDataResources @JvmOverloads constructor(
+	private val locations: MutableList<KLocation> = ArrayList(),
+	private val routePointIndexes: MutableList<Int> = ArrayList()
+) {
+
+	private val rules: MutableMap<RouteTypeRule, Int> = LinkedHashMap()
+
+	private val pointNamesMap: MutableMap<RouteDataObject, Array<IntArray?>> = HashMap()
+
+	private var currentSegmentStartLocationIndex: Int = 0
+
+	fun getRules(): MutableMap<RouteTypeRule, Int> = rules
+
+	fun getLocations(): MutableList<KLocation> = locations
+
+	fun getRoutePointIndexes(): MutableList<Int> = routePointIndexes
+
+	fun getCurrentSegmentLocation(offset: Int): KLocation {
+		val locationIndex = currentSegmentStartLocationIndex + offset
+		if (locationIndex >= locations.size) {
+			throw IllegalStateException("Locations index: $locationIndex out of bounds")
+		}
+		return locations[locationIndex]
+	}
+
+	fun getCurrentSegmentStartLocationIndex(): Int = currentSegmentStartLocationIndex
+
+	/**
+	 * Moves the cursor past the segment just handled. Consecutive segments share their meeting
+	 * point, except where a route point falls on it, so the step is usually one short of the length.
+	 */
+	fun updateNextSegmentStartLocation(currentSegmentLength: Int) {
+		val routePointIndex = routePointIndexes.indexOf(currentSegmentStartLocationIndex + currentSegmentLength)
+		val overlappingNextRouteSegment = !(routePointIndex > 0 && routePointIndex < routePointIndexes.size - 1)
+		currentSegmentStartLocationIndex += if (overlappingNextRouteSegment) {
+			currentSegmentLength - 1
+		} else {
+			currentSegmentLength
+		}
+	}
+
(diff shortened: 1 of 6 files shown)


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/routing/RouteDataResources.kt
```diff
@@ -0,0 +1,55 @@
...
+
+	fun getCurrentSegmentStartLocationIndex(): Int = currentSegmentStartLocationIndex
+
+	/**
+	 * Moves the cursor past the segment just handled. Consecutive segments share their meeting
+	 * point, except where a route point falls on it, so the step is usually one short of the length.
+	 */
+	fun updateNextSegmentStartLocation(currentSegmentLength: Int) {
+		val routePointIndex = routePointIndexes.indexOf(currentSegmentStartLocationIndex + currentSegmentLength)
+		val overlappingNextRouteSegment = !(routePointIndex > 0 && routePointIndex < routePointIndexes.size - 1)
```
utafrali, line 46: The double negation makes the condition hard to verify by inspection. `!(routePointIndex > 0 && routePointIndex < routePointIndexes.size - 1)` is true for -1 (not found), 0 (first route point), and `size - 1` (last route point). The Kotlin range form is easier to read:
```kotlin
val isSharedEndpoint = routePointIndex !in 1 until (routePointIndexes.size - 1)
```
Also worth a comment explaining why the last route-point index is excluded from the full-step case. When processing the final segment, this branch is never read again so the current behaviour is harmless, but a future reader won't know that without context.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/util/StringBundle.kt
```diff
@@ -1,11 +1,24 @@
...
 
-	private val map: MutableMap<String, Item<*>> = LinkedHashMap()
+	private val map: MutableMap<String, Item<*>>
+
+	constructor() {
+		map = LinkedHashMap()
+	}
+
+	/** Shares the contents of [bundle], so a subclass can be built around an existing bundle. */
+	protected constructor(bundle: StringBundle) {
```
utafrali, line 16: The protected copy constructor shares the live backing `map` reference, not a snapshot of it. Any `put*` call on the original `StringBundle` after wrapping it in a `RouteDataBundle` will be visible through the wrapper, and vice versa. The doc comment says "shares" which is accurate, but the consequence (full shared mutability) is easy to miss. Strengthening the comment to say something like "both references see each other's writes" would prevent surprises for future maintainers, or a brief note in `RouteDataBundle`'s secondary constructor that the bundle is not expected to be written after wrapping.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/util/StringBundle.kt
```diff
@@ -172,11 +263,11 @@ class StringBundle {
 		}
```
utafrali, line 263: The newly activated `putMap(key, KTIntObjectMap)` overload has no test coverage in either new test file. The `forEach` lambda here uses the positional `{ k, v -> }` form rather than the destructuring `{ (k, v) -> }` that the old commented-out version used for `Map`, which implies a different functional interface. A round-trip test (put a small map, call `getBundle`, read the keys back) would confirm the lambda wires up correctly and that `k.toString()` produces parseable keys.

OsmAnd-shared/src/jvmTest/kotlin/net/osmand/shared/util/StringBundleFormatJvmTest.kt
```diff
@@ -0,0 +1,71 @@
...
+					2 -> random.nextFloat()
+					else -> (random.nextFloat() - 0.5f) * 2000
+				}
+				assertEquals(reference.format(value), formatted(value, maxDigits), "$value at $maxDigits digits")
+			}
+		}
+	}
+
+	@Test
+	fun testOutsideTheFormattedRange() {
```
utafrali, line 63: `testOutsideTheFormattedRange` checks NaN and infinity against the new implementation's own output, not against `DecimalFormat`. The class doc says the implementation should produce the same text as `DecimalFormat`, but this edge path is never cross-checked. Moving `Float.NaN`, `Float.POSITIVE_INFINITY`, and `Float.NEGATIVE_INFINITY` into the `values` array in `testMatchesDecimalFormatOnFixedValues` would close the gap. `DecimalFormat` with a US `DecimalFormatSymbols` does produce `"NaN"`, `"∞"`, and `"-∞"` for those values, so the test would pass, but right now there is no assertion proving that.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/util/StringBundle.kt
```diff
@@ -150,13 +238,16 @@ class StringBundle {
 		return (map[key] as? StringItem)?.asIntArray(defaultValue) ?: defaultValue
```
utafrali, line 238: Changing `putArray` / `getIntIntArray` from `Array<IntArray>` to `Array<IntArray?>` is a source-level breaking change for any existing Kotlin callers in the shared module that currently pass or receive non-nullable row arrays. If the module has no callers of these methods yet, a short note in the PR description would confirm that. If there are callers, they need to be updated here or in a follow-up that lands before this merges.

utafrali (commented):
Well-structured port with a smart compat-test strategy. The float-formatting rewrite is thorough and the cross-module test in StringBundleCompatTest is a good pattern for this series. The main gaps are: no test for the newly activated putMap(KTIntObjectMap) overload, the updateNextSegmentStartLocation condition is hard to verify by inspection, and the Array<IntArray?> signature change may break existing callers.

PR #26014, opened 2026-09-13
Copy RoutingConfiguration to OsmAnd-shared

alex-dev-neo:

`RoutingConfiguration`, its `Builder` and the `routing.xml` parser, reading the bundled file through `BundledRoutingXml` (jvm and android open it from the classpath, iOS has none). Three comparisons come with it:

- `RoutingConfigurationCompatTest`: both parsers read the same `routing.xml`; every profile, every attribute, every parameter and every configuration built with no parameters, every switch on and every choice at its first value.
- `GeneralRouterCompatTest`: every profile in the three parameter states asked about every real road in the test obf files - acceptance, one-way, speeds, priorities, obstacles at every point.
- `DirectionPointsCompatTest`: the same points through both builders, and the flattened form the C++ router gets.

Supersedes #25920.

One more commit since the review started: `RoutingConfiguration.DEFAULT_PENALTY_FOR_REVERSE_DIRECTION` is exported to Objective-C as `defaultPenaltyForReverseDirection`. The iOS app includes the C++ core's headers next to `OsmAndShared.h`, and it is a macro there, so with the Kotlin name in the framework header the app's own sources stop compiling. Only the selector the framework publishes changes (`@ObjCName`); the Kotlin name, and so the side-by-side comparison with java, stays.

### DIFF
--- a/OsmAnd-java/src/test/java/net/osmand/router/RoutingConfigurationDefaultTest.java
+++ b/OsmAnd-java/src/test/java/net/osmand/router/RoutingConfigurationDefaultTest.java
@@ -1,0 +1,42 @@
+package net.osmand.router;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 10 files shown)


alex-dev-neo:

PR #26015, opened 2026-09-13
Romanise names the same way on every platform

alex-dev-neo:

`KTransliterationHelper` on Junidecode's tables in common code, so iOS romanises a name the way the jvm does. Before this the iOS actual applied two Foundation transforms and agreed with the jvm on 62% of names. `KTransliterationTablesTest` pins the tables against Junidecode.

Supersedes #25947.


alex-dev-neo:

PR #26016, opened 2026-09-13
Copy the turn and time preparation to OsmAnd-shared

alex-dev-neo:

`TurnPreparation` and `RoundaboutTurn`: `prepareTurnResults`, `calculateTimeSpeed`, `getTurnInfo`, the U-turn justification, the merging and muting of turns. This is what android tells the driver from, and with it iOS can say the same.

`TurnPreparationCompatTest` runs both preparations over every route of the routing tests - the 114 turn lane cases and the 85 routing cases in all three road directions - found by the java planner and handed to both raw, and compares turn, exit, angle, lanes, time, speed and distance of every segment. What comes before the shared part (the area combine and the split with attached roads) needs the routing context's tiles and stays java on both sides. Descriptions are not compared: java's `addTurnInfoDescriptions` formats text nothing in the apps reads, and the copy does not carry it.

Supersedes #25963.


alex-dev-neo:

PR #26025, opened 2026-09-13
Measure the shared planner against java and C++, and take the lock off the tile width cache

alex-dev-neo:
Part of the routing-to-OsmAnd-shared series. With #24 the shared planner calculates the same routes as the java one; this step asks the question the port is for - is it fast enough on Kotlin/Native to replace the C++ router on iOS - and takes the first thing the answer pointed at.

Two benchmarks over the same routes on real maps (Noord-Holland, Upper Bavaria, Lower Austria; car profile, the base route first and then the detailed search along it, 256 MB for every planner): `RoutePlannerBenchmarkTest` in OsmAnd-java runs the java planner, the C++ router over JNI and the shared planner on the jvm; its namesake in `commonTest` runs the shared planner wherever Kotlin runs. Both are outside a normal run - the java one behind `@Ignore`, the shared one behind `OSMAND_ROUTING_BENCHMARK`, since a Kotlin/Native test binary cannot be told to run an ignored test.

What they found first: `KMapUtils.getTileWidth`, which every distance the routing measures goes through, kept its per-row cache in a stately `ConcurrentMutableMap<Int, Double>` - a lock and two boxed lookups per call. On Kotlin/Native that and the GC were the top of the profile. It is now a `DoubleArray` indexed by the row, no lock, no boxing; the race between two threads computing the same row writes the same value. The java original keeps a `TIntObjectHashMap` under a lock, which is why the shared planner on the jvm is now faster than the java one.

Wall time in ms, best of two after a warmup, on the same Mac (Native is the release simulator binary; the C++ column is the full pipeline over JNI, search in C++ and preparation in java):

| route | java, jvm | C++ over JNI | shared, jvm | shared, Native before | shared, Native after |
|---|---|---|---|---|---|
| Amsterdam to Schiphol 17 km | 562 | 1 181 | 495 | 1 144 | 1 073 |
| Amsterdam to Haarlem 20 km | 652 | 1 298 | 553 | 1 341 | 1 255 |
| Amsterdam to Den Helder 80 km | 801 | 1 078 | 521 | 4 783 | 1 422 |
| Munich to the airport 35 km | 634 | 976 | 457 | 2 515 | 1 125 |
| Munich to Rosenheim 65 km | 1 578 | 2 174 | 1 025 | 8 818 | 2 785 |
| Munich to Garmisch 90 km | 598 | 853 | 395 | 3 242 | 1 112 |
| Vienna to Schwechat 20 km | 1 398 | 2 093 | 1 007 | 5 061 | 2 628 |
| St. Pölten to Krems 30 km | 407 | 528 | 272 | 1 748 | 744 |
| St. Pölten to Wiener Neustadt 70 km | 706 | 788 | 416 | 3 694 | 1 160 |

The routes are identical on the java and shared side to the segment and the visited count; C++ settles a few percent fewer segments, as its port has always done. After this one change Kotlin/Native sits at 0.9 to 1.5 times the C++ router on the simulator, from 1.0 to 4.7 as copied, and the shared planner on the jvm is 1.1 to 1.7 times faster than java's. What is left in its profile is allocation and the collector: the reader's temporaries per tile, the `RouteSegment` per road point, the boxed `HashMap` lookups in the vehicle router's rule cache - the next things to tune, and the reason a device measurement is still owed before the gate is called.


alex-dev-neo:

PR #26026, opened 2026-09-13
Take the allocations out of the planner's hot paths for Kotlin/Native

alex-dev-neo:
Part of the routing-to-OsmAnd-shared series. #26025 measured the copied planner and left a profile; this takes the next items off it. Nothing here changes what the planner computes - `RoutePlannerCompatTest` still finds the same 365 routes segment for segment, with the same visited counts - only what it allocates and how it looks things up on the way.

Five changes, each one line of the Kotlin/Native profile:

- **`PrecalculatedRouteDirection.getIndex`** asked a `KQuadTree<Int>` with a fresh `KQuadRect` per call and collected boxed `Int`s. `KIntQuadTree` is the same tree - same nodes, same insertion, same traversal, so the same nearest point - over `KTIntArrayList`s, with the box passed as four numbers.
- **`RoutingContext.loadRouteSegment`** made a `KTLongObjectMap` per call for the duplicates it excludes. One scratch map on the context, cleared per call.
- **The reader** made three `KTIntArrayList`s and two lists of them per road, a boxed `MutableList<Long>` of ids and a restriction map per block. Scratch lists on the adapter, a `KTLongArrayList` for the ids, and the point type and point name lists only when a road has them.
- **The vehicle router's cache** made an `IntHolder` key per lookup; hits now go through one reusable key, only a stored key is fresh. Its rule matching did java's three bit set operations on a scratch set to ask "is every filter type among the road's types" - `KBitSet.containsAll` asks it in one pass without the scratch set - and the rules are iterated as an array rather than an `ArrayList` with a bounds check per element.

Two things that did not help, so they are not in: a fixed heap target for the Kotlin/Native collector (`GC.autotune = false`, `targetHeapBytes` at 512 MB or 2 GB) made every route five to seven times slower - the adaptive scheduler is the right one; and there is nothing left to take out of `getIndex` itself, which is the same nearest-point search java runs on every heuristic call.

Wall time in ms, best of two after a warmup, one Mac, runs one after another (the C++ column is the whole pipeline over JNI; Native is the release simulator binary; "before" is #26025):

| route | java, jvm | C++ over JNI | shared, jvm | shared, Native before | shared, Native after |
|---|---|---|---|---|---|
| Amsterdam to Schiphol 17 km | 584 | 1 265 | 532 | 1 073 | 800 |
| Amsterdam to Haarlem 20 km | 675 | 1 389 | 505 | 1 255 | 976 |
| Amsterdam to Den Helder 80 km | 811 | 1 138 | 438 | 1 422 | 951 |
| Munich to the airport 35 km | 647 | 988 | 375 | 1 125 | 768 |
| Munich to Rosenheim 65 km | 1 590 | 2 172 | 830 | 2 785 | 1 886 |
| Munich to Garmisch 90 km | 618 | 900 | 317 | 1 112 | 750 |
| Vienna to Schwechat 20 km | 1 364 | 2 230 | 868 | 2 628 | 1 927 |
| St. Pölten to Krems 30 km | 427 | 540 | 222 | 744 | 492 |
| St. Pölten to Wiener Neustadt 70 km | 703 | 851 | 358 | 1 160 | 788 |

Kotlin/Native on the simulator is now at 0.6 to 0.9 of the C++ router as android runs it (22 to 34% less time per route than after #26025), and the shared planner on the jvm is 1.1 to 2.0 times faster than java's. What is left in the Native profile is the collector at about a fifth of the main thread - the `RouteSegment` per road point and the `RouteDataObject` per road, which are the data structure - and the distance arithmetic, which is the algorithm. Measured inside the iOS app as well (an XCTest next to `OARouteProvider`, Release, Designed for iPad on the same Mac; the C++ router as the provider drives it against the shared planner from `OsmAndShared.framework`, same routes, one regional map each, 256 MB): the shared planner takes 0.8 to 1.3 of the C++ router's time - Schiphol 858 against 1 114 ms, Rosenheim 1 951 against 1 656, Vienna 1 838 against 1 890, Krems 514 against 392 - with the same segments, distances and visited counts as on the jvm and the simulator, while the C++ router there settles 5 to 20% fewer segments on the routes where it is faster.


alex-dev-neo:

Issue #17773, opened 2023-07-31
Add possibility to use Ferry in the "Public Transport" routing


zipav:
# 🚀 feature request

### Description

At the moment ferry is not represented in OsmAnd as a form of public transport, adding it to the profile would improve the travelling experience.
Our user from Norway suggested adding a ferry to the Public Transport profile, as it currently exists only in the Avoid roads function in the Pedestrian profile.

### Describe the solution you'd like
Add Ferry to the Public Transport profile

DmitryAlexei:
<table style="width:100%">
  <tr>
    <th>vehicle routing</th>
    <th>pedestrian routing</th>
    <th>public transport routing</th>
    <th>Public transport routing settings</th>
  </tr>
  <tr>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/116079541/73995a15-b550-4179-ad49-3ffb4129f836"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/116079541/7cb0433d-69b3-40c5-96fd-1e2864817739"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/116079541/6dc165d1-800e-4f4a-bab5-a00407536b55"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/116079541/2e6a4f4f-2327-423c-bba3-9ce0f9cd5940"  width="300" height="auto"/> </td>
  </tr>
</table>

vshcherb:
https://www.openstreetmap.org/way/4361073#map=15/58.30134/11.51962

Issue #14922, opened 2022-07-24
Ferry: add penalty for transition and decrease speed


Yogurt4:
# 🐞 routing report

### Routing engine

- [X] OsmAnd's in-app offline routing
- [ ] Any online routing provider (YOURS, openrouteservice, OSRM, etc.)

### Routing Profile

Car

### Start and end points

[Bijela - Stoliv](https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=42.45431%2C18.65474%3B42.46629%2C18.72917#map=14/42.4595/18.6915) (Montenegro)

### Actual and expected routes

The route **path** itself is OK (there is no other viable solution, anyway).
However, OsmAnd tells this 7.4 km will take 11 minutes. This **time** were true only
1. if there was no transition time to embark/disembark the ferry, AND
2. if the ferry travelled at 40 km/h.

OSRM (see the link above) estimates 23 minutes, GraphHopper estimates 27 minutes. I find the latter to be closer to reality.

### Is this a regression?

I haven't installed previous versions to check.

## 🌍 Your Environment

**OsmAnd Version:**

<pre><code>
OsmAnd 4.2.7
</code></pre>

**Device and Android/iOS version:**

Redmi 8
Android 10 QKQ1.191014.001
MIUI Global 12.5.2

**Maps used (online or offline):**

- [X] Offline maps offered within the OsmAnd app for download.
Montenegro @ 1 July 2022.
- [ ] Online (tile / raster) maps 

**Anything else relevant?**

So it's basically two issues:
1. Add at least 5...10 minutes for waiting/embarking and 2..5 minutes for disembarking.
2. The average speed of the ship is less than 10 km/h, or rather appr. 6 km/h including mooring manoeuvres.

For this specific ferry, [their website](http://www.ferry.co.me/EN/RedVoznje.html) suggests the entire transport is about 15 minutes. Add 9+ minutes for the land trips in the above example.

Issue #1645, opened 2015-09-02
Cannot compile project 


chathura1:
I am trying to compile the project via this instructions for Mac OSX 10.10:
https://code.google.com/p/osmand/wiki/GradleCommandLineBuildEnvironment

With
../gradlew --refesh-dependencies clean assembleFullLegacyFatDebug

I am getting following error message:

FAILURE: build failed with an exception.
*What went wrong:
A problem occurred configuring project ':OsmAnd'.
Could not resolve all dependencies for configuration ':OsmAnd:_freqQtcoredebugArmv7DebugCompile'.
   Count not find net.osmand:OsmAndCore_androidNativeDebug:0.1-SNAPSHOT.
       Searching in the following locations:
       https://repo1.maven.org/maven2/net/osmand/OsmAndCore_androidNativeDebug/0.1-SNAPSHOT/maven-metadata.xml
       https://repo1.maven.org/maven2/net/osmand/OsmAndCore_androidNativeDebug/0.1-SNAPSHOT/OsmAndCore_androidNativeDebug-0.1-SNAPSHOT.pom
      https://repo1.maven.org/maven2/net/osmand/OsmAndCore_androidNativeDebug/0.1-SNAPSHOT/OsmAndCore_androidNativeDebug-0.1-SNAPSHOT.aar

vshcherb:
qtcore is not compiled remotely, please use old legacy

PR #26037, opened 2026-09-14
Ferry routing bug 2

nnngrach:
[issue](https://github.com/osmandapp/OsmAnd/issues/17773)
[related tools pr](https://github.com/osmandapp/OsmAnd-tools/pull/1711)
[relatinve core-legacy pr](https://github.com/osmandapp/OsmAnd-core-legacy/pull/365)
[related resources pr](https://github.com/osmandapp/OsmAnd-resources/pull/1479)
[relater core pt](https://github.com/osmandapp/OsmAnd-core/pull/1146)
[relater ios pt](https://github.com/osmandapp/OsmAnd-iOS/pull/5829)

---

### **Gullmarsleden**

58.30263 11.50282
58.29961 11.53579
[https://www.openstreetmap.org/relation/7841117](https://www.openstreetmap.org/relation/7841117)

<details><summary>Spoiler</summary>

Before
Screenshot 2026-08-24 at 14 55 54   Screenshot 2026-08-24 at 14 56 05

After

Screenshot_20260917_005335     Screenshot_20260917_115345    Screenshot_20260917_005359    Screenshot_20260917_115356   

Screenshot_20260917_115610    Screenshot_20260917_115619

time = 5 + 5 + 10+ 2,5 = 22,2 min
- ⏰ 5 min Waiting (default ferryBoardingTime=300 from routing.xml) 
- ↗️ 5 min Boarding (ferryTerminalTime=300 from routing.xml)
- 🚢 10 min Sailing (osm tag  "duration=00:10")
- ↗️ 5/2 min Disembarking (ferryTerminalTime=300 in routing.xml)

</details>

---

### **Sandbanks Ferry**

50.67782 -1.95123
50.68372 -1.94832
https://www.openstreetmap.org/way/147985346

<details><summary>Spoiler</summary>

Before
Screenshot 2026-08-24 at 19 21 23      Screenshot 2026-08-24 at 19 21 59

After
50.68001 -1.95001
50.68314 -1.94872

shorter variant:
Screenshot_20260917_122921    Screenshot_20260917_123100    Screenshot_20260917_123109   Screenshot_20260917_123114

time = 10 + 5 + 4 + 2.5 = 21,5 min
- ⏰ 10 min Waiting (osm tag   "interval=00:20" / 2) 
- ↗️ 5 min Boarding (ferryTerminalTime=300 in routing.xml)
- 🚢 4 min Sailing (osm tag   "duration=00:04")
- ↗️ 5/2 min Disembarking (ferryTerminalTime=300 in routing.xml)

</details>

---

### **Nordoleden**

57.75831 11.61556
57.77241 11.61966
https://www.openstreetmap.org/way/16794766

<details><summary>Spoiler</summary>

Before
Screenshot 2026-08-24 at 20 09 36    Screenshot 2026-08-24 at 20 09 52

After

Screenshot_20260917_123350      Screenshot_20260917_123356      Screenshot_20260917_123407     Screenshot_20260917_123414

This is a complicated example:
Screenshot 2026-09-17 at 12 42 11

time = 5 + 5 + 8 + 2.5 = 20,5 min
- ⏰ 5 min Waiting (default ferryBoardingTime=300 from routing.xml) 
- ↗️ 5 min Boarding (ferryTerminalTime=300 in routing.xml)
- 🚢 8 min Sailing (osm tag   "duration=00:20" * percentage of passed way)
- 🔗 0 min Junctions in the water (zero cost)
- ↗️ 5/2 min Disembarking (ferryTerminalTime=300 in routing.xml)

</details>

---

### **Kungshamn**

58.36137 11.24880 
58.35333 11.22505
https://www.openstreetmap.org/way/189582586 
https://www.openstreetmap.org/way/189584079

<details><summary>Spoiler</summary>

After

Screenshot_20260917_125545   Screenshot_20260917_125553    Screenshot_20260917_125601    Screenshot_20260917_125607

This is another complicated route with 2 connected ways:
Screenshot 2026-09-17 at 12 49 30

time = 5 + 5 + 1.5 + 0 + 5.5 + 2.5 = 19,5 min
- ⏰ 5 min Waiting (default ferryBoardingTime=300 from routing.xml) 
- ↗️ 5 min Boarding (ferryTerminalTime=300 in routing.xml)
- 🚢 1.5 min Sailing Way 1 (calculated with ferry speed)
- 🔗 0 min Junctions in the water (zero cost)
- 🚢 5.5 min Sailing Way 2 (calculated with ferry speed)
- ↗️ 5/2 min Disembarking (ferryTerminalTime=300 in routing.xml)

</details>

---
---

### **Bijela**

This is a route from github [github issue](https://github.com/osmandapp/OsmAnd/issues/14922#issue-1315913000) .
User says:

<details><summary>Spoiler</summary>

```
OsmAnd tells this 7.4 km will take 11 minutes.
GraphHopper estimates 27 minutes. I find the latter to be closer to reality.
the entire transport is about 15 minutes. 
Add 9+ minutes
```

[OSRM link](https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=42.45431%2C18.65474%3B42.46629%2C18.72917#map=14/42.4595/18.6915)

Screenshot 2026-09-17 at 13 06 46

</details>

---

Full way before / after:

42.45431,18.65474
42.46629,18.72917

Now it is closer to asked 27 minutes.

<details><summary>Spoiler</summary>

Screenshot 2026-09-17 at 13 06 29

Screenshot_20260917_132129

</details>

---

Ferry only way before / after:

42.466581,18.674241 
42.465587,18.68559
https://www.openstreetmap.org/way/145625735

<details><summary>Spoiler</summary>

Before

Screenshot 2026-09-17 at 13 05 25

After

Screenshot_20260917_130344   Screenshot_20260917_130412   Screenshot_20260917_130420   

time = 5 + 5 + 5+ 2,5 = 17,5 min
- ⏰ 5 min Waiting (default ferryBoardingTime=300 from routing.xml) 
- ↗️ 5 min Boarding (ferryTerminalTime=300 from routing.xml)
- 🚢 5 min Sailing (osm tag  "duration=00:05")
- ↗️ 5/2 min Disembarking (ferryTerminalTime=300 in routing.xml)

</details>

---

Some routes with ferry a bus variants:

<details><summary>Spoiler</summary>

Screenshot 2026-09-17 at 18 43 10
Screenshot 2026-09-17 at 18 44 53
Screenshot 2026-09-17 at 18 48 37

</details>

---
---

### **IOS Realisation**

<details><summary>Spoiler</summary>

Screenshot 2026-09-21 at 18 22 10   Screenshot 2026-09-21 at 18 22 14

Screenshot 2026-09-21 at 18 22 57    Screenshot 2026-09-21 at 18 23 04

Screenshot 2026-09-21 at 18 23 32   Screenshot 2026-09-21 at 18 23 35

Screenshot 2026-09-21 at 18 24 11   Screenshot 2026-09-21 at 18 24 20

</details>


nnngrach:
## Ferry time model

### Before
- **Car, bicycle, pedestrian:**
  - A ferry was treated as an ordinary road. The crossing time was `distance / speed`, where the speed came from the profile's `route=ferry` rule in `routing.xml`: car 15, bicycle 10, public transport 30, pedestrian and moped had no rule (the profile's default speed was used).
  - That speed was then clamped to the profile's min/max speed. For example, a car ferry actually "sailed" at the car's `minSpeed` of 20 km/h, and a pedestrian's at walking speed.
  - The OSM tags `duration` and `interval` were ignored.
  - Route search added a fixed `amenity=ferry_terminal` obstacle (300 s) at every terminal node passed, even when a car only drove past the terminal. The estimated time didn't include it, so the search cost and the estimated time were different.
  - No waiting for the ferry, no boarding, no disembarking time.
- **Public transport:**
  - Only ferries with a route relation were supported; `route=ferry` ways without a relation were ignored.
  - Waiting was `interval / 2`, otherwise the generic boarding time (180 s). Every stop added 30 s, the speed was 30 km/h, and `duration` was ignored.
  - A bus route crossing water by a ferry got no ferry time at all.

### After
One model for all profiles (car, bicycle, pedestrian, public transport), used in both route search and the estimated time. The crossing time now comes from the OSM tags `duration` and `interval`, not from the vehicle's speed.

- ⏰ **Waiting:** half of the OSM `interval` tag, otherwise `ferryBoardingTime` (a `routing.xml` attribute, 300 s).
- ↗️ **Boarding:** `ferryTerminalTime` (a `routing.xml` attribute, 300 s), the time the ferry stands at the terminal.
- 🚢 **Sailing:**
  - if the OSM `duration` tag is set (and gives a realistic 1–100 km/h), the part of `duration` proportional to the passed part of the way;
  - otherwise the ferry speed from `routing.xml` (15 km/h in all profiles, not clamped by the profile's min/max speed) plus `ferryTerminalTime` for every terminal passed on board, including a terminal joining two ferry ways;
  - terminals inside a way with `duration` are already included in it.
- 🔗 **Junctions in the water** (two ferry ways joined at an untagged node, e.g. Kungshamn) cost nothing.
- ↘️ **Disembarking:** `ferryTerminalTime / 2` (150 s), since vehicles leave without waiting for the ferry's stop to end.
  - **Whole crossings only:** waiting, boarding and disembarking are paid only for a whole crossing (land → ferry → land).  
  
A route that starts or ends on a ferry (at its terminal, or when the route is recalculated during the crossing) only sails.
- **Boat profile excluded:** a boat sails along the ferry line by itself.
- **`routing.xml`:** `ferryBoardingTime` and `ferryTerminalTime` are global attributes that a profile may override. The per-profile `amenity=ferry_terminal` obstacles were removed.

**Public transport specifics:**
- `route=ferry` ways without a relation now become ferry routes.
- A bus route crossing a ferry gets waiting + boarding + sailing + disembarking on that hop.
- Consecutive ferry stops with the same name are treated as berths of one terminal.
- Time on ferries isn't compared with the straight-line walking time.
- The app drops routes whose real walking leg needs a ferry.

nnngrach:
Example for  [way](https://www.openstreetmap.org/way/147985346)

before:

- ⏰ 5 min (ferry_terminal=300)
- 🚗 Driving (ferry speed was calculated with current router car/bike/foot speed)
- ⏰ 5 min (ferry_terminal=300)

after:

- ⏰ 10 min Waiting (osm "interval=20" / 2) 
  - (otherwise use default 5 min: ferryBoardingTime=300 from routing.xml)
- ↗️ 5 min Boarding (ferryTerminalTime=300 in routing.xml)
- 🚢 10 min Sailing (osm duration=00:10)
  - (otherwise calculate route time with ferry)
- ↗️ 5/2 min Disembarking (ferryTerminalTime=300 in routing.xml)

OsmAnd-java/src/test/java/net/osmand/router/TmpFerryNativeCheckTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * TODO #17773 temporary: checks that the native library (core-legacy/binaries) has the ferry changes.
+ * Run it alone (the init script lets tests see the library):
+ * ./gradlew -I ../native-tests.init.gradle :OsmAnd-java:cleanTest :OsmAnd-java:test --tests 'net.osmand.router.TmpFerryNativeCheckTest' -i
+ * and look for "FERRYNATIVE" lines. It fails if native results differ from Java, and it's skipped if the library
+ * isn't found (without the init script too).
+ * Remove before the PR.
+ */
+public class TmpFerryNativeCheckTest {
```
mohithvardhan002: This test is documented as temporary and the comment says “Remove before the PR.” Should TmpFerryNativeCheckTest be removed before merging this PR?

OsmAnd-java/src/main/java/net/osmand/router/TransportRoutePlanner.java
```diff
@@ -228,6 +248,7 @@ public List<TransportRouteResult> buildRoute(TransportRoutingContext ctx, LatLon
 			updateCalculationProgress(ctx, queue);
 			
 		}
+		results.removeIf(r -> !TransportFerryHelper.usesFerry(r));
```
mohithvardhan002: This filtering is marked as temporary in TransportFerryHelper (“show only routes with ferries”). Should this remain enabled in the merged PR, or should it be removed before merging?

nnngrach:
(ai disclamer)

# PR texts for issue #17773 (ferries in public transport + one ferry time model)

Six PRs, one branch name `ferry_routing_bug_2` in every repo. The main body goes to the **android** PR;
the other five carry a short body and link back to it.

---

## android — main PR body

### Problem

https://github.com/osmandapp/OsmAnd/issues/17773

In OSM most ferries are only `route=ferry` ways, often with `duration`, `interval` and `ferry=*`, and without a
`type=route` + `route=ferry` relation. The OBF transport index is built from route relations only, so the public
transport profile did not know these ferries at all: "walk → ferry → walk" or "bus → ferry → bus" could not be
planned. Maintainer's example: way https://www.openstreetmap.org/way/4361073 (Gullmarsleden).

On top of that, the time a ferry costs was different in every profile and even inside one profile. For the same
2 km crossing the app showed about 5 min by car, 11 min by bicycle, 30 min on foot and 20 min by public
transport, and the number used to *choose* the route was not the number shown as the estimated time.

### Change

One ferry model for every profile, used identically in the route search (`BinaryRoutePlanner`), in the estimated
time (`RouteResultPreparation`) and in the public transport planner (`TransportRoutePlanner`). The crossing time
comes from the OSM tags `duration` and `interval` and from two `routing.xml` attributes, not from the speed of
the vehicle standing on the deck.

| | Before | After |
|---|---|---|
| Waiting | none | half of the OSM `interval` tag, otherwise the `ferryBoardingTime` attribute (300 s) |
| Getting on | none | `ferryTerminalTime` (300 s) — the ferry stands its whole stop |
| Sailing | `distance / speed` of the profile's `route=ferry` rule (car 15, bicycle 10, public transport 30, pedestrian and moped had no rule), clamped to the profile's min/max speed, so a car ferry "sailed" at 20 km/h and a pedestrian's at walking speed | the OSM `duration` tag proportional to the passed part of the way when it gives 1–100 km/h, otherwise 15 km/h in every profile, not clamped, plus `ferryTerminalTime` for every terminal passed on board |
| Getting off | none | `ferryTerminalTime / 2` — vehicles leave without waiting for the stop to end |
| Terminals | a fixed `amenity=ferry_terminal` obstacle of 300 s in the search at every terminal node passed, never in the estimated time, so the two differed | the obstacle is removed; the stops of the model are charged in both |

Getting on and off is paid only for a whole crossing land → ferry → land: a route that starts or ends on a ferry,
at its terminal or after a recalculation on board, only sails. Junctions of ferry ways in the water cost nothing.
The `boat` profile is excluded — a boat sails the line by itself.

**resources** — `routing/routing.xml` is the source of truth. Two global `<attribute>` entries a profile may
override, `ferryBoardingTime` and `ferryTerminalTime`, read through `RoutingConfiguration.Builder.getAttribute`;
the five per-profile `<select value="300" t="amenity" v="ferry_terminal"/>` obstacles are removed; the
`route=ferry` speed is 15 in car, bicycle, pedestrian, horsebackriding, public_transport and moped.
`obf_creation/rendering_types.xml` gets `<routing_type tag="interval" mode="text"/>` so the router can read the
`interval` tag of a ferry way, the way it already reads `duration`.

**tools** — new `TransportFerryIndexHelper`, called from `IndexTransportCreator` (and one hook in `IndexCreator`
for the step that has to run after all ways are read). A `route=ferry` way that is not a member of a ferry
relation becomes a direct and a backward `TransportRoute`, fed straight into the existing route pipeline through
`EntityParser.parserRoute(Entity, ref)` — no synthesised relation, so no route gets an "open in OSM" link to an
object that does not exist. Its stops are the `amenity=ferry_terminal` nodes along the way plus both way ends.
A way end that is not a tagged terminal is a generated stop; a generated stop whose node is shared by two or more
ferry ways and by no other way is a junction in the water. Ferry relations are built through
`processTransportRelationV1` so that both directions exist (`processTransportRelationV2` returns one direction
for a single bidirectional way). `ref` falls back to the first four letters of a one-word name, which used to
drop such routes silently ("Nordöleden").

**android** — `FerryRoutingHelper` holds the model for the generic profiles: `getBoardingTime`,
`getAlightingTime`, `parseDuration`, `getRoutingSpeed`, `getTransitionTime`, `getStopTime`, `updateSegmentTimes`,
`hasCrossing`. It is called with one line each from `BinaryRoutePlanner.calcRoutingSegmentTimeOnlyDist`,
`calculateRouteSegmentTime` and `processOneRoadIntersection` (direction-aware: in a reverse search the "next"
segment is the previous one), from `RouteResultPreparation.calculateTimeSpeed` and from
`RoutePlannerFrontEnd.makeStartEndPointsPrecise`. `GeneralRouter.limitSpeed` keeps ferries out of the min/max
speed clamp. `TransportFerryHelper` holds the public transport part: stop flags, travel speed from `duration`,
stop time, `isSameTerminal`, `getAlightingTime`, `getCrossingTime` and `mergeJunctionSegments`.
`TransportRoutePlanner.buildTransportRoute` uses them; time spent on ferries is accumulated in
`TransportRouteSegment.ferryTime` and excluded from the `maxTravelTimeCmpToWalk` comparison, otherwise a real
ferry route always loses to a straight line drawn across the water.

On the app side `TransportRoutingHelper` drops public transport routes whose real pedestrian walking leg needs a
ferry (`FerryRoutingHelper.hasCrossing`) and recalculates the walks of the next candidates until nothing is
removed; `PublicTransportCard`, `RouteDetailsFragment` and `TransportRoutingHelper` show the real walking time
(`RouteCalculationResult.getWholeTime()`) instead of `getRoutingTime()`, which is the search cost and has been a
debug value since 2012 — that is what produced "Walk 20 min, 361 m"; `TransportStopsLayer` hides generated stops
(`TransportFerryHelper.isSyntheticStop(TransportStop)` reads the stop's own `osmand_ferry_synthetic` tag,
`getVisibleStops(TransportRoute)` reads the route tag for the stops of one selected route).

**core-legacy** — the same classes and function names in the same order: `ferryRoutingHelper.h`,
`transportFerryHelper.h`, `RoutingIndex::ferry` and `RouteDataObject::containsType`, `GeneralRouter::limitSpeed`,
the search and estimated-time hooks, the public transport planner, the junction filter in
`TransportRoutingContext::getTransportStops`, the merge of junction segments in
`TransportRoutePlanner::prepareResults`, the route tags passed to Java through
`NativeTransportRoute.tagKeys/tagValues`, and a `TransportRoutingConfiguration` constructor taking
`RoutingConfigurationBuilder` so that iOS, which builds that configuration itself, gets the two global
`routing.xml` attributes (`RoutingConfigurationBuilder::getAttribute`).

**core** (OsmAnd-core, the map renderer of the iOS app) — `ObfTransportSectionReader_P::readTransportStop` reads
`additionalNamePairs` into a new `TransportStop::localizedNames`, the way `Amenity` and `Address` already hold
theirs, and `TransportStop::isSynthetic()` answers from it; `TransportStopSymbolsProvider_P::obtainData` skips
such stops, which is what `TransportStopsLayer` does on Android.

**ios** — `OATransportRoutingHelper` builds the configuration with
`TransportRoutingConfiguration(builder, router, params)`, so the two `routing.xml` attributes reach it;
`hasWalkOverWater` / `hasFerryCrossing` / `walkFrom:to:` / `removeRoutesWithWalkOverWater:` and the
`do { calculateWalkingRoutes } while (…)` loop mirror Android's `TransportRoutingHelper.RouteRecalculationTask`;
`OARouteCalculationResult.getWholeTime` replaces `routingTime` at the seven public transport call sites
(`OAPublicTransportShieldCell`, `OATransportDetailsTableViewController`, `OATransportRoutingHelper`), and the
walk after a segment is `nextSegment->walkDist` as in `RouteDetailsFragment`.

Public transport geometry: parallel ferry ways of different berths are merged by `mergeForwardWays` into one way
that goes there and back, so the part of it between two stops can run backwards. `TransportRouteResultSegment.getGeometry`
(Java and native) now cuts that part in either direction instead of falling back to a straight line between the
stops.

### Data format

No protobuf change and no new fields. What an OBF gains:

- the transport index gains the ferry routes generated from orphan `route=ferry` ways — that is the feature;
- two tags per such route in the **existing** transport route tag dictionary:
  `osmand_ferry_stops` — indexes of the generated stops, `index:j` for a junction in the water, e.g. `0,3:j,5`;
  `osmand_ferry_crossings` — `stop:interval:duration:length` for a non-ferry route that goes over a ferry way
  before that stop (seconds and meters, 0 = unknown), which is what lets a bus pay the ferry it crosses;
- `osmand_ferry_synthetic=yes` on the generated stop itself (written by `TransportFerryHelper.markSyntheticStop`,
  read by `TransportFerryHelper.isSyntheticStop(TransportStop)`), in the **existing** `additionalNamePairs` of
  `TransportStop`, the per-stop key/value channel the generator already writes and every reader already parses.
  The route tag answers "is stop number 3 of this route generated" and serves the router; this one answers "is
  this stop generated" without the routes and serves the map layers, which draw stops straight from the stops
  tree and would otherwise have to load every route of every stop of every tile. Measured on the test maps:
  33 bytes per map (two stops), the string `osmand_ferry_synthetic` being stored once per section;
- one routing tag rule for `interval`, whose value is stored only on ways that carry the tag.

Maps without these tags keep working unchanged: every reader of them returns "absent" and the model falls back to
the `routing.xml` attributes.

### Results

Minutes, first route of each scenario, measured on the test maps of this PR
(`resources/test-resources/routing/ferry_*.obf`, forward / backward):

| Route | Car | Pedestrian | Public transport |
|---|---|---|---|
| Gullmarsleden | 22.67 / 23.40 | 24.99 | 25.78 |
| Sandbanks, next to the terminals | 21.54 | 21.95 | 22.11 |
| Sandbanks, from the streets | 24.21 / 22.09 | 27.18 | 28.05 |
| Sandbanks, bus over the ferry | 24.26 / 21.99 | 27.10 | 28.86 / 28.33 |
| Nordöleden | 21.19 | 23.60 | 24.29 |
| Kungshamn (bridge is faster than the ferry) | 6.23 / 6.20 | 21.30 | 19.34 / 19.29 |
| Bijela | 17.56 | 17.94 | 16.92 |

Every number is explainable by the model. Gullmarsleden by car: 5 min waiting + 5 min getting on + 10 min
`duration` + 2.5 min getting off ≈ 22.7 min. Before this change the same crossing was 10.40 min by public
transport, of which 7.1 min of travel, and the profiles disagreed with each other. The car and bicycle numbers of
Sandbanks stay slightly above the estimated time because that ferry is `access=customers`, which gives it a car
priority of 0.15 in the search — a preference, not a time.

### Testing

- `OsmAnd-java/src/test/java/net/osmand/router/PublicTransportRouteTestingTest` — seven tests, one per mechanism
  (ferry of a relation, orphan ferry way, bus over a ferry together with a straight walk over the water,
  intermediate terminal and `ref` fallback, junction in the water, berth geometry, and the generated stops the
  map hides), each checking both directions, plus `checkFerryRules` on every result: no two ferry segments in a
  row and no change of segments at a junction in the water.
- Whole `:OsmAnd-java:test` passes, and so do the native `RouteResultPreparationNativeTest` and
  `ApproximationNativeTest` against a freshly built `libosmand`.
- Every scenario was run as a JVM test in Java mode and with the native library, on iOS through core-legacy, and
  in the apps themselves, in both directions, for car, pedestrian and public transport: Java, native and iOS give
  the same times, and each app matches its own engine.
- The generated stops the map hides were read back from both readers and are the same: the Java reader in
  `testPTFerryHiddenStops` and `OsmAndCore` in the running iOS app report `58.353142 11.232276` and
  `58.360966 11.248391` on Kungshamn, `57.768681 11.624694` and `57.748488 11.640079` on Nordöleden, and
  nothing on the three maps without generated stops.
- Checked in the app: the ferry ride is drawn along the ferry and not as a straight line (Bijela), two ferry ways
  joined in the water are one ride without a transfer (Kungshamn), generated stops stay off the map, no shown
  route has a walking leg that crosses the water.

Known limitations, left as they are: a ferry relation without a `duration` tag sails at 15 km/h although its
member ways carry `duration` (Bijela), because way tags do not reach the transport index; the app validates the
real walking legs of the first four routes only; a zero-length ferry piece at the very end of a route is charged
getting on in the search, which does not change the ranking because every path to that point pays it; HH routing
picks the new costs up only after HH data is regenerated.

### Related PRs

- android — this PR
- core-legacy — <link>
- core — <link>
- tools — <link>
- resources — <link>
- ios — <link>

One difference is left between the two map renderers: when a route is opened from a stop's menu, Android draws
its stops through `TransportFerryHelper.getVisibleStops` and hides the generated ones, while
`TransportStopSymbolsProvider` on iOS draws every stop of the route. The stops of a route carry no name tags, so
closing it means teaching `core` to read the route tag dictionary and to parse the stop index list — a third copy
of that parser. It is not in this pull request.

## AI disclaimer

The AI was used to investigate the issue, analyse diagnostic output, explore the existing OsmAnd
transport-indexing and routing code, propose implementation options, assist with refactoring, and measure the
results of every engine and app against each other.

Prompts used (summarised, in chronological order over five sessions):

**Session 1 — diagnosis, and a first approach that was dropped**

* Investigate why a public transport route containing a ferry crossing could be calculated correctly from point A to point B, but no route was found in the reverse direction from point B to point A.
* Add detailed diagnostic logging to `TransportRoutePlanner.java` to compare the forward and reverse calculations, including loaded stops, available route segments, route IDs, stop indices, visited segments, and reasons why route candidates were rejected.
* Investigate and explain the `Could not calculate route: f != java.lang.Integer` error encountered while adding diagnostic logging.
* Analyse forward and reverse routing logs and explain the execution flow in both cases, including where the successful and unsuccessful calculations diverged.
* Compare the problematic ferry with another ferry crossing that worked in both directions.
* Take into account that some test routes were calculated by the native transport-routing implementation and therefore did not pass through the modified Java `TransportRoutePlanner`.
* Add equivalent diagnostic logging to the native transport-routing implementation.
* Analyse logs containing four routing tests: ferry crossing forward, ferry crossing reverse, routing between neighbouring stops forward, routing between neighbouring stops reverse.
* Analyse the generated OBF with `BinaryInspector` and confirm that the problematic ferry was stored as only one transport route, while the forward and reverse bus routes were stored separately.
* Trace the ferry route through the OBF generation pipeline and add logging to the relevant map-generation functions.
* Investigate `processTransportRelationV1()` and `processTransportRelationV2()`, explain how they process public transport relations, and identify the principal differences between the two implementations.
* Explain where the `public_transport:version=2` tag comes from and how it affects route processing.
* Investigate OpenStreetMap public transport tagging rules and explain why bus routes are commonly represented by separate directional relations while a ferry crossing may be represented by only one `route=ferry` way or relation.
* Determine how OsmAnd should recognise a ferry object that is usable in both directions even when OSM does not provide two directional transport relations.
* Add support for ferry ways that have no corresponding public transport relation by creating a temporary synthetic relation during OBF generation.
* Load and attach the source way and its terminal nodes correctly to the synthetic relation so that it can pass through the existing relation-based transport indexing pipeline.
* Reuse existing ferry relations when available and avoid generating duplicate transport routes for the same ferry crossing.
* Support both cases: an OSM extract containing a ferry way and an existing ferry relation, and an OSM extract containing only a ferry way.
* Analyse generated OBF output and confirm that the ferry route is present in both directions without duplicate entries.
* Investigate the `IllegalStateException: too long id` error thrown by `TransportRoutePlanner.TransportRouteSegment.getId()`.
* Trace how an OSM entity ID is transformed through `EntityParser`, `TransportRoute`, OBF serialization, OBF reading, and `TransportRouteSegment.getId()`.
* Explain why directly reusing the internally encoded way ID as a relation ID produced a route ID that was too large after the additional transport-routing bit shifts.
* Search the OsmAnd and OsmAnd-tools codebases for existing examples of generated IDs for synthetic nodes, ways, relations, propagated nodes, split objects, and transport routes.
* Initially implement a synthetic relation ID generator with collision checking and a bounded ID range.
* Analyse successful routing logs after introducing the generated synthetic ID.
* Refactor the generated-ID logic to make it easier to understand and reduce unnecessary bitwise operations.
* Address the review concern that the generated relation ID did not correspond to a real OSM relation and could cause the UI to create an invalid `openstreetmap.org/relation/<id>` link.
* Ensure that synthetic ferry routes retain enough information to recover the original OSM ferry way ID.
* Make `ObfConstants.getOsmUrlForId()` return a link to the real source way `https://www.openstreetmap.org/way/<original-way-id>` instead of linking to a synthetic or unrelated OSM relation.
* Update `getOsmObjectId()` and `getOsmEntityType()` so that a synthetic ferry `TransportRoute` is exposed as the original OSM `WAY`.
* Restrict synthetic ferry ID detection to `TransportRoute` objects where appropriate, so ordinary map nodes, ways, amenities, and rendered objects are not accidentally interpreted as ferry routes.
* Preserve the existing direction encoding used by `IndexTransportCreator` (`directRouteId = baseRouteId << 1`, `backwardRouteId = (baseRouteId << 1) + 1`).
* Explain that the lowest bit of a transport route ID represents its direction and that shifting right by one removes this direction bit; replace unclear division by `2` with `>> 1` to match the existing encoding.
* Simplify the implementation after earlier versions became too verbose and introduced multiple marker bits, multiple synthetic ID representations, and several constants.
* Reduce the final design to one synthetic relation offset and a reversible arithmetic transformation (`syntheticRelationId = offset + originalOsmWayId`, `transportRouteId = syntheticRelationId * 2 + direction`, `originalOsmWayId = (transportRouteId >> 1) - offset`).
* Keep synthetic relation IDs in a range that does not overlap real transport relation IDs, has enough space for current OSM way IDs, stays safe when `TransportRouteSegment` appends departure-time and stop-index bits, and fails explicitly instead of silently overflowing.
* Remove obsolete complexity from the earlier implementation: the `2^26`–`2^27` synthetic ID range, `Math.floorMod`, tracking all used relation IDs, collision-probing loops, separate source-relation and route marker bits, multiple synthetic-ID shift constants, and synthetic tags that would require additional OBF metadata.
* Add tests covering creation of a synthetic ferry relation, forward and backward route IDs, detection of synthetic ferry route IDs, recovery of the original OSM way ID, generation of the correct `openstreetmap.org/way/<id>` URL, and that a real relation in the previously used synthetic range is not classified as synthetic.
* Preserve unrelated local and uncommitted changes in the working trees.
* Keep the patch small and focused, avoid broad changes to the existing transport-routing pipeline, and prefer short, reversible, self-explanatory ID conversion functions.
* Do not run the project build or test suite; perform only static checks such as checking the diff, searching usages, and running `git diff --check`.

The synthetic-relation approach of this session was rejected in review (OsmAnd-tools PR #1645) and is **not**
part of this pull request: the generator now feeds the ferry way straight into `IndexTransportCreator`, so no
route gets an ID or an "open in OSM" link of a relation that does not exist.

**Session 2 — the implementation this pull request contains**

1. Add public transport ferry tests (Gullmarsleden, Sandbanks, Nordöleden, Kungshamn, later Bijela), forward and backward, on small test maps; comment out the backward test that returns a wrong result and describe what has to be fixed.
2. Make public transport routing work over `route=ferry` ways that have no route relation; analyse Kungshamn–Smögen backward again, earlier decisions may have been wrong.
3. Separate the synthetic stops we generate in the water from real ferry terminals, and hide the synthetic ones from the map (`TransportStopsLayer`).
4. Two ferry ways joined by a synthetic terminal must be shown and counted as one ride, not as a transfer; fix the route details screen (`RouteDetailsFragment`) that showed a wrong walking time.
5. Explain how the time calculation works and in which cases the bug appears; explain the difference between route search cost and estimated time, and whether such a difference is itself a bug.
6. Instead of a new OBF flag, store the ferry data in the existing route tag dictionary if that gives a smaller change.
7. Why do car, bicycle, pedestrian and public transport give so different times on the same crossing? Which ferry speeds exist in `routing.xml`? Set one ferry speed, and don't clamp ferries by the profile min/max speed.
8. Take the `duration` OSM tag into account, proportionally for a way with several stops, and parse it together with the sanity check of the value.
9. `routing.xml` is the source of truth: one global `ferryBoardingTime` attribute a profile can override, no constants in the code; reuse the existing `ferry_terminal` value as `ferryTerminalTime` instead of keeping per-profile copies.
10. Model of a ferry trip: waiting, getting on, sailing, getting off — implement this one model for route search, estimated time and public transport, and make sure a whole crossing is what pays getting on and off.
11. Walking in a straight line over the water in a public transport route is a bug — fix it.
12. Debug the public transport route on Bijela: the map draws a straight start–finish line instead of the ferry geometry.
13. Port everything to native `core-legacy` and check that Java and native give the same results; add a temporary log/test to prove the app really runs the new native code.
14. Clean the code for the commit: same names, structure and algorithms in Java and native, remove temporary filters, logs and tests, keep only the public transport tests, propose a commit name.
15. Explain and reproduce the remaining problems: the straight walk over water, a route ending on a 0 m ferry piece, and why Bijela ignores the `duration` tag of its way.
16. If something is already known to be missing for the future iOS use of this cross-platform code, do it now.
17. Write the pull request description.

**Session 3 — refactoring, cleanup and this text**

1. This code is much longer than necessary, changes too many existing places, adds too much to the OBF format and does not read like good architecture — refactor it or write better code, without losing functionality, readability or performance; check yourself with the existing tests and skills.
2. Run the full check with the test maps regenerated and the HTML report built.
3. Analyse whether anything is left before submitting; if it is ready, clean the code, remove temporary solutions and TODOs, and delete the tests that only helped during development, keeping the ones needed for the public transport change.
4. Yes — finish the remaining core-legacy items and write the PR text.
5. Complete the list of prompts in this disclaimer with the ones of the earlier sessions.

**Session 4 — the same task restated in a fresh chat, and the routing test utility**

1. Make public transport routing use `route=ferry` ways that are not members of a ferry relation (issue #17773), and give all profiles one ferry crossing time model.
2. The model: waiting (half the `interval`, otherwise a global default), getting on, sailing (the `duration` tag when it is realistic, otherwise the profile ferry speed), getting off for half the terminal time; the route search cost and the estimated time must agree.
3. Keep `routing.xml` the source of truth: one global attribute per constant, overridable per profile, no numbers or defaults in code.
4. Keep the diff minimal, collect the new logic in one helper per area, uniform names, concise code, no regressions.
5. Mirror every routing change in core-legacy with the same names, structure and algorithms, prove Java and native agree, and prepare the native code for iOS; list what the iOS app itself must still do.
6. Explain and reproduce the remaining problems: public transport walking legs drawn straight over water, a zero length ferry piece at the end of a route paying for boarding, and Bijela sailing at the default speed because the relation has no `duration`.
7. Remove the temporary code, clean up before the commits, keep only the public transport tests in the repository, and propose commit names.
8. Separately (local tooling, not part of these pull requests): build a routing test utility instead of checking routes by hand in the app, then verify that it really reads the native results, turn the manual app checks into assertions, and add the same for iOS.

**Session 5 — iOS, the generated stop icon on the map, and the last refactoring**

1. I tested iOS: the times are far from the Android reference, and inside iOS car, pedestrian and public transport disagree with each other. Are Java, native and iOS really synchronised — in architecture, names and results, so that the code looks like one function written in several languages? Write that requirement into the project rules. And the stop icon in the middle of the water on Nordöleden and Kungshamn has to be invisible, as it is on Android. Redo the code and check it with the new tests I added.
2. The branch is created and the changes are committed. Describe the possible solutions of the stop icon in the water in more detail, compare them, and look for how the project already solves such problems, in the current code or in the git history.
3. About option B: is the idea that the flags that used to be helper values on the stop now move into the other dictionary the stop has, the one iOS can read?
4. Implement option B; make sure the code works and reads the same in Java, native and iOS; run the tests when you are done.
5. I tested this code. Now refactor it, with the same goals as before: one algorithm across the repositories, brevity, clarity, readability.
6. Did you run the tests? Is the HTML report updated?
7. Update this disclaimer with the prompts of the earlier chats.

Decided by the agent, not requested explicitly: the split of the new logic into `FerryRoutingHelper` (all
profiles, route search and estimated time), `TransportFerryHelper` (public transport) and
`TransportFerryIndexHelper` (map generator); excluding the `boat` profile from the model and keeping ferries
out of the profile minimum and maximum speed clamp; the format of the new tags — one `osmand_ferry_stops` carrying
the indexes of the generated stops with `index:j` for a junction in the water, `osmand_ferry_crossings` as
`stop:interval:duration:length`, and `osmand_ferry_synthetic=yes` on the stop itself, written and read by
`TransportFerryHelper` alone so that the tag string stays private; treating consecutive ferry stops with the
same name as berths of one terminal (`isSameTerminal`), which Bijela needs, and leaving that rule in the
router instead of collapsing such stops during map generation; the 1–100 km/h sanity range that makes a
`duration` tag believable; dropping public transport routes whose real walking leg crosses water in
`TransportRoutingHelper`; drawing the ferry ride by teaching the existing
`TransportRouteResultSegment.getGeometry` to cut the part of a merged way that runs backwards, instead of a
separate ferry-only geometry function; keeping the junction-transfer design rather than chaining ferry ways
into one route in the generator, because ferry ways branch at hubs (Nordöleden); header-only native helpers
(`ferryRoutingHelper.h`, `transportFerryHelper.h`) so no build lists change, with the function that needs
`TransportRouteResultSegment` defined in `transportRouteResultSegment.cpp`; adding
`RoutingConfigurationBuilder::getAttribute` in core-legacy to mirror the Java builder; replacing the plain
`ferry_*.osm` test sources with `.osm.gz`, following `Routing_test_68_69.osm.gz`; local test and scenario
tooling that is not part of this pull request; marking the generated stop in the existing
`additionalNamePairs` of the stop instead of adding a protobuf field or joining stop → routes while rendering;
the `osmand_` prefix for the new tags, which is what the project uses for its own non-OSM tags
(`osmand_access`, `osmand_vehicle`); removing `isSyntheticStop(TransportStop)` and `getVisibleStops` from
core-legacy again, because the iOS map is drawn by `core` and nothing called them; naming the new field of
`core::TransportStop` `localizedNames` after `Amenity`; leaving the stops of a route opened from a stop's menu
unfiltered on iOS instead of teaching `core` the route tag dictionary and a third copy of the stop index
parser.

---

## core PR body

Map renderer part of https://github.com/osmandapp/OsmAnd/issues/17773: the public transport stops the map
creator generates at ferry way ends are not OSM objects and must not be drawn.
`ObfTransportSectionReader_P::readTransportStop` reads `TransportStop.additionalNamePairs` into a new
`TransportStop::localizedNames`, the way `Amenity` and `Address` already keep theirs;
`TransportStop::isSynthetic()` answers from it and `TransportStopSymbolsProvider_P::obtainData` skips such
stops — what `TransportStopsLayer` does on Android.

Main PR with the description of the model: <link to the android PR>.

## ios PR body

iOS part of https://github.com/osmandapp/OsmAnd/issues/17773. `OATransportRoutingHelper` builds the public
transport configuration with `TransportRoutingConfiguration(builder, router, params)`, so the
`ferryBoardingTime` and `ferryTerminalTime` attributes of `routing.xml` reach it: without them a crossing cost
only its sailing time and every public transport route was 6–13 min short. `hasWalkOverWater`,
`hasFerryCrossing`, `walkFrom:to:`, `removeRoutesWithWalkOverWater:` and the
`do { calculateWalkingRoutes } while (…)` loop mirror Android's `TransportRoutingHelper.RouteRecalculationTask`,
so a route whose real walking leg needs a ferry is dropped. `OARouteCalculationResult.getWholeTime` replaces
`routingTime` at the seven public transport call sites (`OAPublicTransportShieldCell`,
`OATransportDetailsTableViewController`, `OATransportRoutingHelper`), and the walk after a segment is
`nextSegment->walkDist` as in `RouteDetailsFragment`.

Main PR with the description of the model: <link to the android PR>.

## core-legacy PR body

Native mirror of the ferry time model of https://github.com/osmandapp/OsmAnd/issues/17773, used by the Android
app through JNI and compiled into the iOS app: `ferryRoutingHelper.h` and `transportFerryHelper.h` with the same
class and function names as `FerryRoutingHelper` / `TransportFerryHelper` in `OsmAnd-java`, `RoutingIndex::ferry`
and `RouteDataObject::containsType`, `GeneralRouter::limitSpeed`, the hooks in `binaryRoutePlanner.cpp`,
`routeResultPreparation.cpp` and `routePlannerFrontEnd.cpp`, the public transport planner and its junction
handling in `transportRoutePlanner.cpp` and `transportRoutingContext.cpp`, the route tags passed to Java in
`java_wrap.cpp`, and a `TransportRoutingConfiguration` constructor taking `RoutingConfigurationBuilder` so iOS
gets the two global `routing.xml` attributes. `TransportRouteResultSegment::getGeometry` now also cuts the part
of a way that runs backwards between two stops, which is what merged parallel berth ways need.

Main PR with the description of the model: <link to the android PR>.

## tools PR body

Map generation part of https://github.com/osmandapp/OsmAnd/issues/17773: `TransportFerryIndexHelper`, called from
`IndexTransportCreator`, turns a `route=ferry` way that is not a member of a ferry relation into a direct and a
backward `TransportRoute` fed straight into the existing pipeline, marks the generated stops and the junctions in
the water, and records the ferry a non-ferry route crosses. Ferry relations go through
`processTransportRelationV1` so both directions are built, and `ref` falls back to the first four letters of a
one-word name. Flags are two tags in the existing transport route tag dictionary, there is no protobuf change.

Main PR with the description of the model: <link to the android PR>.

## resources PR body

`routing/routing.xml`: two global attributes of the ferry model of
https://github.com/osmandapp/OsmAnd/issues/17773, `ferryBoardingTime` and `ferryTerminalTime` (300 s each, a
profile may override them); the per-profile `amenity=ferry_terminal` obstacles are removed, and the `route=ferry`
speed is 15 km/h in every profile. `obf_creation/rendering_types.xml` gets the `interval` routing tag so the
router can read a ferry's interval the way it already reads `duration`. Five small test maps for the routing
tests, sources gzipped like `Routing_test_68_69.osm.gz`.

Main PR with the description of the model: <link to the android PR>.

---

## AI disclaimer (for the five PRs other than android)

The android PR carries the full disclaimer with the prompts of all five sessions; repeating sixty lines of them
in every repository is noise, so the others carry this and link back:

```markdown
## AI disclaimer

The AI was used to investigate the issue, analyse diagnostic output, explore the existing OsmAnd
transport-indexing and routing code, propose implementation options, assist with refactoring, and measure the
results of every engine and app against each other. The prompts, and the decisions the agent made without being
asked, are listed in the main pull request: <link to the android PR>.
```

ivanPyrohivskyi:
(gen by AI)

## Review: overall impact

The code is structured well after #26168. The main risks now are about **what reaches users, and when**. Map data and HH data are delivered independently of the app version, so they affect every installed app. Test coverage also has gaps.

### 1. New maps and existing apps (OsmAnd-tools#1711)

Maps are downloaded independently of the app version. As soon as maps are generated with OsmAnd-tools#1711, **every installed Android and iOS version** will read the new transport data with the **current native PT router** (`runNativePTRouting` is the default on Android; the Java planner runs only in PT safe mode). That router has none of the new ferry rules:

- **`route=ferry` ways become regular PT routes.** Old clients use them with the generic `ferry` route type timing from their bundled `routing.xml`. They don't get the boarding, terminal or `duration` model, and ferry crossings of other routes are not paid.
- **Junction stops in the water become regular stops.** The old router can "walk" to them from the shore, start or finish a route there, and change routes there. The flag that forbids this (`osmand_ferry_stops` with `j`) is a route tag that old clients don't read.
- **Synthetic stops are visible on the map in old versions.** Old clients don't filter these generated stops: stops at ferry way ends that aren't `amenity=ferry_terminal`. The flag is stored as a name in the pseudo-language `osmand_ferry_synthetic` with the value `yes`. Please check where old clients may show or search all names of a stop (context menu, search). `yes` may appear there as a name.

Suggestion: don't generate the new transport data for production maps until an app version that supports it is widely installed. Otherwise the data should be made safe for old clients first. At least junction stops must not be usable as boarding, alighting or walking points, and synthetic stops need a proper name or should not be visible.

### 2. HH regeneration changes route choice on all platforms

`HHRoutingShortcutCreator` (OsmAnd-tools) builds HH shortcuts with `BinaryRoutePlanner` from OsmAnd-java. When HH files are generated with this code and the new `routing.xml`, all the new ferry costs go into the HH data:
- `calculateRoadChangeTime`: getting on and off a ferry;
- `calculateStopTime`: a ferry standing at a terminal on the way;
- the ferry speed from `duration`, and ferry speeds no longer limited by the profile min/max speed;
- the removed `amenity=ferry_terminal` obstacle.

HH is the default route calculation method whenever it is available (`RouteProvider`). So **the choice of long routes will change on every device and platform, including iOS and old Android versions.** Their ETA still comes from the old model: the ETA is computed by the app, not taken from HH data. For one route, the choice will then follow the new model while the displayed time follows the old one.

Suggestion: treat the HH regeneration as a separate, deliberate release step. Before switching, compare routes for regions with many ferries (e.g. Norway, western Sweden, Croatia, Greece), for example with RandomRouteTester, old versus new HH data.

### 3. Tests

**3 of 7 tests fail with the committed maps.** With the `.obf` files in OsmAnd-resources#1479 these tests fail:
- `testPTFerryHiddenStops`: expected 2 synthetic stops, none found;
- `testPTFerryRoutingKungshamn`;
- `testPTFerryRoutingSandbanksBusOverFerry`: 9.42 min instead of 30.92 min, the bus doesn't pay for the ferry.

The committed maps have no `osmand_ferry_stops` / `osmand_ferry_crossings` / `osmand_ferry_synthetic` data. They were generated before the last generator changes. I regenerated all 5 maps from the committed `.osm` files with OsmAnd-tools#1711, and then all 7 tests pass. Please regenerate and commit them.

**The tests depend on the locale.** The expected strings come from `TransportRouteResult.toString()`, which uses `String.format` without a `Locale`. On a machine with a non-English locale (e.g. `uk_UA`) distances are printed as `76,5 m`. Then **6 of 7 tests fail**; only the geometry test passes. Use `Locale.US` there, or compare numbers instead of formatted text.

**There are no tests for car, bicycle, pedestrian, moped and horse routing.** `PublicTransportRouteTestingTest` covers only public transport. Nothing checks route search or ETA over ferries for the other profiles, although the PR changes both:
- boarding, alighting and terminal stops;
- the `duration` speed and speeds not limited by the profile min/max speed;
- `updateSegmentTimes`.

Please add tests that check the ferry segment times per profile on the same 5 maps, in both directions. They are needed now, and they will be the reference for the OsmAnd-core-legacy and iOS ports; otherwise there is nothing to compare the native implementation with.

PR #26041, opened 2026-09-14
Fix unclickable spatial search markers on map (street/city/other objects)

### LINKED ISSUE #25488: [Search] "Show on map" should mark all the current results (streets + POIs), matching the active tab
### Summary

The blue **Show on map** control in search only ever hands off to a POI-name filter over the amenity index. For anything that isn't a POI — most importantly **streets** — that filter finds nothing relevant, so the results you are looking at are not the results you get on the map. It should instead mark **the results you already have, matching the tab you searched in**: Explore → everything, Categories → POIs, Address → streets/addresses; streets with their real geometry, POIs (and other located results) with their normal icons.

This is the still-open *display* half of #25373 (the crash half is fixed), generalised to the address/mixed cases.

### Current behaviour

- **Plain street name** (e.g. "Mittelstraße"): `PoiFiltersHelper.getShowOnMapFilter()` builds a `SearchByNameFilter` that queries only the amenity/POI index. Streets live in the address / routing index, so the streets are never shown — instead you get unrelated POIs whose *name* contains the string (a shop at "…straße 5" etc.). Reported repeatedly, e.g. #25373.
- **Address tab, several matches** (e.g. "Zossener" in Berlin → a bridge + streets in different districts): only a **single marker** is dropped near the search centre, not the matches.
- **Mixed result set**: streets and POIs are never marked together.
- **Entry point is itself inconsistent**: v5.3.x had a *"Show on map" bar*, v5.4 a *button*; the old bar no longer appears for a city search under the Address tab, whereas the button does — so behaviour now differs by how you got there.

### Proposed behaviour

**Show on map marks the located results that correspond to the active tab** — i.e. the same kind of results the tab is for:

- **Explore / free search** → everything found: streets **and** POIs and other located results.
- **Categories** → the POIs (stays POIs-only, #23089 / #23713).
- **Address** → streets, addresses and the like.

In every case:

1. **Streets** → highlighted with their real geometry (a thin line just off each side so the street, its name and its type stay readable between the lines; the two lines merge into one when zoomed out). Geometry from the routing index, stitched across the junction-split ways so it is one continuous street. **A highlighted street is tappable and opens its details / context menu, exactly like a POI** — selecting it from the list and tapping it on the map should behave the same.
2. **POIs and other located results** → their normal, tappable icons (tap → the usual context menu / details).
3. **Trigger** whenever typed text is driving a multi-result list. A single *selected* result still just opens its details (unchanged).
4. **Persistence**: the marks stay until the search bar is closed via **X** — the same behaviour as the Categories / POI overlay. No tap-anywhere-to-dismiss.
5. **Display radius** around the search centre (map centre or My Location): it covers the current view with some margin, floored so nearby matches aren't cut at a city-sized view, and capped so it never spreads country-wide. (The exact "few vs many results → wider vs tighter" heuristic is a detail worth designing together.)

### Why it needs to be careful (won't freeze)

A naïve "mark everything" freezes the app — an early prototype ANR'd rebuilding native map symbols per item, and the current POI-filter path re-searches the amenity index on every pan/zoom. The approach here marks **exactly the already-computed results** (no per-viewport re-search) in **one bounded, capped snapshot** (a fixed number of streets / points), built off the UI thread and published once. That is what keeps it responsive.

### Prototype

I have a working, device-tested prototype of the street half and the POI markers, freeze-safe (routing-index street geometry + a transient highlight layer with a dark casing so it reads against the map's own orange roads + POI markers).

*Searching "Mittelstraße" marks the street with two thin orange rails, the name still readable between them:*

<table style="width:100%">
  <tr>
<th> </td>
  </tr>
</table>

Happy to turn this into a focused PR for the rewrite-independent part (how to draw a matched street on the map is needed by any search UI), or to fold it into the ongoing search redesign — whichever fits.

### Relation to ongoing work

The search UI is being rebuilt (#25331) with a design in #24708; a broader "appetite for a search overhaul" thread is #10060. This proposal is deliberately scoped to the **map-display** side and is meant to inform that work rather than pre-empt it. It is also the **map counterpart** of the list-refinement discussion in #17039: for @sonora's "locate a specific place whose exact tagging you don't know" case, simply seeing *all* the name-matching results on the map lets you recognise the right one by location, without guessing the category.

### References

#25373 · #17039 · #10060 · #24708 · #25331 · #23089 · #23713

*(Is there a preferred direction — a PR for the street-display core, or aligning it with the #24708 design — before I propose code?)*

Corwin-Kh:
## Summary
- Development plugin's spatial search draws a marker for street/city/other search results via `POIMapLayer.customObjectsDelegate`, but tapping it did nothing: the native-renderer draw path never called `data.queryNewData(tileBox)`, so the layer's tap hit-testing (`collectAmenitiesFromPoint`) always saw an empty/stale results cache.
- Non-`Amenity` results (street, city, house, etc.) are wrapped into a synthetic placeholder `Amenity` for rendering, so even after fixing hit-testing, tapping would open a generic "Other POI" menu instead of the real object's context menu.
- Added `POIMapLayer.CustomObjectClickListener` (consulted first in `showMenuAction`) and wired `QuickSearchDialogFragment` to resolve the synthetic marker back to its original `SearchResult` and open the context menu the same way `ExplorePlacesFragment` does when an item is selected from its list.

Refs #25488

## Test plan
- [x] `:OsmAnd:compileNightlyFreeOpengldebugArm64DebugJavaWithJavac` builds successfully
- [ ] Manual: enable "Spatial text search" in the Development plugin, search for a street/city, tap "Show on map", verify the marker is clickable and opens the correct context menu (matching what ExplorePlacesFragment's list would show)
- [ ] Manual: verify real POI (Amenity) results from spatial search still open their normal context menu unaffected


Chumva:
Reviewed the diff and the tap pipeline around it. The diagnosis holds and the fix mirrors `ExplorePlacesFragment.showPointInContextMenu`, so list tap and marker tap now behave the same. Checked both tap entry points (single hit and "What's here?"), the `IdentityHashMap` lookup, listener cleanup on rotation, and the Android Auto delegate path. No blockers; the notes below are non-blocking.

1. *Hit-test the delegate list directly.* `data.queryNewData(tileBox)` starts an `AsyncTask` (plus a `refreshMap()` in `layerOnPostExecute`) on every queried-box miss. If the search is closed while one is in flight, `Task.onPostExecute` puts the marker list back into `data` after `setCustomMapObjects` cleared it, and with no POI filter nothing clears it again, so taps at the old marker positions still open menus. Reading `customObjectsDelegate.getMapObjects()` in `collectAmenitiesFromPoint` when the delegate is set avoids both the task and the window.

2. *Multi-select rows reverse-geocode.* `AmenityMenuController.needStreetName()` is true for subtype `""`, so rows read "Street, Mittelstraße" / "City, <nearest street>". Returning `false` when `SYNTHETIC_TYPE_LABEL` is set fixes it.

3. *`showSpatialSearchAmenityContextMenu` duplicates `ExplorePlacesFragment.showPointInContextMenu`.* One shared helper would keep them in sync.

4. *Follow-up: carry the `SearchResult` with the marker* (an Android-side `Amenity` subclass, or a payload on `CustomMapObjects`). Then `POIMapLayer` can resolve the tap itself without the fragment-bound listener, and `SYNTHETIC_TYPE_LABEL` does not need to live in the shared `Amenity`.

PR #26042, opened 2026-09-14
Fix address search skipping cities of maps sharing a region name

### LINKED ISSUE #25993: Address search skips cities of maps that share the same region name
**Problem**

Legacy address search (`SearchCoreFactory.SearchAddressByNameAPI.initAndSearchCities`) loads CITY/TOWN objects into `townCitiesCache` once per `BinaryMapIndexReader.getRegionName()`. When two installed maps have the same address region name, only the file searched first in the session contributes cities; the cities of the other file are never loaded until the app restarts. Streets and POIs of that file are still found, so the city silently disappears from the results.

The check was added in 4643c1a69b (#24536, 2026-03-23) and is present in r5.4 (5.4.4) and master.

**Who is affected**

Users who still have the combined German state maps from 2025 (Hessen, Niedersachsen, and most likely Bayern, Baden-Württemberg, Nordrhein-Westfalen — replaced by regional maps and marked deleted in July 2026). Their address region name is `Germany`, so with several of them installed only one state's cities are searchable. Reported via support: "München" and "Marburg" are not found although the Bayern and Hessen maps are installed (OsmAnd+ 5.4.4, Android 16).

**Steps to reproduce (no device needed)**

1. Take `Germany_niedersachsen_europe_2.obf` and `Germany_hessen_europe_2.obf` from `https://data.osmand.net/backup-maps/2025-07/maps/` — both report `getRegionName() == "Germany"`.
2. Run `SearchUICore` the way `SearchUICoreTest` does, with both files as offline indexes:
   - search "Hannover" at 52.3759, 9.7320, then "Marburg" at 50.8090, 8.7705 with the same `SearchUICore` → no CITY "Marburg", only the street and POIs;
   - reversed order → no CITY "Hannover";
   - "Marburg" in a fresh `SearchUICore` → CITY "Marburg" is the first result.

**Expected**

Cities of every installed map are searchable, regardless of the region name stored in the file.

**Suggested fix**

Key the cache per file (for example file name + date created) instead of per region name.

**Related**

- #25285 — wrong closest city shown for villages. The same user also saw a hamlet "Marburg" attributed to Caen / Aachen with a distance of about 6,400 km; not reproduced yet.
- The spatial search (`USE_SPATIAL_TEXT_SEARCH`, off by default) finds no objects at all in maps built before 2026-06-01 (name index format change between the 2026-05-01 and 2026-06-01 builds). Separate concern if spatial search becomes the default.

(gen by AI)

Corwin-Kh:
Fixes #25993

### Problem

`SearchCoreFactory.SearchAddressByNameAPI.initAndSearchCities` loaded CITY/TOWN objects into `townCitiesCache` once per `BinaryMapIndexReader.getRegionName()`. When several installed files report the same region name, only the file searched first in the session contributed cities — the cities of the other files were never loaded until the app restarted. Streets and POIs of those files were still found, so the city silently disappeared from the results.

Affects users with the combined German state maps of 2025 (Hessen, Niedersachsen, Bayern, ...): their region name is `Germany`, so with several of them installed only one state's cities are searchable.

### Fix

Key `townCitiesCache` per file (absolute path + `getDateCreated()`) instead of per region name. Several readers over the same file keep the same key, so cities are still loaded only once per file, and an updated file gets a new key.

### Reproduced and verified

`Germany_niedersachsen_europe_2.obf` + `Germany_hessen_europe_2.obf` (https://data.osmand.net/backup-maps/2025-07/maps/), both report `getRegionName() == "Germany"`, driven through `SearchUICore` the way `SearchUICoreTest` does:

| | before | after |
|---|---|---|
| "Hannover", then "Marburg" in the same `SearchUICore` | no CITY Marburg, only street and POIs | CITY Marburg is the first result |
| "Marburg", then "Hannover" | no CITY Hanover | CITY Hanover is the first result |
| "Marburg" in a fresh `SearchUICore` | CITY Marburg found | CITY Marburg found |

Added `TownCitiesCacheTest.testCachePerFileNotPerRegionName` — two copies of one obf under different names report the same region name, the cache has to keep them apart. It fails if the key goes back to the region name, or if it drops the file path. The real maps are too large to ship as a test resource, so the end-to-end case above was checked manually.

### No duplicate cities

The fix changes how many files contribute cities: one file per region name before, every address file in the search radius now. So the fair question is whether several files of one region lying next to each other — a map and its live updates, named by day — can publish the same city twice.

Live update diffs never reach this code: they carry no address index, and `SearchPhrase.getOfflineIndexes(rect, ADDRESS)` requires `containsAddressData()`. Checked 21 daily and monthly diffs across 5 regions (Andorra, Monaco, Malta, Liechtenstein, Germany bremen-city) — `containsAddressData() == false` in every one. With a map and 5 of its diffs given to the search, 6 files go in and exactly 1 reaches `initAndSearchCities`.

Real diffs therefore cannot be used to test it, so `testNoDuplicateCitiesWithSeveralFilesOfSameRegion` takes the worst case instead: 4 copies of one address-carrying obf named the live-updates way (`same_region_2.obf`, `same_region_26_09_12.obf`, ...). All 4 are loaded and the city is published 8 times (4 files × the short and long distance `SearchAddressByNameAPI`), the result dedup collapses it to one row. The test asserts one row, searching twice so the second search runs against the already filled cache; with dedup turned off it fails with `expected:<1> but was:<8>`.

Two address files of one region do overlap in practice — a full map plus the roads-only map of the same region, roads maps carry cities (`Andorra_europe_2.road.obf`: 8 CITY/TOWN, region name `Andorra`). Checked manually with both installed: raw results double, the final list still has one row per city. Note that before this change that case was broken the same way as #25993 — whichever of the two files came first won, and the cities of the other one were dropped.

`./gradlew :OsmAnd-java:test --tests 'net.osmand.search.*'` — 74 tests, 0 failures.


OsmAnd-java/src/main/java/net/osmand/search/core/SearchCoreFactory.java
```diff
@@ -1877,12 +1878,20 @@ public static class TownCitiesCache {
...
+		public void add(BinaryMapIndexReader reader) {
+			townCitiesInit.add(getKey(reader));
+		}
+		
+		// files could share the same region name (ex. old combined German state maps are all "Germany"),
+		// so cities must be cached per file, otherwise only the first file of the region is loaded
+		private String getKey(BinaryMapIndexReader reader) {
+			File file = reader.getFile();
+			String name = file != null ? file.getAbsolutePath() : reader.getRegionName();
+			return name + "_" + reader.getDateCreated();
```
Chumva, line 1894: Could this lead to duplicates with live or map+road files?

Issue #3331, opened 2016-12-07
Missing "Use location" menu item


Norbert80:
Hi,

in recent builds (OsmAnd+ 2 5.4) I am missing the "Use location" menu item. Has it been removed intentionally? If yes I vote for re-adding this item.
An alternative solution would be an implemenation similar to MAPS.ME. There a use location function is triggered by just clicking on the position arrow. I consider this an intuitive solution.

Best Regards,
Norbert

njohnston:
sonora commented on this on a [separate issue](https://github.com/osmandapp/Osmand/issues/3272#issuecomment-261790515).

To summarise, you can tap the current position indicator. On the context menu you can then add a favourite or marker from the current location, or press the three dots button for more actions.

Norbert80:
@sonora do you talk about very fresh development builds?

Norbert80:
@sonora: Thanks for the information. I see this already works as requested in OsmAnd. Sorry for the confusion. Please close this issue.

sonora:
Yes I know, it s a little paradigm change, but now with the fully interactive maps we have I guess that mid term it is even more intuitive like we have it now, the "Use location" menu was really just an interim solution.

PR #26052, opened 2026-09-15
Open read-only GPX database connections read-only (OsmAnd-Issues #3331)

### LINKED ISSUE #22641: Track appearance change not synced to cloud
### Description

Local changes of track appearence (e.g. changed color) are not detected by cloud sync.

Changes of other track attributes (e.g. description) are detected by sync.

### Steps to reproduce

- Change the color of a track
- Apply the change
- Open Menu → Settings → OsmAnd Cloud

### Actual result

- No local change is detected
- "Sync now" doesn't sync anything

### Expected result

- Track with previously changed appearance is shown under "Local changes"
- Track can by synced to cloud

```
OsmAnd Version: 5.0.5
Android version: 14
Device model: 
Crash-Logs: ?
```

Chumva:
> Merged at 9830a42043 with the file-check fallback (`canOpenReadOnly`) and three tests, as described below. The review suggestions (probe the read-only connection and fall back on failure, a WAL-guarantee test, the import scenario as a test) are in the follow-up PR #26055.

Fixes osmandapp/OsmAnd-Issues#3331 (milestone 5.4-android). Targets `r5.4`: the crash is reported on a `master` nightly (5.5.0#8743m) and reproduces on `r5.4` as well, and the changed file is identical on both branches, so `master` gets it with the regular merge-up.

## Problem

Importing an `.osf` with ~1 000 tracks kills the app with

```
android.database.sqlite.SQLiteDatabaseLockedException: database is locked (code 5 SQLITE_BUSY): , while compiling: PRAGMA journal_mode
    at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
    at android.database.sqlite.SQLiteConnection.executeForString(SQLiteConnection.java:790)
    at android.database.sqlite.SQLiteConnection.setJournalMode(SQLiteConnection.java:407)
    at android.database.sqlite.SQLiteConnection.open(SQLiteConnection.java:260)
    at android.database.sqlite.SQLiteConnectionPool.tryAcquireNonPrimaryConnectionLocked(SQLiteConnectionPool.java:989)
    at android.database.sqlite.SQLiteDatabase.getVersion(SQLiteDatabase.java:1094)
    at net.osmand.shared.api.SQLiteAPIImpl$SQLiteDatabaseWrapper.getVersion(SQLiteAPIImpl.kt:35)
    at net.osmand.shared.gpx.GpxDatabase.openConnection(GpxDatabase.kt:91)
    at net.osmand.shared.gpx.GpxDatabase.updateGpxParameters(GpxDatabase.kt:173)
    at net.osmand.shared.gpx.GpxDbHelper.updateDataItem(GpxDbHelper.kt:244)
    at net.osmand.plus.settings.backend.backup.items.GpxSettingsItem.updateGpxParams(GpxSettingsItem.java:131)
    at net.osmand.shared.gpx.GpxDbHelper.onProgressUpdate(GpxDbHelper.kt:522)
    at net.osmand.shared.gpx.GpxReader.onProgressUpdate(GpxReader.kt:67)
    at net.osmand.shared.KAsyncTask$publishProgress$1.invokeSuspend(KAsyncTask.kt:36)
```

The same exception has a second fatal path: the import thread does the same opens (`GpxSettingsItem$1.readFromStream` → `GpxDbHelper.add`), and `ImportFileItemsTask.doInBackground` only catches `IllegalArgumentException | IOException`, so `AsyncTask` rethrows it. That is the variant reproduced below. The `GpxReader` threads hit it as well but swallow it (`doReading()` logs and stops).

## Cause

`GpxDatabase.openConnection(readonly = true)` asks the Android `SQLiteAPIImpl` for a read-only connection, and never got one: `Context.openOrCreateDatabase()` is always read-write, and the flag only dropped `MODE_ENABLE_WRITE_AHEAD_LOGGING`. [bbc14193c5](https://github.com/osmandapp/OsmAnd/commit/bbc14193c535160cdb2c35b17fb16c81f4197062) (2018-05-02, "Update target sdk to 26") translated the earlier, wrong `SQLiteDatabase.OPEN_READONLY` (that value is `MODE_WORLD_READABLE` to a `Context`, forbidden from targetSdk 24) into `MODE_PRIVATE | (readOnly ? 0 : MODE_ENABLE_WRITE_AHEAD_LOGGING)`, and the KMP port [02387538cf](https://github.com/osmandapp/OsmAnd/commit/02387538cfd6f1ed54027c73bfbbc1795227ed29) (2024-05-30) copied the expression. Without the WAL flag the framework applies its default journal mode (`TRUNCATE`) to the connection on open. So:

- a "read-only" open that finds no other connection open switches the shared `gpx_database` file from WAL to a rollback journal (checkpoint, WAL delete, header rewrite, under an exclusive lock);
- the next writable open switches it back (header rewrite, under an exclusive lock);
- while the file is in rollback mode every transaction blocks all readers, and every new connection needs a shared lock just to compile `PRAGMA journal_mode` — the statement in the trace.

The GPX database never opened read-only before the KMP migration (the Java `GPXDatabase` had 10 writable opens and none read-only). Read-only opens arrived with [0cabf8d2ca](https://github.com/osmandapp/OsmAnd/commit/0cabf8d2cafd71eb42fd012e7e3c4777f19008b2) (2024-09-18, startup batch load — the source of the occasional "Could not change the database journal mode" warnings), then [25b5ed8158](https://github.com/osmandapp/OsmAnd/commit/25b5ed81584b295cd9a3abc1e09ab8cfb95eb2ce) (2025-07-05, "Fix #22641") added `updateAppearanceTimestamp`, a read-only open after **every** write, and [a02bf59444](https://github.com/osmandapp/OsmAnd/commit/a02bf594441f7ce104a0a4fd071dbbce08666dbd) (2025-12-27, #24196, "Change GPXdatabase mode to read-only where no write is needed") converted seven more read paths. During an import the database is opened ~14 times per file from up to six threads — the import thread (`add`, `updateGpxParams`), up to four `GpxReader`s and the main-thread progress callback — about six of them read-only, so the file flips modes thousands of times and the lock wait exceeds the framework busy timeout of 2.5 s. Whichever thread loses gets `SQLITE_BUSY`.

## Reproduction (Pixel 8 Pro, Android 17 / SDK 37, `gplayFreeOpenglArm64Debug`)

A throw-away instrumented harness (not part of this PR) generates an `.osf` with 1 000 tracks of 300 points, each with a colour and width, and imports it through the real path (`FileSettingsHelper.collectSettings` → `importSettings` → `ImportFileItemsTask` → `SettingsImporter` → `GpxSettingsItem` → `GpxReader`s → main-thread callback), then checks every track in `GpxDbHelper`.

| | `r5.4` @ [246314b98e](https://github.com/osmandapp/OsmAnd/commit/246314b98e6c0c8b445be4f7e61fc14b3d8725b5) | this branch |
|---|---|---|
| Import of 1 000 tracks | **process crashed after 2 min 55 s, 872 files in** — `SQLiteDatabaseLockedException … while compiling: PRAGMA journal_mode` in `GpxDatabase.openConnection` ← `GpxDbHelper.add` on `AsyncTask #1` | completed, 157 s, `ok=1000 missing=0 wrongColor=0 notAnalysed=0` |
| `Could not change the database journal mode … from 'wal' to 'TRUNCATE'` in logcat | 2 222 | 0 |
| `database is locked` lines in logcat | 4 451 | 0 |

The same import also crashed twice on a `master` 5.5.0 debug build on the same device (813 files / 2 min 44 s; 53 s), matching the issue's build.

## Fix

Make `readOnly` mean read-only in the Android `SQLiteAPIImpl.getOrCreateDatabase`: open with `SQLiteDatabase.OPEN_READONLY`, exactly as `openByAbsolutePath()` already does. The framework skips every journal-mode, sync-mode and metadata-table step for a read-only connection, so a read path can no longer change the journal mode of the file; the connection cannot write; and it is cheaper to open (one pooled connection, none of the open-time PRAGMAs). Writable opens keep `MODE_ENABLE_WRITE_AHEAD_LOGGING`, so the file stays in WAL and readers never wait for writers — the state the Java `GPXDatabase` always had and the state iOS has (its `SQLiteAPIImpl` ignores `readOnly`; SQLiter opens everything in WAL).

Two cases need a writable open instead (`canOpenReadOnly`): the database file does not exist yet (a read-only open cannot create it; `GpxDatabase.openConnection` then sees version 0 and initialises it as before), and a hot `-journal` (> 0 bytes) left by a rollback-journal transaction of the old build that was interrupted — for instance by this crash — which a read-only connection cannot roll back and would fail on with `SQLITE_READONLY_ROLLBACK` at every start.

Why here: this file is where the 2018 translation lost the read-only meaning; `getOrCreateDatabase` of this implementation has a single caller, `GpxDatabase.openConnection`, whose reopen-as-writable branch for creation and upgrade now serves its purpose; nothing in `commonMain`, the JVM implementation (`?mode=ro`) or iOS changes. All `openConnection(true)` callers only run `SELECT`s. The app's other databases go through the separate Java `net.osmand.plus.api.SQLiteAPIImpl`, untouched.

## Verification

`OsmAnd/test/java/net/osmand/test/junit/GpxDatabaseOpenConnectionTest.kt`, three instrumented tests:

- `readOnlyOpenKeepsWriteAheadLogging` — the journal mode of the file must survive a read-only `openConnection()`.
- `readOnlyOpenIsReadOnly` — `openConnection(true)` returns a read-only connection, `openConnection(false)` a writable one.
- `openIsNotBlockedByTransactionOnReadOnlyConnection` — a transaction on a connection opened the read-only way must not block another thread's `openConnection()`. Before the fix that connection is in rollback mode and the open waits for the busy timeout; the test then repeats the framework open and fails with the exception it throws, so the report reads like the issue. After the fix the connection refuses the transaction.

```
./gradlew :OsmAnd:assembleGplayFreeOpenglArm64Debug :OsmAnd:assembleGplayFreeOpenglArm64DebugAndroidTest
adb -s 37211FDJG001E7 install -r -d OsmAnd/build/outputs/apk/gplayFreeOpenglArm64/debug/OsmAnd-gplayFree-opengl-arm64-debug.apk
adb -s 37211FDJG001E7 install -r OsmAnd/build/outputs/apk/androidTest/gplayFreeOpenglArm64/debug/OsmAnd-gplayFree-opengl-arm64-debug-androidTest.apk
adb -s 37211FDJG001E7 shell am instrument -w -e class net.osmand.test.junit.GpxDatabaseOpenConnectionTest net.osmand.test/androidx.test.runner.AndroidJUnitRunner
```

Before (`r5.4` build, test APK from this branch):

```
Tests run: 3,  Failures: 3
org.junit.ComparisonFailure: expected:<[wal]> but was:<[truncate]>
java.lang.AssertionError                                       (readOnlyOpenIsReadOnly: isReadOnly() is false)
android.database.sqlite.SQLiteDatabaseLockedException: database is locked (code 5 SQLITE_BUSY): , while compiling: PRAGMA journal_mode
    at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
    at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:1145)
    at android.database.sqlite.SQLiteConnection.executeForString(SQLiteConnection.java:845)
    at android.database.sqlite.SQLiteConnection.setJournalMode(SQLiteConnection.java:452)
    at android.database.sqlite.SQLiteConnection.setJournalFromConfiguration(SQLiteConnection.java:374)
    at android.database.sqlite.SQLiteConnection.open(SQLiteConnection.java:291)
    Suppressed: java.lang.AssertionError: could not open the database after 2539 ms while another thread held a transaction
```

After: `OK (3 tests)` in 3.4 s.

The two fallbacks, on the `Pixel_10_Pro` emulator (Android 16 / API 36, x86_64, `nightlyFreeLegacyX86Debug` as a fresh install of `net.osmand.dev`):

- no `databases/` directory at all at first start → the read-only requests of `loadItems()` fell back to the writable open, `gpx_database` was created, `OK (3 tests)`, no "Failed to get or create database" in logcat;
- a non-empty `gpx_database-journal` planted next to the database with the app stopped → start-up rolled it back through the writable open (file gone afterwards), `OK (3 tests)`, no read-only errors. Without the fallback the same file makes a read-only open fail its first read with `SQLiteReadOnlyDatabaseException: attempt to write a readonly database (code 776 SQLITE_READONLY_ROLLBACK)` (checked with a throw-away probe opening the file with `OPEN_READONLY` directly) — i.e. a crash at every start until a writable open happened first, which start-up never does.

The same on **Android 7.0 (API 24, the minSdk)**, `Api24` x86 emulator, fresh install of `net.osmand.dev`: database created through the fallback at first start, `OK (3 tests)`; planted hot journal rolled back at the next start, `OK (3 tests)`; no read-only errors either time. Both open paths resolve the same file on every version: `Context.openOrCreateDatabase()` is `SQLiteDatabase.openDatabase(getDatabasePath(name).getPath(), …)` plus a `chmod`, on API 24 as on API 35, and the read-only branch uses the same `getDatabasePath()`.

Not verified: the merge into `master` (the changed file is identical there, and all `openConnection(true)` users on `master` are `SELECT`-only as well); an iOS build (androidMain only, iOS unaffected by construction).

## Known gap, not addressed here

- `GpxSettingsItem.updateGpxParams` still runs its database writes on the main thread from the reader's progress callback (four opens and two `UPDATE`s per imported file) — no longer a crash source, but jank on a large import; the issue's expected result asks for it to move off the main thread, which is a separate change.
- `ImportFileItemsTask.doInBackground` catches only `IllegalArgumentException | IOException`, so any other runtime failure of an item import is still fatal.
- The Java `net.osmand.plus.api.SQLiteAPIImpl` keeps the same `readOnly ? 0 : MODE_ENABLE_WRITE_AHEAD_LOGGING` expression for the other databases; low traffic there, left alone.
- Every `GpxDatabase` operation still opens and closes its own `SQLiteDatabase`; a long-lived instance would remove those opens altogether, but that is a shared-code refactor.

## AI disclaimer

Prompts used (summarised):
1. Analyse the crash from the issue and how to fix it without breaking anything.
2. Whether the WAL change affects other places; whether it can be fixed in `GpxSettingsItem` or `GpxDbHelper`/`GpxDatabase`; keep the fix local, on a branch from 5.4; why the crash appeared now when there were only warnings before.
3. Make the local fix on 5.4 without committing.
4. A/B test on the Pixel with the `gplayFreeOpenglArm64Debug` variant: crash without the fix, working with it.
5. Make the change simple and small; whether passing `openConnection(true)` for the track import alone would do.
6. Explain the crash in simple terms; why iOS has no such crash; why the `readOnly ? 0 : MODE_ENABLE_WRITE_AHEAD_LOGGING` expression was added; what the JVM implementation does.
7. Create a branch from 5.4 with the fix and a test, no worktrees; make the test reproduce the crash with the exception from the issue; use the elvis operator; open a draft PR onto `r5.4` following `AGENTS.md`; link commit hashes.
8. Whether opening read-only connections with `SQLiteDatabase.OPEN_READONLY` would work; whether read-only has upsides; do the proper read-only implementation and fix the crash with it.

- The first version of the fix (always open writable in `GpxDatabase.openConnection`, superseded here) and, for this version, the two fallbacks to a writable open (missing database file, hot rollback journal) and their placement in `canOpenReadOnly`.
- The design of the regression tests (journal-mode check, read-only check, blocked-open reproduction that repeats the framework open to attach the exception) and their waits for application initialisation and idle readers.
- The synthetic 1 000-track `.osf` harness used for the before/after import, its parameters, and installing the r5.4 build over the newer build on the device with `adb install -r -d`.
- The git history investigation, branch name, commit message and PR wording.


alex-dev-neo:
Reviewed the diagnosis, the change and the surrounding code. The diagnosis holds and the fix is in the right place; I found no blockers. Notes below are all non-blocking.

**What I checked**

- The stack in the issue fails on a **writable** open (`updateGpxParameters` → `openConnection(false)`), inside `tryAcquireNonPrimaryConnectionLocked` — i.e. while the framework compiles `PRAGMA journal_mode` for the *second* pooled connection. That statement can only wait if the file is in a rollback journal at that moment and someone holds an exclusive lock, which is exactly what the alternating read-only/writable opens produce. Android's `beginTransaction()` is `BEGIN EXCLUSIVE`, and in rollback mode that blocks readers, while in WAL it does not. So the explanation matches the reported trace, not just a similar one.
- `getOrCreateDatabase` of the shared `androidMain` implementation really does have a single caller (`GpxDatabase.openConnection`); every other database goes through `net.osmand.plus.api.SQLiteAPIImpl`.
- All `openConnection(true)` call sites are `SELECT`-only — 8 on `master` and all 10 on `r5.4`, including the parallel batch loader in `getGpxDataItems()`, which is where `r5.4` puts even more pressure on the lock than `master`.
- Not touching the Java implementation is right: `SQLiteTileSource.createDataBase()` calls `getOrCreateDatabase(name, true)` and then runs `CREATE TABLE` / `INSERT` on it, so real read-only semantics would break tile databases there.
- The missing `-wal` fallback is not needed: WAL recovery needs to write the `-shm`, not the database, and the app owns those files, so `SQLITE_READONLY_RECOVERY` cannot happen here.

**Suggestions**

1. *Fall back on failure, not only on a pre-check.* `canOpenReadOnly()` decides up front from `file.exists()` and the `-journal` size. If the read-only open fails for a reason it does not model (a race with file removal, corruption), `getOrCreateDatabase` returns `null` and the caller silently returns an empty list of tracks, where the old writable open would have succeeded. `PlacesDatabaseHelper.openConnection()` already has the idiom for this: `api.openByAbsolutePath(path, readOnly) ?: if (readOnly) api.openByAbsolutePath(path, false) else null`. Strongest form: open read-only, read `version` inside the same `try`, and on any `SQLiteException` close and reopen writable — that covers the hot journal too (it surfaces on the first read, not on the open) and lets the file checks go away.

2. *The issue's acceptance criterion is not met literally.* "progress is updated without performing database transactions on the main thread" — `GpxSettingsItem.updateGpxParams()` still does 4 opens and 2 `UPDATE`s per file on the main thread, and it runs inside `readerSync.synchronize {}` in `GpxDbHelper.onProgressUpdate()`, so the main thread holds the reader lock across that database I/O. The PR lists this as a known gap, which is fine, but the checkbox in the issue cannot be ticked by this PR alone.

3. *No hard guarantee against the crash.* WAL stops readers from waiting, but writers still serialise and the busy timeout is 2.5 s, with the import thread, up to four `GpxReader`s and the main thread all writing. Nothing catches `SQLiteDatabaseLockedException` on the main thread, and `ImportFileItemsTask.doInBackground` still catches only `IllegalArgumentException | IOException`. 1 000 tracks pass empirically; a catch/retry would make it independent of timing.

4. *The third test degenerates after the fix.* `openIsNotBlockedByTransactionOnReadOnlyConnection` returns early at `if (transactionRefused.get())`, so on the fixed build it asserts the same thing as `readOnlyOpenIsReadOnly`. A test that holds a transaction on a **writable** connection and asserts another thread can still open and read would check the WAL guarantee directly, and would fail if `MODE_ENABLE_WRITE_AHEAD_LOGGING` is ever dropped again.

5. *Same defect elsewhere (separate PR).* `net.osmand.plus.api.SQLiteAPIImpl` keeps `readOnly ? 0 : MODE_ENABLE_WRITE_AHEAD_LOGGING` for backup, markers, search history, POI, travel and tiles. The semantics cannot change there (see `SQLiteTileSource`), but always passing the WAL flag is safe and removes the same flapping. Backup is not low traffic in this respect: `CollectLocalFilesTask` holds a "read-only" connection for the whole scan, so the backup database stays in rollback mode for its duration.

Nit: the PR title undersells the change — the commit headline ("Open read-only GPX database connections read-only") describes it better.

Chumva:
@sonora, this change could also improve GPX DB speed in some cases by 30% 
image

Issue #19715, opened 2024-04-30
Issues with Olympiiskaya subway station


vshcherb:
## Description 
Fix issues related to visibility of public transport stops.

<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/1042025/e962adcc-661c-4294-8515-cbdb770a643c"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/1042025/3ad95686-c387-4a63-989d-e153d5911033"  width="300" height="auto"/> </td>
    <td> <img src="https://github.com/osmandapp/OsmAnd/assets/1042025/b1474569-8bb9-46f7-95db-147b451b26d7"   width="300" height="auto"/></td>

  </tr>
</table>   

* Coordinates `50.43214, 30.51630`

- [ ] Too many same titles Olympiiskaya - should be only one visible ! Idea: Probably apply mechanism with same title in radius (@xmd5a2 )- https://github.com/osmandapp/OsmAnd-core-legacy/blob/1d39a1a0fde5b25afbfcf75010ac2a80ef714fb5/native/src/textdraw.cpp#L547C20-L547C50 (debugText..SameName). Compare not final name ```text == text ``` but compare ```baseText == baseText``` if not empty (where baseText doesn't contain nameTag2)
- [x] **2 Issues with bus stop** : No nearby transport in 150m (UI issue)
- [ ] **2 Issues with bus stop** : No transport at bus stop (data issue)

- - - 

<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <td><img src="https://github.com/osmandapp/OsmAnd/assets/9365308/7796a880-f280-41f8-bddb-ee44ea153808"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/osmandapp/OsmAnd/assets/1042025/caef3812-b947-48db-a72d-da18ebc6ff65"  width="300" height="auto"/> </td>
      </tr>
    </table>
</details>  

- [x] Olympiiskaya (3) - has different icons (test with June maps / Kyiv)

Dima-1:
- https://github.com/osmandapp/OsmAnd/pull/19797
<table style="width:100%">
  <tr>
[...]

PR #26053, opened 2026-09-15
Move transport stop matching for an amenity to OsmAnd-java

alisa911:
Follow-up for osmandapp/OsmAnd-Issues#3263 (context menu for vector tile objects on the web).

The web server has to show the same transport stop for an amenity that the Android context menu shows. The review of osmandapp/OsmAnd-tools#1694 asked not to duplicate the Android matching and its constants on the server, so the matching moves to OsmAnd-java and both sides call it.

`net.osmand.data.TransportStopMatcher` holds the constants, `isPublicTransportStop`, `isSubwayEntrance`, `getSearchRadius`, `findBestStopForAmenity` and `aggregateStopsForAmenity` with the five private matching helpers. Reading the stops stays platform-specific: Android keeps `findTransportStopsAt` over `ResourceManager`, the server uses its own readers, and the shared code takes a ready `List<TransportStop>`.

Android behaviour is unchanged. The moved method bodies are identical to the previous ones, the order stays radius → read → sort → branch → `local[0] ?? nearby[0]`, and the `stop.setTransportStopAggregated(...)` side effect that `TransportStopController` relies on is preserved. `SHOW_STOPS_RADIUS_METERS_UI` moved from `TransportStopController` into the shared class so that `SHOW_STOPS_RADIUS_METERS = SHOW_STOPS_RADIUS_METERS_UI * 6 / 5` keeps a single source, and its two UI call sites were updated. `AmenityMenuController` now calls `TransportStopMatcher.isPublicTransportStop` instead of its own inline copy.

`MapPoiTypes.getPublicTransportTypes()` walks the same transportation → public_transport → types path and caches the result, but it is not the same set: it also includes the `water_transport` filter and the poi additionals of every type. Reusing it here would make ferries and water transport count as public transport stops, which is a behaviour change for the Android context menu, so this check keeps its own walk over the `public_transport` filter only.

The server side is osmandapp/OsmAnd-tools#1694 and needs this merged first.

## AI disclaimer

Written with Claude Code (Claude Opus 5), driven by [@alisa911](https://github.com/alisa911). Summary of the requests it worked from, in order:

1. Study osmandapp/OsmAnd-Issues#3263 together with the web PR osmandapp/web#1996 and the server PR osmandapp/OsmAnd-tools#1694, and list what the review still asks for. Two comments by @RZR-UA on the server PR drive this change: `TransportStopsService.java:177` (does the port match Android, why not all of it, consider sharing `findBestTransportStopForAmenity` between Android and Web) and `:35` (do not duplicate constants, place them in OsmAnd-java).
2. Check whether the code can be shared at all before writing anything. Confirmed: OsmAndServer already builds `:OsmAnd-java` as a subproject pointing at the android checkout, every type the matching uses lives in `net.osmand.data`, and the only platform-specific part is reading the stops. OsmAnd-java has no `androidx.annotation` on its classpath, so `@NonNull`/`@Nullable` are dropped in the moved code.
3. Move the code, on the explicit condition that nothing in Android may break or change behaviour. `TransportStopHelper` keeps `findTransportStopsAt`, both public entry points and the route helpers and delegates the rest; the constants and the five private matching methods moved as they were; `SHOW_STOPS_RADIUS_METERS_UI` moved out of `TransportStopController` with its two UI call sites updated; the inline public-transport check in `AmenityMenuController` was replaced by the shared method.
4. Asked why the port had one difference from the original. The subway branch had been rewritten as an early return, so the aggregate no longer needed a separate declaration. On request the original `if/else` shape was restored, after which a mechanical comparison of every moved method body against master reports all of them identical.
5. Pointed out that `TransportStopAggregator` and the existing `TransportStopAggregated` are hard to tell apart in a file list, asked what the existing class holds and why a separate class was preferred over static factories on it. Renamed to `TransportStopMatcher`.
6. Remove the class comment as unnecessary.

7. Go through three findings about the new class. Two were confirmed and are about the contract of a now public API: `findBestStopForAmenity` relies on the caller having read the stops within `getSearchRadius`, and the list and its stops are modified in place. Filtering by the radius inside was rejected because the stops are read over a bbox, not a circle, so a circular cutoff would change which stop Android returns; instead the parameter became `stopsInSearchRadius` and the mutation is stated on `aggregateStopsForAmenity`. The third finding, about `MapPoiTypes.getPublicTransportTypes()`, is answered in the description above.

The agent compiled `:OsmAnd:compileAndroidFullLegacyArm64DebugJavaWithJavac` and `:OsmAndServer:compileJava`, both successful, and compared the moved method bodies against master mechanically. It did not run the app and did not run the Android tests. The single commit was made by the agent at the author's explicit request.


OsmAnd-java/src/main/java/net/osmand/data/TransportStopMatcher.java
```diff
@@ -0,0 +1,176 @@
...
+import net.osmand.util.Algorithms;
+import net.osmand.util.MapUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransportStopMatcher {
+
+	public static final int SHOW_STOPS_RADIUS_METERS_UI = 150;
+	public static final int SHOW_STOPS_RADIUS_METERS = SHOW_STOPS_RADIUS_METERS_UI * 6 / 5;
```
RZR-UA: Please test https://github.com/osmandapp/OsmAnd/issues/19715 with 150 meters and then remove SHOW_STOPS_RADIUS_METERS_UI and keep only SHOW_STOPS_RADIUS_METERS = 150 in case of no regression.

RZR-UA (approved):

PR #26054, opened 2026-09-15
Spatial search: offer the street when the map lacks the house number

vshcherb:
A query for a house the map does not have listed the same number on same-named streets of other towns first. Example: "17bis Rue des Pavillons" at a point on that street in Nanterre (the house was added to OSM after the map build) gave "17 bis Rue des Pavillons" in Villejust 22 km away, and with "Nanterre" in the query house 17 in Puteaux 1.7 km away. `calcBuilding` skipped the street + number combination because the street has no such house, and the street alone matches fewer tokens, so it ranks under any house.

Now the combination stays as a street row when the street is within 300 m of the location (street bbox), the stage covers the whole query and no POI takes part. These rows are dropped when the house itself is found: exactly within 5 km, or partially ("315" for "315 B") within 300 m. The result keeps the street atom, so the row is a STREET.

### Tests

Golden `france_nanterre_rue_des_pavillons` goes to resources separately: it fails on master and passes with this change in Pipeline, Score and Combinations.

All 84 golden tests with full row dumps, master vs this change (the spatial package is the same as on c12e6f46ff, where it was measured):
- `france_nanterre_rue_des_pavillons`: 4 phrases fixed.
- `turkey_sokak`: 7 of 15 phrases change and fail. The number-named street "2.Sokak" at the point comes first for "2 2 Sokak", "6178/2.Sokak", "2/6. Sokak", "2 6 Sokak", "1/2 7029 Sokak", and "Küçük Sanayi Sokak" (460 m) for "2. Sanayi 6. Sokak". Under review, golden not updated.
- `ukraine_boiarka_33_line`: passes; rows 2-6 of "вулиця 33-я Лінія" are now neighbouring "Лінія" streets instead of houses 33.
- The other 81 tests: rows unchanged.

Real maps (1 Sep build): the street at the point is now first for 17bis / 27 bis Rue des Pavillons (with and without Nanterre), 18 Rue des Petits Pois, 241 South Fort Lauderdale Beach Blvd, 2807 East Oakland Park Boulevard and Plaza de Callao 1. "Plaza de Callao 1 Madrid" still starts with "1 <Madrid>" at 4.7 km.

## AI disclaimer

Prompts used (summarised):
1. Create a golden test "suggest street when house number is missing" for 17bis Rue des Pavillons, ignored for now.
2. Suggest a simple fix that does not harm exact search: when only the house number is missing, show the street as if the house were there; if the house is found, the street is not needed.
3. Show the run of all tests and how much code changes, with minimal code; the row must be a STREET; make a review page with OK buttons.
4. Put 3-5 expected rows into the test.
5. Open a draft PR with the android changes only; resources are committed separately.

Decided by the agent, not requested explicitly: the 300 m street radius, the 5 km exact / 300 m partial house thresholds, the whole-query and no-POI conditions (each added after a golden test broke without it), keeping the street atom in `SpatialSearchResult` so the row is a STREET.


vshcherb:
Closed: a simpler version is being tried in the workspace first. (gen by AI)

PR #26094, opened 2026-09-18
Fixed npe when getting icon for spatial search result in show on map

Corwin-Kh:
### Problem

`RenderingIcons.getResId()` returns an `Integer` and only knows rendering icons (`mx_`, `h_`, small icons). Address icons (`ic_action_street_name`, `ic_action_building`, …) and the `gpx_icon` values of travel amenities are ordinary drawables, so it returns `null` — and that `null` was unboxed straight into the `int iconId` parameter of `PointImageUtils.getOrCreate(...)`, giving an NPE in `POIMapLayer` (legacy canvas path) and `POITileProvider`.

On top of that, every non-POI object shown on the map by spatial search got the street icon: the helper in `QuickSearchDialogFragment` only distinguished `HOUSE` from everything else.

### Change

- `RenderingIcons.getPointIconId(ctx, id, defIconId)` — resolves a rendering icon first, then a plain drawable by name, then falls back to the default. It never returns 0, so `getResourceEntryName()` inside `getOrCreate()` stays safe.
- `POIMapLayer` (2 call sites) and `POITileProvider` use it. The OpenGL path also stops silently degrading to `mx_special_marker` for names that are ordinary drawables.
- `QuickSearchListItem.getAddressIconName()` — icon name per address `ObjectType` (city/town, village, postcode, street, house, intersection), `null` for anything else.
- `QuickSearchDialogFragment` uses it and sets `GPX_ICON` only when a name is returned, so non-address `MapObject`s (transport stops and the like) no longer get a street icon — they fall back to the neutral marker.

Behaviour: `HOUSE` and `STREET` markers are unchanged; city/village/postcode/intersection get their own icons; an unknown icon name renders the neutral marker instead of crashing.

### AI disclaimer

Written with Claude Code (Opus 5), pair-programming style.

**Human:** identified the problem and the source of the fix (an unfinished local "show streets" experiment), decided which part of it to land, reviewed every step, caught both design problems listed below, opened the PR, and is responsible for building and testing the result.

**AI:** searched the history to locate the earlier work, transplanted the icon-resolution code into the current tree, wired it up, did the nullability/semantics refactor requested in review, and audited the final diff for behavioural regressions. The AI did not build or run the app.

**Key prompts** (translated from Russian, condensed):

1. *"Find the commit where we added showing streets found by spatial search on the map."* → history research; the relevant work turned out to be uncommitted, sitting in a local stash.
2. *"Extract from that stash the code for getting icons for address objects. I think we also fixed a crash there for the case when `ic_action_street_name` was not found."* → the actual scope of this PR.
3. *"`getAddressIconName` does not seem to be used anywhere now."* → review catch; the call site in `QuickSearchDialogFragment` had not been transplanted.
4. *"So now all icons for spatial search come from `getAddressIconName`. That is semantically wrong — not every spatial search result is an address."* → review catch; the method became `@Nullable`, `STREET` got an explicit case instead of being the catch-all default, and the caller now guards on `null`.
5. *"Did we introduce any regression?"* → diff audit of all five files.

Everything else from the stash (the address-object search itself) was deliberately left out of this PR.


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/25488 (main Issue for PR)
https://github.com/osmandapp/OsmAnd/issues/25485 (connected issue)

This PR fixes the loop crash caused by incorrect loading of a "special" `ic_action_street_name` (and similar) icon.

```
08:37:03.448  E  OsmandMapTileView Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference {RenderingBaseImage}
                 java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
                   at net.osmand.plus.views.layers.POIMapLayer.onPrepareBufferImage(POIMapLayer.java:915)
                   at net.osmand.plus.views.OsmandMapTileView.refreshBaseMapInternal(OsmandMapTileView.java:1121)
                   at net.osmand.plus.views.OsmandMapTileView.lambda$refreshBufferImage$2(OsmandMapTileView.java:1395)
                   at net.osmand.plus.views.OsmandMapTileView.$r8$lambda$MsdWOkmbW9fZXnb-I1VO7XRb_zs(Unknown Source:0)
                   at net.osmand.plus.views.OsmandMapTileView$$ExternalSyntheticLambda2.run(D8$$SyntheticClass:0)
                   at android.os.Handler.handleCallback(Handler.java:995)
                   at android.os.Handler.dispatchMessage(Handler.java:103)
                   at android.os.Looper.loopOnce(Looper.java:248)
                   at android.os.Looper.loop(Looper.java:338)
                   at android.os.HandlerThread.run(HandlerThread.java:85)
```

PR #26103, opened 2026-09-18
Fix dash path effect cache never hitting in OsmandRenderer

vshcherb:
`OsmandRenderer.getDashEffect` caches `PathEffect` in a `Map<float[], PathEffect>`. Array keys use identity `equals`/`hashCode`, and the key array is allocated fresh on every call, so `containsKey` never matches. Every dashed line drawn by the legacy renderer allocates a new `OsmandDashPathEffect` (finalizable, with a native peer) and adds an entry to a map that is never cleared — unbounded growth for the process lifetime.

Fix: key by the dash pattern contents and the phase.

Measured on a heap dump (nightly, few minutes of use): 4357 `OsmandDashPathEffect` instances, only **23 distinct patterns**, `dashEffect` map size 4357 — a 0% hit rate; one pattern was allocated 3175 times.

Test with the method copied verbatim, 10 000 calls with one identical pattern:

```
current code : objects created = 10000, map size = 10000
content key  : objects created = 1,     map size = 1
```

Scope: this only affects the legacy (non-OpenGL) renderer path, which is used while `mapRenderer == null` (`OsmandMapTileView:1374`). With OpenGL on I could not make the map grow at all: 3 min of panning adds 0 entries, and 20 MapActivity recreations add 0 entries. All the 5.4.6/5.5.0 ANR dumps under investigation run the OpenGL core, so this is **not** claimed to be the cause of the large heaps there — it is a real unbounded leak on the legacy path, worth fixing on its own.

Not compiled: the app module does not build in this checkout (missing generated icon resources).

## AI disclaimer

Prompts used (summarised):
1. Analyse the ANR/GC dumps collected from 5.4.6 and explain the huge heaps.
2. Where could that many finalizable objects come from — confirm with a test.
3. Reproduce it on the emulator; it is a bug either way, open a request.

Decided by the agent, not requested explicitly: keying by `Arrays.toString(dashes) + '_' + st` (the phase was not part of the old key at all); leaving the surrounding code untouched.


sonora:
@vshcherb  What a catch! 
- We should likely grep the entire projects for more occurrenes of
```
Map<float[], Something>
Map<int[], Something>
Map<double[], Something>
```

- Same for `Set`?

- And perhaps there are more types risky as Map keys? Like
```
PointF
RectF
Paint
List<Float> / List<Integer> ?
```

EDIT: I've checked the render subsystem and it looks like _pathEffect_ was the only thing affected...

**But it looks like _BinaryMapDataObject.java_ may suffer from a,similar issue?**

vshcherb:
This old rendering nothing critical found small fix

vshcherb:
it was a bug only here found

PR #26108, opened 2026-09-19
Copy the map section of the obf reader into OsmAnd-shared

alex-dev-neo:
Second step of the obf reader port for [OsmAnd-Issues#2935](https://github.com/osmandapp/OsmAnd-Issues/issues/2935), which brings the MAP and POI sections into OsmAnd-shared so that iOS can search an obf without the C++ core.

**Stacked on #26107.** Its base is `kmp-obf-1-collator`, because the map search needs the generic `SearchRequest<T>` that PR introduces. Review that one first; once it lands this diff stands on its own.

### What is in it

`net.osmand.shared.binary` gains the map section, copied from `BinaryMapIndexReader` and `BinaryMapDataObject` in OsmAnd-java with the java method names kept, so the two can be read side by side:

- `MapIndex` - the table that turns a tag and a value into a number and back, the r-tree roots, and `adoptMapObject`, which restates an object in another section's numbers.
- `MapTree` / `MapRoot` - the boxes, their zoom ranges and where their bytes are.
- `BinaryMapDataObject` - one map object with its points, type numbers and names.
- `SearchFilter` - accepts or rejects an object from its type numbers alone, before the rest of it is parsed.
- In the reader: `readMapIndex`, `readMapEncodingRule`, `readMapLevel`, `readMapTreeBounds`, `searchMapIndex`, `readMapDataBlocks`, `searchMapTreeBounds`, `readMapDataObject`, and the `containsMapData` overloads.
- In `SearchRequest`: the `searchFilter` field and the `buildSearchRequest` factory.

Android is untouched: it keeps its java classes, and nothing outside OsmAnd-shared and the test changes.

### Deliberate differences from java

- The read statistics behind `READ_STATS` are left out. They serve the obf inspection tools, which stay in OsmAnd-java.
- `MapIndex.positiveLayers` / `negativeLayers` are plain sets of boxed ints rather than a primitive set copied in from trove; they hold a handful of ids each.
- `readMapEncodingRule` does not intern the tag and the value. Java's interning shares the few dozen distinct tags of a file between its sections; Kotlin/Native has no interning, so each row keeps its own.
- `BinaryMapDataObject.toString` prints the raw id, because turning it back into an osm id belongs to `ObfConstants`, which is not copied yet.
- Where java hands back a null it read - a name the string table had nothing for, a type its section has no rule for - the copy answers `""` or leaves the entry out instead of throwing. The reader fills names and their order together, so no object read out of a file reaches those paths.

### Tests

`MapSectionCompatTest` opens every obf the tests ship with in both readers and compares what they read. Both are opened fresh per case, so the state they build up lazily - the encoding table and the box trees - is compared too.

| test | what it compares |
| --- | --- |
| `sectionsAreTheSame` | sections, names, offsets, centres, and every r-tree root with its zoom range and box |
| `coverageIsTheSame` | `containsMapData` for corners, centres and a point outside, per root and zoom |
| `objectsAreTheSame` | every object of every zoom level: id, area, points, inner rings, types, names, ordered names, layer, label, decoded tags; then the encoding tables both sides ended up with; and `compareBinary` on object pairs at three precisions, which exercises the shape simplification |
| `adoptedObjectsAreTheSame` | objects restated into a seeded table on both sides, and the table that results |
| `filteredSearchIsTheSame` | a filter is offered the same types in the same order, and turns down the same objects |
| `emptyBoxIsTheSame` | a box the file does not reach, and the ocean and land flags |

94 786 map objects compared across the 26 test files.

Also green: `:OsmAnd-java:test`, `:OsmAnd-shared:jvmTest`, `:OsmAnd-shared:compileTestKotlinIosSimulatorArm64`, `:OsmAnd-shared:compileDebugKotlinAndroid`, `:OsmAnd-shared:linkDebugFrameworkIosSimulatorArm64` (no ObjC export name collisions), `:OsmAnd:compileAndroidFullLegacyArm64DebugKotlin`.


alex-dev-neo:
Added the benchmarks, since the port swaps trove for the shared collections and drops java's `intern()` on the encoding table, and none of that had a number on it.

Reading every object of every zoom level of the 26 test obf files once, 47 393 objects:

| phase | java (jvm) | copy (jvm) | copy (native) |
| --- | --- | --- | --- |
| open, 26 files | 4.6 ms | 3.8 ms | 2.4 ms |
| search | 15.8 ms | 9.5 ms | 22.0 ms |
| per object | 0.33 us | 0.20 us | 0.46 us |

Kotlin/Native reads an object in about the time the java reader takes on the jvm, and in about twice the time this same code takes there - the same ratio the routing port measured. Nothing here is a regression gate: the java side asserts only that both readers found the same objects, and the shared one does nothing at all unless `OSMAND_MAP_BENCHMARK` is set, which is how `RoutePlannerBenchmarkTest` is gated too.

Two notes on reading the table. The open column is not a like for like, because java also reads the address, poi and transport headers, which the copy skips. And the copy being the faster of the two on the jvm was not chased further - the lazy encoding table, where java interns, accounts for about 1.4 ms of it, and the rest is in the object loop; nothing in the port depends on the answer.

PR #26110, opened 2026-09-19
Group a spatial search marker with the place on the map, and fix its icon

Corwin-Kh:
Follow-up to the spatial search "show on map" work. Three things, all visible when a search result is shown on the map.

### 1. A city was listed twice

Tapping the marker of a found city offered two identical "Київ" rows in the multi-select: the place the map draws and our marker. The marker never merged with it, for two independent reasons:

- the marker had no OSM id of its own, so `BaseDetailsObject.overlapsWith()` had nothing to compare;
- the place drawn by the map arrives from `MapSelectionHelper.selectObjectsFromOpenGl()` as a ready-made `BaseDetailsObject`, and `getOsmId()` only understands `Amenity` and `MapObject` — a group's ids were invisible, so it could never merge with anything.

Now the marker carries the id of its address object and `overlapsWith()` compares the id sets of two groups. As a side effect the general "the same POI was collected by a layer and by the renderer" duplicate collapses as well.

Two guards came with it, both needed because the marker is a placeholder that is collected before the renderer and therefore merges first: a real POI category is no longer shadowed by the placeholder's "other" category, and an empty subtype is not appended (it used to turn `city` into `;city`, which then broke `getIconNameForAmenity()`).

### 2. The marker icon

`RenderingIcons.getResId()` only knows rendering icons (`mx_`, `h_`, small). Address icons are ordinary drawables, so it returned `null` — which was unboxed straight into the `int iconId` parameter of `PointImageUtils.getOrCreate()` (NPE), and on the OpenGL path fell back to `mx_special_marker`, so a found city showed up under a meaningless icon.

This is the same fix as #26094, which is already merged into master. **r5.4 does not have it**, so this PR carries it: expect a small conflict in `QuickSearchListItem` when r5.4 is merged into master — the two copies differ only in the CITY branch described below.

On top of it: a city now uses `mx_place_city` instead of the generic `ic_action_building2`, in the search list and on the map alike. The icon existed in the resources but was referenced nowhere; only TOWN had its own icon, which looks like an oversight.

### 3. Address drill-down

`CITY`, `VILLAGE` and `STREET` had been moved into the "open the context menu" branch unconditionally, which broke the Address tab: a city stopped listing its streets and a street stopped listing its houses. The guard from #26068 is back, together with the `selectSearchResult()` call and the `showResult` flag that `QuickSearchDialogFragment.searchFinished()` reads.

Also in here: the context-menu code for a search result moved from `ExplorePlacesFragment` into `ContextMenuLayer.showContextMenu(SearchResult)`, so the list and the map marker share one path, and `AmenitySearcher` no longer matches a city, street or building by name — that match used to return an unrelated POI, e.g. a stop named after the city.

### 4. Added after the review

- `RenderingIcons.getPointIconId()` renamed to `getResIdOrDefault()`, which says what it does (9d01159605).
- An address object keeps its `wikidata` when a detailed amenity is looked up; only the match by name stays off, so a city, street or building with a wikidata tag is still enriched, while a stop named after the city is not picked up any more (5a5dc68ddf).
- The multi-select row of an address marker read **"User defined"**. The marker is an `Amenity` with the "other" category and no subtype, so `AmenityMenuController` fell back to translating the category. It is a `SearchResultAmenity`, which already carries its `SearchResult`, so the type now comes from the same `QuickSearchListItem.getTypeName()` the result list uses — no extra field in the data model. Tapping the marker alone was always correct, because that menu opens for the `City` itself (ddbdcc63c2).

### AI disclaimer

Written with Claude Code (Opus 5), pair-programming style.

**Human:** found and reported every problem here, tested each iteration on a device, decided the approach at each fork, and is responsible for building and verifying the result. Two of the catches below are what kept this PR from shipping a regression.

**AI:** searched the history, wrote the code, ran a review of its own diff, and fixed what the review found. The AI did not build or run the app.

**Key prompts** (translated from Russian, condensed):

1. *"Find the commit where we added showing streets found by spatial search on the map."* → the relevant work turned out to be uncommitted, in a local stash.
2. *"Extract the icon code for address objects from that stash. I think we also fixed a crash there when `ic_action_street_name` was not found."* → became #26094; re-applied here because r5.4 lacks it.
3. *"So now all icons come from `getAddressIconName`. That is semantically wrong — not every spatial search result is an address."* → review catch; the method became `@Nullable` and address-only.
4. *"I search for a city, the marker lands where the city already is, and tapping it shows two cities in the multi-select — the city is not grouped into one object."* → the report that drove the main fix.
5. *"Review the current changes, I think there is a regression."* → the review found that the first attempt handed the raw `City` to the selection, where `addObjectNoCombine()` swallows it without storing it; the city silently disappeared whenever more than one object was under the finger. Reworked to pass the marker amenity instead.
6. *"Why is the icon a building? Is there no icon for a city?"* → `mx_place_city`.


OsmAnd-java/src/main/java/net/osmand/data/BaseDetailsObject.java
```diff
@@ -159,6 +159,11 @@ private Long getOsmId(Object object) {
 	}
 
 	public boolean overlapsWith(Object object) {
+		if (object instanceof BaseDetailsObject detailsObject) {
```
alex-dev-neo, line 162: Needs to be checked by backend team

OsmAnd-java/src/main/java/net/osmand/search/AmenitySearcher.java
```diff
@@ -75,6 +78,12 @@ public Request(MapObject mapObject) {
                 names = stop.getOtherNames();
                 wikidata = null;
                 names.add(stop.getName());
+            } else if (mapObject instanceof City || mapObject instanceof Street
```
alex-dev-neo, line 81: Needs to be checked by backend team

OsmAnd/src/net/osmand/plus/render/RenderingIcons.java
```diff
@@ -260,6 +262,15 @@ public static String getBigIconName(@NonNull Integer iconId) {
 		return null;
 	}
 
+	@DrawableRes
+	public static int getPointIconId(@NonNull Context ctx, @Nullable String id, @DrawableRes int defIconId) {
```
alex-dev-neo: Why `getPointIconId` name and not `getResId`?

OsmAnd-java/src/main/java/net/osmand/search/AmenitySearcher.java
```diff
@@ -75,6 +78,12 @@ public Request(MapObject mapObject) {
...
+                    || mapObject instanceof Building) {
+                // matching by name alone picks up an unrelated POI, e.g. a stop named after the city
+                latLon = mapObject.getLocation();
+                wikidata = null;
+                names = Collections.emptyList();
```
ivanPyrohivskyi: Redundant block. After get by name [always filter by osmId](https://github.com/osmandapp/OsmAnd/blob/1d52658a50610f6e34e05fc19799735f422b052b/OsmAnd-java/src/main/java/net/osmand/search/AmenitySearcher.java#L330)

OsmAnd-java/src/main/java/net/osmand/data/BaseDetailsObject.java
ivanPyrohivskyi: Here is Ok

ivanPyrohivskyi (approved):

alex-dev-neo (approved):

PR #26113, opened 2026-09-19
Copy the travel model and the track optimizer into OsmAnd-shared

alex-dev-neo:
Sixth step of the obf reader port for [OsmAnd-Issues#2935](https://github.com/osmandapp/OsmAnd-Issues/issues/2935).

**Stacked on #26112**, itself on #26111, #26109, #26108 and #26107.

### What is in it

`net.osmand.shared.travel` is a new package:

- `TravelArticle` - one wikivoyage article, and the base of anything the travel screens show;
- `TravelGpx` - a track stored in an obf file: what the poi section said about it, and what each of its points is written into a gpx as;
- `TravelArticleIdentifier` - what names an article across the files it may appear in, comparing coordinates with a tolerance because the same route read out of two files does not land on the same point;
- `WikivoyageOSMTags` - the tags an article point carries over into its gpx.

`net.osmand.shared.gpx.TravelObfGpxTrackOptimizer` joins a route back together after the indexer cut it up per region: the pieces overlap by a few points near each boundary and arrive in whatever order the files were read in.

Android is untouched; nothing outside OsmAnd-shared and the tests changes.

### Deliberate differences

- `Parcelable` is left out of the identifier: it is how android passes it between screens.
- `TravelArticle.getTravelBook` is left out; it turns a path into a display name through the app's directory layout.
- `DefaultColors` is a twelve entry table inside `TravelArticle` rather than the app's palette class, which also carries the names for the interface. The values are written into gpx files, so they are frozen.
- `wpt_extra_tags` is read with kotlinx.serialization rather than Gson. A value that is not a json object makes Gson throw and the whole gpx build fail; here the point simply comes back without its extra tags.

### Two things the tests found in the java original

Neither is changed here - a port that quietly fixes what it copies is a port nobody can review - but both are worth a look:

1. **`TravelGpx` never uses the English title.** It reads `amenity.getName("en")` into a local and then assigns `title = isEmpty(title) ? amenity.getName() : enTitle`. At that point `title` is always empty, so the English name is computed and thrown away.
2. **`TravelArticleIdentifier` equality and `hashCode` disagree.** Equality allows the two coordinates to be about a metre apart - which is the point of it - while the hash is over the raw doubles. Two identifiers that are equal can therefore hash differently, so the type is unsafe as a hash key. The copy behaves the same.

### Tests, and one gap

`TravelOptimizerCompatTest` hands the same track to the java optimizer and to the copy and compares the result point for point: pieces arriving out of order (A D C B E and two other orders), two pieces of one route sharing the points near a region boundary, an overlap of eleven points, an overlap displaced sideways, pieces that meet exactly, closed loops that must not be joined onto anything, pieces too far apart, the empty cases, and 400 random tracks whose pieces touch, overlap, reverse or miss each other. Where the answer can be reasoned about, the test also says what it should be rather than only that the two agree.

**`TravelGpx` and `TravelArticle` have no java-versus-copy test.** Their originals live in the android app module, which `OsmAnd-java`'s tests cannot see, so there is nothing to compare against from there; the app's own tests are instrumented and need a device. `TravelModelTest` in `commonTest` instead pins the behaviour against expectations read off the originals by hand - 16 cases over tag parsing, the radius letter, shortlink tiles, both kinds of gpx point, the file name, the image url and the identifier - and it runs on the jvm and on Kotlin/Native. If the team would rather have a real comparison, the way to get one is to move these two classes into OsmAnd-java, and I am happy to do that as a follow-up.

Also green: `:OsmAnd-java:test`, `:OsmAnd-shared:jvmTest`, `:OsmAnd-shared:iosSimulatorArm64Test`, `:OsmAnd-shared:compileDebugKotlinAndroid`, `:OsmAnd-shared:linkDebugFrameworkIosSimulatorArm64` (no ObjC export name collisions), `:OsmAnd:compileAndroidFullLegacyArm64DebugKotlin`.


alex-dev-neo:
Added benchmarks for the two things here that run once per track opened and grow with its length.

A route of 20 000 points in 8 pieces, which is about what the 736 km Cesta hrdinov SNP comes to:

| what | java (jvm) | copy (jvm) | copy (native) |
| --- | --- | --- | --- |
| join the pieces of a route | 2.3 ms | 2.9 ms | 6.4 ms |
| build the points of a track | - | 6.5 - 10.5 ms | 8.7 ms |

Fifteen milliseconds on Kotlin/Native for both together is nothing against the second the issue allows for opening a route; that second goes on reading the files.

Two notes on reading the table. Building the points has no java column because its original is `TravelGpx` in the android app, which the java test module cannot see - the same gap as for the tests. And its jvm figure swings by half from run to run: each round allocates twenty thousand points with their tag maps and the garbage collector lands inside the measured window, while Kotlin/Native holds steady, which is why it ends up the faster of the two there. I left the range in rather than picking the flattering end of it.

Both benchmarks stay out of a normal run, as in the steps before.

PR #26114, opened 2026-09-19
Copy the travel helper and the amenity repository into OsmAnd-shared

alex-dev-neo:
Last step of the obf reader port for [OsmAnd-Issues#2935](https://github.com/osmandapp/OsmAnd-Issues/issues/2935): `TravelObfHelper`, the gpx builder behind it, and the `AmenityIndexRepository` both search through. With this, iOS can read `Travel.obf` through OsmAnd-shared instead of OsmAnd-core. **Android keeps its own classes and is not switched over.**

Based on #26113 (A6). Review that one first.

## What moved

| shared | copied from |
|---|---|
| `binary/AmenityIndexRepository` | `net.osmand.search.core.AmenityIndexRepository` (OsmAnd-java) |
| `binary/BinaryAmenityIndexRepository` | `AmenityIndexRepositoryBinary` (app) |
| `binary/ObfReaderSupplier` | new: where the repository gets its reader from |
| `travel/TravelObfHelper` | `TravelObfHelper` (app) |
| `travel/TravelObfGpxBuilder` | `TravelObfGpxFileReader` (app) |
| `travel/PopularArticles`, `travel/WikivoyageSearchResult` | same names (app) |
| `travel/TravelObfContext`, `travel/TravelBookmarks` | new: the platform hooks |

## What stayed behind

Deciding **which obf files to search**, in what order, and which the user has switched off depends on downloads, live updates and map settings; it stays in `ResourceManager` behind `TravelObfContext`. **Saved articles** stay in `TravelLocalDataHelper`'s sqlite behind `TravelBookmarks`, which has a `NONE` for a platform with no storage yet.

Not ported: the `AsyncTask` and its progress dialog (the builder is a plain call, the caller picks the thread and the `GpxReadCallback` is gone), `acquireGpxFileHeightData`, which approximates altitudes against the routing graph, `openTrackMenu`, which is android UI, and the poi-category cache the app keeps per file — its computation, `calculateDeltaSubcategories`, is here, only the storage is not.

Two places differ on purpose, both marked in the code:

- **`search()` builds its matcher from `KCollatorStringMatcher`**, not from a `SearchPhrase`, because the search core is not in shared yet. It treats the query as an unfinished word, which is what java does while the user is typing. The matcher is a parameter, so once the search core lands the caller passes the real one and nothing here changes.
- **Waypoints are ordered by a hand-written natural-order comparison.** Java uses the ICU collator in numeric mode, which no shared platform has. "Station 9" still comes before "Station 15"; two names the collator calls equal may come out the other way round.

## Tests

`TravelRepositoryCompatTest` runs the repository against the 26 obf files the tests ship with: **196 box searches, 92 name searches, 25 485 amenities compared**, plus `checkContains`, `isPoiSectionIntersects`, the subtype prefix search and a path search. The java side of each comparison is the reader call `AmenityIndexRepositoryBinary` makes, written out with the same arguments — the app class needs an `OsmandApplication` and cannot be reached from the java test module. That is the point of the test: the repository's methods take top/left/bottom/right and have to build requests out of left/right/top/bottom, and a wrapper that swapped a pair would still return amenities, only the wrong ones.

`TravelObfHelperTest` (19) and `TravelObfGpxBuilderTest` (13) run on the jvm and on Kotlin/Native. **No obf the tests ship with holds travel data**, so their input is built by hand through a `FakeAmenityIndexRepository`. As with A6, there is no java-versus-copy test for the helper or the builder — the originals are in the app module.

One thing worth a look, copied as it stands: **`loadPopularArticlesForLang` stops one short of the end** (`foundAmenitiesIndex < foundAmenities.size() - 1`), so the last amenity found waits for the next widening of the search. Against a real file hundreds arrive at a time and nobody notices; `theLastAmenityFoundIsLeftForTheNextPass` pins the behaviour.

## Speed

Second commit adds a row to the travel benchmark — the whole build, from what the sections hand over to a finished gpx file:

```
what                          java (jvm)   copy (jvm)   copy (native)
join the pieces of a route       2.3 ms       2.9 ms          6.4 ms
build the points of a track         -      6.5 - 10.5 ms      8.7 ms
build the gpx file of a route       -       4.4 - 5.6 ms      7.9 ms
```

Eight milliseconds on Kotlin/Native for a 736 km route is nothing against the budget for opening one: that time goes on reading the files.

## Carried over from r5.4

Third commit brings [#26120](https://github.com/osmandapp/OsmAnd/pull/26120) across, since this copy has the same loop: `searchMapIndex` reads the encoding rules of a map section before it looks at the zoom and the box, so the first tap on a route read every installed file. `AmenityIndexRepository` gains `isMapSectionIntersects`, answered from the roots that are read when the file is opened, and the helper skips the files a tap cannot be in.

The other half of the java fix does not apply here. Android asks its *shallow* reader because `BinaryMapReaderResource` opens a file handle per section, and the guard would otherwise open the file it means to skip; in shared `ObfReaderSupplier` hands out the one open reader.

`mapSectionIntersectsIsTheSame` puts the new method against `BinaryMapIndexReader.containsMapData` over the 26 obf files at three zooms, and `aFileTheTapIsNotInIsNotRead` pins that the helper does not ask a file that does not cover the tap.

## The corpus switch

Fourth commit is test infrastructure for the whole stack, put here rather than in #26111 so as not to rebase six branches. The obf files the tests ship with are small and were built years ago, and none carries poi tag groups or a top index, so those branches of the poi reader were never compared to java. `TestObf` now reads `OSMAND_OBF_CORPUS` - the files to read instead of the ones in `src/test/resources` - and `OSMAND_TEST_HEAP` sets the heap that holding every object of a regional file from both readers needs. Neither changes a normal run.

```
OSMAND_OBF_CORPUS=/path/Ukraine_transcarpathia_europe.obf OSMAND_TEST_HEAP=8g \
  ./gradlew :OsmAnd-java:cleanTest :OsmAnd-java:test --tests "*CompatTest"
```

What it found is in [#26111](https://github.com/osmandapp/OsmAnd/pull/26111): everything passes, with 2 242 682 map objects, 100 445 roads and 90 223 amenities compared, 45 237 of the amenities carrying tag groups.

## The lock on iOS

`KLock` on iOS becomes an `NSRecursiveLock`. It stands in for a java monitor, and the copied code re-enters it - `searchTravelGpx` by route types calls `searchTravelGpx` by route ids - where java lets a thread re-enter a monitor it already holds. A plain `NSLock` deadlocks there instead, on the main thread, with no error; that is what the first build of the iOS side did.

It is not free: an uncontended lock/unlock pair costs 11.0 ns against 5.8 ns for `NSLock` (Foundation on arm64, 20M pairs). `KLock` is also what the routing code locks with, so the question is how often that one is taken. `GeneralRouter.RouteAttributeContext.evaluate` is behind the result cache in `GeneralRouter`, which answers per road-type set rather than per segment, so it is a cache-miss path: **187 312** calls for the whole of `RoutePlannerCompatTest`, every route of every profile in it. At +5.2 ns that is under a millisecond for the entire suite. The other holders - the amenity repository, the travel helper, the places database - lock once per search call.

## Verified

`:OsmAnd-java:test`, `:OsmAnd-shared:jvmTest`, `:OsmAnd-shared:iosSimulatorArm64Test` (the 9 failures there are the same 9 as on the base branch), `:OsmAnd-shared:compileDebugKotlinAndroid`, `:OsmAnd-shared:linkDebugFrameworkIosSimulatorArm64` (no ObjC export name collisions), `:OsmAnd:compileAndroidFullLegacyArm64DebugKotlin`.


alex-dev-neo:
Benchmark numbers, for the record. Both sides measure the same code over a route of 20 000 points cut into 8 pieces, with 500 waypoints; `join` and `build the points` are the two parts on their own at full length, `build the gpx file` is the whole thing at the mix a real route has.

```
what                          java (jvm)   copy (jvm)   copy (native)
join the pieces of a route       2.3 ms       2.9 ms          6.4 ms
build the points of a track         -      6.5 - 10.5 ms      8.7 ms
build the gpx file of a route       -       4.4 - 5.6 ms      7.9 ms
```

The last two have no java column: `TravelGpx` and `TravelObfGpxFileReader` are in the app module, which the java test module cannot see.

To reproduce, from `android`:

```
OSMAND_TRAVEL_BENCHMARK=1 ./gradlew --no-daemon :OsmAnd-shared:jvmTest --tests "*TravelBenchmarkTest" -i
```

```
./gradlew :OsmAnd-shared:linkReleaseTestIosSimulatorArm64
cd OsmAnd-shared && SIMCTL_CHILD_OSMAND_TRAVEL_BENCHMARK=1 xcrun simctl spawn --standalone <udid> \
  build/bin/iosSimulatorArm64/releaseTest/test.kexe --ktest_filter='net.osmand.shared.travel.TravelBenchmarkTest.*'
```

Native is measured on the release binary — the debug one says nothing useful about speed.

alex-dev-neo:
Two more commits, both for the iOS side that builds on this branch.

**`6ddc13fbfd`** — a Kotlin default argument exports only the full form to Objective-C, so `search()` and `readGpxFile()` could not be called from Swift without building a `KStringMatcher` or a `Cancellable` by hand. Both are now explicit overloads. No behaviour change.

**`f6be99f5ca`** — `KAlgorithms.simplifyFileName` / `compareFileVersions`, copied from the private `Algorithms.simplifyFileName` and `Algorithms.getStringVersionComparator`. iOS has to build its repository list in the same order android consults its obf files in, and reimplementing that in Swift would be a second place to get it wrong. `FileVersionOrderCompatTest` sorts real and generated names with both comparators and compares every pair both ways round.

PR #26128, opened 2026-09-20
Crash report: say where the resident and native memory went

vshcherb:
A report could say a process held 1.1 GB and not say what of. Two additions, both from data the report already collects.

### Where the resident memory went

`summary.native-heap` is one number covering the allocator, the mapped libraries and everything the core allocated. `/proc/self/smaps` names every mapping, including the anonymous ones Android labels, so the sample now carries resident megabytes per category:

```
rss=582 nat=148/716/564
smaps=dalvik:154,native:229,gpu:1,so:63,code:80,font:4,stack:1,other:29
```

Read together, those two say the allocator has mapped 716 MB, holds 229 MB resident and has 148 MB handed out — retention that no `Debug` counter separates, and the question behind "the native heap plateau keeps rising".

Cost, measured on a Pixel 9 Pro XL where smaps is 2.6 MB and 88k lines:

| | first walk | later walks |
|---|---|---|
| total | 269 ms | 60-100 ms |
| kernel (`read`) | 23 ms | 38-65 ms |
| parsing | 246 ms | 20-35 ms |

The first one runs interpreted, later ones settle at about what `Debug.getMemoryInfo()` already costs (median 68 ms on this device). So it is scanned as bytes, only two line kinds out of the twenty-five a mapping prints are looked at, it runs on the sampling thread once every ten summaries, and it switches itself off if it ever costs more than 600 ms.

Graphics mostly does not appear there — EGL and GL buffers are counted by mtrack rather than mapped into the process — so `summary.graphics` stays the source for that, and the `gpu` category is near zero on this device by design rather than by bug.

**Only categories are written, never the names of mapped files**: those are the maps somebody downloaded, which is where they live and where they travel.

### Which classes hold native memory

Every object owning a native buffer registers it with `NativeAllocationRegistry`, which records what one allocation costs. The histogram reader now picks up the registries and their thunks on the pass it already makes for array owners, and names a registry after the class whose static field holds it:

```
--- native registered 41077K in 7963 allocations, 121 registries
40207   82      unattributed
327     3387    android.graphics.Paint
161     1674    android.graphics.RenderNode
78      807     android.graphics.Path
```

41 MB of a 157 MB native heap. The useful part of that is the negative: most native memory is raw C++ allocation that no Java object accounts for, so a Java-side leak hunt is the wrong place to look. The 82 unattributed allocations average about 490 KB each, which is the shape of tile bitmaps.

Field names are read from the dump, so a platform that renamed them leaves the section out instead of reporting invented numbers. Cost on a 138 MB dump: 797 ms against 821 ms without the change.

### Testing

Both were built and run on a Pixel 9 Pro XL (Android 16, SDK 37) against `net.osmand.dev`. The smaps numbers above come from that device's `memory_log.txt`; the histogram numbers come from a dump taken with "Collect & analyze now" on it (`dump=80MB dumpMs=1998 analyzeMs=1211`), and the reader was also run on two earlier dumps from the same app on a desktop JVM under `-Xmx160m`.

Two bugs in the first version of the smaps walk were found only on the device and are fixed here: the mapping name was taken as the last field, which breaks on `[anon:dalvik-main space]` because it contains a space, and unnamed mappings then had their inode read as a name. Both sent the whole Java heap into `other`.

## AI disclaimer

Prompts used (summarised):
1. On release phones, can we collect anything that finds where native memory and RSS go, the way heapprofd did in the lab?
2. Do it — start with smaps in the memory log, it is the cheapest, and verify it on the connected device.
3. Then the `NativeAllocationRegistry` attribution from the heap dump; one PR for both.

Decided by the agent, not requested explicitly: the category list and that only categories are written; scanning smaps as bytes after measuring that the framework's own walk costs a third as much; the once-every-ten-summaries cadence and the 600 ms self-disable; naming a registry after the class whose static holds it and stripping the `$NoImagePreloadHolder` wrapper; leaving registries held in local variables unattributed rather than adding a third pass for them.


OsmAnd/src/net/osmand/plus/feedback/HeapHistogram.java
```diff
@@ -232,22 +248,120 @@ private void readClassDump(@NonNull Reader in, boolean secondPass) throws IOExce
...
+		if (!isRegistry && !isThunk || size > Integer.MAX_VALUE) {
+			return false;
+		}
+		int length = (int) size;
+		if (instance.length < length) {
+			instance = new byte[length];
+		}
+		in.readFully(instance, length);
+		if (isRegistry && registrySizeOffset + 8 <= length) {
+			entryFor(objectId)[0] = readLong(instance, registrySizeOffset);
```
alex-dev-neo: **`size` is read without masking `IS_MALLOCED`, so the per-class rows are AOSP's "unknown size" sentinel, not measured bytes.**

`libcore.util.NativeAllocationRegistry` documents the field as: *"The least significant bit is one iff the object was allocated primarily with system malloc(). This field is examined by ahat and other tools."* And `createMalloced(classLoader, freeFunction)` — the two-arg form, for allocations whose size is not known — stores `DEFAULT_SIZE = 98`, commented in AOSP as *"Assumed size for malloced objects that don't specify a size"*. Combined, the field holds `99`.

The sample output in the PR description shows this is what is being reported:

| row | bytes / allocations |
|---|---|
| `327  3387  android.graphics.Paint` | 334848 / 3387 = **99** |
| `161  1674  android.graphics.RenderNode` | 164864 / 1674 = **99** |
| `78   807   android.graphics.Path` | 79893 / 807 = **99** |

All three named classes are placeholder registries; the 566 KB attributed to them is `98|1` multiplied out. The negative result the PR describes ("most native memory is raw C++ allocation that no Java object accounts for") still holds, but the rows that name classes do not mean what they read as.

Two further consequences of the unmasked bit:
- `new NativeAllocationRegistry(cl, f, 0)` stores `0|1 = 1`, so those allocations pass the `bytes <= 0` filter and join the `count` total at 1 byte each.
- A non-malloced registry with a genuine size of 0 is dropped by that same filter and its allocations are not counted at all.

So one "7963 allocations" figure mixes measured, sentinel and malloc-unknown allocations. Suggest masking bit 0 before use, and reporting a masked `98` on a malloced registry as unknown rather than as bytes.

OsmAnd/src/net/osmand/plus/feedback/MemoryLog.java
```diff
@@ -437,6 +461,158 @@ private static void appendProcessSummary(@NonNull StringBuilder sb) {
...
+			}
+			while (i < length && header.charAt(i) == ' ') {
+				i++;
+			}
+		}
+		String name = i < length ? header.substring(i) : "";
+		if (name.isEmpty()) {
+			return SMAPS_ANON;
+		}
+		if (name.startsWith("[anon:dalvik") || name.startsWith("[anon:Zygote")) {
```
alex-dev-neo, line 590: **On API 24-28 this rule misses the Java heap entirely and line 596 books it as `gpu`.**

`[anon:dalvik-…]` naming needs the kernel's anon-VMA naming, which arrives with Android 9/10. `osmand_minSdk` is 24 (`versions.gradle:2`), and on Android 7.0-8.1 the ART spaces appear in smaps under their ashmem names:

```
/dev/ashmem/dalvik-main space (region space)
/dev/ashmem/dalvik-large object space
/dev/ashmem/dalvik-zygote space
/dev/ashmem/dalvik-LinearAlloc
/dev/ashmem/dalvik-jit-code-cache
```

None starts with `[anon:dalvik`, so the ladder falls through to `name.startsWith("/dev/ashmem")` on line 596 and returns `SMAPS_GPU`. A 200 MB Java heap on such a device is written as `gpu:200` with `dalvik` absent — the same shape of bug the description says was already found twice on the device ("Both sent the whole Java heap into `other`"), for the older half of the supported range.

The same devices have no `[anon:…]` naming for malloc either, so `native` reads 0 and the allocator lands in `anon`. The line is at its least trustworthy on the platform where an OOM is most likely.

alex-dev-neo (commented):
Reviewed the two files at the head commit. Three of these would change what a report is read to mean or add a new stall; the rest are smaller.

Worth fixing before merge:
- **The per-class native attribution is sentinel arithmetic.** All three classes named in the description's sample output work out to exactly 99 bytes per allocation — that is AOSP's `DEFAULT_SIZE = 98` with the `IS_MALLOCED` bit, i.e. "size not known", reported as measured bytes.
- **The smaps categoriser misses the Java heap on API 24-28** and books it as `gpu`, via the `/dev/ashmem` clause. `minSdk` is 24.
- **The walk runs inside the monitor that main-thread `onTrimMemory` takes**, ungated and with its budget checked only after the walk returns.

The other nine are a misclassification each for gralloc libraries and `/memfd:` mappings, a wasted second pass, a String per mapping, a failure path that never latches off, and three cleanups.

The two features themselves read well, and the privacy line — categories only, never mapped file names — holds in the code.

OsmAnd/src/net/osmand/plus/feedback/MemoryLog.java
```diff
@@ -437,6 +461,158 @@ private static void appendProcessSummary(@NonNull StringBuilder sb) {
...
+	 * <p>
+	 * Only the categories are written, never the names of the mapped files: those are the maps
+	 * somebody downloaded, which is where they live and where they travel.
+	 */
+	private static void appendSmaps(@NonNull StringBuilder sb) {
+		if (!smapsAffordable || !summaryDue || smapsCount++ % SMAPS_EVERY != 0) {
+			return;
+		}
+		long start = SystemClock.uptimeMillis();
+		String categories = readSmapsByCategory();
```
alex-dev-neo, line 484: **This walk holds the monitor that main-thread `onTrimMemory` blocks on.**

`sample()` is `static synchronized`, and so is `onTrimMemory(int)`, which `OsmandApplication.onTrimMemory` (`OsmandApplication.java:577`) calls on the main thread. Everything `buildSample` does is inside that monitor.

Before this PR the longest hold an ordinary user could reach was `Debug.getMemoryInfo` under `SUMMARY_BUDGET_MS = 200` — `maybeCollectHistogram` is gated behind the development plugin's `AUTO_HEAP_HISTOGRAM`, so the multi-second path never runs for them. `appendSmaps` is ungated, its budget is 600 ms, and the budget is checked *after* the walk returns:

- on the 1.1 GB process this feature exists to explain, the first walk is the slow one — tens of thousands of mappings, page tables under `mmap_sem`, while the kernel is reclaiming;
- `smapsAffordable` only goes false once that walk has already finished;
- a `TRIM_MEMORY_RUNNING_CRITICAL` arriving meanwhile blocks the main thread for its full duration.

Reading smaps outside the monitor, or checking elapsed time inside the read loop and abandoning the walk rather than after it, decouples the two.

OsmAnd/src/net/osmand/plus/feedback/HeapHistogram.java
```diff
@@ -232,22 +248,120 @@ private void readClassDump(@NonNull Reader in, boolean secondPass) throws IOExce
 		int statics = in.readShort() & 0xffff;
 		for (int i = 0; i < statics; i++) {
 			in.skip(idSize);
-			in.skip(valueSize(in.readByte() & 0xff));
+			int type = in.readByte() & 0xff;
+			if (secondPass && type == 2) {
+				// a registry is usually a static of the class whose objects it accounts for
+				rememberRegistryOwner(in.readId(idSize), classId);
```
alex-dev-neo: **This records every non-null object static in the dump, not only registries.**

The branch fires for each static of type 2 in every class dump, and `registryOwners` keeps them all. Boot classpath plus app classes contribute on the order of 10⁴-10⁵ non-null object statics; each entry is a `HashMap` node plus a boxed `Long`, roughly 50-60 bytes, so several MB. `appendNativeAllocations` then reads about 121 of those keys.

Where that lands matters: `maybeCollectHistogram` (`MemoryLog.java:690`) only calls `HeapDump.collect` once `used >= min(350 MB, 0.8 × maxMemory)`, and `HeapHistogram.analyze` runs in that same process. The map is built precisely when the heap has the least room, and the class javadoc's "Only class names and counters are kept, a few tens of thousands of small entries" stops being true.

The ordering problem is real — a registry instance can be read after the class dump that holds it — but it can be bounded: prune against `registries` periodically, or resolve owners from a class-dump-only pass once `registries` is complete.

OsmAnd/src/net/osmand/plus/feedback/MemoryLog.java
```diff
@@ -437,6 +461,158 @@ private static void appendProcessSummary(@NonNull StringBuilder sb) {
...
+				|| name.startsWith("[anon:GWP-ASan") || name.startsWith("[anon:bionic")) {
+			return SMAPS_NATIVE;
+		}
+		if (name.startsWith("[stack") || name.startsWith("[anon:stack_and_tls")
+				|| name.startsWith("[anon:thread")) {
+			return SMAPS_STACK;
+		}
+		if (name.startsWith("/dev/kgsl") || name.startsWith("/dev/mali") || name.startsWith("/dev/dri")
+				|| name.startsWith("/dev/dma_heap") || name.startsWith("/dmabuf")
+				|| name.contains("gralloc") || name.startsWith("/dev/ashmem")) {
```
alex-dev-neo, line 603: **`contains("gralloc")` is a substring test on the whole path, and it runs before the `.so` rule.**

Most devices map gralloc HAL libraries into every app process:

```
/system/lib64/libgralloctypes.so
/vendor/lib64/hw/gralloc.<board>.so
/vendor/lib64/android.hardware.graphics.mapper@4.0-impl.so
```

`libgralloctypes.so` contains `gralloc`, so it returns `SMAPS_GPU` here and never reaches `name.endsWith(".so")` on line 599. Library text and data — exactly what the `so` bucket is for — is reported as GPU memory, in a section whose stated expectation is that `gpu` sits near zero because graphics buffers are not mapped into the process.

Anchoring the rule to the buffer paths it means (`/dev/…gralloc`, `/dmabuf`) or moving it below the `.so` test fixes it.

OsmAnd/src/net/osmand/plus/feedback/HeapHistogram.java
```diff
@@ -77,7 +90,7 @@ public static String analyze(@NonNull File hprof) throws IOException {
 		try (InputStream in = new BufferedInputStream(new FileInputStream(hprof), 256 * 1024)) {
 			histogram.read(new Reader(in), false);
 		}
-		if (!histogram.bigArrays.isEmpty()) {
+		if (!histogram.bigArrays.isEmpty() || histogram.thunkRegistryOffset >= 0) {
```
alex-dev-neo, line 95: **The guard should also require `registrySizeOffset >= 0`.**

Nothing is produced from the second pass unless *both* offsets were found. If a platform renames or drops `NativeAllocationRegistry.size` while `CleanerThunk.this$0` is still present — the graceful-degradation case this design is built around — then `registrySizeOffset` stays -1, every `isRegistry` test is false, all entries keep `[0] == 0`, `bytes <= 0` skips them, `count` stays 0 and `appendNativeAllocations` writes nothing.

On a dump with no big arrays, that is a full second read and parse of the file for no output: ~800 ms and 138 MB of I/O on the numbers in the description.

```java
if (!histogram.bigArrays.isEmpty()
        || (histogram.thunkRegistryOffset >= 0 && histogram.registrySizeOffset >= 0)) {
```

OsmAnd/src/net/osmand/plus/feedback/MemoryLog.java
```diff
@@ -437,6 +461,158 @@ private static void appendProcessSummary(@NonNull StringBuilder sb) {
...
+						i++;
+						continue;
+					}
+					// only two lines out of the twenty-five a mapping prints are worth looking at,
+					// and the rest are told apart by their first byte alone
+					byte first = buffer[lineStart];
+					if (first == 'R' && lineStart + 1 < i && buffer[lineStart + 1] == 's') {
+						totals[category] += parseKb(buffer, lineStart + 4, i);
+					} else if (first >= '0' && first <= '9' || first >= 'a' && first <= 'f') {
+						category = categoryOf(new String(buffer, lineStart, i - lineStart,
```
alex-dev-neo, line 525: **This decodes a String per mapping, which is the cost the method's javadoc gives as the reason for scanning bytes.**

`readSmapsByCategory` opens with *"it is scanned as bytes: decoding it into strings costs several times more than the kernel spends producing it"* — and then decodes the single longest line of each mapping, the one carrying the full path, plus one `substring` per mapping inside `categoryOf`.

On the Pixel 9 Pro XL figures in the description (88k lines ≈ 3.5k mappings) that is ~7k short-lived Strings and a few hundred KB of garbage per walk, in a process the sampler only walks because memory is short. It is most of the 20-35 ms steady-state parse cost in the table.

Every test in `categoryOf` is a prefix or suffix comparison, so it can run against `buffer[lineStart..i)` directly with no String at all.

OsmAnd/src/net/osmand/plus/feedback/MemoryLog.java
```diff
@@ -437,6 +461,158 @@ private static void appendProcessSummary(@NonNull StringBuilder sb) {
...
+				|| name.endsWith(".art")) {
+			return SMAPS_CODE;
+		}
+		if (name.endsWith(IndexConstants.BINARY_MAP_INDEX_EXT)) {
+			return SMAPS_OBF;
+		}
+		if (name.endsWith(".ttf") || name.endsWith(".otf")) {
+			return SMAPS_FONT;
+		}
+		return SMAPS_OTHER;
```
alex-dev-neo, line 620: **`/memfd:` mappings have no rule, so they all land here.**

`/dev/ashmem` is the old name; on Android 10+ the same things are memfd-backed:

- `/memfd:jit-cache (deleted)` and `/memfd:jit-zygote-cache (deleted)` — the ART JIT code cache, tens of MB on a long-running map session;
- `/memfd:dmabuf` and other `/memfd:` names for shared graphics and IPC buffers.

Both fall through to `SMAPS_OTHER`, while `code` — the bucket meant to hold executable mappings — under-reports by the JIT share. Some of the `other:29` in the sample line is this.

OsmAnd/src/net/osmand/plus/feedback/MemoryLog.java
```diff
@@ -437,6 +461,158 @@ private static void appendProcessSummary(@NonNull StringBuilder sb) {
...
+			return;
+		}
+		long start = SystemClock.uptimeMillis();
+		String categories = readSmapsByCategory();
+		long spent = SystemClock.uptimeMillis() - start;
+		if (categories != null) {
+			sb.append(" smaps=").append(categories);
+			sb.append(" smapsMs=").append(spent);
+		}
+		if (spent > SMAPS_BUDGET_MS) {
```
alex-dev-neo, line 490: **A failing read never latches the feature off.**

`readSmapsByCategory` catches `IOException | RuntimeException`, calls `log.error(e)` and returns null. A failing open is fast, so `spent > SMAPS_BUDGET_MS` is false and `smapsAffordable` stays true.

On any device or ROM where the read is refused, that is a full stack trace in the log every twentieth sample for the rest of the process's life, and the reader is never given up on — unlike `summaryAffordable`, which at least latches. Worth telling failure apart from cost and setting `smapsAffordable = false` after the first failed read.

OsmAnd/src/net/osmand/plus/feedback/HeapHistogram.java
```diff
@@ -377,6 +491,49 @@ private String className(long classId) {
...
+		for (Map.Entry<Long, long[]> entry : registries.entrySet()) {
+			long[] registry = entry.getValue();
+			long bytes = registry[0] * registry[1];
+			if (bytes <= 0) {
+				continue;
+			}
+			total += bytes;
+			count += registry[1];
+			String owner = registryOwners.get(entry.getKey());
+			long[] sum = byOwner.get(owner != null ? owner : "unattributed");
```
alex-dev-neo: Minor: the ternary is evaluated twice for the same entry (here and on line 513) — a change to the fallback label has to be made in both places.

Also, `count`, `attributeArrays`, `entryFor` and now this method each spell out the same get / null-check / `new long[2]` / put sequence. Hoisting the key into a local and reusing one helper covers all four sites.

OsmAnd/src/net/osmand/plus/feedback/HeapHistogram.java
```diff
@@ -232,22 +248,120 @@ private void readClassDump(@NonNull Reader in, boolean secondPass) throws IOExce
...
+			return;
+		}
+		// the registry is a static of a holder class that exists only to delay class loading;
+		// the class people would recognise is the one it is nested in
+		String name = className(holderClassId);
+		int holder = name.indexOf("$NoImagePreloadHolder");
+		registryOwners.put(candidate, holder > 0 ? name.substring(0, holder) : name);
+	}
+
+	private long readLong(@NonNull byte[] bytes, int offset) {
```
alex-dev-neo, line 349: Minor: this is `readId(byte[], int)` on line 405 with the loop bound fixed at 8 instead of `idSize`, and `Reader.readId`'s `idSize == 8` branch is a third copy of the same accumulate loop.

One `readValue(byte[] bytes, int offset, int width)` covers the two array variants — `readId` passing `idSize`, `readLong` passing 8.

OsmAnd/src/net/osmand/plus/feedback/MemoryLog.java
```diff
@@ -437,6 +461,158 @@ private static void appendProcessSummary(@NonNull StringBuilder sb) {
...
 
+	/**
+	 * The Debug counters say how much, /proc/self/smaps says where: every mapping is named, and
+	 * Android names the anonymous ones too, so "native" stops being one number and becomes scudo,
+	 * the GPU driver, the mapped libraries and the maps a person opened.
+	 * <p>
+	 * Only the categories are written, never the names of the mapped files: those are the maps
+	 * somebody downloaded, which is where they live and where they travel.
+	 */
+	private static void appendSmaps(@NonNull StringBuilder sb) {
```
alex-dev-neo, line 479: Worth weighing: `appendSmaps` only runs when `summaryDue`, which is exactly when `appendProcessSummary` has already called `Debug.getMemoryInfo(info)` — itself a full smaps walk inside the framework — and emitted `summary.code`, `summary.stack` and `summary.graphics`.

The sample line then carries both `code=… stk=… gfx=…` and `smaps=…code:80,stack:1,gpu:1`, differing only in PSS vs RSS, and the process walks its own page tables twice in a row.

What is genuinely new is the dalvik / native / so / obf / anon / other split — and that is the argument the description makes. Dropping the three overlapping categories would make the new information stand out and halve the kernel cost of a due sample.

OsmAnd/src/net/osmand/plus/feedback/HeapHistogram.java
```diff
@@ -239,15 +257,103 @@ private void readClassDump(@NonNull Reader in, boolean secondPass) throws IOExce
...
+			if (registries.size() >= MAX_REGISTRIES) {
+				return null;
+			}
+			entry = new long[2];
+			registries.put(registryId, entry);
+		}
+		return entry;
+	}
+
+	/**
```
alex-dev-neo, line 344: **Leftover javadoc: this documents the deleted `rememberRegistryOwner`, not `readLong`.**

`rememberRegistryOwner` went away with `registryOwners`, but its comment block stayed behind and now sits directly above `readLong(byte[], int)` — so a big-endian accumulate loop is documented as "Remembers which class holds a registry. A thunk points at its own registry and would name itself…".

Just delete the block; `readLong` needs no javadoc.

OsmAnd/src/net/osmand/plus/feedback/HeapHistogram.java
```diff
@@ -239,15 +257,103 @@ private void readClassDump(@NonNull Reader in, boolean secondPass) throws IOExce
...
+			}
+		}
+		return true;
+	}
+
+	@Nullable
+	private long[] entryFor(long registryId) {
+		long[] entry = registries.get(registryId);
+		if (entry == null) {
+			if (registries.size() >= MAX_REGISTRIES) {
```
alex-dev-neo, line 335: **The cap truncates silently — a capped run and a complete one look identical in the report.**

Past `MAX_REGISTRIES` this returns null, so later registries get no size and later thunks are not counted, but nothing downstream knows: `appendNativeAllocations` still prints

```
--- native registered <total>K in <count> allocations, <registries.size()> registries
```

with `registries.size()` pinned at 4096 and the totals presented as whole. On the platform this valve exists for — one where the field layout is read wrong and arbitrary ids come out as registries — that is exactly the case where the numbers are meaningless, and the report gives no way to tell.

A boolean set where the cap fires, and a word on the line when it did (`… 4096 registries, truncated`), is enough to keep the section honest.

alex-dev-neo (commented):
Re-read at `819d8318fc`. The two lock-free rewrites hold up: `onTrimMemory`'s packed `AtomicLong` round-trips correctly — the `+1` / `-1` on the level preserves the old `trimLevel = -1` semantics, and the count can only be read together with the level it belongs to — and `sample` is now the only `static synchronized` method, reached only from the diagnostic thread. Dropping `registryOwners` and the static-field read takes the second pass back to allocating nothing per class.

Two small things the rewrite left behind, inline.

Side note, not code: the description still documents the "Which classes hold native memory" section with the Paint / RenderNode / Path table, which `7321e7f0c4` replaced with power-of-two size buckets. Worth updating before merge — and the three rows in that table were exactly the placeholder registries (99 bytes each = AOSP's `DEFAULT_SIZE | IS_MALLOCED`), so the new bucketing reads better anyway, though `size` is still used unmasked.

PR #26130, opened 2026-09-20
Reduce the heap a GPX point takes: 670 -> 163 bytes

vshcherb:
Three commits, all about what one shown GPX point costs in Java heap.

**Keep speed out of the point extensions map.** A parsed point kept speed twice - in `WptPt.speed`
and as a string in its extensions map - and `writeWpt()` pushed speed and heading back into that map
on every save, together with a deferred copy and an extensions writer per point. The writer
regenerates `<speed>` from the field anyway. `VehicleTrackPointsAnalyser` also read a missing OBD tag
as `0f` instead of `NaN`, so a `VehicleData` was allocated for every point of any track that carries
extensions; iOS already defaults it to NaN.

**Keep point extensions in one flat array.** They lived in a `LinkedHashMap` - about 40 bytes per
entry plus the table - and every value was its own `String`, although a recording repeats the same
handful of them (`gps`, `1.4`, `171.6`). A point now keeps one array of `[key, value, key, value …]`
sorted by key with a binary search over it; `GpxExtensionsMap` is only a view on that array. Tag
names are shared through one table and the parser keeps one instance of every distinct value per
file.

**The same in the deprecated Java parser**, which the server and the tools still use. It had the same
duplicated speed, and its writer was worse than the shared one: `getExtensionsToRead()` handed out
the live map, so writing a file removed `profile`, `trkpt_idx` and the group colours from the points
in memory and left an extensions writer on every point.

Measured on a whole personal collection - 867 gpx files, 471 MB, 2.53M points in memory:

| | before | after |
|---|---|---|
| everything loaded | 1249 MB, 518 B/point | 384 MB, 159 B/point |
| the 444 recordings alone (410 MB) | 1133 MB, 679 B/point | 283 MB, 170 B/point |
| the same through the Java parser | 1148 MB, 688 B/point | 298 MB, 179 B/point |
| tracks without point extensions | 119 B/point | 118 B/point |
| parsing, 10 passes over 34 recordings | 2219 ms | 2075 ms |
| reading 5 tags on every point, 3 passes | 359 ms | 187 ms |

On the emulator (same debug build, one 100k point track shown on the map): Java heap 33 -> 102 MB
before, 35 -> 57 MB after, so **685 -> 217 bytes per displayed point**.

**No regressions.** Every one of the 867 files was dumped and compared before and after, through
both parsers: the same points (every field and every extension), the same `PointAttributes` for
every point, the same `GpxTrackAnalysis`, and the file written back **byte for byte identical**
(only the order of the tags inside `<extensions>` can differ, and it was never stable - the writer
puts them into a plain `HashMap`). `bearing` and `heading` are deliberately left in the extensions
map: moving them into the fields saves another 15 bytes per point but reformats those two tags on
save (`251` becomes `251.0`).

Tests: `:OsmAnd-shared:jvmTest` 48 -> 62 green (15 new, including what a recording plugin attaches
to a point and what survives a save), `:OsmAnd-java:test` gains 3 (2 fail on the old writer, 1 on the
writer of this branch before the follow-up below), and
its 11 pre-existing environment failures are unchanged. Android and iOS targets compile, the app
builds and runs, and the Kotlin/Native tests fail exactly as they do on a clean checkout
(`xmlFactoryApi has not been initialized`, all the gpx-loading ones).

Not touched: `GpxFile.buildGeneralSegment` still clones every point of a multi-segment track
(`GpxTrackAnalysis` writes `point.distance` and caches `point.attributes` on the point), and `WptPt`
itself is 116 bytes.

**Review follow-up** - c38be4c040: the copied map went in *after* the field values, so a `speed`
string kept in the map (a non-positive or unparseable `<speed>`) overrode a speed set on the field
later, e.g. by `WebGpxParser.addSegmentsToTrack` on loaded points. The copy now goes in first and the
two fields override it, which is the old precedence exactly. Tests in both parsers fail on e3989d524f.

Known contract change, left as is: `GpxExtensionsMap.put()` carries Kotlin's parameter null check,
where the old `LinkedHashMap` stored a null value and the writer skipped it. Every Java `put`/`putAll`
site in the app passes non-null values; the Gson-decoded `extra_tags` maps in `TravelObfGpxFileReader`
are the one place that could only do so if the travel data carried a JSON `null`.

Verified on a Pixel 8 Pro (Android 17, `gplayFreeOpenglArm64Debug`) before the follow-up: a track
with every kind of extension loads, renders and analyses; a rewrite through a waypoint edit is
byte-identical to the r5.4 writer except that file-level `<extensions>` come out in key order instead
of file order; recording with the `plugins` writer, the in-memory current track written twice, and the
regular save all keep `provider`, `bearing`, `speed` and `heading`; Java heap 690 -> 219 bytes per
displayed point for a 100k-point track. After the follow-up: `:OsmAnd-shared:jvmTest` 61 green,
`GpxWriteTest` 3 green, the app builds.

Three smaller ones after a second pass:

- e052453c87: `equals`/`hashCode` for the entries of both extension maps, so `map.hashCode()` is not
  identity based.
- 0670ef7255: `copyExtensions()` copies the array once instead of a map copy plus an insert per key -
  `buildGeneralSegment` clones every point of a multi-segment track through it.
- d4e9bc11ce: `getExtension(key)` searches the array directly; the `WptPt` getters the map layer calls
  per point per frame no longer build a map view for it.

Checked once more on the Pixel 8 Pro with the whole branch: the GPX instrumented tests (7) pass; a
100k-point track in 4 segments opens, joins, splits into intervals and is written back complete, at
84 MB of Java heap where r5.4 reaches 180 MB and hangs with an ANR; route planning, favourites and
map markers write their files as before; no exception in the log.

## AI disclaimer

Prompts used (summarised):
1. How to fix this "a shown GPX point costs ~800 bytes of Java heap" analysis with very small changes in 5.4.
2. Check the crash reports for any hint that the problem is real.
3. Build a tag dictionary and keep two int arrays per point instead of a map - maybe a dictionary for the values too, then one array - with a binary search on get.
4. Can we get down to 40-50 bytes per point?
5. Here are real tracks to test with (a personal OsmAnd export).
6. Prepare a draft pull request against r5.4, everything in one, as separate commits, and take the smallest variant right away, with tests.
7. Update the Java code too, deprecated as it is; and why is `GPXExtensions` in Java not changed as well, for the tests and to keep the two in sync.
8. Count all the tracks of the export - where is it worst and best per point.
9. Measure what displaying 100k track points really costs on the emulator.
10. Write tests over those tracks that compare everything before and after (read, write, analysis), without committing them, so that the release has no regressions.
11. Analyse the pull request for issues, bugs and regressions; test and build on a device.
12. Fix the two findings with minimal changes, one commit per fix, in this pull request.
13. Fix the smaller findings locally, simplify them, keep the Java parser's original behaviour, then push what remained.
14. Re-check the whole branch on a device for regressions and crashes; drop the two commits that only covered hypothetical inputs.

Decided by the agent, not requested explicitly: splitting the work into three commits; keeping the
extensions entry when speed is not positive so the file round-trips exactly; dropping only the
extension writers that the writer itself created, and keeping the one a recording plugin put there
(the first version of this deleted it and the new test caught it); interning the tag names instead
of numbering them, after measuring that a reference and an int both cost 4 bytes and that the win
comes from one array instead of a hash table; a value pool scoped to one load call instead of a
global value dictionary; leaving `bearing`, `heading` and `buildGeneralSegment` alone; keeping the
comparison harness and the benchmarks out of the branch; in the follow-up, copying the point's map first
instead of guarding only `speed`.


OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
```diff
@@ -821,12 +829,25 @@ object GpxUtilities {
 				extensions.remove(BACKGROUND_TYPE_EXTENSION)
 			}
 		}
-		assignExtensionWriter(p, extensions, "extensions")
+		assignExtensionWriter(p, extensions, "extensions", false)
 		writeExtensions(serializer, null, p, null)
+		// the writers above exist only for this serialization; keeping them would attach two maps
+		// and a writer object to every point of the file on every save
+		p.removeExtensionsWriter("extensions")
+		p.removeExtensionsWriter(GPXTPX_TRACK_POINT_EXTENSION)
```
alex-dev-neo: `writeWpt` created the `"extensions"` writer just above, but not necessarily the `gpxtpx:TrackPointExtension` one — `SavingTrackHelper` registers under that same key for external-sensor values:

```java
// SavingTrackHelper.java:538 (and :420 for the DB path)
GpxUtilities.INSTANCE.assignExtensionWriter(wptPt, extensions, "plugins");
```

That call puts the regular keys under `"plugins"`, but the gpxtpx ones (`gpxtpx:hr`, `gpxtpx:cad`, `gpxtpx:power`, `gpxtpx:atemp`) under `GPXTPX_TRACK_POINT_EXTENSION`. Those points go into the long-lived in-memory current track.

On save, `assignExtensionWriter(p, extensions, "extensions", false)` above does **not** overwrite that writer (the point's own extensions map holds no gpxtpx keys), `writeExtensions` emits it correctly — and then this line deletes it. A second save of the same `GpxFile` object writes no `<gpxtpx:TrackPointExtension>` at all. `SaveCurrentTrackTask.doInBackground` already calls `writeGpxFile(fout, gpx)` once per collected date on the same `gpx`, so two dates are enough to lose the sensor data on the second write.

Suggestion: rather than removing by key, don't route serialization through the point at all. `writeExtensions(serializer, extensions, p, progress)` already accepts an explicit map, so the gpxtpx/regular split can stay local to `writeWpt` and no writer has to be installed and then undone. That also drops the need for the `keepDeferred` flag, and makes it exception-safe — as written, if `writeExtensions` throws, lines 836-839 never run and every point of the file keeps the writers, which is exactly the leak this PR removes.

Same removal in `GPXUtilities.java:1317`.

OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt
```diff
@@ -1741,24 +1778,42 @@ object GpxUtilities {
...
+							return
+						}
 					} catch (_: NumberFormatException) {
 					}
 				}
 
 				POINT_BEARING -> {
 					try {
 						target.bearing = toDouble(value).toFloat()
+						if (!target.bearing.isNaN()) {
```
alex-dev-neo: This early return also takes `bearing` out of the extensions map, and there is a reader that looks only there:

```java
// LocationSimulationUtils.java:134
String br = location.getExtensionsToRead().get("bearing");
if (!Algorithms.isEmpty(br)) {
    l.setBearing((float) Algorithms.parseDoubleSilently(br, 0));
}
```

Simulating a position from a recorded GPX now yields locations with no bearing at all. The speed lookup 8 lines above survives only because it checks `location.getSpeed() > 0` first; bearing has no such fallback. It should read the `WptPt.bearing` field instead.

Related, from the same move of `bearing` into the field: `WptPt(WptPt)` copies `speed` and `heading` but not `bearing`. Until now bearing survived a clone through `copyExtensions`; field-only, every clone drops it — `ReversePointsCommand:40`, `ChangeRouteModeCommand:59`, `MeasurementEditingContext:627`, `GpxFile.buildGeneralSegment`. `SharedUtil.jWptPt`/`kWptPt` do copy `bearing`, so the copy constructor looks like a plain omission.

Chumva:
Two behaviour differences against `r5.4` found while reviewing. Both are backed by a JVM test that passes on `801ef87` and fails on `e3989d5`; everything else I exercised (load/render/analysis, file rewrite, recording with plugin writers, the in-memory double save, regular save) checked out on a Pixel 8 Pro, and the heap claim reproduces there: 690 → 219 bytes per displayed point for a 100k-point track.

**1. `writeWpt`: a `speed`/`heading` string left in the point's map now overrides the field.**
The old code pushed the field values *into* the map, so the field won. The new code fills the local map from the fields and then `putAll(p.getExtensionsToRead())` — https://github.com/osmandapp/OsmAnd/blob/e3989d524fb544d544509ebc66f80ed52bae0ca1/OsmAnd-shared/src/commonMain/kotlin/net/osmand/shared/gpx/GpxUtilities.kt#L801-L808 and https://github.com/osmandapp/OsmAnd/blob/e3989d524fb544d544509ebc66f80ed52bae0ca1/OsmAnd-java/src/main/java/net/osmand/gpx/GPXUtilities.java#L1489-L1496 — so the map wins. Trigger: a point parsed with a non-positive or unparseable `<speed>` (the string is kept in the map by design) whose `speed` field is set afterwards, e.g. `WebGpxParser.addSegmentsToTrack` does `filePoint = point.ext; filePoint.setSpeed(point.speed)` on loaded points; the file then gets `<osmand:speed>0</osmand:speed>` instead of the real value. Fix: `putAll` first, then the two field writes — that restores the old precedence exactly, including the case where a non-positive string must round-trip.

**2. Null values from Java callers now throw in `put`/`putAll`.**
`GpxExtensionsMap.put(key: String, value: String)` carries Kotlin's parameter null check; the old `LinkedHashMap` stored the null and the writer skipped it. Every Java `put`/`putAll` site in `OsmAnd/src` passes non-null values as far as I can tell, except the Gson-decoded `Map<String, String>` in `TravelObfGpxFileReader` (https://github.com/osmandapp/OsmAnd/blob/e3989d524fb544d544509ebc66f80ed52bae0ca1/OsmAnd/src/net/osmand/plus/wikivoyage/data/TravelObfGpxFileReader.java#L271-L278), where a JSON `null` in `extra_tags` would now be an NPE. A cheap guard is a `putAll` override in `GpxExtensionsMap` (and in the Java `ExtensionsMap`) that skips null values; otherwise the contract change is worth a line in the description.

**Update:** 1 is fixed on this branch in c38be4c040, with tests that fail on e3989d524f. 2 stays as a known contract change - no in-app caller passes a null value unless the travel data carried a JSON `null`, so it is noted in the description instead of guarded in the map.

(gen by AI)

PR #26145, opened 2026-09-20
Read the map target without copying the renderer state on per-frame paths

Chumva:
## Problem

`MapRendererView.getTarget()` in the core wrapper is not a getter but four allocations:

```java
// OsmAnd-core r5.4, wrappers/android/src/net/osmand/core/android/MapRendererView.java:1224
public final PointI getTarget() {
    PointI fixedPixel = _mapRenderer.getState().getFixedPixel();      // MapRendererState copy + PointI
    if (fixedPixel.getX() >= 0 && fixedPixel.getY() >= 0) {
        return _mapRenderer.getState().getFixedLocation31();          // MapRendererState copy + PointI
    } else {
        return _mapRenderer.getState().getTarget31();
    }
}
```

`IMapRenderer.getState()` copies the whole `MapRendererState` natively under `_requestedStateMutex` (provider maps included) and wraps it in a Java proxy; `getFixedPixel()` / `getFixedLocation31()` copy a `PointI` the same way. Every one of these proxies has `finalize()`, so each call is four native allocations, four `FinalizerReference`s for every collection to walk, and four objects for the finalizer daemon to delete later. `getAzimuth()` and `getElevationAngle()` copy the whole state once more each.

On the Android side this ran on every frame while the map follows the location, and on every tick of the Java-side animations and every gesture:

1. `AnimateDraggingMapThread.animatingMapAnimator()` - the loop that mirrors the native animation into the Java tile box, one tick per redraw: `getTarget()` + `getState().getTarget31()` + `getAzimuth()` + `getElevationAngle()` = 8 finalizable objects per tick.
2. `PointLocationLayer.onUpdateFrame()` - `getTarget()` for the marker linked to the map target, then `marker.getPosition()` to read back the value it had just set.
3. `RouteLayer.drawLocations()` - `getTarget()` per frame when the route is drawn from the map centre (`ANIMATE_MY_LOCATION`, the default).
4. `OsmandMapTileView.setRotateImpl()` / `setZoomAndAnimationImpl()` per tick of the rotate and zoom animations and per rotate or pinch gesture, both `moveTo()` overloads per fling tick or key press, `AnimateDraggingMapThread.startMoving()` / `startDragging()` once per animation.

This is the mechanism behind the SWIG proxies in the field heap histograms of 5.5.0#8870m (Pixel 9 Pro XL, 2026-09-20): between two in-app histograms taken at most 30 minutes apart, 7,956 `MapRendererState`, 11,444 `PointI` and 28,878 `FinalizerReference` objects appeared, for an app with one renderer. The ANR traces of the same period show the main thread parked in `WaitingWeakGcRootRead` while the collector processes exactly those references.

Measured on the base build with heapprofd Java-heap sampling (`heaps: com.android.art`, 256-byte interval, 20 s) during car navigation, SWIG getter proxies created per second:

| call site | Pixel 8 Pro, Android 17 | emulator, Android 16 |
|---|---|---|
| `AnimateDraggingMapThread.animatingMapAnimator` | 1,304 | 172 |
| `RouteLayer.drawLocations` | 75 | 65 |
| `PointLocationLayer.onUpdateFrame` | 10 | 20 |
| everything reaching `MapRendererView.getTarget` | 713 | 154 |
| all SWIG getter proxies, any site | **1,458** | **1,018** |

## Fix

`IMapRenderer.getMapTargetLocation(PointI&)` already exists in the core and returns the same value as `getTarget()` - `fixedPixel.x < 0 || fixedPixel.y < 0 ? target31 : fixedLocation31`, read under the same mutex (`src/Map/MapRenderer.cpp:3583` on r5.4) - into a caller-owned `PointI`, with no copy of the state and no new proxy. `MapRendererView` does not expose it or its renderer, so `MapRendererContext`, which the app already keeps in `net.osmand.core.android`, gets a static `getTarget31()` that reads `_mapRenderer` through package access, in two forms:

- `getTarget31(mapRenderer, point)` fills the given point and returns it: the three per-frame paths keep one `PointI` and pass it every frame;
- `getTarget31(mapRenderer)` returns a new point: a drop-in for `getTarget()` on the per-tick and per-gesture paths, one `PointI` instead of two state copies and two points.

The animation loop takes azimuth, elevation angle and the flat target from one `MapState` per tick (`MapRendererView.getState()` is `_requestedState.getMapState()`, which carries `azimuth`, `elevationAngle`, `target31`, `fixedPixel` and `fixedLocation31` verbatim - `src/Map/MapRendererState.cpp:52`), instead of four `MapRendererState` copies plus one `MapState`. `PointLocationLayer` sets the location circle from the value it just assigned to the marker instead of reading it back.

Per tick or frame the three per-frame paths now create 2 finalizable objects (one `MapState`, one `PointI` for the flat target) instead of 17; `setRotateImpl()` goes from 8 to 5 per tick, `setZoomAndAnimationImpl()` from 9 to 6. The reused points are created on the frame path, not in field initializers: the layers are constructed in `OsmandApplication.onCreate()` before `NativeCore` is loaded, and a version of this change with a `PointI` field initializer crashed at start with `UnsatisfiedLinkError` in `OsmAndCoreJNI.swig_module_init()`.

No behaviour changes: the values read are the same fields of the same requested state, `MapMarker_P::setPosition()` stores the point unchanged, and a null renderer throws the same `NullPointerException` `getTarget()` threw.

## Verification

Builds are `gplayFullOpenglFatDebug` 5.4.0: base `22c369e7c3` (origin/r5.4 at the time) and this change, `0f94a24674`, rebased onto `c1a198edef`. Two earlier shapes of the change (per-frame paths only; then everything, with the target read into reused points on the gesture paths too) were measured the same way before being simplified into this commit; their numbers are consistent with the ones below and are not repeated.

- Google Pixel 8 Pro, Android 17, arm64 (fixes through `cmd location providers set-test-provider-location gps`, no speed/bearing in the fixes).
- Emulator Pixel_10_Pro, Android 16, x86_64 (fixes through `adb emu geo fix`).

Scenario, identical for every run: cold start at zoom 17 with the map linked to the location, car navigation Kyiv -> Bila Tserkva (86 km) started with `osmand.api://navigate?dest_lat=49.7961&dest_lon=30.1104&profile=car&force=true`, voice muted, GPS fed at 20 m/s along the route, 45 s to settle, then 20 s of heapprofd sampling, 120 s of ART GC log, and one `am dumpheap` (without `-g`) about 12 s after a collection to count the objects created since it.

```
adb shell "run-as net.osmand.plus sed -i 's/name=\"last_known_map_zoom\" value=\"[0-9]*\"/name=\"last_known_map_zoom\" value=\"17\"/' shared_prefs/net.osmand.settings.xml"
adb shell am start -n net.osmand.plus/net.osmand.plus.activities.MapActivity
adb shell am start -a android.intent.action.VIEW -d 'osmand.api://navigate?dest_lat=49.7961&dest_lon=30.1104&dest_name=BilaTserkva&profile=car&force=true' net.osmand.plus
adb shell am start -a android.intent.action.VIEW -d 'osmand.api://mute_navigation' net.osmand.plus
adb shell 'cat /data/local/tmp/heap.cfg | perfetto -c - --txt -o /data/misc/perfetto-traces/java_alloc.pftrace'   # android.heapprofd, heaps: "com.android.art", sampling_interval_bytes: 256, duration_ms: 20000
adb logcat -d --pid=$(adb shell pidof net.osmand.plus) | grep "GC freed"
adb shell am dumpheap net.osmand.plus /data/local/tmp/nav.hprof
```

### SWIG getter proxies per second (heapprofd, allocations inside `net.osmand.core.jni` frames, scaled bytes / 24), Pixel 8 Pro

| call site | before | after |
|---|---|---|
| `AnimateDraggingMapThread.animatingMapAnimator` | 1,304 | **314** |
| `RouteLayer.drawLocations` | 75 | 34 |
| `PointLocationLayer.onUpdateFrame` | 10 | 18 |
| reaching `MapRendererView.getTarget` | 713 | **0** |
| reaching `MapRendererContext.getTarget31` (the allocating form) | - | 1 |
| all SWIG getter proxies, any site | 1,458 | **465** |

What remains in the animation loop is the two objects per tick it is designed to create; the per-second figure moves between runs with how much the map animated in the sampled 20 s. What remains in `RouteLayer.drawLocations` is `QListVectorLine.get()` from `buildActionArrows()`, not target reads. On the emulator the same scenario gave 1,018 -> 94 proxies per second, but its animation loop happened to be idle during the sampled window of the last run, so the phone is the figure to read; on both devices nothing reaches `getTarget()` any more.

### Objects created since the last collection, by class (heap dump, unreachable from roots), Pixel 8 Pro, per second over a ~12 s window

| class | before | after |
|---|---|---|
| `net.osmand.core.jni.PointI` | 1,220 | **466** |
| `net.osmand.core.jni.MapRendererState` | 943 | **2** |
| `net.osmand.core.jni.MapState` | 129 | 144 |
| `net.osmand.core.jni.IMapRenderer` (core director callbacks) | 646 | 673 |
| `net.osmand.core.jni.VectorLine` (RouteLayer arrows) | 349 | 395 |
| `java.lang.ref.FinalizerReference` | 2,083 | **1,295** |

The emulator's dump windows are not usable for this table: with the synthetic feed it recalculates the route continuously and collects every 0.6 s (200 collections in 120 s on the base build), so a dump taken "12 s after a collection" covers less than a second there.

### Allocation rate

ART GC log over 120 s on the Pixel: 6 collections and 311 MB freed before, 4 collections and 354 MB after. The byte rate is dominated by route recalculation, tile loading and the route arrows; the 1,458 -> 465 proxies per second are ~25 KB/s and cannot be seen in it. The object counts above are the measure of this change.

Fling, keyboard scrolling (`KEYCODE_DPAD_*`) and the zoom buttons were exercised on the emulator with the final build: the map moves and zooms as before, no crash on either device across the runs.

Not verified: the legacy (non-OpenGL) renderer build - the changed lines run only with a `MapRendererView`; Android Auto; the phone was fed fixes without bearing, so its map did not rotate to the bearing (the emulator's did); no instrumented test - the change removes allocations from a per-frame path and there is no JVM or instrumented harness here that observes them. `git diff --check` is clean.

## Known gap, not addressed here

- `OsmandMapTileView.refreshMapInternal()` still copies the state once per refresh to read `getFixedPixel()` (no out-parameter for it in the core), and the touch-down handlers read `getTarget()` once per gesture.
- `PointLocationLayer.onUpdateFrame()` still reads `marker.getPosition()` once per frame for `lastMarkerLocation`, and its two other branches go through `containsLatLon()` (two `PointI` and a tile-box copy per frame).
- The core side: `MapRendererView.getTarget()`, `getAzimuth()`, `getElevationAngle()` and the master-branch `getZoomLevel()` should read fields instead of copying the state, `MapRendererView` should expose `getTarget(PointI)` so the package-level helper here can go, and the director callbacks create one `IMapRenderer` proxy per callback (673 per second on the Pixel above).
- `RouteLayer` rebuilds arrows and lines per frame (`VectorLine` ~400 per second), which is a separate change.

## AI disclaimer

Prompts used (summarised):
1. Audit all map layers, plugins and widgets for memory leaks and unnecessary allocations; list the problems by priority.
2. Try to reduce the memory churn from `PointI` in the layers with minimal changes; then also the remaining `getTarget()` callers on the rotate, zoom and drag paths.
3. Build `gplayFullOpenglX86Debug` to measure before and after (explicit permission for the Gradle builds); target r5.4; also measure on the attached Pixel 8 Pro.
4. Make it a single commit; make the change simpler, smaller and more readable.

Decided by the agent, not requested explicitly: using `IMapRenderer.getMapTargetLocation()` through a package-level helper in `MapRendererContext` instead of a core change; taking the animation loop's values from one `MapState`; the two helper forms (reused point on the per-frame paths, allocating drop-in elsewhere); leaving the core-side copies to follow-ups; building the `Fat` variant so one APK served both devices.


vshcherb:
https://github.com/osmandapp/OsmAnd-core/pull/1142/changes#diff-bc2b8caafd694e005af65b8e334cf37b11c36707cd6fd9753a4fab1d0a9abebb

PR #26158, opened 2026-09-21
Add backslash to illegal file name characters and update error string

### LINKED ISSUE #25795: Saving a route as a track, using invalid characters in the track name doesnt work and it doesnt warn user
### Description

when saving a track from a route, and using special characters such as `>` in the track name,
the track doesnt appear in the tracks list 

### Steps to reproduce

-make a route, go to details, save with a name containing `>` symbol.

### Actual result

-go to the sidebar, tracks, new track **doesn't** appear
AND
-while naming the new track, it **doesn't** tell you if the text field contains disallowed characters 

### Expected result

-go to the sidebar, tracks, new track **does** appear 
OR easier to implement?:
-while naming the new track, it **should** tell you if the text field contains disallowed characters, the same way it does when renaming in the tracks menu

```
OsmAnd Version: 5.3.10
Android version: 12
Device model: SM-G950F (codename: dreamlte)
```

kraativ:
Fixes osmandapp/OsmAnd#25795.

---

## Summary of Changes
- Added backslash `\` to `ILLEGAL_FILE_NAME_CHARACTERS` and `ILLEGAL_PATH_NAME_CHARACTERS` in `FileUtils.java`.
- Added backslash `\` to `illegalFileNameCharacters` in `TravelObfGpxBuilder.kt`.
- `file_name_containes_illegal_char` now lists the disallowed characters: the resource is `Name can\'t contain: %1$s` and the list is passed from code (second commit, see below).

---

### Problem
The first commit (a1bc323cd2) wrote the list into the resource as `Name can\'t contain: / \\ : * ? " &lt; &gt; |`. aapt2 treats an unescaped `"` in a string resource as a quote toggle, not as a character, and drops it silently - no build error or warning - so the compiled string was `Name can't contain: / \ : * ?  < > |`, eight of the nine characters (visible in the second screenshot below). The same escaping (`\"`, `\\`; and `\ ` becomes a plain space) would have had to be reproduced correctly in all 55 translations.

Also, `AddNewTrackFolderBottomSheet` and `TrackFolderOptionsController` validate with `ILLEGAL_PATH_NAME_CHARACTERS`, which deliberately allows `/` (a new folder `a/b` is created with `mkdirs()`), but showed the same message, so they claimed `/` was forbidden.

### Fix
`FileUtils` gets two display strings and builds both patterns from them, so the check and the message have one source:

```java
public static final String ILLEGAL_FILE_NAME_CHARS = "/ \\ : * ? \" < > |";
public static final String ILLEGAL_PATH_NAME_CHARS = "\\ : * ? \" < > |";
public static final Pattern ILLEGAL_FILE_NAME_CHARACTERS = Pattern.compile("[" + Pattern.quote(ILLEGAL_FILE_NAME_CHARS.replace(" ", "")) + "]");
public static final Pattern ILLEGAL_PATH_NAME_CHARACTERS = Pattern.compile("[" + Pattern.quote(ILLEGAL_PATH_NAME_CHARS.replace(" ", "")) + "]");
```

The resource becomes `Name can\'t contain: %1$s`; the seven call sites pass the file list or, in the two folder dialogs and in `checkRenamePossibility(dirAllowed = true)`, the path list.

Translations of `file_name_containes_illegal_char` are removed from all 55 `values-*/strings.xml` (one line each). Those locales fall back to the English message, which shows the characters, instead of the old generic text without them; Weblate will offer the new source with the placeholder for translation.

### Verification
- Derived patterns vs. the hand-written regexes of a1bc323cd2: identical `find()` result for every code unit U+0000..U+FFFF, checked on OpenJDK 21 and on ART (emulator, Android 16 / API 36, ICU-backed `java.util.regex.Pattern`).
- `aapt2 dump resources` on the built APK: `file_name_containes_illegal_char` has only the default config, value `Name can't contain: %1$s` (a neighbouring untouched string has 61 configs).
- Xiaomi Redmi Pad Pro (2405CRPFDG), Android 16 / API 36, `gplayFreeOpenglArm64Debug`, system locale Ukrainian (a locale whose translation was removed), error text read from the view hierarchy with `uiautomator dump`:

| Dialog | Input | Shown |
|---|---|---|
| Plan route -> Save as new track (`SaveAsNewTrackBottomSheetDialogFragment`, the flow from the issue) | `Mon 21 Sep 2026 "` | `Name can't contain: / \ : * ? " < > \|`, Save disabled |
| My Places -> Tracks -> Rename (`RenameFileBottomSheet`) | `Mon 21 Sep 2026 \` | `Name can't contain: / \ : * ? " < > \|`, Save disabled |
| Tracks -> Add new folder (`AddNewTrackFolderBottomSheet`) | `Hike*` | `Name can't contain: \ : * ? " < > \|`, Add disabled; `Hike/` clears the error |
| Tracks -> folder -> Rename -> Apply (`TrackFolderOptionsController`) | `import*` | toast `Name can't contain: \ : * ? " < > \|`, folder not renamed |

- Not verified on a device: `SaveGPXBottomSheet` (track recording) and `EditTrackGroupBottomSheet` (waypoint group), which received the same one-argument change; dark theme and RTL. `:OsmAnd-java:test` was skipped for the local build: `SearchUICoreTest[13: depo]` fails on this branch independently of the change (fixed on master by c7a24d6ef0; this module is untouched here).

---

## Screenshots (first commit)

The second one shows the dropped `"` described above.

<p align="center">
  Screenshot_20260921-142807
  &nbsp;&nbsp;
  Screenshot_20260921-145049
</p>

---

## AI disclaimer

Produced with Android Studio AI Agent (Gemini 3) - first commit; Claude Code (Opus 5) - review, second commit and this description.

Prompts used (summarised):

Android Studio AI Agent (Gemini 3):
1. Analyze issue #25795 regarding track naming with invalid characters and missing warnings.
2. Implement the requested UI review changes (adding backslash to illegal filename characters and updating the error string).
3. Test build, verify on device, commit to a new branch, and prepare PR.

Claude Code (Opus 5):
1. Analyze this pull request for issues, bugs, regressions and possible improvements.
2. Make the message a `%1$s` format string and pass the characters from code.
3. Make it even simpler and smaller.
4. Comment on the pull request about the problem found, verify the fix on the tablet or emulator, update the pull request, and update or remove the translated strings so they get translated properly.

- Gemini 3: extended the validation to `ILLEGAL_PATH_NAME_CHARACTERS` and the KMP shared builder `TravelObfGpxBuilder.kt` to ensure complete cross-platform consistency.
- Claude Code: building the patterns from the display strings (`Pattern.quote` of the list without spaces) instead of keeping two hand-maintained copies; passing the path list without `/` in the folder dialogs; deleting the translations rather than editing them; verifying the derived patterns against the previous regexes on OpenJDK and ART; leaving the KMP copy in `TravelObfGpxBuilder.kt` as it is.


Chumva:
(gen by AI)

Two things found while reviewing the resource change.

**1. The `"` is silently dropped from the new message.**

`file_name_containes_illegal_char` now contains a bare double quote:

```xml
<string name="file_name_containes_illegal_char">Name can\'t contain: / \\ : * ? " &lt; &gt; |</string>
```

aapt2 treats an unescaped `"` in a string resource as a whitespace-preserving quote toggle, not as a character, so it is removed without any build error or warning. Compiled the exact element in isolation with aapt2 35.0.0 (`aapt2 compile` + `aapt2 link` + `aapt2 dump strings`):

```
Name can\'t contain: / \\ : * ? " &lt; &gt; |     ->   Name can't contain: / \ : * ?  < > |
Name can\'t contain: / \\ : * ? \" &lt; &gt; |    ->   Name can't contain: / \ : * ? " < > |
```

The second screenshot in the description shows the same result on the device: `/ \ : * ?  < > |` with the `"` missing and a double space in its place. So the message lists eight of the nine forbidden characters. The rest of `strings.xml` escapes literal quotes as `\"` (92 occurrences).

**2. The folder dialogs now claim `/` is forbidden while accepting it.**

`AddNewTrackFolderBottomSheet` and `TrackFolderOptionsController` validate with `ILLEGAL_PATH_NAME_CHARACTERS`, which deliberately allows `/` (a new folder name `a/b` creates nested folders via `mkdirs()`), but they show the same string, so a user who types `*` there is told that `/` is not allowed either. With the previous generic text this did not matter.

Both points, plus the risk of the same escaping mistake being repeated in every translation (`\\` and `\"` would have to be typed correctly in 55 locales, and aapt2 turns `\ ` into a plain space just as silently), go away if the message is `Name can\'t contain: %1$s` and the list is passed from the same `FileUtils` constants the regexes are built from. A follow-up commit doing that is being pushed to this branch; details in the next comment and in the updated description.

Chumva:
(gen by AI)

Pushed 56f875074d to this branch: the resource is now `Name can\'t contain: %1$s`, the list comes from `FileUtils` (`ILLEGAL_FILE_NAME_CHARS` / `ILLEGAL_PATH_NAME_CHARS`), and both patterns are built from those same strings. The 55 translations of the string are removed so those locales fall back to English with the list until Weblate delivers a translation with the placeholder.

Verified on a Redmi Pad Pro (Android 16, `gplayFreeOpenglArm64Debug`): save-as-new-track and rename dialogs show `Name can't contain: / \ : * ? " < > |`, the new-folder dialog and the folder-rename toast show `Name can't contain: \ : * ? " < > |`. Details and what was not verified are in the updated description.

PR #26181, opened 2026-09-21
Survive the renderer stopping under the development plugin's sampler

Chumva:
Follow-up to #26167.

## Problem

Since #26167 (`4593d91144`, 21 Sep) the development plugin's `AvgStatsEntry` is built on its own `HandlerThread`. Its frame counters are plain fields of `MapRendererView`, but `getBasicThreadsCPULoad()` dereferences the native renderer proxy:

```
// net.osmand.core.android.MapRendererView (OsmAndCore_android 5.4, from the class file)
public final float getBasicThreadsCPULoad() { NativeCore.checkIfLoaded(); return _mapRenderer.getBasicThreadsCPULoad(); }
```

`_mapRenderer` is set to `null` by the main thread in `stopRenderer()` / `releaseRendering()` when the map activity goes away or the rendering engine is switched, while `OsmandMapTileView.getMapRenderer()` keeps returning the view. The `renderer != null` check at the top of the constructor does not cover it. While the sampler ran on the main looper the two could not overlap; now a sample taken in that window throws a `NullPointerException` on the collector thread, and an uncaught exception there ends the process.

## Fix

The one native call is wrapped; a sample taken while the renderer stops keeps a CPU load of zero. The frame counters above it stay as they are. Seven lines.

## Verification

Not reproduced: the window is the few instructions between the `null` check and the call, once every ten seconds. Read from the core's `MapRendererView` class (the proxy is only nulled, never deleted, so the failure is a Java `NullPointerException` and not a native crash, and a `catch` is the right shape).

The fixed build (`gplayFullOpenglFatDebug`, origin/r5.4 `4593d91144` plus the memory commits of this series; emulator `Pixel_10_Pro`, Android 16) ran with the development plugin enabled and its widgets on screen through the idle, driving and navigation scenarios of the series: the FPS and memory widgets kept updating (screenshots), 0 crashes. The activity-destroyed-while-the-process-lives case was exercised only with the agent's own earlier version of the collector thread, not with the code of #26167.

## AI disclaimer

Prompts used (summarised):
1. Fix locally everything memory-related that can be done with small changes, as separate commits.
2. Re-check the pull requests vshcherb opened today - maybe some of these were already fixed.
3. Open separate pull requests for each fix, with a description and a results table.

Decided by the agent, not requested explicitly: this guard is the one part of the agent's own version of the collector change that #26167 does not have; the rest of that version was dropped in favour of the merged pull request.


OsmAnd/src/net/osmand/plus/plugins/development/OsmandDevelopmentPlugin.java
```diff
@@ -405,8 +405,13 @@ private AvgStatsEntry(OsmandApplication app) {
 				this.idle1k = renderer.getIdleTimePartLast1K();
 				this.gpu1k = renderer.getGPUWaitTimePartLast1K();
 
-				float cpuBasic = renderer.getBasicThreadsCPULoad();
-				this.cpuBasic = cpuBasic > 0 ? cpuBasic : 0; // NaN
+				try {
```
vshcherb, line 408: NPE?

